PR #476, opened 2018-04-16
Add 2nd order time integrator [add-2ndorder-timeint-dev]

IdoAkkerman:
This implements some second order time integrators such as Newmark/Gen-alpha/HTT/WBZ.
Another time dependent operator has been defined. This is not a very elegant extension, as the "operator" requires 2 vectors as input (solution and rate) to obtain the acceleration.

Two routines to check time integrators are included in the mini apps/tools directory.

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge| 
 | --- | --- | --- | --- | --- | --- | --- | 
| [#476](https://github.com/mfem/mfem/pull/476) | @IdoAkkerman | @tzanio | @cjvogl + @stefanozampini | 01/13/20 | 02/19/20 | 02/27/20 | |

PR #1262, opened 2020-01-24
Adjustments to the branch-history script [branch-history-dev]

tzanio:
The [branch-history script](https://github.com/mfem/mfem/blob/master/tests/scripts/branch-history) may be too conservative currently.

In this PR we want discuss adjustments suggested by the team as they use it.

These adjustments, include:
- Adding filtering 
- Changing the default ["acceptable" values](https://github.com/mfem/mfem/blob/master/tests/scripts/branch-history#L66-L77)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -199,5 +199,9 @@ miniapps/gslib/pfindpts
 tests/unit/output_meshes
 tests/unit/unit_tests
 
+# Test script output
+tests/scripts/branch-history.err
+tests/scripts/branch-history.out
+
 # VPATH builds
 build-*/*
--- a/tests/scripts/branch-history
+++ b/tests/scripts/branch-history
@@ -64,11 +64,11 @@ EOF
 esac
 
 # Maximum number of acceptable commits in the branch
-branch_max_commits=100
+branch_max_commits=200
 # Maximum number of acceptable files changed in any commit in the branch
-commit_max_files_changed=100
+commit_max_files_changed=50
 # Maximum number of acceptable lines changed in any commit in the branch
-commit_max_lines_changed=3000
+commit_max_lines_changed=2000
 # Maximum acceptable size (in K) of any file changed in any commit in the branch
 file_max_size=200
 # Maximum number of acceptable lines added to any file changed in any commit in the branch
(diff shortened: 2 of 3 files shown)


tzanio:
@mfem/reviewers, please comment here what changes you'd like to see in the new `branch-history` script.

The initial commit in this PR will filter out changes in the existing source files.

tzanio:
See also #1267 -- do we want to incorporate part of the scripts in makefile targets, or should we encourage folks to run the scripts themselves and document them better?

IdoAkkerman:
The script checks all commits, not the final changes.
For instance  in #476 I changed fe.cpp somewhere and changed it back later (debug stuff I guess).
It now gets flagged as exceptionally large file.
Is it reasonable to check every commit? Instead of just the final changes?

File size check will flag changes to fe.cpp and mesh.cpp. Why would the file size be an issue?
Shouldn't that mean that these files need to be split in smaller ones?

tzanio:
> The script checks all commits, not the final changes.
> For instance in #476 I changed fe.cpp somewhere and changed it back later (debug stuff I guess).
> It now gets flagged as exceptionally large file.
> Is it reasonable to check every commit? Instead of just the final changes?

It checks all commits that are not in master, which is very much intentional. In the settings of a GitHub PR if a large file has been added in one commit and deleted in another the reviewers will not see it (unless they review every single commit). 

> File size check will flag changes to fe.cpp and mesh.cpp. 

These two files will be addressed

IdoAkkerman:
If not done already:
tests/scripts/*.err
tests/scripts/*.out
should be added to .gitignore

IdoAkkerman:
Now realize, it was perhaps out of place to make this commit.
If I overstepped here please let me know!!

tzanio:
> Now realize, it was perhaps out of place to make this commit.
> If I overstepped here please let me know!!

In general it is good to ask, but that particular commit is fine.

It should be `tests/scripts/*.err` and `tests/scripts/*.out` though...

tzanio:
We should update the [Automated Testing]( https://github.com/mfem/mfem/blob/master/CONTRIBUTING.md#automated-testing) section of `CONTRIBUTING.md` with description of the travis checks which in particular explains that failures in `branch-history` are acceptable.

pazner:
We can also think about incorporating something like this script: https://github.com/mfem/mfem/issues/1255#issuecomment-586528331

The goal would be to reduce the number of false positives resulting from small modifications to large files like `mesh.cpp`.

tzanio:
Closed in favor of #1267

PR #1263, opened 2020-01-24
Minimal surface example [minimal-surf-dev-rewritten]

camierjs:
This PR adds a new `meshing` `miniapp`, which solves Plateau's problem: the Dirichlet problem for the minimal surface equation. 

It adds the `SDIM=3` kernels for `diffusion` (setup) and `vecdiffusion` (setup & apply), and supports two types of solvers (`ByVDIM` or `ByNODES`), on all `devices` and w/ or w/o `AMR`.

Two problems can be run: 
 - problem 0 solves the minimal surface equation of parametric surfaces,
 - problem 1 solves  `-div( q grad u) = 0, with q(u) = (1+|∇u|²)^{-1/2}` for surfaces of the form `z=f(x,y)`, using Picard iterations.   

Here are the different parameterizations available so far:

- `Costa`
Costa

Costa-amr

Costa-AMR-HO

CostaAMRLoop_s

- `Mobius`
Mobius

- `Catenoid`
Catenoid

- `Helicoid`
Helicoid

- `Enneper`
Enneper

- `Scherk`
Scherk

- `Hold`
Hold

| [#1263](https://github.com/mfem/mfem/pull/1263) | @camierjs | @tzanio | @mlstowell + @stefanozampini + @rcarson3 | 04/01/20 | 04/15/20 | 04/21/20 | |


adrienbernede:
@tzanio @camierjs 
For the record, this branch was rewritten using `git rebase --interactive --rebase-merges`, rewriting the operations done by git to preserve the rest of the history, and going through the same merge conflicts.

I would say this is the hard way. But I could understand what I was doing and why it was safe.

However, @camierjs found the [filter-repo](https://github.com/newren/git-filter-repo) project and tried to rewrite the history with it unfortunately without success. I thought there might be a way to do it right with this tool, so I created [an issue to ask for help](https://github.com/newren/git-filter-repo/issues/50). It led to a very interesting conversation about how this works, and how to choose between rebase and filter repo.

In the end I tried and was able to clean the branch history with this command in just seconds and no hassle because it matches all the pre-requisites.

I remain convinced that developers should use git-hooks to perform these checks at commit time to avoid the hassle/risk of rewriting history when the PR is finally ready.

tzanio:
Thanks @adrienbernede for all your help!

For completeness, can you post the recipe for using the filter-repo tool (when it works)?

adrienbernede:
Attempt for a recipe:
`filter-repo` will rewrite the history applying the desired change _as long as it does not require conflict resolution_. This is a bit of a shortcut, and is better explained [here](https://github.com/newren/git-filter-repo/issues/50#issuecomment-578340676). We could also say _as long as the changes are at file level_ (compare to the usual line level).

In particular, it means that it is particularly efficient to delete a file, because it will "simply" impose that the file be removed from each commit. To better understand the nuance: if this file has been renamed at some point, `filter-repo` won’t catch it, and the file with the new name will persist in history as if it were added when it was renamed.

The working syntax is the following:
`git filter-repo --refs ^master feature --invert-paths --path <path-to-delete>`

Which reads: filter the repository for commit attainable from `feature` but not from `master` by taking from the existing history everything except the `path-to-delete`.

Note: `filter-repo` is much more versatile, and can be used for example to split an existing project is separate repositories, or extract a subset, change directory names, ... anything that could strongly impact the history and make it incompatible with the original one.

adrienbernede:
@camierjs In travis I manually triggered the optional to verify its output. Otherwise it wouldn’t have been run due to the failure in the first stage.

The failure in the first stage is due to a failing test not a gitignore issue. That’s because we need to run the tests to perform the gitignore check. This workflow will be improved in the future.

camierjs:
LGTM

camierjs:
LGTM

tzanio:
Awesome miniapp!

miniapps/meshing/mesh-minimal-surface.cpp
```diff
@@ -0,0 +1,1283 @@
+// Copyright (c) 2010-2020, Lawrence Livermore National Security, LLC. Produced
```
tzanio, line 1: I suggest renaming this to simply `minimal-surface`.

camierjs: `done`

camierjs:
LGTM

tzanio:
@mlstowell, @stefanozampini and @rcarson3 -- the reivews of this PR are due soon

fem/bilininteg.cpp
```diff
@@ -2162,36 +2162,29 @@ void VectorDiffusionIntegrator::AssembleElementMatrix(
       }
    }
 
-   elmat = 0.0;
```
mlstowell, line 2165: We should probably still initialize `elmat` to zero so ensure that its off-diagonal bocks are properly initialized.

miniapps/meshing/minimal-surface.cpp
```diff
@@ -0,0 +1,1302 @@
...
+   }
+
+   virtual void Refine()
+   { for (int l = 0; l < opt.refine; l++) { UniformRefinement(); } }
+
+   virtual void Snap()
+   {
+      GridFunction &nodes = *GetNodes();
+      for (int i = 0; i < nodes.Size(); i++)
+      { if (std::abs(nodes(i)) < EPS) { nodes(i) = 0.0; } }
```
stefanozampini, line 202: is this example conforming to MFEM coding conventions? (i.e. by running `astyle`)

miniapps/meshing/minimal-surface.cpp
```diff
@@ -0,0 +1,1302 @@
...
+               }
+               // project the delta vector in radial direction
+               const double ndotd = (ni*di) / (ni*ni);
+               di.Set(ndotd,ni);
+               // set global vectors
+               for (int d = 0; d < SDIM; d++) { delta(d*ndof + i) = di(d); }
+            }
+            add(*nodes, delta, *nodes);
+         }
+         // x = λ*nodes + (1-λ)*x
```
stefanozampini, line 372: Do we really need these special characters (lambda, delta above) in comments?

miniapps/meshing/minimal-surface.cpp
```diff
@@ -0,0 +1,1302 @@
...
+            const int NQ = ir->GetNPoints();
+            for (int q = 0; q < NQ; q++)
+            {
+               eTr->SetIntPoint(&ir->IntPoint(q));
+               const DenseMatrix &J = eTr->Jacobian();
+               CalcAdjugate(J, Jadj);
+               Jadjt = Jadj;
+               Jadjt.Transpose();
+               const double w = Jadjt.Weight();
+               minW = fmin(minW, w);
```
stefanozampini, line 400: A matter of taste, but I always prefer to explicitly use the `std::` namespace (for `fmin` in this case) and explicitly use the proper include inside the compilation unit

miniapps/meshing/minimal-surface.cpp
```diff
@@ -0,0 +1,1302 @@
...
+   }
+   S->Create();
+   S->Solve();
+   delete S;
+   return 0;
+}
+
+// Problem 1: solve the Dirichlet problem for the minimal surface equation
+//            of the form z=f(x,y), using Picard iterations.
+static double u0(Vector &x) { return sin(3.0 * PI * (x[1] + x[0])); }
```
mlstowell, line 1135: I believe `FunctionCoefficients` based on this type of function have been deprecated.  The argument `const Vector &x` is preferred.

miniapps/meshing/pminimal-surface.cpp
```diff
@@ -0,0 +1,1302 @@
...
+                  "--no-vis-mesh", "Enable or disable mesh visualization.");
+   args.AddOption(&opt.by_vdim, "-c", "--solve-byvdim",
+                  "-no-c", "--solve-bynodes",
+                  "Enable or disable the 'ByVdim' solver");
+   args.AddOption(&opt.print, "-print", "--print", "-no-print", "--no-print",
+                  "Enable or disable result output (files in mfem format).");
+   args.Parse();
+   if (!args.Good()) { args.PrintUsage(mfem::out); MPI_Finalize(); return 1; }
+   MFEM_VERIFY(opt.lambda >= 0.0 && opt.lambda <= 1.0,"");
+   MFEM_VERIFY(!opt.vis_mesh, "Option not available in parallel!");
```
mlstowell, line 1289: If this option is not available then why offer it?

miniapps/meshing/minimal-surface.cpp
```diff
@@ -0,0 +1,1302 @@
...
+   bool pa = true;
+   bool vis = true;
+   bool amr = false;
+   bool wait = false;
+   bool print = false;
+   bool radial = false;
+   bool by_vdim = false;
+   bool snapshot = false;
+   bool vis_mesh = false;
+   double tau = 1.0;
```
mlstowell, line 106: Just out of curiosity, why is `tau` not available in parallel?

fem/bilininteg.cpp
camierjs: We've switched from an `+=` to an `=` for the last step where `elmat` is assigned: that's why I removed it. I think we should fill all of it -- if I'm not missing something.

miniapps/meshing/minimal-surface.cpp
camierjs: The `make style` runs fine; but I agree, I could skip lines there.

miniapps/meshing/minimal-surface.cpp
camierjs: I removed them, right.

camierjs: `done`

miniapps/meshing/minimal-surface.cpp
camierjs: I added new lines with for-loop statements.

camierjs: `done`

fem/bilininteg.cpp
mlstowell: True but you only fill `sdim` blocks along the diagonal which each have a size of `dof x dof`.  So, in 2D you fill 2 blocks on the diagonal and this leaves 2 off-diagonal blocks of size `dof x dof` untouched, in 3D there are 3 filled blocks and 6 untouched blocks.  To put it another way the nested loop only makes `sdim*dof*dof` assignments but there are `sdim*sdim*dof*dof` entries.

camierjs: Ok, thanks -- I'll put it back.

camierjs: `done`

miniapps/meshing/minimal-surface.cpp
camierjs: sure, `done`

miniapps/meshing/minimal-surface.cpp
camierjs: I've added the `const`, right.

camierjs: `done`

miniapps/meshing/pminimal-surface.cpp
camierjs: You're right - `removed`

miniapps/meshing/minimal-surface.cpp
camierjs: I was using only in serial to create the gif, I pushed it to the parallel version too.

camierjs:
LGTM

camierjs:
I wanted to thank @pazner and @v-dobrev for their patience answering all of my questions: it was a big help!

miniapps/meshing/minimal-surface.cpp
```diff
@@ -0,0 +1,1326 @@
...
+   for (int i = 0; i < opt.niters; i++)
+   {
+      b = 0.0;
+      uold = u;
+      BilinearForm a(&fes);
+      if (opt.pa) { a.SetAssemblyLevel(AssemblyLevel::PARTIAL); }
+      const double q_uold = qf(order, AREA, mesh, fes, uold);
+      MFEM_VERIFY(std::fabs(q_uold) > EPS,"");
+      ConstantCoefficient q_uold_cc(1.0/sqrt(q_uold));
+      a.AddDomainIntegrator(new DiffusionIntegrator(q_uold_cc));
```
rcarson3, line 1252: Here and also in the parallel version the ```BilinearForm```, ```ConstantCoefficient```, and ```CGSolver``` objects could probably be hoisted out of the loop, so they aren't constantly created each iteration. However, this problem appears to run fairly fast, so I'm not too sure it's that big of a deal for this miniapp.

camierjs: Thanks for the review @rcarson3 , the `BilinearForm` is using the `DiffusionIntegrator` with the `ConstantCoefficient` built from `uold` so I kept it in for-loop body, but I moved the CG out of it as you suggested.

camierjs: `done`

tzanio:

PR #1208, opened 2019-12-21
Laghos GPU addons [laghos]

camierjs:
This PR brings:

- a couple of optimizations from @pwang234 (`SmemPAMassApply3D` & `ElementRestriction::Mult`),
- few new `QuadratureInterpolator` tensor functions:
   - `Values`: interpolate the values of the E-vector at quadrature points,
   - `Derivatives`: interpolate the derivatives of the E-vector,
   - `PhysDerivatives`: interpolate the derivatives in physical space of the E-vector,
- a way to propose small linear algebra code useful on the GPU now in `linalg/blas.hpp`.

---
Before:
[plot2_MFEM_cuda_bp1_linux_gcc_N001_std.pdf](https://github.com/mfem/mfem/files/3990866/plot2_MFEM_cuda_bp1_linux_gcc_N001_std.pdf)

After:
[plot2_MFEM_cuda_bp1_linux_gcc_N001_all.pdf](https://github.com/mfem/mfem/files/3990869/plot2_MFEM_cuda_bp1_linux_gcc_N001_all.pdf)

---
- [x] `NVTX` cleanup
- [x] move code into `linalg/blas.hpp` 

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge| 
 | --- | --- | --- | --- | --- | --- | --- | 
| [#1208](https://github.com/mfem/mfem/pull/1208) | @camierjs | @tzanio | @artv3 + @v-dobrev | 01/24/20 | 02/26/20 | 03/10/20 | |

PR #1270, opened 2020-01-29
More extensions for Laghos

artv3:
PR https://github.com/mfem/mfem/pull/1208 has some useful functions, I would like to propose a few more methods that would be useful in applications. 

~~This PR is a work in progress, working towards having this complete by Feb 3rd~~


artv3:
@v-dobrev @camierjs , I added the methods I needed for my application namely``PhysDerivatives`` . 

This function calculates the vector gradient by first calling ``Derivatives`` then a second method ``ApplyInvJac`` which applies the inverse of the Jacobian taking the solution from the reference space to the physical space. Currently this requires launching two kernels, is there a preference in terms of having a monolithic kernel that does everything in single kernel call? I can also do some more optimization.

tzanio:
> **This PR is a work in progress, working towards having this complete by Feb 3rd**

Can you remove this if it is no longer relevant?

tzanio:
I don't think we need the formal review process for PRs like this.

@camierjs, feel free to review and merge directly if you agree.

PR #1274, opened 2020-01-31
[WIP] Fix shared edge/face orientation in NC meshes [proj-disc-coef-fix]

### LINKED ISSUE #1230: Bug in ProjectDiscCoefficient with NCMESH and HDIV/HCURL spaces
Attached is an MWE that, when run with 4 processes, exercises the bug

BUG

Note that, if I don't call EnsureNCMesh, the output is fine.
[test.cpp.txt](https://github.com/mfem/mfem/files/4052748/test.cpp.txt)
[test.mesh.txt](https://github.com/mfem/mfem/files/4052749/test.mesh.txt)

jakubcerveny:
This is work in progress to fix #1230. The problem is that when accumulating values on shared edges/faces in `GridFunction::AccumulateAndCountZones`, the conforming code somehow manages to keep the local edge/face orientation consistent across processors. NCMesh does not produce a globally consistent vertex numbering and thus edge/face orientations on processor boundary varies. This is not a problem normally because the P matrix handles this. However in `ProjectDiscCoefficient`, element DOFs are used as-is and are assumed to be the right orientation. I'm not sure how this is handled in the conforming code, either the Mesh carefully numbers the vertices to keep edge/face orientations or there must be something in the way the communicator and shared elements are ordered -- it's hard to tell from the code.

I changed the repro code to produce circular contours (so the right solution is easier to tell). This is how the bug looks like if NC mesh is turned on:
proj-disc-bug

ParNCMesh is now modified to order the vertices along the global SFC element order. This seems to fix the problem in 2D but not in 3D: 
proj-disc-bug-3d

I'll look more next week.


jakubcerveny:
@tzanio, I'm stuck and might need a bit of help with this:

I'm trying to figure out why the function `ParGridFunction::ProjectDiscCoefficient` (on a vector coefficient) projects face DOFs with an incorrect sign (for a shared face).

Here is when the test program `stefano` is run on a conforming mesh on two processors (this is a simple vector coefficient that I defined):
proj-disc1

If the line `stefano.cpp:17` is uncommented (`EnsureNCMesh`), the four shared DOFs on the right processor (this is RT order 1) are negative and they sum up to zero incorrectly in `pgridfunc.cpp:554` (`gcomm.Reduce<double>`):
proj-disc2

The negative DOF values are obtained in `GridFunction::AccumulateAndCountZones` on line 1945 (`fes->GetFE(i)->Project`). 

I have trouble figuring out where do the negative signs come into play. The physical vertices are the same in both the conforming/NC runs. The shared face in `ParMesh::shared_quads` has the same vertex indices in the same order. The shared local DOFs on both processors are the same regardless of C/NC (they are (8,9,10,11) on rank 0 and (17,16,19,18) on rank 1).

There must be some normal somewhere that is flipped. Something with the `face_info` (I'm always confused how these should behave on shared faces).

jakubcerveny:
So it really is the `ldof_sign`, which is not applied in the NC case, although it is calculated with the same result in both C/NC. We definitely do not want to apply it every time in the NC case because the P matrix doesn't need it in NC. I need to think more about this and when to apply the signs.

jakubcerveny:
I was thinking more about this and the quick and dirty solution is to apply `ldof` signs in `GridFunction::AccumulateAndCountZones` in the case of an NC mesh. But this is a hack and not really the solution of the original problem.

I'll probably rework the way faces are initialized in the Mesh, if NCMesh is present. Right now face orientation is controlled by the Mesh class which doesn't know about other processors, unlike the NCMesh. If faces were generated by (Par)NCMesh, we could choose a globally consistent orientation according to the SFC-based ordering of vertices, much like this already works for edges thanks to #1392. Then the `ldof_sign` array will always be positive in the NC case.

jakubcerveny:
I wrote a unit test for the current fix of `ProjectDiscCoefficient` and it revealed a deeper problem with NC meshes and `ldof` signs. Years ago when I implemented the P matrix for NC meshes, I thought the `ldof` signs were just an artifact of the P matrix being boolean, and avoided the explicit sign flips in `ParFiniteElementSpace::GetElementDofs`. But it seems that in many places in `ParGridFunction`, e.g. in L2 error calculation or vector-valued projection, the sign flips are necessary, as they are really part of the space. I'm thinking now of always applying the signs and adjusting the P matrix construction...

tzanio:
Maybe I'm missing something here, but what do you mean by "the sign flips are really part of the space". My understanding is also that they are needed just in conjunction with `P`.

jakubcerveny:
I'm also still confused by this, but since for a conforming mesh the method `ParFESpace::GetElementDofs` always flips the sign of some of the shape functions (so that they fit across the CPU boundary), it seems to me that the inverted shape is meant to be how the space is defined. Right now in NC mode we get the correct solution when applying P (when coming from true DOFs to all DOFs), but this only hides the problem for RT spaces. Then in places that don't use P, like `ParGridFunction`, the space is incorrectly defined and it shows in `CalcL2Error`, `ProjectCoefficient`, etc. 

The problem is partially fixed by the new global vertex ordering due to `par-ncmesh-saving-dev`, so edges are consistently oriented, but for faces the same thing doesn't seem possible, so we're still stuck with the `ldof` signs.

PR #1278, opened 2020-02-03
Small fixes when running a matrix-free code (MATSHELL)

stefanozampini:
We need a couple of fixes when the user (in this case myself) wants to use a `MATSHELL`, i.e. by specifying as jacobian type for the nonlinear solvers either `Operator::ANY_TYPE` (in combination with a Jacobian operator which is not a `PetscParMatrix`) or `Operator::PETSC_MATSHELL`

First, we need to tell the DM we are using a matrix type (`MATSHELL`) for which `DMSHELL` has poor support. Second, we need to change the way the `MATSHELL` uses its context
Since it is barely impossible to do reference counting on void* memory, this branch reverts back the context of the `MATSHELL` as it was initially for MFEM, only a pointer to the `mfem::Operator`.

| [#1278](https://github.com/mfem/mfem/pull/1278) | @stefanozampini | @tzanio | @tzanio + @jandrej | 02/03/20 | 02/19/20 | 02/23/20 | |

### DIFF
--- a/linalg/petsc.cpp
+++ b/linalg/petsc.cpp
@@ -3819,6 +3819,14 @@ static PetscErrorCode __mfem_ts_ijacobian(TS ts, PetscReal t, Vec x,
    // Jacobian reusage
    ierr = PetscObjectStateGet((PetscObject)P,&ts_ctx->cached_ijacstate);
    CHKERRQ(ierr);
+
+   // Fool DM
+   DM dm;
+   MatType mtype;
+   ierr = MatGetType(P,&mtype); CHKERRQ(ierr);
+   ierr = TSGetDM(ts,&dm); CHKERRQ(ierr);
+   ierr = DMSetMatType(dm,mtype); CHKERRQ(ierr);
+   ierr = DMShellSetMatrix(dm,P); CHKERRQ(ierr);
    PetscFunctionReturn(0);
 }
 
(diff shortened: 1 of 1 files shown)


stefanozampini:
A question: is there any plan to add smart pointers support in MFEM? This would be nice to have, at least for the `mfem::Operator` class

linalg/petsc.cpp
```diff
@@ -3819,6 +3816,14 @@ static PetscErrorCode __mfem_ts_ijacobian(TS ts, PetscReal t, Vec x,
...
    ierr = PetscObjectStateGet((PetscObject)P,&ts_ctx->cached_ijacstate);
    CHKERRQ(ierr);
+
+   // Fool DM
+   DM dm;
+   MatType mtype;
+   ierr = MatGetType(P,&mtype); CHKERRQ(ierr);
+   ierr = TSGetDM(ts,&dm); CHKERRQ(ierr);
+   ierr = DMSetMatType(dm,mtype); CHKERRQ(ierr);
+   ierr = DMShellSetMatrix(dm,P); CHKERRQ(ierr);
```
jandrej: Maybe I don't understand the context here, but since this is needed repeatedly it does look like it is a limitation in PETSc right now. Do you want to keep this here as a workaround until there is a better option in PETSc or we do reference counting?

stefanozampini: The alternative is for MFEM to manage the `DM` , but this requires coding the class `DMMFEM` and it is not a priority now; another aspect is that PETSc uses matrices as containers, which the users can modify by setting values etc, while MFEM concept of getting back the Jacobian is somewhat different (you almost always can get back a different matrix every time). Also, this fix is specific for `DMSHELL` ( which is the default if the user does not provide any) and is a no-op call for any other `DM` type. So it is fine if it stays here (forever....()

tzanio:
@jandrej can we discuss this in person next week?

jandrej:
Absolutely!

tzanio:

tzanio:

tzanio:

PR #1290, opened 2020-02-10
Adding element types to test_fe.cpp [test-fe-dev]

mlstowell:
Adding tests for SEGMENT, TRIANGLE, QUADRILATERAL, TETRAHEDRON, and WEDGE elements for H1, ND, RT, and L2 spaces where applicable.

Also, added loops over the order so that we can verify that the number of degrees of freedom changes as expected.

| [#1290](https://github.com/mfem/mfem/pull/1290) | @mlstowell | @tzanio | @acfisher + @akgillette | 02/16/20 | 03/16/20 | 03/22/20 | |


gillette7:
Looks reasonable.  A couple comments:

1) I'm not so sure how to "test" this test.  
2) In the loops like this 
`for (int p = 1; p<=2; p++)`
I think `p` represents the order of the element?  But sometimes `p` is capped at 2 and other times at 3... why not go higher?

mlstowell:
Hi, @gillette7 ,

This is certainly a very basic sanity check so "testing" it does seem trivial.

You make a good point about the order.  I think this test originally used a fixed order but I wanted to make sure the number of DoFs was being computed as expected.  In 2D three values should be sufficient and we should use four in 3D.  Not sure why I picked 2 and 3.  I'd be happy to change them.  What are your thoughts?

Best wishes,
Mark

gillette7:
I think up to p=3 in 2D and p=4 in 3D is sufficient - looks like that would require a few small changes.  Otherwise looks good!

tzanio:
@acfisher, do you need more time to review this?

tzanio:

tzanio:
LGTM

PR #1299, opened 2020-02-17
Fad -   Forward AD native MFEM class with unit tests

bslazarov:
This pull request consists of a simple automatic differentiation template class and several unit tests. It is the first of several pull-requests for adding new features to accommodate AD-related implementations in MFEM. The class is templated in fdual.hpp and the unit tests are implemented in tests/unit/linalg/test_fdual.cpp. The implementation is based on dual numbers.

### DIFF
--- a/linalg/fdual.hpp
+++ b/linalg/fdual.hpp
@@ -10,156 +10,155 @@
 // Software Foundation) version 2.1 dated February 1999.
 
 
-
 #ifndef FDUAL_H
 #define FDUAL_H
-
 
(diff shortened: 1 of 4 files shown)


tzanio:
@bslazarov can we wait until this is used inside the library before reviewing it? Right now it is a standalone capability, but I think it will be better to review it when it is somewhat integrated... Any objections?

bslazarov:
We can wait. It is better to have an example where the AD is utilized. We
will have one soon.

Issue #1052, opened 2019-08-30
SLI requires SPD preconditioner


ben-s-southworth:
The stationary linear iteration solver, i.e., fixed-point iteration, requires an SPD preconditioner because it prints the residual/checks stopping tolerance in the preconditioned norm, $$\langle B\mathbf{r},\mathbf{r}\rangle$$. If $$\langle B\mathbf{r},\mathbf{r}\rangle < 0$$, the iteration stops immediately. This is the simplest/most general iterative method though and should not have such a constraint.

Also, in SLI and CG (maybe others too, haven't checked), each iteration prints the current residual/error in inner-product form, which is the norm squared. Because stopping tolerance is usually based on the norm (not squared), I think this is misleading and suggest changing to printing $$\|\mathbf{r}\|_B$$ as opposed to $$\langle B\mathbf{r},\mathbf{r}\rangle$$ (or in the case of SLI, either $$\|\mathbf{r}\|$$ or $$\|B\mathbf{r}\|$$. 

I'm happy to make these changes and submit PRs, but want to check w/ MFEM team first.

pazner:
I generally agree, and have also found the squared tolerances to be a source of confusion in the past (e.g. how the PCG wrapper function sets the tolerances to be the square root of the values that are passed in).

It seems reasonable that the SLI solver should measure convergence in something like the 2-norm of the residual vector.

tzanio:
> Also, in SLI and CG (maybe others too, haven't checked), each iteration prints the current residual/error in inner-product form, which is the norm squared. Because stopping tolerance is usually based on the norm (not squared), I think this is misleading and suggest changing to printing $$|\mathbf{r}|_B$$ as opposed to $$\langle B\mathbf{r},\mathbf{r}\rangle$$ (or in the case of SLI, either $$|\mathbf{r}|$$ or $$|B\mathbf{r}|$$.

I generally agree with this, I'm just concerned that users may forget to update all their tolerances.

> The stationary linear iteration solver, i.e., fixed-point iteration, requires an SPD preconditioner because it prints the residual/checks stopping tolerance in the preconditioned norm, $$\langle B\mathbf{r},\mathbf{r}\rangle$$. If $$\langle B\mathbf{r},\mathbf{r}\rangle < 0$$, the iteration stops immediately. This is the simplest/most general iterative method though and should not have such a constraint.

What does *hypre* do? The SPD checks can be useful for debugging, but I agree that SLI should be general.

[...]

PR #1303, opened 2020-02-19
Modified SLI solver to work w/ non-SPD operator and preconditioner

ben-s-southworth:
Stopping tolerance is based on the norm of residual or preconditioned residual, and no longer requires that the preconditioner induces a norm. 

| [#1303](https://github.com/mfem/mfem/pull/1303) | @ben-s-southworth | @tzanio | @delyank + @barker29 | 02/22/20 | 03/10/20 | 03/17/20 | |


delyank:
I don't know if there was a discussion on this issue. If there is one, please point me to it.

I understand the idea behind the PR, but here is my concern.

Consider the SPD case and say that A is the matrix and B is the preconditioner but the very special case when B = A^(-1) exactly. Also, e is the current error and r is the current residual. So Ae = r and Br = e.

Then, the measure (B r, r) = (e, A e). So it is the energy (squared) of the error and it usually is the most natural measure for a problem that aims at minimizing the energy from the get go. Conversely, || Br || = || e ||. That is, this the Euclidean norm of the error. I know this is more general. So the difference is in the norms and they are not precisely equivalent. The difference (when relative errors are considered) is like measuring the error in L2 vs H1 (excluding the sqrt part). For an elliptic problem H1 is the natural choice.

So it makes sense to have both somehow, I guess.

ben-s-southworth:
@delyank There was a brief discussion here: https://github.com/mfem/mfem/issues/1052. I don't disagree that for an elliptic/H1 problem, the A-norm may be the correct norm to measure in. However, I don't think it makes sense for the *default* behavior of the most basic and general linear fixed-point iteration to be designed for/require H1/SPD problems. Having both options makes sense and would not b difficult with a flag or two separate classes, I guess the question is which to default? SPD default would not affect existing code, but still seems non-intuitive to me, and vice-versa for non-SPD default.

tzanio:
Let's postpone this merge until the mfem-4.1 release is complete...

delyank:
Ben, I agree that assuming SPD in the default is not the best. Adding the flag is not the main thing now and Andrew has a good point. Probably, it's very rare to use this class for SPD problems considering that CG is always an option. So let's leave it with your general modification without flags.

tzanio:

tzanio:
Can you please merge in master and document this in CHANGELOG?

PR #1304, opened 2020-02-19
Support for simplices with findpts in MFEM

kmittal2:
This PR enables use of findpts with tri/tet/prism meshes. Since findpts is designed for quad/hex meshes, each simplex is split into quads/hexes. These quad/hex elements that are generated through this process are of the same order as the input mesh.  More sample runs have been added to findpts.cpp and pfindpts.cpp that demonstrate interpolation with findpts for tri, tet, prism, and AMR meshes.

simplex

| [#1304](https://github.com/mfem/mfem/pull/1304) | @kmittal2 | @tzanio | @jakubcerveny + @vladotomov | 02/22/20 | 03/12/20 | 03/22/20 | |


fem/gslib.cpp
```diff
@@ -184,35 +174,270 @@ void FindPointsGSLIB::Interpolate(Array<unsigned int> &codes,
...
+   meshsplit->SetNodalFESpace(&nodal_fes);
+
+   Vector irlist;
+   const int NE = meshsplit->GetNE(),
+             dof_cnt = nodal_fes.GetFE(0)->GetDof(),
+             pts_cnt = NE * dof_cnt;
+   irlist.SetSize(dim * pts_cnt);
+
+   const TensorBasisElement *tbe =
+      dynamic_cast<const TensorBasisElement *>(nodal_fes.GetFE(0));
```
jakubcerveny: Check for NULL

fem/gslib.cpp
```diff
@@ -60,51 +61,40 @@ void FindPointsGSLIB::Setup(Mesh &m, double bb_t, double newt_tol, int npt_max)
...
-   const GridFunction *nodes = mesh->GetNodes();
-   const FiniteElementSpace *fes = nodes->FESpace();
-
-   dim = mesh->Dimension();
-   const int NE = mesh->GetNE(),
-             dof_cnt = fes->GetFE(0)->GetDof(),
-             pts_cnt = NE * dof_cnt;
-   gsl_mesh.SetSize(dim * pts_cnt);
+   dim  = mesh->Dimension();
+   unsigned dof1D = mesh->GetNodalFESpace()->GetFE(0)->GetOrder() + 1;
```
jakubcerveny: Could we factor out `mesh->GetNodalFESpace()->GetFE(0)` (here and below) as `FiniteElement *fe = ...`, please

fem/gslib.cpp
```diff
@@ -184,35 +174,270 @@ void FindPointsGSLIB::Interpolate(Array<unsigned int> &codes,
...
+         }
+         pt_id++;
+      }
+   }
+   NEsplit = 1;
+}
+
+void FindPointsGSLIB::GetSimplexNodalCoordinates() // tri/tet/wedge
+{
+   int gt = mesh->GetNodalFESpace()->GetFE(0)->GetGeomType();
```
jakubcerveny: Factor out `mesh->GetNodalFESpace()->GetFE(0)` (possible elsewhere too).

fem/gslib.hpp
```diff
@@ -28,14 +28,17 @@ class FindPointsGSLIB
 {
 protected:
    Mesh *mesh;
+   IntegrationRule *simir;
```
jakubcerveny: `simir` should be initialized to `NULL` in the constructor.

jakubcerveny:
Aside from this PR, I suggest these improvements:
* It's a little strange in C++ to require the user to call `FindPointsGSLIB::FreeData` explicitly, otherwise a memory leak will occur. The method should be called automatically by the destructor. 
* The method can still be useful if the user wants to free up some memory early (before actual destruction). In that case the method should also free the member `gsl_mesh`.
* Similarly, the `Setup` method should really be the constructor, there seems to be little reason for the object to be able to exist in an "un-setup" state
* It would be really nice if the miniapp could also visualize the result of the interpolation (on a uniform mesh).

kmittal2:
> Aside from this PR, I suggest these improvements:
> 
> * It's a little strange in C++ to require the user to call `FindPointsGSLIB::FreeData` explicitly, otherwise a memory leak will occur. The method should be called automatically by the destructor.
> * The method can still be useful if the user wants to free up some memory early (before actual destruction). In that case the method should also free the member `gsl_mesh`.
> * Similarly, the `Setup` method should really be the constructor, there seems to be little reason for the object to be able to exist in an "un-setup" state
> * It would be really nice if the miniapp could also visualize the result of the interpolation (on a uniform mesh).

Thanks @jakubcerveny . @vladotomov  and I had discussed changes to the interface of the `FindPoints` and `Interpolate` method [here](https://github.com/mfem/mfem/pull/1232#discussion_r386615778). Since we now have FindPoints dependencies in TMOP and LAGHOS, it makes sense to change the interface of FindPoints (not requiring call to Setup etc.) in one single PR. We should decide if we should change the interface in this PR or another PR?

kmittal2:
@jakubcerveny an issue in moving the call to `FreeData()` in the destructor is that `FreeData()` calls `findpts_free_D` which calls `comm_free`. Since `comm_free` has to be called prior to `MPI_Finalize`, we can't really have `FreeData()` as part of the destructor. Do you know of any workarounds to this issue?

jakubcerveny:
I would change the interface in a separate PR. As for the destructor, I wouldn't worry about `MPI_Finalize`. There's several options:

1. Allocate the class dynamically and delete explicitly, just like e.g. the mesh and space in `ex1`
2. Instantiate as a local variable, but enclose in a `{` `}` scope (block) that ends before `MPI_Finalize` is called.
3. Change `pfindpts.cpp` to use the class `MPI_Session` (see `ex18p`). The instances are destroyed in reverse order they are constructed and `MPI_Session` calls `MPI_Finalize` when it's destroyed. This is IMO the cleanest and most "C++" solution.

kmittal2:
> I would change the interface in a separate PR. As for the destructor, I wouldn't worry about `MPI_Finalize`. There's several options:
> 
> 1. Allocate the class dynamically and delete explicitly, just like e.g. the mesh and space in `ex1`
> 2. Instantiate as a local variable, but enclose in a `{` `}` scope (block) that ends before `MPI_Finalize` is called.
> 3. Change `pfindpts.cpp` to use the class `MPI_Session` (see `ex18p`). The instances are destroyed in reverse order they are constructed and `MPI_Session` calls `MPI_Finalize` when it's destroyed. This is IMO the cleanest and most "C++" solution.

Sounds good @jakubcerveny . Since this might affect our other dependencies, I will open a new PR for effecting the interface changes along with the constructor and destructor.

tzanio:
Adding a bit more time to review...

fem/gslib.hpp
```diff
@@ -28,14 +28,17 @@ class FindPointsGSLIB
...
    Vector gsl_mesh;
    struct findpts_data_2 *fdata2D;
    struct findpts_data_3 *fdata3D;
-   int dim;
+   int dim, NEsplit;
```
vladotomov: `NEsplit` can now be made local variable in `GetSimplexNodalCoordinates`.

fem/gslib.hpp
vladotomov: I don't like the name, how about `ir_simplex`?

vladotomov:
Very cool addition Ketan, great job. Went over the code and tested it, all looks good.
- [ ] mention in CHANGELOG.
- [ ] make the `NEsplit` local to the `GetSimplexNodalCoordinates`.
- [ ] go over styling, maybe rename `simir`.

tzanio:

fem/gslib.cpp
```diff
@@ -184,35 +181,273 @@ void FindPointsGSLIB::Interpolate(Array<unsigned int> &codes,
...
+{
+   const FiniteElement *fe   = mesh->GetNodalFESpace()->GetFE(0);
+   const Geometry::Type gt   = fe->GetGeomType();
+   const GridFunction *nodes = mesh->GetNodes();
+   Mesh *meshsplit           = NULL;
+   const int NE              = mesh->GetNE();
+   int NEsplit;
+
+   // Split the reference element into a reference submesh of quads or hexes.
+   if (gt == Geometry::TRIANGLE)
```
tzanio: I wonder if this could be of use in other parts of the code? We don't have to change it now, but we may consider moving it into an auxiliary Mesh function if there are other uses...

tzanio:
LGTM

PR #1353, opened 2020-03-13
Partial assembly for H(div)

dylan-copeland:
PA is implemented for Raviart-Thomas (RT) elements in H(div). Examples 4 and 5 have been extended to use PA with Jacobi preconditioners. For example 5, this involves assembling the diagonal of BDB^T, where D is the diagonal of M. This capability has been added to MixedBilinearForm. 

Also a new test problem is added to ex24, demonstrating VectorFEDivergenceIntegrator with PA.

There are unit tests for the RT mass and div-div bilinear forms, as well as the mixed H(div)-L_2 VectorFEDivergenceIntegrator.

| [#1353](https://github.com/mfem/mfem/pull/1353) | @dylan-copeland | @v-dobrev | @YohannDudouit + @psocratis | 03/18/20 | 04/24/20 | 04/30/20 | |


tzanio:
Please resolve the above merge conflicts

tzanio:
@YohannDudouit and @psocratis -- reminder that the reviews for this PR are due this week.

dylan-copeland:
> Please resolve the above merge conflicts

@tzanio What does this mean? Is there is an automatic check that merges master into the	branch and tries to build? I don't see any report saying there is a failed build. I only see that AppVeyor fails, but there are no details (I cannot click on AppVeyor and get a report). Is this	something I can find internally in GitLab? If so, are there any instructions on how to view GitLab reports?

tzanio:
It means that `master` has changed and there are conflicts between these changes and the changes in the PR (so the PR cannot be merged without resolving them). It also means that Travis will not run, because it runs the tests on the merge state, which now cannot be determined.

GitHub catches and automatically reports this at the bottom, see the "This branch has conflicts that must be resolved
Use the web editor or the  to resolve conflicts." section.

You can try using the GitHub interface to resolve the conflicts (the "Resolve conflicts" button) but I recommend that you just pull master, merge it in your branch, resolve the conflicts and commit the changes here. Let me know if you need help with that.

Tzanio

dylan-copeland:
@psocratis asked a good question about `ElementRestriction::MultUnsigned`. This function is similar to `MultTransposeUnsigned` (it seems I copied and pasted it), but why is it not more similar to `Mult`? The main difference is that `Mult` uses `gatherMap`, whereas `MultUnsigned` and `MultTransposeUnsigned` use `offsets` and `indices`. Another difference is the loop over `dof*ne` vs. `ndofs`. 

I suppose the purpose of `gatherMap` is to use one array in the `MFEM_FORALL` rather than two (`offsets` and `indices`). @YohannDudouit @v-dobrev Is this correct, and are there any significant differences or reasons to make `MultUnsigned` more like `Mult`? Do you recommend changing `MultUnsigned` to use `gatherMap`?

fem/bilininteg_hdiv.cpp
```diff
@@ -0,0 +1,1995 @@
...
+         const double J12 = J(q,0,1,e);
+         const double J22 = J(q,1,1,e);
+         const double c_detJ = W[q] * coeff(q, e) / ((J11*J22)-(J21*J12));
+         /*
+              // (c/detJ) J
+              y(q,0,e) = c_detJ * J11; // 1,1
+              y(q,1,e) = c_detJ * J21; // 2,1
+              y(q,2,e) = c_detJ * J12; // 1,2
+              y(q,3,e) = c_detJ * J22; // 2,2
+         */
```
psocratis, line 56: Is this commented piece of code needed here?

dylan-copeland: Removed.

camierjs:
> I suppose the purpose of `gatherMap` is to use one array in the `MFEM_FORALL` rather than two (`offsets` and `indices`). @YohannDudouit @v-dobrev Is this correct, and are there any other significant differences or reasons to make `MultUnsigned` more like `Mult`? Do you recommend changing `MultUnsigned` to use `gatherMap`?

Using `gatherMap` was one optimization introduced with the #1208 , I think we should try to propagate this in the `MultUnsigned`, yes.

fem/bilininteg_hdiv.cpp
```diff
@@ -0,0 +1,1987 @@
...
+         const double J31 = J(q,2,0,e);
+         const double J12 = J(q,0,1,e);
+         const double J22 = J(q,1,1,e);
+         const double J32 = J(q,2,1,e);
+         const double J13 = J(q,0,2,e);
+         const double J23 = J(q,1,2,e);
+         const double J33 = J(q,2,2,e);
+         const double detJ = J11 * (J22 * J33 - J32 * J23) -
+         /* */               J21 * (J12 * J33 - J32 * J13) +
+         /* */               J31 * (J12 * J23 - J22 * J13);
```
psocratis, line 87: comment leftovers?

fem/bilininteg_hcurl.cpp
```diff
@@ -419,11 +431,39 @@ static void PAHcurlMassAssembleDiagonal3D(const int D1D,
...
+                                       mapsO->B, mapsC->B, pa_data, diag);
+      }
+      else if (fetype == mfem::FiniteElement::DIV)
+      {
+         PAHdivMassAssembleDiagonal2D(dofs1D, quad1D, ne,
+                                      mapsO->B, mapsC->B, pa_data, diag);
+      }
+      else
+      {
+         MFEM_ABORT("Unknown kernel.");
```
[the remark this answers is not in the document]
dylan-copeland: It would make sense to have a file for VectorFE, but so far there would be a small amount of code in it. I am fine with doing it either way, but would think it is fine as is for now. @YohannDudouit @v-dobrev Does anyone else have a preference?

fem/bilininteg_hdiv.cpp
dylan-copeland: This was copied and pasted from `VectorMassIntegrator::AssemblePA` which still has these empty comments. It is also in bilininteg_hcurl.cpp, and I will remove it there too for this PR. @jandrej might want to remove them from bilininteg_vecmass.cpp at some point, or I can do it in this PR if desired.

dylan-copeland:
> Using `gatherMap` was one optimization introduced with the #1208 , I think we should try to propagate this in the `MultUnsigned`, yes.

@camierjs In the master branch, `gatherMap` is used in `Mult` but not `MultTranspose`. It appears to me that is because `MultTranspose` needs to use `indices` and `offsets`, so this optimization cannot be done there. I will use `gatherMap` in `MultUnsigned` but not `MultTransposeUnsigned`.

examples/ex4.cpp
```diff
@@ -167,32 +186,41 @@ int main(int argc, char *argv[])
...
+   }
+   else
+   {
+      GSSmoother M((SparseMatrix&)(*A));
+      PCG(*A, M, B, X, 1, 10000, 1e-20, 0.0);
+   }
 #else
-   // 10. If compiled with SuiteSparse support, use UMFPACK to solve the system.
+   // 11. If compiled with SuiteSparse support, use UMFPACK to solve the system.
    UMFPackSolver umf_solver;
```
psocratis, line 211: This checks should be changed like example 1, i.e if pa is enabled it shouldn't allow the direct solve and run by default the CG solver

fem/bilininteg_hdiv.cpp
```diff
@@ -0,0 +1,1987 @@
...
+static void PADivDivAssembleDiagonal3D(const int D1D,
+                                       const int Q1D,
+                                       const int NE,
+                                       const Array<double> &_Bo,
+                                       const Array<double> &_Gc,
+                                       const Vector &_op,
+                                       Vector &_diag)
+{
+   MFEM_VERIFY(D1D <= HDIV_MAX_D1D, "Error: D1D > HDIV_MAX_D1D");
+   MFEM_VERIFY(Q1D <= HDIV_MAX_Q1D, "Error: Q1D > HDIV_MAX_Q1D");
```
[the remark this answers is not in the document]
dylan-copeland: Good question. In the H(curl) case, we have this check only for 3D Apply functions, not Assemble functions or any 2D functions. My understanding is that this check prevents using too much memory on the GPU, so the 3D case is more restrictive. What I don't understand is that the array sizes depend on the number of elements as well as Q1D and D1D, so is it only the number of DOF's or quadrature points per element that matters, so that one element at a time can be handled? This is a question for @camierjs or @YohannDudouit. Also, should there be a 2D max Q1D and D1D? Should the checks be done also in setup functions or just apply functions?

examples/ex4.cpp
dylan-copeland: Thanks, I overlooked that case since I did not test with SUITESPARSE.

fem/bilininteg_hdiv.cpp
vladotomov: I think these are used to trick the `make style`, which would otherwise move the code to the lest.

dylan-copeland: Thanks @vladotomov, I didn't realize that. I just tried `make style` and confirmed that it does move the code to the left, which makes it less readable. I will keep the empty comments.

examples/ex24.cpp
```diff
@@ -23,11 +24,15 @@
...
 //
 // Description:  This example code illustrates usage of mixed finite element
-//               spaces. Using two different approaches, we project a gradient
-//               of a function in H^1 to H(curl). Other spaces and example
-//               computations are to be added in the future.
+//               spaces, with two variants:
 //
-//               We recommend viewing examples 1 and 3 before viewing this
+//               1) (u, grad p) for u in H(curl) and p in H^1
+//               2) (q, div v) for v in H(div) and q in L_2
```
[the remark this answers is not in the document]
dylan-copeland: Fixed.

fem/restriction.cpp
```diff
@@ -168,6 +168,29 @@ void ElementRestriction::Mult(const Vector& x, Vector& y) const
...
 }
 
+void ElementRestriction::MultUnsigned(const Vector& x, Vector& y) const
+{
+   // Assumes all elements have the same number of dofs
+   const int nd = dof;
+   const int vd = vdim;
+   const bool t = byvdim;
+   auto d_offsets = offsets.Read();
+   auto d_indices = indices.Read();
```
psocratis, line 178: These 2 variables are not used anymore

examples/ex4p.cpp
```diff
@@ -269,7 +290,7 @@ int main(int argc, char *argv[])
       sol_sock << "solution\n" << *pmesh << x << flush;
    }
 
-   // 17. Free the used memory.
+   // 18. Free the used memory.
    delete pcg;
    delete prec;
    delete hfes;
```
psocratis, line 296: There are some unnecessary comments in ex4 lines: 272, 289 and ex4p lines: 319, 336 that I think they should be removed. There are not introduced in this PR though

dylan-copeland: Right, these are already in master, not from this PR. @v-dobrev Should we keep these commented definitions of z in ex4 and ex4p?

fem/restriction.cpp
dylan-copeland: Good catch.

fem/bilininteg_hcurl.cpp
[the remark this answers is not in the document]
v-dobrev: Ultimately, as we add more integrators that work with both H(div) and H(curl), we will need to create a new file probably. However, I don't think we need to do it right now.

fem/bilininteg_hdiv.cpp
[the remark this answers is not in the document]
v-dobrev: In this particular kernel, only `HDIV_MAX_Q1D` seems to be used. It is used just for the size of a 1D array, so this is not that important here from the point of view of memory usage. The check just makes sure that we do not try to use array entities beyond what is allocated, see the array `double a[HDIV_MAX_Q1D];` below.

Looking at the code below, it caught my attention that when being set `a` is accessed with `a[qx]` with `0<=qx<Q1D` but then it is read using `a[dx]` with `0<=dx<D1Dx`. This seems a little suspicious.

examples/ex4p.cpp
v-dobrev: I think these are left there for the reader (users) in case they want to try another solution that depends on `z`. Do you find them more confusing than helpful? 😄

dylan-copeland: @v-dobrev Thanks for clarifying. This may be the first time I've seen commented code in an example. It is helpful and a little confusing, but it might be more helpful it were less confusing. Maybe we should add a comment saying something like "uncomment if f depends on z". I am fine with leaving these lines as they are, but a comment might be a good idea.

v-dobrev: I agree, adding a comment as you suggest will be an improvement.

dylan-copeland: Comments are added.

fem/bilininteg_hdiv.cpp
```diff
@@ -0,0 +1,2007 @@
...
+               }
+            }
+
+            for (int dx = 0; dx < D1Dx; ++dx)
+            {
+               for (int qx = 0; qx < Q1D; ++qx)
+               {
+                  const double wx = (c == 0) ? Gc(qx,dx) : Bo(qx,dx);
+                  diag(dx + (dy * D1Dx) + osc, e) += div[qx] * wx * wx;
+               }
```
YohannDudouit, line 1007:
```suggestion
               double val = 0.0;
               for (int qx = 0; qx < Q1D; ++qx)
               {
                  const double wx = (c == 0) ? Gc(qx,dx) : Bo(qx,dx);
                  val += div[qx] * wx * wx;
               }
               diag(dx + (dy * D1Dx) + osc, e) += val;
```

fem/bilininteg_hdiv.cpp
```diff
@@ -0,0 +1,2007 @@
...
+
+               for (int dx = 0; dx < D1Dx; ++dx)
+               {
+                  for (int qx = 0; qx < Q1D; ++qx)
+                  {
+                     const double wx = (c == 0) ? Gc(qx,dx) : Bo(qx,dx);
+                     diag(dx + ((dy + (dz * D1Dy)) * D1Dx) + osc, e) +=
+                        a[dx] * wx * wx;
+                  }
```
YohannDudouit, line 1071:
```suggestion
                  double val = 0.0;
                  for (int qx = 0; qx < Q1D; ++qx)
                  {
                     const double wx = (c == 0) ? Gc(qx,dx) : Bo(qx,dx);
                     val += a[dx] * wx * wx;
                  }
                  diag(dx + ((dy + (dz * D1Dy)) * D1Dx) + osc, e) += val;
```

fem/bilininteg_hdiv.cpp
```diff
@@ -0,0 +1,2007 @@
...
+                     }
+
+                     osc += D1Dx * D1Dy * D1Dz;
+                  }  // loop c
+               }  // loop qz
+
+               for (int i=0; i<3*D1D*(D1D - 1)*(D1D - 1); ++i)
+               {
+                  diag(rx,ry,rz,e) += row[i] * row[i] * D(i,e);
+               }
```
YohannDudouit, line 1895:
```suggestion
               double val = 0.0;
               for (int i=0; i<3*D1D*(D1D - 1)*(D1D - 1); ++i)
               {
                  val += row[i] * row[i] * D(i,e);
               }
               diag(rx,ry,rz,e) += val;
```

fem/bilininteg_hdiv.cpp
```diff
@@ -0,0 +1,2007 @@
...
+                  }
+
+                  osc += D1Dx * D1Dy;
+               }  // loop c
+            }  // loop qy
+
+            for (int i=0; i<2*D1D*(D1D - 1); ++i)
+            {
+               diag(rx,ry,e) += row[i] * row[i] * D(i,e);
+            }
```
YohannDudouit, line 1986:
```suggestion
            double val = 0.0;
            for (int i=0; i<2*D1D*(D1D - 1); ++i)
            {
               val += row[i] * row[i] * D(i,e);
            }
            diag(rx,ry,e) += val;
```

tests/unit/fem/test_pa_coeff.cpp
```diff
@@ -174,141 +174,168 @@ TEST_CASE("Hcurl pa_coeff")
...
+                  {
+                     if (spaceType == 0)
+                     {
+                        assemblyform.AddDomainIntegrator(new CurlCurlIntegrator(*coeff2));
+                     }
+                     else
+                     {
+                        assemblyform.AddDomainIntegrator(new DivDivIntegrator(*coeff2));
+                     }
+                  }
```
YohannDudouit, line 287: This can be factorized with the code above.

tzanio:
Should we keep e.g. Example 5 simple and have the PA version in a separate ex5-pa.cpp example?

dylan-copeland:
There are some comments from @psocratis and @YohannDudouit about organizing the PA code for VectorFEMassIntegrator, and I agree that it is currently not organized perfectly, but this is tricky. Here are some options.

1) (Current implementation) All of the PA kernels for H(curl) mass are in bilininteg_hcurl.cpp and for H(div) in bilininteg_hdiv.cpp. VectorFEMassIntegrator is in bilininteg_hcurl.cpp and calls the kernels for H(curl) or H(div) depending on the space type. This requires having declarations for H(div) kernels in bilininteg.hpp, which is undesirable as pointed out by @YohannDudouit.

2) Move VectorFEMassIntegrator to bilininteg_vecmass.cpp (existing file for vector H^1) or a new file, e.g. bilininteg_vecfe.cpp. Calling kernels in bilininteg_hcurl.cpp and bilininteg_hdiv.cpp would require putting even more declarations in bilininteg.hpp.

2b) Same as (2) but put the required declarations in a new header, bilininteg_vecfe.hpp.

3) Move VectorFEMassIntegrator to a new file, e.g. bilininteg_vecfe.cpp, and move all the mass kernels for H(curl) and H(div) to this file. Then there would be no declarations in bilininteg.hpp, and it would make more sense not to have VectorFEMassIntegrator in bilininteg_hcurl.cpp, but we would have H(curl) and H(div) mass kernels in this file rather than separating them into bilininteg_hcurl.cpp and bilininteg_hdiv.cpp.

It seems to me that (3) best addresses the concerns pointed out in the reviews, but it is also flawed. Does anyone have a preference or a better suggestion? I don't think any of these options is ideal, so I would keep the current implementation (1) by the law of inertia, but I am open to suggestions.

YohannDudouit:
When doing `make test` the `Parallel example [ mpirun -np 4 ex24p ... ]:` test doesn't conclude on my mac, does it work for you @dylan-copeland  and @psocratis ?

psocratis:
Hmm I ran make test on my mac the other day with no issues. I’ll check again and let you know a bit later

examples/ex4.cpp
```diff
@@ -167,32 +186,41 @@ int main(int argc, char *argv[])
...
+   if (pa)
+   {
+      OperatorJacobiSmoother M(*a, ess_tdof_list);
+      PCG(*A, M, B, X, 1, 10000, 1e-20, 0.0);
+   }
+   else
+   {
+      GSSmoother M((SparseMatrix&)(*A));
+      PCG(*A, M, B, X, 1, 10000, 1e-20, 0.0);
+   }
```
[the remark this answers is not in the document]
dylan-copeland: I don't know why, but the code displayed here on GitHub is different from the current version of the code. In the real code, there is PCG in one case and CG in the other case, so I don't this comment still applies.

fem/bilinearform.hpp
```diff
@@ -705,6 +705,8 @@ class MixedBilinearForm : public Matrix
 
    void Assemble(int skip_zeros = 1);
 
+   void AssembleDiagonal_ADAt(const Vector &D, Vector &diag) const;
```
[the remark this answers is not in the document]
dylan-copeland: @YohannDudouit I added a comment describing what the function does, but AssembleDiagonal has an additional remark about AMR meshes which I do not understand and do not know whether it applies to this function as well.

YohannDudouit:
I have these warning when compiling with `nvcc`:
```
linalg/../fem/bilininteg.hpp(2101): warning: overloaded virtual function "mfem::NonlinearFormIntegrator::AssemblePA" is only partially overridden in class "mfem::VectorFEDivergenceIntegrator"

linalg/../fem/bilininteg.hpp(2393): warning: overloaded virtual function "mfem::NonlinearFormIntegrator::AssemblePA" is only partially overridden in class "mfem::DivDivIntegrator"
```

dylan-copeland:
@YohannDudouit I noticed a couple of things that might explain the compiler warning. First, maybe those classes need `using BilinearFormIntegrator::AssemblePA`. Second, only a couple of classes have `AssemblePA` as a protected function. I don't know why that would be, and it seems it should always be a public function. Please let me know if one or both of these changes fixes the compiler warning you get. If so, I will commit them.

YohannDudouit:
I have issues with device samples runs for `ex5` and `ex5p`:
```
[FAILED] ex5 -m ../data/star.mesh -pa -d cuda -no-vis
[FAILED] ex5 -m ../data/beam-hex.mesh -pa -d cuda -no-vis
```
```
[FAILED] mpirun -np 4 ex5p -m ../data/star.mesh -pa -d cuda -no-vis
[FAILED] mpirun -np 4 ex5p -m ../data/beam-hex.mesh -pa -d cuda -no-vis
```
The error is the same for the 4 failed tests:
```
Options used:
   --mesh ../data/star.mesh
   --order 1
   --partial-assembly
   --device cuda
   --no-visualization
***********************************************************
dim(R) = 41280
dim(W) = 20480
dim(R+W) = 61760
***********************************************************


alias not found
Aborted (core dumped)
```

fem/bilininteg_hdiv.cpp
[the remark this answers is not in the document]
dylan-copeland: @v-dobrev Thanks for finding this. It seems to be a bug, i.e. `a[dx]` should be `a[qx]`. The diagonal preconditioner still works (doesn't have to be perfect), so I didn't catch this. I should add unit tests for the diagonal assembly.

dylan-copeland:
@psocratis were you able to run the tests on your Mac? The only command that works on my Mac is ssh, so I can only test in on linux.

psocratis:
> @psocratis were you able to run the tests on your Mac? The only command that works on my Mac is ssh, so I can only test on linux.

Yes I did it yesterday. All the tests passed on my Mac

dylan-copeland:
The `nvcc` compiler warnings are fixed now, and I removed device support from example 5. There seem to be issues with `BlockOperator` on GPU that should be addressed by another PR.

v-dobrev:
@vladotomov, the request for review from you was unintentional -- it was either a bug in the mfem-review-app or an error by me.

v-dobrev:
@YohannDudouit, do you have any additional comment or requests for changes? If not, can you approve the PR?

@dylan-copeland, I have one request: can you add a bullet (or more than one bullet, if you want) in the CHANGELOG about the new things in this PR?

fem/bilininteg.hpp
```diff
@@ -57,6 +57,9 @@ class BilinearFormIntegrator : public NonlinearFormIntegrator
    /// Assemble diagonal and add it to Vector @a diag.
    virtual void AssembleDiagonalPA(Vector &diag);
 
+   /// Assemble diagonal of ADA^T and add it to Vector @a diag.
+   virtual void AssembleDiagonalPA_ADAt(const Vector &D, Vector &diag);
```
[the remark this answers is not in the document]
v-dobrev: I don't understand from the doxygen comment what this method does, specifically it is not clear what `A` is. Where is this method needed?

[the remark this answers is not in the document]
dylan-copeland: `A` is this integrator. It is used in example 5/5p.

fem/bilininteg.hpp
```diff
@@ -2889,6 +2918,62 @@ class VectorInnerProductInterpolator : public DiscreteInterpolator
...
+                                  const Array<double> &_Bc,
+                                  const Vector &_op,
+                                  Vector &_diag);
+
+void PAHdivMassAssembleDiagonal2D(const int D1D,
+                                  const int Q1D,
+                                  const int NE,
+                                  const Array<double> &_Bo,
+                                  const Array<double> &_Bc,
+                                  const Vector &_op,
```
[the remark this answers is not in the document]
v-dobrev: It looks like these functions are here because they are defined in one `*.cpp` file and then used in another `*.cpp` file. I think we can move these declarations to the file where they are used -- they do not need to be visible by when including `mfem.hpp`.

The `constexpr` constants `HDIV_MAX_D1D` and `HDIV_MAX_Q1D` can be moved to the implementation file -- they are probably used just there.

dylan-copeland:
@v-dobrev I updated the CHANGELOG. As far as I know, the only thing remaining is the issue of organizing the code into the most appropriate files. I made a comment on April 7th that lists some options. If anyone has a preference, we can make that change.

fem/bilininteg.hpp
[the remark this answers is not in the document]
v-dobrev: I looked closer at how this method and the method that uses it (`MixedBilinearForm::AssembleDiagonal_ADAt`) are, in turn, used in example 5/5p. The idea is to assemble the diagonal of the approximate Schur complement, `S = B diag(M)^{-1} B^t`. This diagonal is then used as a preconditioner for S. That by itself is not a great preconditioner, so the question is: why do you think this is useful? Do you see this as a necessary component for building a (p- and/or h-) multigrid for S using the diagonals for smoothing?

I think we want to make sure that this new functionality will be ultimately useful, similar to the way `BilinearFormIntegrator::AssembleDiagonalPA` and `BilinearForm::AssembleDiagonal` are useful in example 1/1p (using the diagonals for smoothing as in the multigrid PR) or just for preconditioning high-order mass matrices.

v-dobrev:
> @v-dobrev I updated the CHANGELOG. As far as I know, the only thing remaining is the issue of organizing the code into the most appropriate files. I made a comment on April 7th that lists some options. If anyone has a preference, we can make that change.

Thanks, @dylan-copeland!

I left some comments regarding the declarations in the headers here: https://github.com/mfem/mfem/pull/1353#discussion_r409273589

Regarding moving implementations from `bilininteg_hcurl.cpp` and `bilininteg_hdiv.cpp` to new files: I think that will be a better organization. So far, (except in these two files) we've tried to separate the new PA extensions of different integrators in different files. From this perspective, it makes sense to me to introduce `bilininteg_vec_fe_mass.cpp` (or something like that). Since this integrator essentially supports two different integrators, I think it also makes sense to split the kernels into separate files `bilininteg_hcurl_mass.cpp` and `bilininteg_hdiv_mass.cpp` while keeping the common code in `bilininteg_vec_fe_mass.cpp`. This common code will dispatch to the hcurl/hdiv functions. For example, `VectorFEMassIntegrator` will have a field `fespace_type` (0-hcurl, 1-hdiv) that is set during assembly in `VectorFEMassIntegrator::AssemblePA`; after that the `AssemblePA` will simply dispatch to the protected methods `AssemblePA_Hcurl` or `AssemblePA_Hdiv` which are defined in `bilininteg_hcurl_mass.cpp` and `bilininteg_hdiv_mass.cpp`, respectively. Similar for the other methods: `AddMultPA` and `AssembleDiagonalPA`. Another advantage of this is that the two cases can be compiled in parallel.

fem/bilinearform.hpp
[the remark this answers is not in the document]
YohannDudouit: I'm really not convinced that this method should be part or the interface of a mixed bilinear form, @v-dobrev shouldn't we make this a function instead of a method?

dylan-copeland:
I think the last commit rearranges the code as discussed, with some small differences. One is that I made a new file bilininteg_vectorfe.cpp to contain all the VectorFE functions involving PA with the option of H(curl) or H(div), including MixedVectorGradientIntegrator. That integrator currently supports only H(curl)-H^1, but in the future it could support H(div)-H^1. This new cpp file is small enough that I don't see a need to separate mass and other integrators into separate files (bilininteg.cpp has many integrators in one file).

YohannDudouit:
@dylan-copeland Fix the last appveyor errors and I'll approve :) I can help if you have any trouble.

dylan-copeland:
@YohannDudouit Before the last commit, AppVeyor had a link to a build report where I found errors. Now that I made a commit that might fix one error, it just says it was unable to build, with no report. How can I find the errors now?

YohannDudouit:
@dylan-copeland You need to fix the merge conflicts I think.

v-dobrev:
@dylan-copeland, do you have an answer for https://github.com/mfem/mfem/pull/1353#discussion_r409359674?

@YohannDudouit, @psocratis, is this ready for testing?

v-dobrev:
LGTM

fem/bilininteg.hpp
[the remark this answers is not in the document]
dylan-copeland: @v-dobrev Sorry, I never saw this comment (GitHub doesn't show notifications for each comment, apparently). I agree this diagonal preconditioner is not very effective, but that is the best we have so far. In this example, `S` is an approximation of the Laplacian, similar to example 1. In examples 1 and 3 we use Jacobi preconditioning in the PA case, so the same question could be asked about those examples. One answer, as you mentioned, is the possibility of using this as a smoother in multigrid in the future. Of course, development of better PA preconditioners is a long-term goal, and this is the first step, and I hope this diagonal assembly will be useful later in something like multigrid. For now, it shows how to use Jacobi for a Schur complement, which seems better than no preconditioner.

psocratis:
> @dylan-copeland, do you have an answer for [#1353 (comment)](https://github.com/mfem/mfem/pull/1353#discussion_r409359674)?
> 
> @YohannDudouit, @psocratis, is this ready for testing?

Yes I think it's ready

fem/bilininteg.hpp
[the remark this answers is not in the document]
v-dobrev: OK, thanks @dylan-copeland!

v-dobrev:
I checked the gitlab ci results -- there are baselin differences but most of them are due to other (merged) PRs for which the baseline has not been updated yet.

v-dobrev:

v-dobrev:
Sorry for the new review request, @vladotomov, please igonore it.

v-dobrev:
Two issue uncovered by the internal tests:
* valgrind reports leaks in `main (ex5.cpp:179)` and `main (ex5p.cpp:238)`
* the new sample run `mpirun -np 4 ex5p -m ../data/star.mesh -pa -no-vis` does not converge (not even close) in 1000 iterations; this is not a bug, just needs more iterations.

@dylan-copeland, can you please address these?

dylan-copeland:
@v-dobrev Thanks for finding these issues. I fixed the memory leaks, but I am suspicious of the large number of iterations for ex5p with star.mesh. It may actually be a bug, considering that ex5 converges much faster than ex5p for both star.mesh and beam-hex.mesh. I wonder whether this is a result of recent commits, because I believe I checked the new sample runs at some point and got good results. I will investigate this further on Tuesday.

dylan-copeland:
@v-dobrev There is a very simple explanation for why ex5p was taking so many iterations. It does more mesh refinement than ex5. If the serial mesh refinement is reduced to get the same fine mesh as ex5, then the results are identical for ex5 and ex5p (890 iterations for star.mesh -pa). This difference in problem sizes existed before this PR and makes sense, as the parallel version can solve a larger problem. It raises a question for the PA sample runs, however, since ex5p does not have a good preconditioner and takes too many iterations for large problems. Do we want to  

1. remove the star.mesh -pa run from ex5p;
2. allow many iterations for PA (not sure how many, 3000 was not enough);
3. introduce a serial refinement input parameter for ex5p so that a sample run can be given with the same fine mesh size as ex5 (so that convergence takes 890 iterations for star.mesh)?

I think 3 would be the best option, if you don't mind complicating ex5p a bit more. Option 2 would make the sample runs too slow, so I don't like that. Option 1 would be the simplest, and would still leave the beam-hex.mesh -pa sample run, which is sufficient for demonstrating PA. I am fine with 1 or 3. What do you think?

dylan-copeland:
@v-dobrev The ex5p sample run converges now with a smaller mesh.

v-dobrev:
Re-merged in `next`.

Issue #1348, opened 2020-03-11
4.1: error: cannot initialize a parameter of type 'const char *' with an lvalue of type 'int'


yurivict:
clang-8 complains:

```
[30/84] /usr/bin/c++  -DMFEM_CONFIG_FILE=\"/wrkdirs/usr/ports/math/mfem/work/.build/config/_config.hpp\" -Dmfem_EXPORTS  -O2 -pipe -fstack-protector-strong -fno-strict-aliasing -O2 -pipe -fstack-protector-strong -fno-strict-aliasing -fPIC   -std=c++11 -MD -MT CMakeFiles/mfem.dir/mesh/element.cpp.o -MF CMakeFiles/mfem.dir/mesh/element.cpp.o.d -o CMakeFiles/mfem.dir/mesh/element.cpp.o -c /wrkdirs/usr/ports/math/mfem/work/mfem-4.1/mesh/element.cpp
FAILED: CMakeFiles/mfem.dir/mesh/element.cpp.o 
/usr/bin/c++  -DMFEM_CONFIG_FILE=\"/wrkdirs/usr/ports/math/mfem/work/.build/config/_config.hpp\" -Dmfem_EXPORTS  -O2 -pipe -fstack-protector-strong -fno-strict-aliasing -O2 -pipe -fstack-protector-strong -fno-strict-aliasing -fPIC   -std=c++11 -MD -MT CMakeFiles/mfem.dir/mesh/element.cpp.o -MF CMakeFiles/mfem.dir/mesh/element.cpp.o.d -o CMakeFiles/mfem.dir/mesh/element.cpp.o -c /wrkdirs/usr/ports/math/mfem/work/mfem-4.1/mesh/element.cpp
In file included from /wrkdirs/usr/ports/math/mfem/work/mfem-4.1/mesh/element.cpp:12:
In file included from /wrkdirs/usr/ports/math/mfem/work/mfem-4.1/mesh/mesh_headers.hpp:26:
In file included from /wrkdirs/usr/ports/math/mfem/work/mfem-4.1/mesh/mesh.hpp:25:
/wrkdirs/usr/ports/math/mfem/work/mfem-4.1/mesh/../general/zstr.hpp:81:35: error: cannot initialize a parameter of type 'const char *' with an lvalue of type 'int'
   std::string tmp(p, std::strlen(p));
                                  ^
/usr/include/string.h:97:28: note: passing argument to parameter here
size_t   strlen(const char *) __pure;
                            ^
[...]

PR #1360, opened 2020-03-17
Hotfix for sample runs

YohannDudouit:
This PR fixes `ex6` and `ex6p` to use deterministic backends of libCEED in the sample runs, and fixes `nurbs_ex11p` to print messages only once.

| [#1360](https://github.com/mfem/mfem/pull/1360) | @YohannDudouit | @tzanio | @tzanio + @camierjs | 03/18/20 | 03/18/20 | 03/19/20 | |

### DIFF
--- a/examples/ex6.cpp
+++ b/examples/ex6.cpp
@@ -20,7 +20,7 @@
 //               ex6 -pa -d occa-cuda
 //               ex6 -pa -d raja-omp
 //               ex6 -pa -d ceed-cpu
-//               *ex6 -pa -d ceed-cuda
+//               * ex6 -pa -d ceed-cuda
 //               ex6 -pa -d ceed-cuda:/gpu/cuda/ref
 //
 // Description:  This is a version of Example 1 with a simple adaptive mesh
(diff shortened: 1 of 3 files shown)


examples/ex6.cpp
```diff
@@ -20,7 +20,8 @@
 //               ex6 -pa -d occa-cuda
 //               ex6 -pa -d raja-omp
 //               ex6 -pa -d ceed-cpu
-//               ex6 -pa -d ceed-cuda
+//               * ex6 -pa -d ceed-cuda
+//               ex6 -pa -d ceed-cuda:/gpu/cuda/ref
```
tzanio: Is this the best backend to use, what about `/gpu/cuda/shared`?

YohannDudouit: Good point, thank you.

jandrej:
I hijacked the PR and added the hotfix for #1348.

camierjs:
Shouldn't we switch examples `1` & `1p` to use the `shared` CEED GPU backend also?
Even if we don't see differences now with the tests, we might if we add some new features.

tzanio:
> Shouldn't we switch examples `1` & `1p` to use the `shared` CEED GPU backend also?
> Even if we don't see differences now with the tests, we might if we add some new features.

I am OK either way,  but it may be better to wait until we start seeing the differences (in particular we will know what change exposes them)

tzanio:

camierjs:
Works fine on `pascal` too, thanks @YohannDudouit for the fix!

PR #1370, opened 2020-03-23
PWConstCoefficient documentation fix [pw-const-doc-fix-dev]

### LINKED ISSUE #1363: Trouble with Parameters and Boundary Conditions
I am new to FEM, and am integrating MFEM into software

**Currently:**
- I have .vtk UnstructuredGrid files that I read into a MFEM::Mesh, and then run various FEM simulations on them, converting back to VTK format when done (saving the grid function solution)
- I am using MFEM Example16 as a base to design a Thermal Conductivity FEM simulation
https://github.com/mfem/mfem/blob/master/examples/ex16.cpp

**Questions and Concerns:**
- How would I read in materials from my VTK Scalar Fields and assign them to parameter values at that point in the mesh?
- The process of setting up boundary conditions is very confusing to me, and I don't understand how to set them up. 
- If I was able to define the geometry and values of a boundary condition in my VTK file using scalar fields, would I then be able to use them in MFEM?

mlstowell:
This addresses an issue that was raised in issue #1363.

Resolves #1363.

| [#1370](https://github.com/mfem/mfem/pull/1370) | @mlstowell | @v-dobrev | @v-dobrev + @tzanio | 03/23/20 | 03/24/20 | 03/25/20 | |

### DIFF
--- a/fem/coefficient.hpp
+++ b/fem/coefficient.hpp
@@ -91,7 +91,7 @@ public:
    { constants = 0.0; }
 
    /** c should be a vector defined by attributes, so for region with
-       attribute i  c[i] is the coefficient in that region */
+       attribute i  c[i-1] is the coefficient in that region */
    PWConstCoefficient(Vector &c)
    { constants.SetSize(c.Size()); constants=c; }
 


tzanio:

Issue #1037, opened 2019-08-16
L2 finite element maptype errors & inconsistencies


Dan2997925:
I am interested in solving electromagnetics problems in 2D, with E field in H(curl) and B field in L2. The CurlInterpolator requires that the L2 space use MapType==INTEGRAL. But other parts of MFEM e.g. GridFunction::GetValue require  that the L2 space use MapType==VALUE. And some functions such as VectorFECurlIntegrator do not check the map type at all (it should check the MapType, and divide by |J| for MapType==INTEGRAL).

tzanio:
@v-dobrev, can you help fixing this? It could be something that affects other physics...

mlstowell:
PR #1931 adds support for using `CurlInterpolator` with `MapType == VALUE`.  It also adds support for `MapType == INTEGRAL` in `GridFunction::GetValue`.

This leaves `VectorFECurlIntegrator`, and possibly other oversights, to be dealt with.

mlstowell:
PR #1995 adds several unit tests to verify that the L2 basis functions with `MapType == INTEGRAL` work as expected.  These tests identified at least one issue which was also fixed in that PR.

The trouble with `VectorFECurlIntegrator` is that it is only designed to work in 3D.  In 2D, where the `Curl` is scalar-valued, the correct alternative is to use the `MixedScalarCurlIntegrator` (this only works in 2D).

It is possible that we still don't have unit tests for every integrator and interpolator.  To my knowledge we have not performed a comprehensive audit of these unit tests.  If any more problems arise please let us know.

PR #1375, opened 2020-03-25
[WIP] Axisymmetric examples [cyl-coords-dev]

mlstowell:
This PR adds example code to demonstrate how to setup and solve standard problems on axisymmetric domains.  There is currently an issue with visualizing an L2 field which uses a `MapType` of `FiniteElement::INTEGRAL`.  Resolving this requires small changes to `GridFunction` and `VisualizationSceneSolution` in `GLVis`.  I'm not sure if it's better to place those changes here or in a separate PR.


mlstowell:
@tzanio and @v-dobrev ,

To visualize L2 fields with `MapType INTEGRAL` (or even to project them into a field of a different type) we'll need to make small changes to `GridFunction::GetValues()` (and perhaps similar methods).  Also, GLVis currently calls `GridFunction::GetGradients()` for scalar fields but this won't work (and probably cannot be implemented) for this map type.  I was thinking of adding a convenience method like `GridFunction::HasGradient()`, or something similar, so that GLVis could skip its call to `GetGradients` if it isn't available.

Since it would be best to fix each of the `GetValue` variants, even those not needed by these examples, should I put together a separate PR to address this issue?  I'd be happy to implement these changes either way.  Just let me know your preferences.

Best wishes,
Mark

tzanio:
Mark, is this related to https://github.com/mfem/mfem/issues/1037?

mlstowell:
@tzanio ,
Yes, it is a related issue although I had forgotten about that.

PR #1378, opened 2020-03-27
[WIP] krylov monitors

barker29:
The main idea here is an abstract `IterativeSolverMonitor` class that allows for more flexibility in how residuals etc. are reported in various `IterativeSolver`s.

The dream was to implement several `Legacy` monitors that would duplicate exactly the output from the current solvers, and then over the next couple versions of MFEM to mark those as deprecated and move to a more unified interface, and also one where users could write and use their own monitors.

After a little work, I decided this interface / framework was not workable. In particular, I could not figure out a unified interface for different *kinds* of residual information, for example in `BiCGSTAB` which prints two different kinds of residuals at each iteration. But perhaps someone can learn from my mistakes now that something like this is being considered.


barker29:
Closed because this functionality is implemented in #1390

PR #1391, opened 2020-04-01
[barker29/cg-converged-bugfix] add checks for indefinite preconditioner in CGSolver

barker29:
Current if `CGSolver` has an indefinite preconditioner, it may get a negative `(Br, r)` inner product, compare that to desired tolerance, and conclude that the iteration has converged. This PR adds a check to catch that and correctly report `GetConverged() == 0`.

| [#1391](https://github.com/mfem/mfem/pull/1391) | @barker29 | @tzanio | @delyank + @dylan-copeland | 04/01/20 | 04/07/20 | 04/09/20 | |

### DIFF
--- a/linalg/solvers.cpp
+++ b/linalg/solvers.cpp
@@ -388,7 +388,6 @@ void CGSolver::Mult(const Vector &b, Vector &x) const
                 << nom << (print_level == 3 ? " ...\n" : "\n");
    }
 
-   r0 = std::max(nom*rel_tol*rel_tol, abs_tol*abs_tol);
    if (nom < 0.0)
    {
       if (print_level >= 0)
@@ -401,6 +400,7 @@ void CGSolver::Mult(const Vector &b, Vector &x) const
       final_norm = nom;
       return;
    }
+   r0 = std::max(nom*rel_tol*rel_tol, abs_tol*abs_tol);
    if (nom <= r0)
    {
       converged = 1;
--- a/linalg/solvers.cpp
+++ b/linalg/solvers.cpp
@@ -382,7 +382,6 @@ void CGSolver::Mult(const Vector &b, Vector &x) const
    }
    nom0 = nom = Dot(d, r);
    MFEM_ASSERT(IsFinite(nom), "nom = " << nom);
-
    if (print_level == 1 || print_level == 3)
    {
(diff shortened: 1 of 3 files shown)


linalg/solvers.cpp
```diff
@@ -382,14 +382,25 @@ void CGSolver::Mult(const Vector &b, Vector &x) const
...
+      if (print_level >= 0)
+      {
+         mfem::out << "PCG: The preconditioner is not positive definite. (Br, r) = "
+                   << nom << '\n';
+      }
+      converged = 0;
+      final_iter = 0;
+      final_norm = nom;
+      return;
+   }
```
dylan-copeland, line 403: The calculation of r0 can be moved past this if-statement, since it is not used in the case of a return.

tests/unit/linalg/test_cg_indefinite.cpp
```diff
@@ -0,0 +1,59 @@
...
+   indefinite.Add(1, 0, 1.0);
+   indefinite.Finalize();
+
+   Vector v(2);
+   v(0) = 1.0;
+   v(1) = -1.0;
+   Vector x(2);
+   x = 0.0;
+
+   // check indefinite operator
```
dylan-copeland, line 42: Indefinite operators are still allowed (in which case only a warning is printed), so this comment is a bit confusing. For example, scaling the operator in example 1 by -1 in the PA case (using a Jacobi preconditioned) results in convergence with warnings about an indefinite operator. It might be better to clarify this test with a comment that more precisely describes how it fails.

dylan-copeland:
Does linalg/test_cg_indefinite.cpp need to be added to tests/unit/CMakeLists.txt?

tests/unit/linalg/test_cg_indefinite.cpp
barker29: In my view, CG just doesn't make sense with an indefinite operator. The example you give has a negative definite operator, which is sensible (but a bit unorthodox) for CG. I've tried to clarify the comments to refer to just the example we're checking in the test (which is clearly inappropriate for CG). Let me know if I need to clarify more.

tzanio:
LGTM

tzanio:

Issue #1332, opened 2020-03-03
NCMesh saving


januszpigonia:
How can I in example 6p save the configuration (ncmesh and solution) after a few refinement steps and then restore the results and continue ?

Janusz

jakubcerveny:
This is something that's high on my TODO list. We need to extend the MFEM 1.1 mesh format to allow for saving a parallel NC mesh, which is unfortunately not possible right now.

tzanio:
Please see #1392 and provide feedback there on the new NCMesh format.

Issue #1326, opened 2020-02-28
Use an (static) adaptive mesh after it is saved and loaded


tangqi:
Hi,
I have some issues to use the adaptive mesh after it is saved.
![Screen Shot 2020-02-27 at 5 26 39 PM](https://user-images.githubusercontent.com/1406660/75499141-613b7e00-5986-11ea-8016-b171ccbf3fe9.png)

This mesh is obtained by a simple local refinement through mesh-explorer.cpp. (It is periodic in one direction). 

Now if I do a simple `ProjectCoefficient` to H1 space, this is what I got:
![Screen Shot 2020-02-27 at 5 30 29 PM](https://user-images.githubusercontent.com/1406660/75499355-f9d1fe00-5986-11ea-92a1-152a8c0179de.png)
But the function should be 
![Screen Shot 2020-02-27 at 5 31 13 PM](https://user-images.githubusercontent.com/1406660/75499407-1b32ea00-5987-11ea-8422-e3697d035cea.png)

Do I miss something when I loaded this mesh? The test code is very simple
```c++
   Mesh *mesh = new Mesh(mesh_file, 1, 1);
   ParMesh *pmesh = new ParMesh(MPI_COMM_WORLD, *mesh);
...
   ParGridFunction psi(&fespace);
   FunctionCoefficient psiInit(func);
   psi.ProjectCoefficient(psiInit);
```
 I think if I do not save this mesh and instead generate it on the fly, it seems okay with projections. But it would nice if I can directly use a mesh like this. Thanks!

jakubcerveny:
Hi @tangqi, could you please attach the mesh that you created? Thanks

tangqi:
Thanks Jakub, @jakubcerveny

See mesh:
https://github.com/mfem/mfem/blob/tds-mhd-dev/miniapps/mhd/Meshes/amr-mesh-new.mesh

It is created by mesh-explorer by a simple local refinement based on the coarse mesh:
https://github.com/mfem/mfem/blob/tds-mhd-dev/miniapps/mhd/Meshes/xperiodic-new.mesh

Here is a simple serial test code:
```c++
#include "mfem.hpp"
#include <memory>
#include <iostream>
#include <fstream>

using namespace std;
using namespace mfem;

double InitialPsi(const Vector &x)
{
   double beta=1e-3, lambda=.5/M_PI, ep=.2;
   return -lambda*log( cosh(x(1)/lambda) +ep*cos(x(0)/lambda) )
[...]

PR #1392, opened 2020-04-02
ParNCMesh I/O and new NC mesh format [par-ncmesh-saving-dev]

jakubcerveny:
There are currently several problems with our nonconforming mesh format:
- `ParNCMesh` cannot be saved to a file, because the `MFEM mesh v1.1` format does not support storing element rank assignments (#1332).
- `v1.1` does not support storing root element orientation (`NCMesh::root_state`). This could be causing bugs like #1326, if the orientation is lost.
- The `v1.1` format was designed as a minimal extension to the conforming format `v1.0`, which in hindsight was not fortunate. The format does not mirror well the `NCMesh` internal data structure: elements and vertices need to be renumbered when saving/loading, and empty slots in the element/node arrays are not supported. As a result, the `NCMesh` is not identical when saved and loaded again.
- The parallel `v1.2` format completely ignores NC meshes and is in fact another branch off `v1.0`. A potential `v1.3` for parallel nonconforming meshes does not make sense.

This PR proposes the following:
- A new `MFEM NC mesh v1.0` format is introduced, independent of the conforming formats `MFEM mesh v1.0` and `v1.2`.
- The `v1.1` format is deprecated, although it can still be loaded, for backward compatibility.
- The new nonconforming format better mirrors the actual data structure and removes known inconsistencies. It works both in serial and in parallel, as the only parallel information is the optional `rank` section and the element `rank` member. Everything else in `ParNCMesh` is calculated from the ghost layer.
- The new format is still human readable and editable and is similar in style to the conforming format.
- Example 6p now illustrates how checkpoints can be saved and later loaded to restart the computation.

### Example of the new format

The following file was obtained by converting a 1x1 inline quad mesh to an NC mesh and refining it once:

```shell
MFEM NC mesh v1.0

# NCMesh supported geometry types:
# TRIANGLE    = 2
# SQUARE      = 3
# TETRAHEDRON = 4
# CUBE        = 5
# PRISM       = 6

dimension
2

# optional rank for parallel files, defaults to 0
rank
0

# rank attr geom ref_type nodes/children
elements
6
0 1 3 3 1 2 3 4     # this is a non-leaf (refined) element
0 1 3 0 0 4 8 7
0 1 3 0 4 1 5 8
0 1 3 0 8 5 3 6
0 1 3 0 7 8 6 2
0 1 -1              # empty slot (freed ID in NCMesh::elements)

# attr geom nodes
boundary
8
1 1 0 4
4 1 7 0
1 1 4 1
2 1 1 5
2 1 5 3
3 1 3 6
3 1 6 2
4 1 2 7

# vert_id p1 p2
vertex_parents
5
4 0 1
5 1 3
6 2 3
7 0 2
8 4 6

# root element orientation
root_state
1
0

# top-level node coordinates
coordinates
4
2
0 0
1 0
0 1
1 1

mfem_mesh_end
```
Notes:
- Both refined (`ref_type` != 0) and leaf elements (`ref_type` == 0) are listed in the `elements` section. This replaces the problematic `coarse_elements` section of `v1.1`. Refined elements reference child element IDs instead of node IDs, as in `NCMesh`. Element indices correspond 1:1 to the array `NCMesh::elements`. Empty entries can be represented by setting `geom` to -1.
- `boundary` and `vertex_parents` sections are the same as in `v1.1`. Vertex parents can be listed in any order and the IDs do not have to be contiguous. The IDs correspond 1:1 to `NCMesh` vertex nodes (i.e., not `Mesh::vertices`). Edge nodes are not stored and are assigned any of the free IDs on load.
- `root_state` stores the first N entries of `NCMesh::root_state`, the remaining entries (if N < M) are assumed to be zero. Here M is the number of root elements, automatically detected in the `elements` section (the first M elements must not have a parent). 
- The `coordinates` section replaces `vertices` of `v1.1` and only lists the coordinates of the top-level nodes (coordinates of other nodes are calculated from the parent relations in `NCMesh`). Only the first N node IDs should be top-level nodes (these have no parents and are thus not listed in `vertex_parents`). The section can be longer but we check that it covers all top-level nodes. When loading vertices from `v1.1`, we now trim the array to only cover the top-level nodes. 
- If the `coordinates` section is missing, we expect to find a `nodes` section instead of it. The Nodes are loaded the same way as in `v1.0` (it's just not a subsection of `coordinates`/`vertices`).
- The sections `rank`, `boundary`, `vertex_parents` and `root_state` are all optional (their order must be kept though, if present). Exacly one of `coordinates` or `nodes` must be present. 
- The file ends with `mfem_mesh_end` as in `v1.2`, which is not strictly necessary, but could be useful.

More implementation notes:
- `v1.1` can still be loaded as a `Mesh`, but not as a `ParMesh`. Although technically possible (all ranks would load the same file and then Rebalance could be called), it would probably be confusing.
- Changes in `ParMesh` and `ParNCMesh` are in fact minimal. Everything is handled by the serial `NCMesh` and `ParNCMesh` just calculates all its structures from the ghost layer.
- All parallel files (like `mesh.000000`) are valid serial NC meshes that can be displayed individually in GLVis. The corresponding `ParGridFunction` files are now also compatible.

UPDATE
- The serial code now knows the necessary minimum about ghost elements (those with `element.rank != MyRank`), ghost vertices, ghost edges and faces, which allows it to skip the ghosts when creating the Mesh representation seen by the rest of MFEM. 
- The member `ParNCMesh::MyRank` is now `NCMesh::MyRank` and it gets loaded from the `rank` section. When loading a mesh in parallel, the rank is checked to equal the actual MPI rank.
- `NCMesh::UpdateVertices` now works the same in both serial and parallel and enumerates vertices in specific order to keep compatibility with conforming meshes, and between serial and parallel NC meshes (see code for details).
- Similarly, `NCMesh::OnMeshUpdated` now handles ghost edges and faces, so that they are indexed the same way in serial and in parallel. 
- As a result, parallel solutions on NC meshes should now display in serial GLVis, both complete and as individual parts. 
- Possible TODO: processor boundary faces are not output anywhere explicitly as in ParMesh::Print, so 3D parts are displayed incomplete, I'm not sure how to handle this (I think GLVis should generate this information itself, but that's not how it works now).

TODO
- [x] Compatibility with partial grid functions?
- [x] Fix on the fly NC initialization (polar-nc)
- [x] Bug? Empty `coarse_elements` in legacy format required?
- [x] Mesh::SetCurvature must remove NCMesh::coordinates
- [x] `ParNCMesh::GetFaceNeighbors` crashes on curved meshes (`coordinates.Size() == 0`)
- [x] `ExchangeFaceNbrData` on curved meshes: "Nodes are not ParGridFunction!"
- [x] Convert `v1.1` meshes in the repo.
- [x] Fix loading curved v1.1 meshes
- [x] Crash when restarting ex6p in NC tet mesh (escher)
- [ ] Document the new format.
- [x] Update `ex6p` description.
- [x] Update CHANGELOG.

| [#1392](https://github.com/mfem/mfem/pull/1392) | @jakubcerveny | @v-dobrev | @mlstowell + @tzanio | 10/06/20 | 12/29/20 | 02/03/21 | |


jakubcerveny:
There's some more TODO items but people may start looking at this and comment on the format...

jakubcerveny:
I was looking into the problem of making partial ParGridFunctions compatible with partial NC mesh files, but it's more tricky than I thought. I will have to modify both NCMesh and ParNCMesh to make sure the elements and vertices are enumerated exactly the same way in parallel and serial.

@tzanio can we postpone the reviews please?

tzanio:
> @tzanio can we postpone the reviews please?

Done. Mark as `ready-for-review` when it is ready again.

termi-official:
Hi guys,

first off, thanks for the effort! I have two questions on this branch.

1. Is it possible to reset the internal tree data structure for an [Par]NCMesh? 
To elaborate on what I mean, I have written a tool on top of MFEM, which is somewhat similar to the Shaper miniapp, to generate a (nonconforming) mesh from voxel image data with the property of a minimum number of elements. I want to fix these elements to be the new roots for my mesh.

2. Is there an alternative way or a hotfix to store and load nodal information (in contrast to the not yet fully functional [Par]GridFunction), which can be made available as a coefficient for the integrator classes?
My imaging data also contains the spatially varying coefficients for my PDE, basically giving me information for a spacially varying diffusion coefficient field, therefore I also must save associated information. I my current implementation I realize this by storing a ParGridFunction and hand it to the DiffusionIntegrator via a VectorGridFunctionCoefficient.

Best regards,
Dennis

jakubcerveny:
Hi Dennis,

1. A function to flatten the NCMesh does not exist, but could be written. Just create a new `NCMesh::elements` that contains the leafs only, and resize (and zero-out) `NCMesh::root_state`. Oh and you may need to renumber the element references in `NCMesh::Face::elem`. I'm not sure if this actually pays off because in a uniformly refined hex mesh the number of coarse elements is ~14% (1/7) of the number of leafs.

2. I have now tested `ex6p` and the solution, when the NC mesh is saved with ParPrint, is broken because the ordering of elements is different. I still need to look into this. In the meantime I suggest you use the master branch, sorry.

Jakub

termi-official:
Hi Jakub,

thanks for the quick answer! And no reason to sorry. :)
The flattening procedure sounds really doable. I will take a look into this. To make the idea in my first point clear, it is not about saving memory, but about having a as coarse as possible mesh with fixed material interfaces. I want these to be kept in their place when running simulations with AMR (since derefinement could mess up the resolution of these material interfaces).

Best regards,
Dennis

termi-official:
Hi again,

I realized an issue related to element attributes with all available file formats. Since this PR redesigns the NC mesh format, this may be relevant.
All formats discard the attribute assignment for lower-dimensional entities, i.e. the attributes for points, edges and faces in a 3d mesh (point/point+edge attributes in 1d/2d respectively) are lost when saving and loading a mesh. This is relevant in simulations of models with multiple materials, in which material interfaces may require a different treatment, depending on non-local properties determined in preprocessing steps.

My proposal here would be to include *optional* sections (e.g. face_attributes, ...) for the attributes of lower dimensional entities. This way the format would stay compatible and if attributes for lower dimensional entities are not used, the saving them could be skipped.

Best regards,
Dennis

jakubcerveny:
Hi Dennis,

I'm not sure I understand. In MFEM you can have element attributes and boundary element attributes (boundary elements, if present, are dim-1 dimensional). We do save these. If you wanted vertex/edge/face attributes, these would have to be supported in MFEM in the first place. Can you describe your use case?

Jakub

termi-official:
Hi Jakub,

thanks for the quick response. I simulate electrophysiologcal models of heart tissues. These tissues are usually either embedded into other biological tissues or in some fluid. Face attributes are useful for storing information about "internal" material interfaces, e.g. for the boundary layers between different parts in the heart or between the heart and its surrounding. I need the face attributes frequently in all steps of my pipeline, spanning preprocessing, the simulation itself (because this information becomes handy in the coupling scheme I use) and postprocessing. 
For me it would be a nice to have, but I can also get away by storing this face information separately and managing it manually. I just wanted to point towards possible issue. If it is intended to not store such information, then it is (at least for me) also perfectly fine.

Regarding the support for face attributes, if I understand the implementation correctly, then I can access face attributes by first getting the face instance via 
 
https://github.com/mfem/mfem/blob/a25138e61179753291e0a128a94fae4e540e2bf9/mesh/mesh.hpp#L804

and then there is the following method

https://github.com/mfem/mfem/blob/62117e96dd3d9f64dc5d679a1b029cc802effa2a/mesh/element.hpp#L54-L58

where I tried to be consistent with the current implementation, i.e.

https://github.com/mfem/mfem/blob/a25138e61179753291e0a128a94fae4e540e2bf9/mesh/mesh.hpp#L1021-L1028


Best regards,
Dennis

jakubcerveny:
You are right, I forgot we do hold faces explicitly in `Mesh::faces`. I'm not sure this is a place for users to store attributes but it is true `Mesh::GetFaceTransformation` works with them. Is this the intended design @v-dobrev ?

My personal guess would be that on material interfaces you should be able to define "boundary" elements inside the mesh and then see their attributes in face integrators etc., but I may be wrong.

v-dobrev:
The idea for what we call "boundary elements" is that some of them can be actually internal faces describing interfaces. How these are actually used is up to the application, e.g. you probably do not want to apply boundary conditions there.

Does merging the interior (interface) faces with the real boundary faces make sense in your application?

One issue with maintaining attributes for all internal faces is that when you refine the mesh you create new faces that are not a refinement of existing faces, so there is no natural attribute assigned to those. That's one reason why to me it only makes sense to save attribute data for subsets of the faces and we just merge those with the real boundary elements.

What we do not currently support is storing attributes for lower-dimensional mesh entities like edges (in 3D) and vertices (in 2D and 3D). Is that something you need too? One option that has come up is to store them as "boundary elements" too -- that way the mesh format does not need to change.

One other thing, we currently do not copy the attributes from the boundary elements to their respective face elements (which are stored as separate objects in the Mesh class). I'm not sure if this will be useful to have. That's why the face attributes are not really used at this point, even though we copy the attributes in some places like `Mesh::GetFaceTransformation`.

termi-official:
Thanks for the elaborate explanation and your notes! That actually helps me to clear up several implementation-related uncertainties I had. I also totally agree on the point, that it really does not make sense to store attributes (d-2)-dimensional entities for the mentioned reason. Luckily I do not need attributes on these entities. Being able to reuse boundary elements for the internal boundaries should already be more than enough for my application. 
Thanks!

mesh/ncmesh.cpp
```diff
@@ -4893,73 +4880,404 @@ void NCMesh::LoadVertexParents(std::istream &input)
...
 
-      out << (int) el.ref_type;
-      for (int i = 0; i < nch; i++)
+   out << "\n# rank attr geom ref_type nodes/children";
+   out << "\nelements\n" << elements.Size() << "\n";
+
+   for (int i = 0; i < elements.Size(); i++)
+   {
+      const Element &el = elements[i];
+      out << el.rank << " " << el.attribute << " ";
```
termi-official, line 5031: Sorry to bother again, but I just noticed that the new storing mechanism does not sync element attributes properly beween the NCMesh elements ant the Mesh elements. Changing element attributes via `Mesh::SetAttribute` does not affect the elements stored in the NCMesh, causing the mentioned issue. 

This should be fixable either through changing  `Mesh::SetAttribute` to also change the corresponding element attribute in the NCMesh element set. Since I could not figure out how to find the corresponding NCMesh element in constant time, I also thought alternatively about changing the signature of `NCMesh::Print` to 
```cpp
void NCMesh::Print(std::ostream &out, Mesh& mesh) const
```
and changing to this line to something like
```cpp
if(el.index == -1)
{
    out << el.rank << " " << el.attribute << " ";
}
else 
{
    out << el.rank << " " << mesh.GetAttribute(el.index) << " ";

}
```
looks also like it might work out as a hotfix.

Best regards, 
Dennis

jakubcerveny: You're right and I think we should go for the first solution: `Mesh::SetAttribute` should call `NCMesh::SetAttribute` which would do `elements[leaf_elements[index]].attribute = attribute` or something similar.

tangqi:
Hi Guys, what's the status of this new mesh? If we use this new mesh, can we save the adaptive mesh and solution, and then restart from there? 

I start to do some multiscale simulations with my solver. The debug has been really challenging now for those hard cases without restart capability.

jakubcerveny:
@tangqi This is is not ready yet, sorry, I had to switch to other things. I will return to this ASAP.

jakubcerveny:
@tangqi I think this should already work for your use case (you may need to merge this branch into your own though). I modifed `ex1p` and added `examples/parprint-test` to demonstrate loading/saving of parallel NC solutions. Whatever you compute with `ex1p` is saved with the new ParPrint and loaded in `parprint-test`, where you can also refine and rebalance the solution, to check that it was loaded correctly and still works when modified.

What doesn't work yet is visualizing the new format in GLVis, in particular visualizing the partial files of a parallel solution, I need to make more changes to support that.

tangqi:
Awesome! Thanks @jakubcerveny  I will run some tests on my branch then.

jakubcerveny:
Merged `ncmesh-serial-ghosts-dev` into this branch. Serial code now knows about ghost elements and the algorithm to assign vertex, edge and face indices has been unified. Serial code (like GLVis) can now load parallel partial meshes and GridFunctions.

jakubcerveny:
I'll fix the crash when loading parallel NC tet meshes next week, but otherwise this is finished.

general/hash.hpp
```diff
@@ -483,6 +495,36 @@ void HashTable<T>::DeleteAll()
...
+template<typename T>
+void HashTable<T>::UpdateUnused()
+{
+   unused.DeleteAll();
+   for (int i = 0; i < Base::Size(); i++)
+   {
+      if (Base::At(i).next == -2) { unused.Append(i); }
+   }
+}
+
```
tangqi, line 527: Is item.next=-1 needed here?

Also, is UpdateUnused just "Reset" the HashTable

mesh/mesh.cpp
```diff
@@ -3535,15 +3573,44 @@ void Mesh::Loader(std::istream &input, int generate_edges,
    // - does not check the orientation of regular and boundary elements
    if (finalize_topo)
    {
-      FinalizeTopology();
+      // don't generate boundary elements for NCMesh, especially in parallel
+      // (FIXME: how can this even work for conforming ParMesh?
+      //  -- we don't want to generate artificial BEs for internal partitions!)
+      bool generate_bdr = (ncmesh == NULL);
+
+      FinalizeTopology(generate_bdr);
```
tangqi, line 3581: The generate_bdr flag was added by @pazner for refining mesh.  https://github.com/mfem/mfem/pull/963

I suspect we should never generate boundary in the mesh loader, what do you think @pazner?

general/hash.hpp
jakubcerveny: No, this just updates the `unused` list. This is a special function that's only needed when repopulating the table e.g. when loading from a file.

mesh/mesh.cpp
pazner: That also sounds reasonable to me — the boundary elements should be created explicitly by the mesh loader, and not by `FinalizeTopology`. However, I'm not very familiar with the mesh loading logic, so I can't say with complete confidence.

jakubcerveny: OK I disabled boundary element generation in any case on load.

termi-official:
Hi Jakub, 
I have two question related to the free id's. In your example you have a mesh file using the new format, generated by utilizing a single 2D element and refining it once.

1. I would expect to see 5 elements (root + four children), so where does the "free" element come from? Or is it manually added for demonstration purposes?
2. I also observed that the boundary elements are numbered from 1 to 4. I can observe the same behavior in my remeshing tools (which are implemented on top of MFEM) and that changes in the boundary attribute assignment are not properly propagated to the ncmesh writer. Probably more related to  #1610 and #1572 , but I just want to ask whether the former behavior is intended.

jakubcerveny:
Hi @termi-official, sorry for the late answer (I have been a bit ill the past few days). 

1. The empty slot in the example is just for illustration (I added it manually). Empty slots only occur if some elements have been derefined and their slots have not been reused yet.

2. You are right that changes in boundary attributes are not propagated from the Mesh to the NCMesh. This is an issue that is not really related to this PR, please feel free to add code that does this (possibly along with other attribute related changes we discussed some time ago) and we can make it another PR.

termi-official:
Hi @jakubcerveny , 
no problem, I hope you are well again! And thanks for the clarification. 

Regarding the boundary attributes I just wanted to double check if this is the same issue as we discussed already or whether this could be something else, because the default attribute assignment looked a bit strange to me. But let us take look into this in pr 1610 after this here is merged.

I am currently testing this branch against my implementation and for now the new IO on non-conforming meshes works like a charm. However, I still need at least a week or two to fully integrate and test this properly. I want to note that the last time I integrated this branch (few months ago already) load balancing broke the attribute assignment. For now I was not able to reproduce this with ex6p and ex21p, so I will update you on this next week.

Best regards,
Dennis

mlstowell:
Hi, @jakubcerveny ,

I would like to suggest a small change to `ex6p`.  Add a command line option to specify the value of `max_dofs`.  This would enable us to test AMR after a restart.  Presumably it will work fine but this will make the restart option a little more meaningful in this example.

Best wishes,
Mark

mlstowell:
The `joule` sample run which uses the new `cylinder-tet.mesh` fails due to a triangular face with an unsupported orientation.   I checked master and this seems to be a new issue in this PR.

Also, the sample run using the `cylinder-hex-q2.mesh` appears to go unstable but, unfortunately, this also happen in `master` so this may have been a previously unnoticed problem.

mlstowell:
The serial unit tests fail however this PR is not the most likely culprit.  Perhaps merging the latest master branch into this one will resolve this.

mlstowell:
I ran the restart feature of `ex6p`, with a couple extra AMR iterations, though valgrind and didn't see any noteworthy output.

mlstowell (changes requested):
@jakubcerveny , your code is very easy on the eyes.  Thank you for that.

I think this is ready for merging once the following tasks are addressed:
- [ ] Consider adding `max-dofs` option to `ex6p`.
- [ ] Resolve issue with `joule` and its `cylinder-tet.mesh` sample run.
- [ ] Make sure the `unit_tests` run correctly.

jakubcerveny:
@mlstowell, I think the problem with `joule` is not related to this PR -- I am actually able to get this error in master too: #1874. 

The reason this is visible now is that in master, `cylinder-tet.mesh` is a `v1.1` mesh without a `vertex_parents` section, so the mesh is actually loaded as conforming. But if you do `mesh->EnsureNCMesh(true)` (force simplices to be NC), you get the same error. I created the above issue to address it. 

In this PR, we should maybe keep `cylinder-tet.mesh` a conforming (`v1.0`) mesh instead of converting it to the new NC format?

jakubcerveny:
I merged latest master and when I do `make pdebug; make test` I don't get any errors...

mlstowell:
> @mlstowell, I think the problem with `joule` is not related to this PR -- I am actually able to get this error in master too: #1874.
> 
> The reason this is visible now is that in master, `cylinder-tet.mesh` is a `v1.1` mesh without a `vertex_parents` section, so the mesh is actually loaded as conforming. But if you do `mesh->EnsureNCMesh(true)` (force simplices to be NC), you get the same error. I created the above issue to address it.
> 
> In this PR, we should maybe keep `cylinder-tet.mesh` a conforming (`v1.0`) mesh instead of converting it to the new NC format?

Yes, I think that's best.  I just double-checked that the original mesh works in this new branch if we just change the version in the header from v1.1 to v1.0.  Should we keep the new mesh?  If it was non-trivial to create it, I think we should keep it somewhere.  I'll leave that up to you, @jakubcerveny.

mlstowell (approved):

v-dobrev:
Saving "empty slots" is kind of strange -- this is a byproduct of how we store things internally. I guess you need it for recovering the internal state which influences future id assignments. That, however, may not be sufficient since we don't know the order in which the "empty slots" are listed in the `free_element_ids` array. Could this be a problem?

We can probably introduce "empty slot removal" step after refinement/derefinement to fill-in the empty slots with ids from the end. This, however, will require updating element ids which may not be that easy.

jakubcerveny:
I think the problem is that we are trying to come up with a human readable format that is also a serialization format. The empty slots are a compromise. Normally they are pretty rare (typically they are immediately reused in dynamic AMR) and if the file is created manually, the user does not have to care or even know about them. 

The alternative is an explicit element ID on each line in the `elements` section, but that's mostly redundant and it would be another deviation from the `v1.0` style. I would also like to avoid any renumbering of the IDs.

mesh/pncmesh.cpp
```diff
@@ -46,12 +46,38 @@ ParNCMesh::ParNCMesh(MPI_Comm comm, const NCMesh &ncmesh, int *part)
...
+   MPI_Comm_size(MyComm, &NRanks);
+
+   int my_rank;
+   MPI_Comm_rank(MyComm, &my_rank);
+
+   int max_rank = 0;
+   for (int i = 0; i < leaf_elements.Size(); i++)
+   {
+      max_rank = std::max(elements[leaf_elements[i]].rank, max_rank);
+   }
```
termi-official, line 63: Shouldn't this be something like 
```cpp
bool owns_elements = true;
for (int i = 0; i < leaf_elements.Size(); i++)
{
    owns_elements &= elements[leaf_elements[i]].rank == my_rank;
}
```
? Or do I misunderstand the intent here?

mesh/pmesh.hpp
```diff
@@ -349,6 +366,19 @@ class ParMesh : public Mesh
...
    void PrintAsOneXG(std::ostream &out = mfem::out);
 
+   /** Print the mesh in parallel PVTU format. The PVTU and VTU files will be
+       stored in the directory specified by @a pathname. If the directory does
+       not exist, it will be created. */
+   virtual void PrintVTU(std::string pathname,
+                         VTKFormat format=VTKFormat::ASCII,
+                         bool high_order_output=false,
+                         int compression_level=0,
+                         bool bdr=false);
```
termi-official, line 376: How does this play together with ParGridFunctions?

mesh/pmesh.cpp
```diff
@@ -5291,10 +5324,17 @@ long ParMesh::ReduceInt(int value) const
 
 void ParMesh::ParPrint(ostream &out) const
 {
-   if (NURBSext || pncmesh)
+   if (NURBSext)
    {
-      // TODO: AMR meshes, NURBS meshes.
-      Print(out);
+      // TODO: NURBS meshes.
```
termi-official, line 5329: Shouldn't we give at least a warning in debug mode here?

mesh/pmesh.cpp
```diff
@@ -1776,6 +1797,28 @@ void ParMesh::SetCurvature(int order, bool discont, int space_dim, int ordering)
    Nodes->MakeOwner(nfec);
 }
 
+void ParMesh::EnsureParNodes()
```
termi-official, line 1800: Is there any reason why we don't simply virtualize and overwrite `EnsureNodes`?

termi-official (approved):
Hi @jakubcerveny ,

load-balancing is broken for me on this branch. This issue seem to be related to the `Rebalance` messing up attribute assignment on relocated elements, is some corner cases. One corner case is when an NC mesh has at least one leaf element has a different attributes than its parents, then rebalancing the element seems to destroy the attribute assignment (at least in some cases). If you want I can try to capture this in a unit test. This is a regression as it worked previously, however I think fixing this should be a follow up PR, as it is again specifically related to attributes.

I also found a feature which is not working correctly. Loading a conforming mesh in parallel and transforming it to a nonconforming mesh does not work, because `EnsureNCMesh` does not setup the pncmesh member correctly (and ncmesh is just instantiated with a the serial NCMesh). This is not a regression introduced by your changes and should also be broken on the master. I still think this would be a nice addition to this PR as it already fixes several issues related to NC meshes.

As a further note, I really cannot figure out how to create a Par(NC)Mesh with N partitions, without loading the full serial mesh on N processes and constructing a Par(NC)Mesh from it, so I could not test this. Creating many partitions this strategy on large meshes turns out to be a bottleneck. The same question is open for me with loading ParGridFunctions from serial files in N partitions, which should be closely related. 

As @v-dobrev brought it up, I noticed that `Trim` does *not* release the introduced "free element slots". I think this should be documented to avoid surprise. Regarding clearing the free slots in trim, I am not sure if it is worth the hassle, as I do not see any application for this.

In addition to this I also left some minor comments and questions on some of your changes. Excepting making a conforming mesh nonconforming for AMR simulations, when loaded in parallel, I cannot find any major issues for general applications. As I think this problem will be fixed quickly and the other mentioned issues are more tangential to this PR I give my approval.

Best regards,
Dennis

mesh/pncmesh.cpp
jakubcerveny: There is no need for a rank to own any elements ("empty" ranks are legal). This is just a simple sanity check when loading a parallel mesh: if there are elements (e.g. in the ghost layer) that are supposed to belong to a rank that is outside the range of the current communicator, it's clear the mesh was created on a larger MPI run and cannot be loaded.

mesh/pmesh.hpp
jakubcerveny: I am not sure, hopefully this is orthogonal, I just rearranged the `Print*` declarations that seemed scattered in the header file.

mesh/pmesh.cpp
jakubcerveny: Perhaps... this wasn't my code.

mesh/pmesh.cpp
jakubcerveny: I believe the two methods are fundamentally different. `Mesh::EnsureNodes` is a public method that makes sure for the user that `Nodes` is defined, even if the mesh has straight edges. On the other hand, `ParMesh::EnsureParNodes` is an internal helper that converts `Nodes` (if they exist) from a `GridFunction` to a `ParGridFunction`.

`EnsureParNodes` is actually a refactor of the messy block in `ParMesh::RebalanceImpl` in master. Maybe I should have chosen a less similar name to `EnsureNodes` (then again, the name implies what actually happens).

Yet another question is whether `Mesh::EnsureNodes` works correctly if called on a `ParMesh`, but this is out of the scope of this PR.

jakubcerveny:
Hi @termi-official, 

Thanks for the review. Load balancing never cared about element attributes and I don't believe this actually works in master. Although this is a valid use case, I never had it in mind when writing the Par/NCMesh code (the assumption was that the whole refinement tree shares the attribute of its root element). Handling NC element attributes correctly will require a project of its own. Do you think you could summarize in an issue all the things that don't work regarding attributes in your use cases (I started losing track of them), and we'll make a PR to fix this.

Converting a conforming ParMesh to an NC ParMesh is not possible. One needs to convert while the mesh is still serial. I added a check in EnsureNCMesh -- good catch.

A parallel NC mesh needs to be created from a single serial Mesh that is assumed to be coarse enough to fit on all MPI ranks. This is of course a strong assumption but it greatly simplifies the parallel NC implementation, so it was accepted at the very beginning.

As for NCMesh::Trim, it would be possible to release unused blocks at the end of BlockArray. I added a TODO comment for the future.

Thanks again for testing this PR thoroughly.

mesh/pmesh.cpp
pazner: @jakubcerveny, `ParMesh::SetCurvature` may already do what you need (it's not exactly the same as your `EnsureParNodes`, but it's close)

termi-official:
Hi Jakub,

glad to read that this one was helpful for you and thanks for the comments. I also want to note that my mentioned issues on the attribute management on NC meshes are currently not too important for me, because I have duck taped some patches which work for me in my project.

I have some more questions regarding restarts, which I forgot in my previous post. 
Since this PR basically unlocks AMR simulations with restarts, do we have `DataCollection`s supporing this feature? I skipped through all subclasses and I think only the VisItDataCollection may be capable of this, correct? If so, are there any future plans to include a HDF5 (or similar) based format? If the VisItDataCollection is also uncapable of handling this, I have some very basic MFEM-specific DataCollection, which I could factor out of my project and contribute. My implementation is not really efficient, but it works.

Being able to load a DataCollection is relevant in workflows where FEM simulations are chained. For example in my current project I use the result of a first FEM simulation as the input for a second FEM simulation.

jakubcerveny:
Hi Dennis,

I am not too familiar with DataCollections, but on a quick look it seems `DataCollection::SaveMesh` is able to use `pmesh->ParPrint` (if `format == PARALLEL_FORMAT`), so my guess would be that the parallel NC format could just work with DataCollections. Did you try this?

termi-official:
Hi Jakub,
output with the DataCollection works fine, but the `Load` method is "not implemented" (i.e. throws an exception), which makes sense, because we are missing meta information at this point. That is why I subclassed the DataCollection to store some additional metainformation, which works.
Going through the Subclasses the SaveMesh is never used and Save is usually overwritten. The ParaViewDataCollection works fine for just outputting data, but loading will not work, because the underlying refinement tree is missing. For the VisItDataCollection loading and storing should work fine, but the class is at least missing the cycle information necessary to load the collection (i.e. the only way I to reconstruct the cycle number seems to be parsing the written folder names, if I have not missed something obvious here). This should be rather easy to fix. For the remaining DataCollections I am not even sure if the output works properly.

jakubcerveny:
OK, sounds like a good topic for another PR, if you feel like contributing your improvements.

tzanio (approved):
Approved with one suggestion below.

Thanks for adding this and illustrating it with the restart option in ex6p, @jakubcerveny!

data/amr-hex.mesh
```diff
@@ -1,241 +1,257 @@
-MFEM mesh v1.1
+MFEM nonconforming mesh v1.0
```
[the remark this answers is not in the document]
jakubcerveny: Sure... I don't have a strong preference.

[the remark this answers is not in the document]
jakubcerveny: Done.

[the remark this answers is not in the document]
tzanio: Thannk Jakub

v-dobrev:
Quoting from https://github.com/mfem/mfem/pull/1392#issuecomment-725731237:

> Saving "empty slots" is kind of strange -- this is a byproduct of how we store things internally. I guess you need it for recovering the internal state which influences future id assignments. That, however, may not be sufficient since we don't know the order in which the "empty slots" are listed in the `free_element_ids` array. Could this be a problem?

@jakubcerveny, I don't think you answered my question above. To clarify a bit: since we are storing the "empty slots" by marking them, we do lose their order as stored internally in `free_element_ids`; that in turn means we are NOT recovering the internal state 100%, hence my question. Isn't such 100% recovery the reason to want to store the "empty slots" in the first place?

jakubcerveny:
@v-dobrev, the main reason to store empty slots is to preserve exact element IDs, which occasionally requires skipping unused indices in `NCMesh::elements`. You are right that the ordering of `free_element_ids` is not preserved (it will be sorted on load), so the recovery is just 99.9% :-) But future element ID assignment is not really important, it's the current set of active IDs that matters.

v-dobrev:
> @v-dobrev, the main reason to store empty slots is to preserve exact element IDs, which occasionally requires skipping unused indices in `NCMesh::elements`. You are right that the ordering of `free_element_ids` is not preserved (it will be sorted on load), so the recovery is just 99.9% :-) But future element ID assignment is not really important, it's the current set of active IDs that matters.

Okay, good. I just wanted to double check if the ordering is important.

v-dobrev:

v-dobrev:
Re-merged in `next`.

jakubcerveny:
The problems in autotest seem to be caused by changes in edge/face orientation, especially the `tesla` miniapp has flipped signs in the linear system. Still debugging...

jakubcerveny:
I was able to reproduce this more simply in the RHS of `ex3p` on `ball-nurbs.mesh` on 2 CPUs. Should be easier to find now.

jakubcerveny:
So it's the `ldof_signs`, again. In this branch, the vertices of an NC mesh are ordered in such a way that if (i,j) is an edge and i<j on some processor, the corresponding vertices on another processor are ordered the same way. This means that the `ldof_sign` array is all positive, which I thought is a good thing. But apparently somewhere in the code the shared edges are broken because of this (for Hcurl and Hdiv spaces).

jakubcerveny:
It was a pain but I think I found the problem. The `ldof_signs` was a false alarm. It is true there is a difference now between the new NC mesh and the master branch in edge orientation, due to the new globally consistent ordering of the vertices. This causes some edge DOFs in H(curl) spaces to be inverted and as a result there are sign flips in the linear system and the RHS. But this is perfectly OK, IMO. 

The problem is actually with `HypreAMS` when it's calling `ParMesh::GetVertex(i)` on NC curved meshes. There is a warning in `Mesh::GetVertex` that it shouldn't be called for high-order meshes. In this branch the vertices are actually uninitialized (I would have never found this without valgrind), which I'll try to fix, but it's partly an old problem. I think we should make sure `GetVertex` always works, even on curved meshes.

jakubcerveny:
I made sure `vertices` are set to valid coordinates even though `Nodes` are present, which is the previous behavior. However, I think the proper solution is to set `vertices` to zero size if `Nodes` are active, and modify `GetVertex` to be able to handle high-order meshes.

tzanio:

tzanio:
Thanks @jakubcerveny. I will take a closer look at `HypreAMS` and this issue next week.

jakubcerveny:
I was looking at the remaining differences in AMR runs in autotest. They are caused by the new ordering of vertices in NCMesh, which I tested by reverting to the old ordering. Then, in a serial run of `ex6`, I reproduce exactly the autotest log. The parallel runs of `ex6p` and `volta` are not exactly the same as the autotest, maybe I'm not running them the same or there are machine differences. But when I compare `par-ncmesh-saving-dev` and `master` on my machine, I get an exact match (`mpirun -n 4 ex6p`):

amr

The old ordering must of course normally be kept `#ifdef`ed out.

tzanio:

examples/ex6p.cpp
```diff
@@ -169,31 +189,30 @@ int main(int argc, char *argv[])
...
    //     The refiner will call the given error estimator.
    ThresholdRefiner refiner(estimator);
    refiner.SetTotalErrorFraction(0.7);
 
-   // 13. The main AMR loop. In each iteration we solve the problem on the
+   // 14. The main AMR loop. In each iteration we solve the problem on the
    //     current mesh, visualize the solution, and refine the mesh.
-   const int max_dofs = 100000;
-   for (int it = 0; ; it++)
+   for (int it = 1; ; it++)
```
v-dobrev, line 215: I think it is more appropriate (and consistent with `ex6`) to call the first iteration "0" because it corresponds to the initial mesh and so no there's no AMR at that point.

v-dobrev:
There was no mention of the "new vertex ordering" in the above descriptions -- I thought this was just adding/changing file i/o. What's the motivation for changing the vertex ordering?

examples/ex6p.cpp
jakubcerveny: I forgot to change `ex6` to match the one-based iteration in ex6p, which I think is more natural for the user and also makes the checkpoint condition simpler (`it % 5 == 0` at the end of `ex6p`). Fixed in `ex6`.

jakubcerveny:
I agree it should have been mentioned more explicitly (it's hidden in the "UPDATE" section). There is however a huge comment in the code in `NCMesh::UpdateVertices`, which explains the reasoning:
- In `master`, the serial NCMesh and parallel ParNCMesh are using a different ordering of vertices. This is unnecessary and inconsistent but more importantly, it prevents the serial code from loading parallel partial GridFunctions e.g. for visualization. This is fixed now in this branch.
- In `master`, if you load a GridFunction (even in serial) and then change the mesh to NCMesh, the GridFunction will break because of the difference of vertex ordering between conforming/nonconforming meshes. This could be considered a bug and this branch fixes that too.
- The new ordering is globally consistent which also makes edge orientations consistent in parallel (I want to use that in #1274).

tzanio:

examples/ex6p.cpp
v-dobrev: That's not the action I was expecting as a result of my comment.

If there is disagreement, I think it is appropriate to resolve it before further propagating the point of disagreement.

v-dobrev: @mlstowell, @tzanio, what is your preference regarding the above dispute?

tzanio: I agree that the two examples should be consistent, other than that I don't have a strong preference.

For what it's worth, Example 21 also starts from `0`: https://github.com/mfem/mfem/blob/master/examples/ex21.cpp#L192

v-dobrev:
One more thing: due to the new vertex ordering, it seems like AMR Mesh + GridFunction generated and saved by the current `master` branch and previous MFEM versions will no longer be readable by this branch. If I understand this correctly, the mesh itself will be fine, however, the GridFunction data will be permuted and with flipped signs in some cases. Is that correct?

If this is the case, this deserves to be mentioned in the `CHANGELOG` in a separate bullet.

This will also cause issues with the MFEM plugin in VisIt which is based on an older MFEM version.

The best solution will be to support reading correctly the previous default AMR mesh format and respective GridFunctions. How hard will this be to implement?

examples/ex6p.cpp
jakubcerveny: I don't think there is a disagreement, it's a negligible issue and both solutions are valid in my view. If there is a precedent in other examples let's change it accordingly.

jakubcerveny:
You are right, this could be a problem. The issue has already been addressed for the embedded Nodes, which are permuted when loading v1.1. We have unfortunately less control over other GridFunctions, but maybe if we remembered that the NCMesh was loaded from the legacy format, we could automatically convert other GridFunctions too. I'll check if this is possible.

jakubcerveny:
Done: all GridFunctions are now converted transparently if they are defined over an NCMesh that was loaded from a `v1.1` file. The same happens for Nodes. Edges are now handled too, including a potential sign flip. Tested on GLVis compiled with this branch and different examples from `master`. Tried `ex1` (with forced NC mesh) with cubic elements both in the solution and in Nodes. This indeed didn't work well, but works nice now. Tested both 2D and 3D and also `ex3`, all seems resolved.

tzanio:

v-dobrev:
Re-merged in `next`.

examples/ex6p.cpp
```diff
@@ -280,32 +299,50 @@ int main(int argc, char *argv[])
...
+      // 24. Inform also the bilinear and linear forms that the space has
       //     changed.
       a.Update();
       b.Update();
+
+      // 25. Save the current state of the mesh every 5 iterations. The
+      //     computation can be restarted from this point. Note that unlike in
+      //     visualization, we need to use the 'ParPrint' method to save all
+      //     internal parallel data structures.
+      if ((it + 1) % 5 == 0)
```
v-dobrev, line 331: An alternative is to use the condition: `if (it > 0 && it % 5 == 0)`. That way we'll get a checkpoints after iterations 5, 10, 15, ... but not for the initial mesh, `it == 0`.

jakubcerveny: Hm, I don't know, the first checkpoint will saved after 6 iterations in this way, I'd rather keep it as it is...

tzanio:
I just noticed that the reordering in this PR breaks previously saved files like the ones in

https://github.com/GLVis/web/tree/gh-pages/data

For example `minimal-surface.saved` and `shaper.saved` do not work correctly with `glvis -saved` or the web version in https://glvis.org/live.

Is there an easy way to update these files to the new ordering?

examples/ex6p.cpp
v-dobrev: I now realize that this checkpoint is at the wrong place, in my opinion. The right place is to save the mesh and solution after the solve on the mesh, not before. That is what I assumed was happening but I had not checked.

v-dobrev:
@tzanio, I thought the issue you are seeing was resolved in bed65fd3 but maybe not. @jakubcerveny can you take another look?

tzanio:
I rebuild everything again with `glvis:master` using 3 mfem hashes:

- `master`
- 8509a8f (current `par-ncmesh-saving-dev`)
- 65c9a62  (`par-ncmesh-saving-dev` before the fix)

Here is what I get for `./glvis -saved minimal-surface.saved`:

#### master
minimal-surface-master

#### 8509a8f
minimal-surface-8509a8f

#### 65c9a62
minimal-surface-65c9a62

jakubcerveny:
The vertex reordering seems OK (or there is none in the case of `shaper.saved`), but I'm suspecting there is another difference in the `root_state` which in `v1.1` is not saved and has to be guessed (this is one of the motivations for the new format). I'll look more.

tzanio:
> I'm suspecting there is another difference in the root_state which in v1.1 is not saved and has to be guessed (this is one of the motivations for the new format). I'll look more.

If it is really complicated, we can try to recreate these specific datafiles, but I'm concerned that users may have saved files like these e.g. from previous runs.

jakubcerveny:
I'll find it and fix it, the ordering of elements is clearly different (Ctrl+O in GLVis). It's strange that the root states, the roots themselves and the entire trees all turned out to be exactly the same. But I'm closing in, there must be some simple explanation.

jakubcerveny:
It's hopefully fixed. The problem was related to yet another inconsistency of the `v1.1` format: since in `master` the loading code for `v1.0` and `v1.1` is intertwined, the leaf elements are loaded directly to the `Mesh`, in the order they appear in the file, regardless of the orientation of the NCMesh root elements. The proper SFC ordering is produced only after the NCMesh is modified and the leaves are exported to the Mesh again. This is of course a buggy behavior which this branch removes (the Mesh is now always generated from the NCMesh, no matter what), but for the purposes of loading the legacy `v1.1` I have reintroduced this "feature".

tzanio:
Thank you Jakub! I can confirm that 41089b9 fixed the issues I was seeing. 👍

v-dobrev:
Re-merged in `next`.

examples/ex6p.cpp
jakubcerveny: Sorry for the late reply. The reasoning was to allow the example to continue from a checkpoint exactly as it would have if it had continued normally, that's why the checkpoint is saved at the end of the iteration and when restarted, the for loop will just restart with the same data. Maybe that's not how this is normally done, but after all it's really just a sketch. Another reason to have it at the end is stylistic -- we probably don't want to clutter the main part of the code, since checkpointing is secondary here. But if there is a consensus, it's not a problem to move the checkpoint after step 20 in the example.

mlstowell: I think you both have valid points.  In a realistic simulation code we would want to save the mesh and solution immediately after the solve as @v-dobrev states.  If AMR is part of the solution process this would mean immediately refining the mesh after a restart.  This would complicate the code slightly but in a realistic simulation code that would be acceptable.

However, this example code is a special case as @jakubcerveny suggests.  The main purpose is to demonstrate AMR.  We could therefore argue that the refinement taking place in step 21 is actually the "solution" that we are really interested in so it makes more sense to save the refined mesh when storing the checkpoint.

Because I view this as a special case, I vote to keep the code as it is.

tzanio: Good points Mark. You convinced me -- I also vote to keep the checkpoint code as is.

v-dobrev: Just for the record: I'm not convinced that this is a good choice.

However, since you all agree, let's proceed without changing this.

v-dobrev:

jakubcerveny:
Hopefully one last thing that might save us trouble in the future: added a new (optional) section `sfc_version` (right after `rank`) that specifies the type of SFC ordering in the mesh file. Currently this needs to be zero (or unspecified), but in the future we might want to improve the SFC curve e.g. for triangles or prisms. Without this section, changing the SFC in MFEM would break older files. With the SFC version stored in the file, we will be able to fall back to an older ordering in the NCMesh class.

v-dobrev:
Re-merged in `next`.

PR #1402, opened 2020-04-07
Remove astyle version check from makefile

henrykrumb:
I was trying to run `make style` for a different PR, following the "pull request check list" in the contribution guidelines. It turned out that the style target in the main makefile checks for version 2.05.1 of astyle to be installed on the system, whereas I have version 3.1 installed.

I propose to remove the version check entirely, because the minimum required version is from 2014 according to astyle [http://astyle.sourceforge.net/notes.html](release notes) and it can be assumed that astyle is backwards compatible.

### DIFF
--- a/makefile
+++ b/makefile
@@ -668,7 +668,6 @@ status info:
 
 ASTYLE_BIN = astyle
 ASTYLE = $(ASTYLE_BIN) --options=$(SRC)config/mfem.astylerc
-ASTYLE_VER = "Artistic Style Version 2.05.1"
 FORMAT_FILES = $(foreach dir,$(DIRS) $(EM_DIRS) config,"$(dir)/*.?pp")
 FORMAT_FILES += "tests/unit/*.cpp"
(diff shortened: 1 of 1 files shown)


tzanio:
>  it can be assumed that astyle is backwards compatible

This unfortunately is not correct.

IdoAkkerman:
Can confirm.... took me a while to figure out why commits did not pass the checks.
I am happy to hear there is a version check!

Verstuurd vanaf mijn iPhone

henrykrumb:
Thank you both for replying so quickly.
My assumption was based on personal experiences with astyle, but double-checking with both versions (2.05.1 and 3.1), I realized that this does not match with MFEM codebase.
Would you think it makes sense to upgrade the version check to the most recent version 3.1?

tzanio:
@henrykrumb, we are discussing upgrading to newer `astyle` or potentially replacing it with `clang-format`.

This will take us a while though 🚧 

In the meantime, I will close this PR, please let us know if you have any further feedback.

PR #1414, opened 2020-04-09
Adding NURBS face dofs [nurbs-face-dev]

IdoAkkerman:
This PR adds the capability of using face integrals on the boundary of NURBS meshes.
This is necessary for weak forms including normal derivatives, such as the Nitsche formulation for weak enforcement of Dirichlect BCs.

The implementation requires two components:
1) The implementation of GetFaceDofs in fespace for NURBS. For NURBS the dofs are not (straightforwardly) associated with geometric entities, so instead a dedicated `face_dof` table is defined and used. This table is not fully populated at the time. It is only populated at the boundaries. In a seperate routine `GenerateFaceDofs` in fespace this table is generated, but bootstrapping it from the already defined `bdrElem_dof` table.

2) In `GetFaceElement()` the correct NURBS knot span needs to be set in the `NURBSextension`.
For this we use the `LoadBE()` routine from `NURBSextension`, in conjunction with `face_to_be` map that is generated in `GenerateFaceDofs`.

What I am wondering is:
1) Is this a solid approach?
2) What is the best place to make a call to `GenerateFaceDofs`? At this stage it is called right before it is needed, with an escape in the routine if it is already defined. This works but is perhaps 
a bit clumsy.

The natural place seems `UpdateNURBS()` but `GenerateFaceDofs` requires the boundary info in the  mesh to be available, which apparently is not yet available at this stage of initialization.
Perhaps this requires some more research from my side to figure out, what is happening....

| [#1414](https://github.com/mfem/mfem/pull/1414) | @IdoAkkerman | @v-dobrev | @pazner + @acfisher | 04/29/20 | 06/09/20 | 06/15/20 | |


tzanio:
@v-dobrev, can you take this one?

IdoAkkerman:
When adding a small 3D testcase, I noticed the fix for empty METIS partitions was only called once.
I added a while loop to apply the fix recursively. Perhaps there should be check to make sure `nparts >= NumOfElements`.

IdoAkkerman:
This is a small fix and one of the testcases can demonstrate the error that is why I did not create a seperate PR.

IdoAkkerman:
NURBS mesh in patch format needs checking. I think the current placement of `GenerateFaceDofsFormBdr` is not adequate.

IdoAkkerman:
There is still an orientation issues.` GenerateFaces` might need a NURBS specific implementation??

IdoAkkerman:
I can not see why appveyor failed. What do I need to do to get that info?

IdoAkkerman:
Had to remove the sort call. This required the faces to be added in the right order,

miniapps/nurbs/CMakeLists.txt
```diff
@@ -9,33 +9,59 @@
 # terms of the BSD-3 license. We welcome feedback and contributions, see file
 # CONTRIBUTING.md for details.
 
-configure_file(${PROJECT_SOURCE_DIR}/miniapps/nurbs/square-nurbs.mesh
-               ${PROJECT_BINARY_DIR}/miniapps/nurbs/square-nurbs.mesh
-               COPYONLY)
+#configure_file(${PROJECT_SOURCE_DIR}/miniapps/nurbs/square-nurbs.mesh
```
acfisher, line 12: Why are these commented out?

miniapps/nurbs/nurbs_ex1.cpp
```diff
@@ -3,6 +3,7 @@
 // Compile with: make nurbs_ex1
 //
 // Sample runs:  nurbs_ex1 -m square-nurbs.mesh -o 2 -no-ibp
+//               nurbs_ex1 -m square-nurbs.mesh -o 2 --weak-bc
```
acfisher, line 6: I'm not sure miniapps is the right place for these examples.  The miniapps are usually more complicated.  Where do folks think we should put these?

v-dobrev: I don't see a reason to move them at this point. Miniapps don't need to be complicated -- there are already a number of simple miniapps.

IdoAkkerman: Perhaps they can be moved to a subdirectory of examples. However, it seems the other subdirs seem to depend on compilation options (package used yes/no).

miniapps/nurbs/CMakeLists.txt
IdoAkkerman: The testcases below now use the full path to the data directory, so there is no need for a copying the files. I will remove the lines all together to prevent confusion.

data/square-nurbs.mesh
```diff
@@ -0,0 +1,57 @@
+MFEM NURBS mesh v1.0
```
pazner, line 1: This same file is duplicated in `miniapps/nurbs`

data/square-disc-nurbs-patch.mesh
```diff
@@ -0,0 +1,155 @@
+MFEM NURBS mesh v1.0
```
pazner, line 1: This same file is duplicated in `miniapps/nurbs`

data/pipe-nurbs-2d.mesh
```diff
@@ -0,0 +1,70 @@
+MFEM NURBS mesh v1.0
```
pazner, line 1: This same file is duplicated in `miniapps/nurbs`

data/cube-nurbs.mesh
```diff
@@ -0,0 +1,73 @@
+MFEM NURBS mesh v1.0
```
pazner, line 1: This same file is duplicated in `miniapps/nurbs`

fem/fespace.hpp
```diff
@@ -112,6 +112,8 @@ class FiniteElementSpace
 
    mutable Table *elem_dof; // if NURBS FE space, not owned; otherwise, owned.
    Table *bdrElem_dof; // used only with NURBS FE spaces; not owned.
+   Table *face_dof; // used only with NURBS FE spaces;
+   Array<int> face_to_be; // used only with NURBS FE spaces;
```
pazner, line 116: Do these member data belong in `FiniteElementSpace` or in `NURBSExtension`?

Because they are going to be NULL unless the mesh is a NURBS mesh, placing them in `NURBSExtension` might be more logical?

It may also simplify some of the constructor/destructor logic, which seems to me to be a bit convoluted.

fem/fespace.cpp
```diff
@@ -1788,59 +1828,66 @@ void FiniteElementSpace::GetBdrElementDofs(int i, Array<int> &dofs) const
...
-
-   // for 1D, 2D and 3D faces
-   nv = fec->DofForGeometry(Geometry::POINT);
-   ne = (dim > 1) ? fec->DofForGeometry(Geometry::SEGMENT) : 0;
-   if (nv > 0)
-   {
-      mesh->GetFaceVertices(i, V);
-   }
-   if (ne > 0)
+   if (face_dof)
```
pazner, line 1831: Perhaps this is more a naming/documentation issue, but reading this code is confusing for me.

The `if (face_dof)` branch is to treat the case of NURBS face dofs separately. However, a reader of the code (who may not be aware that the protected member data `face_dof` is only non-NULL for NURBS meshes) may not make the connection with NURBS meshes.

A comment could help here, or perhaps adding a member function with a more descriptive name (e.g. mentioning NURBS) to check if `face_dof` is not NULL.

mesh/mesh.hpp
```diff
@@ -896,6 +896,9 @@ class Mesh
        Return the face index of boundary element i. (3D) */
    int GetBdrElementEdgeIndex(int i) const;
 
+   // Checks if the info required for the above function is available.
+   bool BdrInfoAvailable() const;
+
```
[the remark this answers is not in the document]
IdoAkkerman: The function `GetBdrElementEdgeIndex` requires the boundary datamembers to be generated otherwise it will fail. The initialization of the NURBS mesh/space is unfortunalty a bit convoluted, resulting in the situation that the  `GetBdrElementEdgeIndex` could be called when they are not defined. This prompts the need for this check.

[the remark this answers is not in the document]
IdoAkkerman: I will update the comment. I am open for beter functionnames. Or another "mechanism" alltogether that does not rely on this check/data.

fem/fespace.cpp
```diff
@@ -2024,11 +2073,13 @@ void FiniteElementSpace::Destroy()
    if (NURBSext)
    {
       if (own_ext) { delete NURBSext; }
+      delete face_dof;
    }
    else
    {
       delete elem_dof;
       delete bdrElem_dof;
+      delete face_dof;
```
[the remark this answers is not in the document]
IdoAkkerman: I agree, however outside the if looked strange, as it resulted in one "floating" delete statement.  I guess I imagined the face-dof table to be owned by the NURBSext...
Personally I would like it keep it as is, assuming the face_dof will eventually be moved to NURBSext.

miniapps/nurbs/nurbs_ex1.cpp
```diff
@@ -147,6 +154,12 @@ int main(int argc, char *argv[])
    args.AddOption(&ibp, "-ibp", "--ibp", "-no-ibp",
                   "--no-ibp",
                   "Selects the standard weak form (IBP) or the nonstandard (NO-IBP).");
+   args.AddOption(&strongBC, "-sbc", "--strong-bc", "-wbc",
+                  "--weak-bc",
+                  "Selects strong or weak enforcement of Dirichlet BCs.");
+   args.AddOption(&kappa, "-k", "--kappa",
+                  "One of the two DG penalty parameters, should be positive."
+                  " Negative values are replaced with (order+1)^2.");
```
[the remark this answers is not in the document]
IdoAkkerman: Indeed only one. Will correct the comment.

data/square-nurbs.mesh
IdoAkkerman: I will delete the duplicate files in the miniapps/nurbs directory

fem/fespace.hpp
IdoAkkerman: I have considered them placing them there just as `elem_dof` and `elemBdr_dof`, this would have been required if the `face_dofs` where generated in the NURBSExt (just as the other Tables). However, this is not the case, the current mechanism requires fespace and mesh data in order to generate the `face_dof` table and `face_to_be` mapping.

Perhaps the comment should say : // currently only used with NURBS FE spaces
Both data members could be used by  fespace types other the NURBS in the future.

So I think it is both natural and convienent to have them in the fespace.

IdoAkkerman: I agree it is constructor/destructor convoluted, but it only becomes worse when it is changed IMO.

The NURBS patch topology is technically a mesh, but all knotspans also provide a mesh (more aligned with the standard notion of a mesh). When calling refine, other constructor/update paths are taken. The fespace for the solution can create its own NURBSext or use the one from the `Mesh->Nodes()` fespace. This makes it non-trivial.
Perhaps a mayor overhaul of NURBS initialization is required at some point.

fem/fespace.cpp
IdoAkkerman: In the future other fespace types might use the `face_dof` table, so I think linking it to hard with the NURBS case is not necessary. I would argue the current format is quite OO in the sense that NURBS or not is "hidden" and irrelevant. The only relevant question here is, I want a face dof, can I use the `face_dof` table or not. Moreover the current implementation is 100% analogous with `GetElementDofs` and `GetElementBdrDofs`.

mesh/mesh.cpp
```diff
@@ -3341,6 +3341,12 @@ void Mesh::Loader(std::istream &input, int generate_edges,
          }
       }
    }
+   else if (NURBSext)
+   {
+      {
+         Nodes->FESpace()->GenerateFaceDofsFromBdr();
+      }
+   }
```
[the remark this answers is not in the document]
IdoAkkerman: Removed.

fem/fespace.hpp
```diff
@@ -6,7 +6,7 @@
 // availability visit https://mfem.org.
 //
 // MFEM is free software; you can redistribute it and/or modify it under the
-// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// terms of the BSD-3 license.  We welcome feedback and contributions, see file
```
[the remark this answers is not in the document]
IdoAkkerman: Removed.

fem/fespace.hpp
```diff
@@ -526,10 +528,17 @@ class FiniteElementSpace
        is preserved. */
    void ReorderElementToDofTable();
 
+   /** @brief  Generates partial face_dof table.
+
+       Table only defined for face on boundary. Uses bdrElem_dof table
+       and the mesh boundary information.*/
+   void GenerateFaceDofsFromBdr();
```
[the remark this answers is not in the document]
IdoAkkerman: Changed to:

```
   /** @brief  Generates partial face_dof table.

        The table is only defined for exterior faces that coincide with a boundary.
        The routine uses the bdrElem_dof table and the mesh boundary information.*/
```

fem/fespace.cpp
```diff
@@ -6,7 +6,7 @@
 // availability visit https://mfem.org.
 //
 // MFEM is free software; you can redistribute it and/or modify it under the
-// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// terms of the BSD-3 license.  We welcome feedback and contributions, see file
```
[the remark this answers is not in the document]
IdoAkkerman: Removed.

fem/fespace.cpp
```diff
@@ -1517,6 +1519,43 @@ void FiniteElementSpace::UpdateNURBS()
    ndofs = NURBSext->GetNDof();
    elem_dof = NURBSext->GetElementDofTable();
    bdrElem_dof = NURBSext->GetBdrElementDofTable();
+   delete face_dof;
+   face_dof = NULL;
+
```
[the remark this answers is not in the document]
IdoAkkerman: Removed.

miniapps/nurbs/nurbs_ex1.cpp
```diff
@@ -278,7 +298,15 @@ int main(int argc, char *argv[])
    if (mesh->bdr_attributes.Size())
    {
       Array<int> ess_bdr(mesh->bdr_attributes.Max());
-      ess_bdr = 1;
+      if (strongBC)
+      {
+         ess_bdr    = 1;
```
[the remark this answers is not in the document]
IdoAkkerman: Removed

miniapps/nurbs/nurbs_ex1.cpp
```diff
@@ -278,7 +298,15 @@ int main(int argc, char *argv[])
...
+      }
+      else
+      {
+         ess_bdr    = 0;
```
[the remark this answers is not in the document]
IdoAkkerman: Removed.

miniapps/nurbs/nurbs_ex1.cpp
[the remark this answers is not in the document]
IdoAkkerman: Changed to:

```
   args.AddOption(&kappa, "-k", "--kappa",
                   "Sets the SIPG penalty parameters, should be positive."
                   " Negative values are replaced with (order+1)^2.");
```

miniapps/nurbs/nurbs_ex1p.cpp
```diff
@@ -292,16 +315,29 @@ int main(int argc, char *argv[])
    if (pmesh->bdr_attributes.Size())
    {
       Array<int> ess_bdr(pmesh->bdr_attributes.Max());
-      ess_bdr = 1;
+      if (strongBC)
+      {
+         ess_bdr    = 1;
```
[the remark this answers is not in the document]
IdoAkkerman: Removed

miniapps/nurbs/nurbs_ex1p.cpp
```diff
@@ -143,21 +143,32 @@ int main(int argc, char *argv[])
...
                   " isoparametric space.");
    args.AddOption(&ibp, "-ibp", "--ibp", "-no-ibp",
                   "--no-ibp",
                   "Selects the standard weak form (IBP) or the nonstandard (NO-IBP).");
+   args.AddOption(&strongBC, "-sbc", "--strong-bc", "-wbc",
+                  "--weak-bc",
+                  "Selects strong or weak enforcement of Dirichlet BCs.");
+   args.AddOption(&kappa, "-k", "--kappa",
+                  "One of the two DG penalty parameters, should be positive."
+                  " Negative values are replaced with (order+1)^2.");
```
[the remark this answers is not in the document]
IdoAkkerman: Similar as above.    Changed to:

```
args.AddOption(&kappa, "-k", "--kappa",
                   "Sets the SIPG penalty parameters, should be positive."
                   " Negative values are replaced with (order+1)^2.");
```

data/square-nurbs.mesh
IdoAkkerman: Removed

data/cube-nurbs.mesh
IdoAkkerman: Removed

data/square-disc-nurbs-patch.mesh
IdoAkkerman: Removed

miniapps/nurbs/nurbs_ex1p.cpp
```diff
@@ -292,16 +315,29 @@ int main(int argc, char *argv[])
...
    {
       Array<int> ess_bdr(pmesh->bdr_attributes.Max());
-      ess_bdr = 1;
+      if (strongBC)
+      {
+         ess_bdr    = 1;
+      }
+      else
+      {
+         ess_bdr    = 0;
```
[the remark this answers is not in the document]
IdoAkkerman: Removed

IdoAkkerman:
@pazner Thank you for the detailed feedback. I think I have addressed all points raised, either by making the requested change or by arguing why things are the way they are.
Please let me know if you still have questions or objections.

fem/fespace.cpp
```diff
@@ -1517,6 +1519,42 @@ void FiniteElementSpace::UpdateNURBS()
    ndofs = NURBSext->GetNDof();
    elem_dof = NURBSext->GetElementDofTable();
    bdrElem_dof = NURBSext->GetBdrElementDofTable();
+   delete face_dof;
+   face_dof = NULL;
+}
+
+void FiniteElementSpace::GenerateFaceDofsFromBdr()
```
pazner, line 1526: Calling this public function on a non-NURBS finite element space results in a segfault. It might be good to at least add a `MFEM_VERIFY` check to give an error message.

fem/fespace.cpp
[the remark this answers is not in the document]
pazner: I'm fine with leaving this as it is, but I am confused by your comment. Do you plan on moving `face_dof` to the `NURBSExtension`? I thought you had decided against that?

fem/fespace.hpp
```diff
@@ -526,10 +528,17 @@ class FiniteElementSpace
...
+
+       The table is only defined for exterior faces that coincide with a boundary.
+       The routine uses the bdrElem_dof table and the mesh boundary information.*/
+   void GenerateFaceDofsFromBdr();
+
    void BuildDofToArrays();
 
    const Table &GetElementToDofTable() const { return *elem_dof; }
    const Table &GetBdrElementToDofTable() const { return *bdrElem_dof; }
+   const Table &GetFaceToDofTable() const { return *face_dof; }
```
pazner, line 541: I'm not sure that returning `Table &` instead of `Table *` is good here.

I understand the motivation for keeping the interface for `GetFaceToDofTable` the same as e.g. `GetBdrElementToDofTable` and `GetElementToDofTable`, but this will just crash if you try to use it on a non-NURBS space.

I think a good rule of thumb is that a pointer type represents a "possibly NULL" value, and a reference type represents a "never NULL" value.

In the case of `GetElementToDofTable`, the `elem_dof` table is formed in `FiniteElementSpace::Constructor`, so (possibly barring some edge cases?), `elem_dof` should be non-NULL. (Obviously, my same concern applies to `GetBdrElementToDofTable`, but I guess it is too late to change that interface. On the other hand, it doesn't look like this function is used anywhere in the MFEM codebase, examples, or miniapps. Maybe it can be changed?)

I'm not sure if consistency with existing APIs here outweighs safety, but I think it's worth thinking about. Another option would be to perform some error checking (e.g. `MFEM_VERIFY`).

IdoAkkerman: I understand you point. I added to function just for "symmetry", I don;t think it is used.
So I think it is correct to change it, or it might be removed as well.

fem/fespace.cpp
IdoAkkerman: Do you mean the `UpdateNURBS` function? Or the `GenerateFaceDofsFromBdr`?
The latter should work for non-NURBS??? ALthough I did not check, and don;t directly see a purpose. For the former I will add it for safetly.

pazner: I meant `GenerateFaceDofsFromBdr`. For a non-NURBS space, `bdrElem_dof` will be NULL, and so it will crash on line 1535.

IdoAkkerman: Instead of accessing the `bdrElem_dof `table, I now make a call to` GetBdrElementDofs`, this should work for all meshes.

pazner: It still segfaults for me when you access `bdrElem_dof` on line 1538.

IdoAkkerman: Correct. Replaced access to `bdrElem_dof` and `mesh` for sizes info with appropriate `Get` function.

fem/fespace.hpp
```diff
@@ -526,10 +528,17 @@ class FiniteElementSpace
...
+       The table is only defined for exterior faces that coincide with a boundary.
+       The routine uses the bdrElem_dof table and the mesh boundary information.*/
+   void GenerateFaceDofsFromBdr();
+
    void BuildDofToArrays();
 
    const Table &GetElementToDofTable() const { return *elem_dof; }
-   const Table &GetBdrElementToDofTable() const { return *bdrElem_dof; }
+   const Table *GetBdrElementToDofTable() const { return bdrElem_dof; }
+   const Table *GetFaceToDofTable() const { return face_dof; }
```
v-dobrev, line 541: I don't see a good reason for changing the API for `GetBdrElementToDofTable()`. I know that right now the pointer can be NULL but that is a temporary state (we can just error out in that case) -- in the future the Table can be generated on-demand. Ultimately, we will need this Table, e.g. for PA on boundaries.

Similar logic applies to the `face_dof` Table -- can be generated if requested.

miniapps/nurbs/CMakeLists.txt
```diff
@@ -9,33 +9,47 @@
...
-               COPYONLY)
-
 add_mfem_miniapp(nurbs_ex1
   MAIN nurbs_ex1.cpp
   LIBRARIES mfem)
 
-add_test(NAME nurbs_ex1_ser
-  COMMAND $<TARGET_FILE:nurbs_ex1> -no-vis)
+add_test(NAME nurbs_ex1_r0_o4_ser
+  COMMAND $<TARGET_FILE:nurbs_ex1> -no-vis -r 0 -o 4)
```
v-dobrev, line 17: Can you add all the new tests to the GNU makefile too? I can help with that if you don't feel comfortable with make. 😄

fem/fespace.hpp
pazner: I agree that generating the tables on demand could be a good solution, and probably better than changing the API (that being said, I believe this function is extremely rarely if ever used in user code).

Erroring out is also fine, except it gives the caller no way to recover, and there isn't a way to query whether the tables are non-NULL.

v-dobrev: Lazy generation for the non-NURBS case is trivial in both cases (very similar to `FiniteElementSpace::BuildElementToDofTable`). I can add it, if you guys are fine with that.

IdoAkkerman: I am fine with that.

miniapps/nurbs/CMakeLists.txt
IdoAkkerman: I'll have a look and let you know if I am not able to do it.

IdoAkkerman: I have added the test, but did not test...

IdoAkkerman: Was able to test the serial test and they seemed to work.

v-dobrev: Thanks! I'll give it a try too.

v-dobrev:
@IdoAkkerman, @pazner, @acfisher, I made some changes in the branch -- described in some detail in the commit message here: e89fb16. Take a look and let me know if you have any comments.

IdoAkkerman:
@v-dobrev you seemed to have three(?) birds with one stone. Looks like a excellent improvement!

v-dobrev:

v-dobrev:
LGTM

v-dobrev:
Re-merged in `next`.

PR #1423, opened 2020-04-16
[WIP] p and hp refinement support [p-refinement-dev]

jakubcerveny:
Work in progress on variable order FE collections and spaces. 

Early comments welcome.

TODO: 
- [x] Variable order FiniteElementCollection and H1_FiniteElementCollection
- [x] Multiple sets of DOFs on edges and faces
- [x] p-FEM constraints in P matrix in 2D
- [x] Variable order L2_FiniteElementCollection
- [x] Visualize variable order solution (prolong to max order for the time being)
- [x] hp-FEM constraints in P matrix in 2D
- [x] Variable orders in 3D (faces)
- [x] hp cases in 3D
- [ ] hp demo miniapp - get correct convergence on benchmark problems
- [ ] Save/load variable order Space
- [ ] Update GLVis for var. order spaces?
- [ ] Parallel space
- [ ] Update ND and RT collections
- [ ] Other updates/fixes in the collections
- [ ] Update CHANGELOG


jakubcerveny:
p refinement seems to work in 2D:
pfem-orders
Solution:
pfem-solution

jakubcerveny:
Modest first attempts at hp-refinement (the easy case):
hp-mesh
hp-sln

jakubcerveny:
The hard case (slave edges are lower-order) of course has cracks:
hard1
Although when increasing the order (e.g., 3rd + 2nd) they quickly become less noticeable:
hard2

v-dobrev:
The pictures look great! 🎉

Regarding the cracks: these maybe just fine -- did you try to test convergence with uniform refinement of such mesh?

jakubcerveny:
I think I figured out a simple way to handle the "hard" case, which might not be that hard after all. Since in p-FEM we already have to handle multiple sets of DOFs on each edge (several elements of different orders can meet at an edge, hence the new table `edge_dof`), and already in p-FEM the lowest order DOFs constrain all the higher order variants within the same edge, we can do something similar in the hp-case. A master edge wil have its natural DOFs assigned but also DOFs corresponding to the lowest degree slave (but still for the complete edge). The lowest order DOFs on the master edge will be the true DOFs (it doesn't matter they don't correspond to any element), that will contrain both the higher order DOFs on the master edge (as in p-FEM) and all the slave DOFs on the refined side.

jakubcerveny:
So it seems to work. The deal.ii paper probably confused us unnecessarily. It's all really simple in the end: an edge can have multiple sets of DOFs assigned. If it's a conforming mesh, the lowest order DOFs are the true DOFs and the higher order sets interpolate the true DOFs. In case of hp-refinement, we check the orders of slave elements and add the lowest order DOFs to the master edge, if lower than the standard master edge DOFs (the master edge will thus have two sets of DOFs, even though it's directly shared by one element only (in 2D)). The lowest order DOFs are then interpolated by both the other master edge DOFs and all the slave DOFs. 

Random mesh and orders in ex1:
hp-ex1-a
hp-ex1-b
One of the basis functions:
hp-ex1-c

jakubcerveny:
Since it's a simple matter of adding or not adding the lowest slave order to the master face DOFs, I added an option to `FiniteElementSpace` for the user to choose if they want strict conformity, or if the relaxed (discontinous) interpolation is OK:

`fespace->SetRelaxedHpConformity(false)` (the default):
hp-strict

`fespace->SetRelaxedHpConformity(true)` (same mesh and orders):
hp-relaxed

tzanio:
> `fespace->SetRelaxedHpConformity(false)` (the default):

This solution doesn't looks quite right, or at least it looks worse than the second one?

v-dobrev:
Awesome job, @jakubcerveny!

I don't quite follow the edge->dofs associations you describe: I guess I don't understand how the variable order case is supported even on a conforming mesh. One approach I see, is to introduce cuts in the space along edges/faces where the orders differ (even if the mesh itself is not cut). This way the extraction of element dofs from a `GridFunction` does not require any transformation/interpolation, just extraction of a sub-vector. The interpolation becomes part of the conforming prolongation -- just like the non-conforming case with fixed order (the difference being that the mesh in that case it topologically cut). Is that how the variable order case is handled? Maybe the edge->dofs association you mentioned is the data you need in the FE space to construct the variable-order prolongation?

By the way, introducing cuts in the FE space independent of the mesh can allow us to handle the nodes of a periodic mesh with less data duplication, compared to using an L2 space (which is cut everywhere).

> This solution doesn't looks quite right, or at least it looks worse than the second one?

I agree, it does look a little strange but maybe it is just "too constraint". The simpler (to implement) "RelaxedHpConformity" option maybe also be the better (for the solution) option in this case.

jakubcerveny:
Yes, I introduce cuts on edges and faces if the orders of the adjacent elements differ and then add interpolation constraints in the P matrix. I reuse the machinery already present for NC meshes, I just needed to add one block of code in `BuildConformingInterpolation()`.

The solution I posted earlier is indeed ugly because the space is ugly - a linear element is constraining two large edges to be linear. I'm convinced the solution is correct and I plan to post the pictures of all the basis functions to prove it.

jakubcerveny:
Here is one more example, with basis functions.

The mesh looks like this:
hptest-ord

The fully H1 conforming solution has dimension 36 and looks like this:
hptest-sol

The 36 basis functions are here:
hptest-bases

When relaxing the constraints the dimension goes up to 39 and the large edges are no longer linear, but cracks appear:
hp-relaxed

fem/fespace.hpp
```diff
@@ -150,27 +157,54 @@ class FiniteElementSpace
    mutable Array<FaceQuadratureInterpolator*> E2IFQ_array;
    mutable Array<FaceQuadratureInterpolator*> E2BFQ_array;
 
-   long sequence; // should match Mesh::GetSequence
+   long sequence; // to detect changes in the mesh: should match Mesh::GetSequence
+   bool orders_changed; // space needs updating (rebuilding) if true
+
+   bool relaxed_hp; // see SetRelaxedHpConformity()
```
pazner, line 163: `relaxed_hp` is not initialized in some of the constructors of `FiniteElementSpace`, which caused me some undefined/nondeterministic behavior.

fem/fespace.hpp
```diff
@@ -308,6 +345,17 @@ class FiniteElementSpace
    bool Conforming() const { return mesh->Conforming(); }
    bool Nonconforming() const { return mesh->Nonconforming(); }
```
pazner, line 346: Should `FiniteElementSpace::Conforming()` return `false` if the space is variable order?

fem/fespace.hpp
jakubcerveny: Thanks, I'll fix that.

fem/fespace.hpp
jakubcerveny: I'm not sure. The P matrix with the p-interface constraints can be built for conforming meshes too. Then you can start doing nonconforming refinements, the P will be rebuilt and will continue to work. But maybe for the sake of simplicity we could require that the mesh is nonconforming when you want to have a variable order space, to limit the number of combinations there can be (C/NC mesh, constant/variable order space).

pazner: I agree that doing something like `EnsureNCMesh` when the space is variable order could be a simple solution here.

jakubcerveny:
It seems 3D constraints won't be as simple as I thought. Adding lower order face DOFs according to the lowest order slave is OK, but the multiple sets cause trouble if there are indirect constraints. 

In this example there are two master faces, where the horizontal is partially constrained by the vertical one. On the horizontal face, we add order 3 DOFs to constrain the native order 4 DOFs, but these should at the same time be constrained by the vertical face. This constraint is skipped which leads to the discontinuity. I may have to rethink the handling of the strict continuity constraints...

hp3d-problem1
hp3d-problem2

Meanwhile I'm trying to make the relaxed case work correctly, but there is some bug too. I'm pretty certain the relaxed case should work, but I haven't found the bug yet.

jakubcerveny:
After some more debugging and inspecting the dependencies of the DOFs, it's clear that 3D is much more complex. Consider this situation on the middle edge of the top face of the domain:

GLVis_s01

1. the vertical master face has order 3 DOFs
2. the large vertical slave face has order 2 DOFs directly dependent on the order 3 master DOFs (I'm using relaxed conformity here)
3. the top edge holds two sets of DOFs, orders 2 and 3 (since elements of order 2 and 3 meet there)
4. as in 2D, the higher order DOFs (3) on the edge depend on the lowest order DOFs (2), but this creates a circular dependency between the two sets!

Maybe when adding face constraints, we only need to reference the inner face DOFs and skip the edges, which will be handled separately. But in the current implementation this would miss edge-face constraints. I'm still trying to wrap my head around this.

jakubcerveny:
After a discussion with Tzanio I tried to skip edges when processing face constraints, unless the edges lie in the interior of the master face. The idea is to process the edges by edge constraints only, if possible. This helped somewhat, and I can now construct relatively complex P matrices without circular constraints. 

The following case still fails, but at least it indicates that edges and faces need to be treated completely separately.

These are the element orders, purple is cubic and blue is quadratic:

hp3d-problem3

Here is the numbering (IDs) of edges:
hp3d-problem4

The problem is that edge 22 (circled) is quadratic, but it also holds cubic DOFs because of the smaller cubic element (I'm going for full conformity here). However the face constraint (left half of the big master face) constrains the cubic DOFs, not the quadratic one, which is left unconstrained. 

But explicitly doing the face-edge constraints, while limiting the face constraints to interior DOFs only, should fix this problem.

jakubcerveny:
Separating edges when processing face constraints helped. There might still be some problems but it's starting to work...
hp3d-works1
hp3d-works2
hp3d-works3

v-dobrev:
It seems that we need a method that checks for conformity at certain sets of points (maybe random?) on faces and edges. Visual inspection has its limitations -- it will still be very helpful once you know where precicely to look for a problem. 😄 

The pictures here: https://github.com/mfem/mfem/pull/1423#issuecomment-635349331 looked very convincing but I guess there was still an issue.

jakubcerveny:
A conformity checking tool would be nice :-)  So far I managed to reveal problems just by cutting and browsing the volume while having the contours on in GLVis. In fact I had a version of that screenshot above which showed a problem, I just didn't upload it :-)  But it's fixed now with the propagation.

jakubcerveny:
We have some first hp-adaptation results on the 2D inner layer problem. It's not as good as it should be yet (the refinement decisions and the convergence should be better), but I'll post a picture anyway:
hp-layer1

jakubcerveny:
BTW we have encountered the problem that if we add nodal interpolation for those extra master edge/face true DOFs in the R matrix, this messes up the code that determines which true DOFs are essential (`R->BooleanMult`). We had to keep the R matrix as is, and add a third "Q" matrix which is similar but contains interpolation for the special DOFs. Maybe there is a better solution, like nodal-projecting Coefficients directly to the special DOFs, not interpolating them from already interpolated data.

jakubcerveny:
Somehow we can't get proper convergence out of the hp meshes... I wonder if it's because of the simple approach (only two choices (h, p+1), (h/2, (p+1/2))). Also there still might be some problem in the P matrix (see kinks in the curve).

hp-layer-converg

jakubcerveny:
Finally some improvement in the `hptest` miniapp. It seems nodal interpolation is not sufficient to estimate whether to h- or p- refine. Now I do the following:
1. Solve for the current mesh.
2. Solve for a h-refined mesh with orders (p+1)/2.
3. Solve for a p-refined mesh (p+1).
4. Calculate the exact per-element error for the current mesh (to know which elements to refine).
5. Compare per-element error in the (h+1, (p+1)/2) and (h, p+1) meshes, select either h- or p- refinement simply by looking at which error is lower.

This way I get the following mesh which has error 0.00102267 at 43476 DOFs:

hp-layer-43476-0 00102267

Which corresponds roughly to the red dot in the graph from the AMR paper (unfortunately I lost the Hermes2D convergence data, I'll need to rerun it). This is still not as good as the Hermes result, but I guess it is to be expected with such a simplistic strategy. To get a better convergence one has to try to save DOFs by carefully distributing the orders in the h-refinement. We can try to solve once more for (h+1, (p+1)/2 + 1) (or -1?) and test more options.

hp-conv

jakubcerveny:
I'm trying to catch remaining bugs in the 2D and 3D hp spaces. We have a tool, `ex1_hp_check`, which creates a random mesh with random orders >= 2 and solves the Poisson problem with a simple solution, `x^2 + y^2`:
relax1
Element orders:
relax2
Since the exact solution lies in the FE space, the problem is solved exactly. The residual is very small:
relax3
But there is a problem with the `SetRelaxedHpConformity` option. When set to true, the small quadratic edges no longer interpolate the cubic master edge exactly. However, the solution should still lie in the discrete space and the problem should be solved exactly. Unfortunately this is not the case, now the residual is large:
relax4
This is caused by one of the basis functions looking like this:
relax5
This is the large master edge constraining the smaller master edge. Since the basis function is not continuous, the solution will also not be perfectly continuous, no matter what happens on the smaller master edge. I didn't expect this, and thought the basis function is wrong, but in fact the P matrix does what we want on the quadratic slaves - the midpoint values are tied to the cubic function:
relax6
I'm beginning to doubt that the "relaxed" option is correct...

jakubcerveny:
Some more tests regarding the above solution of `x^2 + y^2`.

I solve the problem and calculate the exact L2 error. I also project the exact coefficient to the space, multiply with R and P, and calculate the L2 error. Finally I plug the projection into the linear system and calculate the residual.

For the fully conforming case, I get all zeros, which is correct:
```
FE solution L2 error: 3.3055e-16
Nodal projection L2 error: 1.83885e-16
|Az - B|: 9.75162e-15
```
With `SetRelaxedHpConformity` however, I get the wrong solution, even though the projection lies in the FE space:
```
FE solution L2 error: 0.000336285
Nodal projection L2 error: 1.88453e-16
|Az - B|: 0.0158918
```
The FE solution has a small discontinuity which is even visible on the contour lines in GLVis...

jakubcerveny:
Here are the basis functions that contain the 4th-order master edge, for the relaxed hp mode. They seem correctly constructed in the sense that the quadratic slaves interpolate them 4th-order function values in the right points.

basis1
basis2
basis3
basis4
basis5

What do you think about this @v-dobrev ?

tzanio:
Thinking more about https://github.com/mfem/mfem/pull/1423#issuecomment-656153747, the result is actually not unexpected -- the discretization in the `RelaxedHpConformity` case is non-conforming because the space is not a subspace of H^1. The approximation is probably still OK (needs to be estimated with Strang's Lemmas, cc: @pazner) but we can't expect to recover the exact solution.

Specifically:

1. On continuous level we have `(∇u,∇v) = (f,v)` for all `v` in `U=H^1` with `u=x^2+y^2` and `f=-4`. (What about boundary conditions?)
2. On discrete level we have `(∇u_h,∇v_h) = (f,v_h)` where `u_h`	and `v_h` are in the *conforming* subspace given by `U_h = Range(P)`.
3. If `u` is in `U_h` **and** `U_h` is a subspace of `U`, we do indeed get `u_h=u` since we can simply set `u=u_h` and `v=v_h` in the continuous weak form. This is what happens in the conforming prolongation case.
4. In the `RelaxedHpConformity` case we still have that `u` is in `U_h`, but `U_h` is a subspace of `U`and so we can get only part of the `v_h` orthogonality.

Does this make sense?

pazner:
> The approximation is probably still OK (needs to be estimated with Strang's Lemmas, cc: @pazner) but we can't expect to recover the exact solution.

I haven't thought carefully about this specific case, but it might be necessary to add DG-like penalty terms in order to recover the proper convergence rates. This is actually trivial to do in the current implementation — just add the DG diffusion integrators, and it will work automatically.

v-dobrev:
Another perspective on this: if you print the "energy" of the state `u`, `E(u) = 1/2 a(u,u) - (b,u)` (which is minimized by the solution), you should see that the energy of the non-conforming discrete solution is smaller than the one for the exact solution.

If the non-conforming mesh/space is refined uniformly in `h`, this will introduce more conformity (in particular, the refined space will not contain the coarse space) and the energy of the discrete solution should increase and (probably?) tend to the energy of the exact solution.

jakubcerveny:
Thanks everyone for the comments, I think you are right.

I was trying to calculate the energy for the `x^2 + y^2` problem, but had trouble obtaining a reasonable value because of the nonzero Dirichlet BC. I would probably need to use the uneliminated system.

To make things easier, I used the exact solution `u = x(1-x)y(1-y)`. For this problem we have
`du/dx = y(1-y)(1-2x)`, `du/dy = x(1-x)(1-2y)` and `\Delta u = 2x(x-1) + 2y(y-1) = -f`.

Then `a(u,u) = Integrate[x^2 (1-x)^2 (1-2y)^2 + y^2 (1-y)^2 (1-2x)^2, {x,0,1}, {y,0,1}] = 1/45` and `l(u) = \int -f u = Integrate[-(2x(x-1) + 2y(y-1)) x(1-x) y(1-y), {x,0,1}, {y,0,1}] = 1/45`.

So the energy `E(u) = 1/2 a(u,u) - l(u) = -1/90 = -0.011111111`.

This is exactly what I get from `ex1_hp_check` for the fully conforming case:
```
FE solution L2 error: 3.80965e-17
Nodal projection L2 error: 1.01771e-17
|Az - B|: 8.82835e-16
(1/2 w'Aw - B'w): -0.0111111
```
Many times it works like this with `SetRelaxedHpConformity` as well, but eventually I found a random mesh that fails to recover the exact solution, and the energy really is lower as Veselin says:
```
FE solution L2 error: 4.75943e-05
Nodal projection L2 error: 1.10661e-17
|Az - B|: 0.0044145
(1/2 w'Aw - B'w): -0.0111146
```
(When I refine once, the energy grows to `-0.011113`)

The solution has a visible discontinuity:
GLVis_s01
(the large element has order 6 and the two smaller ones are order 2)

It took me some time to accept that the energy really is lower for this solution (the linear term is to blame), but as Tzanio says, when we derive the weak formulation, we assume the H1 space. The relaxed space is bigger than H1 and accepts discontinuous solutions like this, and also the weak formulation is wrong, because when doing the Green Theorem we don't include the jump terms. We should probably add the DG diffusion integrators as Will suggests.

jakubcerveny:
We have an interesting result from the hp miniapp (we're trying to get proper convergence for two benchmark problems).

Lenka tried the other problem, Babuska's L-shaped domain problem. Here it's critical that the hp adaptive strategy makes good decisions at the corner with the singularity. 

Consider this mesh with order 4 elements around the corner, it has 97 DOFs and total error 0.0365:
lshape_s01

Since we have the exact solution we can test a number of refinement candidates, and the best ones in this particular case (errors are evaluated locally) are two p refinements and one h-refinement with orders distributed like this:
lshape_s03

The problem is that even though the refinements, when evaluated individually, are very good, together they influence each other and the total error in fact grows to 0.0825. If all three refinements were p-refinements to order 5, the error would have been 0.02696. 

This is because of the full conformity and the degradation of the 5-th order edges to 2nd order, due to the h-refined quadratic elements. I believe this is the reason we have been struggling to get nice convergence curves. In the following graph the second best curve ("hp - max 8, more candidates") is the best we could get with full conformity:

lshape_convergence

Then we tried enabling "relaxed conformity" and even though the curve is not perfect (probably because of the oscillations discussed in previous comments), it finally looks like a straight line with a nice slope that's better than simple h-adaptivity with constant p. 

The mesh even looks like what we were used to seeing in Hermes2D -- large high order elements in the smooth regions and gradually smaller elements with lower degrees towards the singularity:
lshape_relaxed

So it seems "relaxed hp" will actually be important in practice and we should make sure we include the necessary interface integrals.

(As a side note, in our old Hermes2D code we didn't have this problem, because the master edge always stayed at its native order. The high order in fact spilled to the adjoining lower degree slaves because the slaves had special constrained edge shapes (this is not possible in MFEM). The mathematicians were suspicious because it wasn't clear what spaces we defined on the individual mesh elements, but it worked well in practice.)

v-dobrev:
@jakubcerveny, it seems like it may be worth considering if we can support the Hermes2D approach as well, e.g. as a third "conformity" mode. For that, we'll probably need to introduce enriched FEs where a standard p-order element is enriched by edge and face (in 3D) bubble basis functions extended suitably to the whole element.

Do you think this will complicate things a lot?

tzanio:
If I understand correctly the "Hermes" approach will require us to to introduce finite elements that support different polynomial orders on each of their faces and edges. This sounds disruptive to me, but I agree we should keep it in mind for the future if the current two approaches run into difficulties.

jakubcerveny:
I reworked the DOF assignment on edges and faces and finally all cases now work in 3D -- tested thousands of random configurations like this:
hp3d-check1
```
FE solution L2 error: 4.00109e-14
Nodal projection L2 error: 4.43681e-16
|Az - B|: 1.55111e-13
```
hp3d-check2

jakubcerveny:
Lenka cleaned up the `hptest` miniapp and ran both problems (L-shape, inner layer) again with the best strategy so far. It finally looks reasonable, even with the occasional kinks caused by the `relaxed_hp` oscillations.

hptest-lshape

hptest-layer

fem/fe_coll.cpp
```diff
@@ -1927,82 +1981,87 @@ L2_FECollection::L2_FECollection(const int p, const int dim, const int btype,
          snprintf(d_name, 32, "%s_T%d_%dD_P%d", prefix, btype, dim, p);
    }
 
+   InitOrder(default_p);
+}
+
+
+bool L2_FECollection::HaveOrder(int p) const
+{
+   const auto &array = L2_Elements[Geometry::POINT];
```
dylan-copeland, line 1990: This function has some problems. First, `L2_Elements[Geometry::POINT]` is not even initialized unless `InitOrder` has been called for `dim = 0`. The constructor should initialize all entries of `L2_Elements` to `NULL`. Second, `L2_Elements[Geometry::POINT]` is set by `InitOrder` only for `dim=0`. Consequently, `InitOrder` gets called multiple times for `dim > 0`, resulting in memory leaks. The array to be checked should depend on the dimension. The same issues are also in `H1_FECollection::HaveOrder`.

jakubcerveny: Hi Dylan, the treatment of variable order collections in this branch is obsolete. I will eventually be merging in the branch `var-order-space-dev`, where the handling of variable order is much simplified. You may actually want to switch to using `var-order-space-dev` which is more mature and will hopefully be merged into `master` soon.

v-dobrev:
This looks like a mistake, why was this merged in `master`?

tzanio:
shoot... wrong PR

v-dobrev:
Was this "merged" because the last commint here is part of #1817?

tzanio:
As @v-dobrev said, it looks like the commits here were subset of #1817 and that's why GitHub lists it as merged.

Can we delete this branch and PR now?

jakubcerveny:
The only thing left of value here is the hpfem miniapp (and the comments in the PR, but those will still be accessible). I put the miniapp in a separate branch, `hpfem-miniapp-dev`. I'll delete this branch now.

adam-sim-dev:
Dear developers, any progress of the parallel version variable order space?

PR #1449, opened 2020-04-28
Batch-LU

artv3:
This PR adds a device friendly batch LU routine as mentioned in PR https://github.com/mfem/mfem/pull/1340. 

I branched of https://github.com/mfem/mfem/pull/1340 since the main application is LU decomposition of block diagonal matrices, although it could be made to work independently.


artv3:
Will make this independent...

PR #1499, opened 2020-05-21
[WIP] Adding specialization for HCurl basis in ProjectBdrCoefficientTangent [hcurl-bdr-dev]

mlstowell:
This PR avoids the calls to `AccumulateAndCountBdrTangentValues` and `ComputeMeans` which are not necessary when using Nedelec basis functions.


mlstowell:
Hey, @jakubcerveny , I wonder if you could take a quick look at this and let me know your thoughts?  My thought was that Nedelec basis functions don't require any special averaging on boundary faces to perform the projection.  Clearly, a vector H1 field would require special care at shared edges and nodes on the boundary.  What I'm unsure about is the role of the `NCMesh` and its boundary closure.  Is this only needed for vector H1 or is it also necessary for ND bases?

jakubcerveny:
The boundary closure is needed for cases like this:
hidden-edge
Here the edge AB is not a boundary edge, but it's a master to AC and CB and contains the actual true DOFs (which might also be essential DOFs). It's really annoying but it has to be handled, in H1 and probably in Hcurl and Hdiv too...

jakubcerveny:
Perhaps we could devise a way to propagate the boundary data to the hidden edges algebraically, through the P matrix... that would be a huge simplification in functions like GetEssentialTrueDofs etc.

mlstowell:
Thanks, @jakubcerveny , for the illustration.  That makes perfect sense.

mlstowell:
I think the change would appear in the R matrix.  Currently, I suspect, the row entries for edge AB in R simply copy the edge AB entries from the DoF vector and ignore the AC and BC entries.  Perhaps we could change this so that the edge AB DoFs are ignored and the TDoFs for edge AB are computed as a linear combination of the DoFs on edges AC and BC.

PR #1505, opened 2020-05-22
Add a DeviceVector class that executes on Device by default.

YohannDudouit:
For backward compatibility and performance reasons it was decided that the `Vector` class executes on CPU by default even when MFEM is ran on Device, it is required to use `UseDevice(true)` on the Vector to have it run on GPU.
In practice `UseDevice(true)` is often omitted/forgotten, and developers that were not part of the core team during the GPU port are unaware of the importance to call this method, resulting in __unsafe__ and __slow__ code because the `Vector` code unexpectedly execute partially or fully on CPU.
This PR proposes to add a very simple class `DeviceVector` which is nothing more than a `Vector` that executes on Device by default.
Since `DeviceVector` simply inherit from `Vector` they can be used wherever `Vector` is used, without the need to change anything in MFEM.
I believe it is much more natural and less bug prone for a developper to specify the intent for a `Vector` in the type, rather than using the `UseDevice(true)` method call.
The logic for this `DeviceVector` class could also be extended to the `Array` class into a `DeviceArray` class.

### DIFF
--- a/linalg/vector.hpp
+++ b/linalg/vector.hpp
@@ -377,6 +377,44 @@ public:
 #endif
 };
 
+/// This class is nothing more than a Vector that executes on Device by default.
+class DeviceVector : public Vector
+{
+public:
+   /// Default constructor for DeviceVector. Sets size = 0 and data = NULL.
+   DeviceVector() : Vector()
+   {
+      this->UseDevice(true);
+   }
+
+   /// Copy constructor. Allocates a new data array and copies the data.
+   DeviceVector(const Vector &v) : Vector(v)
+   {
+      this->UseDevice(true);
+   }
+
+   /// @brief Creates vector of size s.
+   /// @warning Entries are not initialized to zero!
+   explicit DeviceVector(int s) : Vector(s)
+   {
+      this->UseDevice(true);
+   }
+
+   /// Creates a vector referencing an array of doubles, owned by someone else.
+   /** The pointer @a _data can be NULL. The data array can be replaced later
+       with SetData(). */
+   DeviceVector(double *_data, int _size) : Vector(_data,_size)
+   {
+      this->UseDevice(true);
+   }
+
+   /// Create a DeviceVector of size @a size_ using MemoryType @a mt.
+   DeviceVector(int size_, MemoryType mt) : Vector(size_,mt)
+   {
+      this->UseDevice(true);
+   }
+};
+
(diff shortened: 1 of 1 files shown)


tzanio:
Changed the "ready-for-review" label to "wip" according to the discussion this Friday.

rw-anderson:
Embedding this in the type would seem to inhibit the ability to write a single implementation for which the decision to use the device or not is a runtime decision.

One possible alternative to reduce "forgot to add method" errors: a static method which changes the default from false to true wrt UseDevice.  Then you don't have to add UseDevice calls everywhere, you only have to add it when you'd like to inhibit use of the device.

YohannDudouit:
@rw-anderson a `DeviceVector` __is a__ `Vector`, so you can write all your function with `Vector`s, it's only when you declare the `Vector` that you could use a `DeviceVector` to specify your intent for the `DeviceVector` to go on Device by default, but any function that takes a `Vector`, by definition, works fine with a `DeviceVector`.
For instance:
```c++
DeviceVector x(nbdofs), y(nbdofs);
y = 0.0; // Done on device with DeviceVector, but on host with Vector (most common perf issue)
A.AddMult(x, y); // the signature is Mult(const Vector &x, Vector &y)
```
Currently you have to do:
```c++
Vector x(nbdofs), y(nbdofs);
y.UseDevice(true);
y = 0.0;
A.AddMult(x, y);
```

YohannDudouit:
@rw-anderson Also if you use `UseDevice(true)` it only runs on device if you also use `-d cuda` for instance, the `UseDevice(true)` work the same as `DeviceVector`, you just express an intent. I was just thinking it's less bug prone to do:
```c++
class Foo
{
  DeviceVector data;
public:
  Foo(const int size): data(size) { } 
};
```
rather than:
```c++
class Foo
{
  Vector data;
public:
  Foo(const int size): data(size) { data.UseDevice(true); } 
};
```

YohannDudouit:
@rw-anderson  I unfortunately don't think we can change the default, because there is many small vectors in MFEM for which we don't want them to go on Device. You would need to do the opposite operation: `UseDevice(false)` everywhere in the legacy code, when you want a `Vector` to stay on the host.

YohannDudouit:
An alternative though, could be to add some constructors to the `Vector` class, where we specify in the constructor what behavior we expect for the `Vector`.

PR #1506, opened 2020-05-23
Full Assembly built on top of Element Assembly

YohannDudouit:
This PR uses the element and face matrices from Element Assembly to assemble a global sparse matrix corresponding to a CG or DG finite element operator. This enables sparse matrix assembly on GPU!
- [x] full assembly for CG
- [x] full assembly for DG
- [x] find an `AssemblyLevel::` name for the legacy full assembly, the legacy full assembly is currently named `LEGACYFULL` and the new full assembly is called `FULL`
- [x] Update CHANGELOG
- [x] Unit tests

| [#1506](https://github.com/mfem/mfem/pull/1506) | @YohannDudouit | @tzanio | @pazner + @barker29 + @camierjs | 05/25/20 | 06/30/20 | 07/14/20 | |


artv3:
Nice! Do you have any node to node performance comparisons 36 MPI vs 4 GPUs?

examples/ex1.cpp
```diff
@@ -168,7 +177,10 @@ int main(int argc, char *argv[])
    //    domain integrator.
    BilinearForm *a = new BilinearForm(fespace);
    if (pa) { a->SetAssemblyLevel(AssemblyLevel::PARTIAL); }
-   a->AddDomainIntegrator(new DiffusionIntegrator(one));
+   if (ea) { a->SetAssemblyLevel(AssemblyLevel::ELEMENT); }
+   if (fa) { a->SetAssemblyLevel(AssemblyLevel::FULL); }
+   // a->AddDomainIntegrator(new DiffusionIntegrator(one));
+   a->AddDomainIntegrator(new MassIntegrator(one));
```
artv3, line 183: Look like these changes where just for testing? perhaps move to unit test?

fem/restriction.cpp
```diff
@@ -260,6 +211,310 @@ void ElementRestriction::BooleanMask(Vector& y) const
...
+         for (int j = 0; j < elt_dofs; j++)
+         {
+            const int j_E = e*elt_dofs + j;
+            const int j_L = d_gatherMap[j_E];
+            const int j_offset = d_offsets[j_L];
+            const int j_nextOffset = d_offsets[j_L+1];
+            const int j_nbElts = j_nextOffset - j_offset;
+            if (i_nbElts == 1 || j_nbElts == 1) // no assembly required
+            {
+               mfemAtomicAdd(I[i_L],1);
```
artv3, line 295: mfem::AtomicAdd?

examples/ex9.cpp
```diff
@@ -439,8 +447,9 @@ FE_Evolution::FE_Evolution(BilinearForm &_M, BilinearForm &_K, const Vector &_b)
 {
    bool pa = M.GetAssemblyLevel() == AssemblyLevel::PARTIAL;
    bool ea = M.GetAssemblyLevel() == AssemblyLevel::ELEMENT;
+   bool fa = M.GetAssemblyLevel() == AssemblyLevel::FULL;
    Array<int> ess_tdof_list;
-   if (pa || ea)
+   if (pa || ea || fa)
```
artv3, line 452: This work for ex9p too right?

artv3:
In parallel do you support  the KeepNbrBlock method?

examples/ex1.cpp
YohannDudouit: The change in `ex1` are going to be discarded, I left them temporarily for testing purpose for the reviewers We have to decide how we want to to show this new feature.

fem/restriction.cpp
YohannDudouit: why not, @v-dobrev any opinion on this?

YohannDudouit:
@artv3 The support for `KeepNbrBlock` is not part of this PR.

examples/ex9.cpp
YohannDudouit: Now it does.

YohannDudouit:
I pushed a few commits to fix a bug in parallel, thanks @artv3 for catching it.

tzanio:
why is Travis failing?

YohannDudouit:
@tzanio It's because the test explicitly uses `fa.SetAssemblyLevel(AssemblyLevel::FULL);`, but we need to discuss how we want the two Full Assembly modes live together. The problem comes from the fact that `SetAssemblyLevel(assemblyLevel::FULL)` currently means that full assembly is achieved the _new_ way, however the `SparseMatrix` is given to the `GSSmoother` the _old_ way: `GSSmoother M_fa((SparseMatrix&)(*A_fa));`.
If I remove `fa.SetAssemblyLevel(AssemblyLevel::FULL);`, the test passes.

YohannDudouit:
@tzanio @v-dobrev @camierjs @pazner @barker29 @artv3 
We need to agree on the name for the "legacy" full assembly in `AssemblyLevel`, we discussed with @tzanio and the name `FULL` should be reserved for the _new_ full assembly.

tzanio:
> @tzanio @v-dobrev @camierjs @pazner @barker29 @artv3
> We need to agree on the name for the "legacy" full assembly in `AssemblyLevel`, we discussed with @tzanio and the name `FULL` should be reserved for the _new_ full assembly.

Here are a few suggestions

- `AssemblyLevel::NULL`
- `AssemblyLevel::OLD`
- `AssemblyLevel::LEGACY`

All of these could be set to equal 0 so we can easily check if it was set in the code.

artv3:
> `AssemblyLevel::LEGACY`

I vote for ``AssemblyLevel::LEGACY``. Out of curiosity is it possible to mix Legacy and Full assembly?

YohannDudouit:
@artv3 the assembly level is at the `BilinearForm` level, not at the `Integrator` level, so it is not possible.

YohannDudouit:
I've named the `AssemblyLevel`:`LEGACYFULL`, I'll rename it to whatever is desired when we have a consensus.

fem/bilinearform.hpp
```diff
@@ -29,6 +29,9 @@ namespace mfem
 /// classes derived from Operator.
 enum class AssemblyLevel
 {
+   /// Legacy fully assembled form, i.e. a global sparse matrix in MFEM,
+   /// Hypre or PETSC format.
+   LEGACYFULL,
```
tzanio, line 34:
```suggestion
   LEGACYFULL = 0,
```

YohannDudouit: Just a reminder that strongly typed enum don't automatically cast to `int`, and vice-versa.
Meaning that `int assembly = a->GetAssemblyLevel()` will not compile, and `a->SetAssemblyLevel(0)` neither.

linalg/sparsemat.hpp
```diff
@@ -578,6 +578,16 @@ class SparseMatrix : public AbstractSparseMatrix
    Type GetType() const { return MFEM_SPARSEMAT; }
 };
 
+inline std::ostream& operator<<(std::ostream& os, SparseMatrix const& mat)
+{
+   // HostRead forces synchronization
+   mat.HostReadI();
+   mat.HostReadJ();
+   mat.HostReadData();
+   mat.Print(os);
```
pazner, line 587: Should `SparseMatrix::Print` force synchronization?

YohannDudouit: Yes, otherwise you get random values if assembled on GPU, that happened to me...

YohannDudouit: I see you want to add this synchronization, inside `Print()`, I would agree that it would be safer than doing it only here. @tzanio @v-dobrev What do you think?

examples/ex1.cpp
```diff
@@ -66,6 +66,8 @@ int main(int argc, char *argv[])
    int order = 1;
    bool static_cond = false;
    bool pa = false;
+   bool ea = false;
+   bool fa = false;
```
pazner, line 70: Might want to use an enum here instead of bools since these options are mutually exclusive.

fem/bilinearform.cpp
```diff
@@ -122,8 +122,7 @@ void BilinearForm::SetAssemblyLevel(AssemblyLevel assembly_level)
    switch (assembly)
    {
       case AssemblyLevel::FULL:
```
pazner, line 124: Should calling `SetAssemblyLevel(AssemblyLevel::LEGACYFULL)` be supported?

YohannDudouit: Done.

fem/prestriction.hpp
```diff
@@ -26,24 +26,10 @@ class ParFiniteElementSpace;
...
-   const L2FaceValues m;
-   const int nfdofs;
-   Array<int> scatter_indices1;
-   Array<int> scatter_indices2;
-   Array<int> offsets;
-   Array<int> gather_indices;
-
 public:
-   ParL2FaceRestriction(const ParFiniteElementSpace&, ElementDofOrdering,
+   ParL2FaceRestriction(const FiniteElementSpace&, ElementDofOrdering,
```
pazner, line 32: I think this change can be reverted now too.

examples/ex1.cpp
barker29: I agree with you, but we had this discussion before on #1340 and the powers that be decided on bools.

fem/bilinearform.hpp
```diff
@@ -29,8 +29,10 @@ namespace mfem
 /// classes derived from Operator.
 enum class AssemblyLevel
 {
-   /// Fully assembled form, i.e. a global sparse matrix in MFEM, Hypre or PETSC
-   /// format.
+   /// Legacy fully assembled form, i.e. a global sparse matrix in MFEM,
+   /// Hypre or PETSC format.
+   LEGACYFULL = 0,
+   /// Fully assembled form, i.e. a global sparse matrix in MFEM format.
```
barker29, line 35: In this comment it is not really clear what the difference between LEGACYFULL and FULL is. The name suggests we are going to remove LEGACYFULL, but the comments make it look like that means we will stop supporting Petsc and Hypre matrix formats. These comments should be more clear.

fem/prestriction.cpp
```diff
@@ -27,68 +27,57 @@ ParL2FaceRestriction::ParL2FaceRestriction(const ParFiniteElementSpace &fes,
...
-     scatter_indices1(nf*dof),
-     scatter_indices2(m==L2FaceValues::DoubleValued?nf*dof:0),
-     offsets(ndofs+1),
-     gather_indices((m==L2FaceValues::DoubleValued? 2 : 1)*nf*dof)
+   : L2FaceRestriction(fes, type, m)
 {
    if (nf==0) { return; }
    // If fespace == L2
-   const FiniteElement *fe = fes.GetFE(0);
+   const ParFiniteElementSpace &pfes =
```
barker29, line 34: I don't have any problem with the changes in this file, but can you explain why they are part of this PR?

fem/bilinearform_ext.hpp
```diff
@@ -114,10 +94,12 @@ class EABilinearFormExtension : public PABilinearFormExtension
 protected:
    int ne;
    int elemDofs;
+   // The element matrices are stored row major
    Vector ea_data;
    int nf_int, nf_bdr;
    int faceDofs;
    Vector ea_data_int, ea_data_ext, ea_data_bdr;
+   bool simplify;
```
barker29, line 102: I think a better name (like `is_dg_space`) would make the rest of the code changes more understandable.

fem/restriction.hpp
```diff
@@ -57,6 +57,14 @@ class ElementRestriction : public Operator
        emulate SetSubVector and its transpose on GPUs. This method is running on
        the host, since the `processed` array requires a large shared memory. */
    void BooleanMask(Vector& y) const;
+
+   /// Fill a Sparse Matrix with Element Matrices.
+   void FillSpMat(const Vector &mat_ea, SparseMatrix &mat) const;
```
barker29, line 62:
```suggestion
   void FillSparseMatrix(const Vector &mat_ea, SparseMatrix &mat) const;
```

fem/restriction.cpp
```diff
@@ -1012,6 +1276,116 @@ void L2FaceRestriction::MultTranspose(const Vector& x, Vector& y) const
...
+   {
+      auto d_indices1 = scatter_indices1.Read();
+      auto d_indices2 = scatter_indices2.Read();
+      auto mat_fea = Reshape(fea_data.Read(), face_dofs, face_dofs, 2, nf);
+      auto mat_ea = Reshape(ea_data.ReadWrite(), elem_dofs, elem_dofs, ne);
+      MFEM_FORALL(f, nf,
+      {
+         const int e1 = d_indices1[f*face_dofs]/elem_dofs;
+         const int e2 = d_indices2[f*face_dofs]/elem_dofs;
+         //TODO avoid recomputing iE1,iE2,jE1,jE2 by storing them?
```
barker29, line 1341: Are these still pending TODOs? If not, remove the comments.

fem/restriction.cpp
```diff
@@ -1012,6 +1276,116 @@ void L2FaceRestriction::MultTranspose(const Vector& x, Vector& y) const
...
+   }
+   else
+   {
+      auto d_indices = scatter_indices1.Read();
+      auto mat_fea = Reshape(fea_data.Read(), face_dofs, face_dofs, nf);
+      auto mat_ea = Reshape(ea_data.ReadWrite(), elem_dofs, elem_dofs, ne);
+      MFEM_FORALL(f, nf,
+      {
+         const int e = d_indices[f*face_dofs]/elem_dofs;
+         //TODO avoid recomputing iE,jE
```
barker29, line 1374: Ditto.

fem/restriction.hpp
```diff
@@ -57,6 +57,14 @@ class ElementRestriction : public Operator
...
+
+   /// Fill the I array of a Sparse Matrix
```
barker29, line 64: This comment (and one below it) are not very useful IMO. Maybe say something like `Fill I array of SparseMatrix corresponding to assembled matrix on L-vectors for this ElementRestriction.`

fem/restriction.hpp
```diff
@@ -75,6 +83,10 @@ class L2ElementRestriction : public Operator
    L2ElementRestriction(const FiniteElementSpace&);
    void Mult(const Vector &x, Vector &y) const;
    void MultTranspose(const Vector &x, Vector &y) const;
+   /// Fill the I array of a Sparse Matrix
```
barker29, line 86: See similar comment above.

tests/unit/fem/test_assembly_levels.cpp
```diff
@@ -44,7 +44,8 @@ void AddConvectionIntegrators(BilinearForm &k, VectorCoefficient &velocity,
    }
 }
 
-void test_ea(Mesh &&mesh, int order, bool dg, const int pb)
+void test_assembly_level(Mesh &&mesh, int order, bool dg, const int pb,
```
barker29, line 47: This is a nice simplification/extension of this test.

examples/ex9.cpp
```diff
@@ -144,6 +144,7 @@ int main(int argc, char *argv[])
    int order = 3;
    bool pa = false;
    bool ea = false;
+   bool fa = false;
```
pazner, line 147: This example doesn't work when I run with `-d cuda` but without any of `-pa`, `-ea`, or `-fa`. Not completely sure if this is related to changes in this PR.

fem/restriction.cpp
```diff
@@ -260,6 +211,312 @@ void ElementRestriction::BooleanMask(Vector& y) const
...
+   int ind = mfemAtomicAdd(I[i_L],1);
+   return ind;
+}
+
+void ElementRestriction::FillJandData(const Vector &ea_data,
+                                      SparseMatrix &mat) const
+{
+   const int all_dofs = ndofs;
+   const int vd = vdim;
+   // const bool t = byvdim; //TODO use vd
```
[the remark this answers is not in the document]
YohannDudouit: Yes, I think they are valuable for future optimizations.

fem/restriction.hpp
```diff
@@ -120,12 +132,24 @@ class L2FaceRestriction : public Operator
...
+   void FillI(SparseMatrix &mat) const;
+   /// Fill the J and Data arrays of a Sparse Matrix
+   void FillJandData(const Vector &ea_data,
+                     SparseMatrix &mat) const;
+   /// This methods adds the DG face matrices to the element matrices.
+   void FactorizeBlocks(Vector &fea_data, const int elemDofs,
+                        const int ne, Vector &ea_data) const;
```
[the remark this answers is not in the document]
YohannDudouit: I'm not sure what would be a better name, the idea of this method is to add the Face matrices to the Element matrices in the DG case (add the same non-zeros entries).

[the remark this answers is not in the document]
pazner: My issue is that `FactorizeBlocks` seems to suggest that this is computing e.g. an LU or QR factorization of the matrix blocks.

How about `AddDGFaceBlocks` or something similar?

YohannDudouit:
@tzanio @v-dobrev The reviewers are confused because I left some code in `ex1` and `ex9` for them to test the new Full Assembly feature, can we decide if we want the new Full Assembly to be featured in any example so that I can revert `ex1` and `ex9`?

examples/ex1.cpp
```diff
@@ -116,13 +122,16 @@ int main(int argc, char *argv[])
       }
    }
 
+   std::cout << "The number of elements is: " << mesh->GetNE() << std::endl;
```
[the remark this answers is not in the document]
YohannDudouit: `ex1` and `ex9` will be reverted, this is for you to test the new feature.

examples/ex1.cpp
```diff
@@ -116,13 +122,16 @@ int main(int argc, char *argv[])
...
+
    // 5. Define a finite element space on the mesh. Here we use continuous
    //    Lagrange finite elements of the specified order. If order < 1, we
    //    instead use an isoparametric/isogeometric space.
    FiniteElementCollection *fec;
    if (order > 0)
    {
       fec = new H1_FECollection(order, dim);
+      // fec = new L2_FECollection(order, dim);
```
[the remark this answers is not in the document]
YohannDudouit: idem

examples/ex1.cpp
```diff
@@ -168,7 +177,10 @@ int main(int argc, char *argv[])
    //    domain integrator.
    BilinearForm *a = new BilinearForm(fespace);
    if (pa) { a->SetAssemblyLevel(AssemblyLevel::PARTIAL); }
+   if (ea) { a->SetAssemblyLevel(AssemblyLevel::ELEMENT); }
+   if (fa) { a->SetAssemblyLevel(AssemblyLevel::FULL); }
    a->AddDomainIntegrator(new DiffusionIntegrator(one));
+   // a->AddDomainIntegrator(new MassIntegrator(one));
```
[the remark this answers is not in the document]
YohannDudouit: idem.

examples/ex1.cpp
```diff
@@ -203,11 +215,11 @@ int main(int argc, char *argv[])
       if (UsesTensorBasis(*fespace))
       {
          OperatorJacobiSmoother M(*a, ess_tdof_list);
-         PCG(*A, M, B, X, 1, 400, 1e-12, 0.0);
+         PCG(*A, M, B, X, 1, 200, 1e-12, 0.0);
```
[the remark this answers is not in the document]
YohannDudouit: idem.

examples/ex1.cpp
```diff
@@ -203,11 +215,11 @@ int main(int argc, char *argv[])
...
       }
       else
       {
-         CG(*A, B, X, 1, 400, 1e-12, 0.0);
+         CG(*A, B, X, 1, 200, 1e-12, 0.0);
```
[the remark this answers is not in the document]
YohannDudouit: idem.

fem/prestriction.cpp
YohannDudouit: Because I realized while working on this PR that this code could be simplified and avoid code duplication.

fem/bilinearform_ext.hpp
YohannDudouit: It's always a DG space in that case, so I don't think this is correct to rename to `is_dg_space`, would something like `factorize_face_terms` be more clear?

fem/restriction.cpp
YohannDudouit: They are pending, they are clues for optimization.

fem/restriction.cpp
YohannDudouit: idem.

fem/restriction.hpp
YohannDudouit: What do you think of `Fill the I array of SparseMatrix corresponding to the sparsity pattern given by this ElementRestriction`?

fem/restriction.hpp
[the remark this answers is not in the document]
YohannDudouit: `AddFaceMatricesToElementMatrices`?

fem/bilinearform.hpp
YohannDudouit: Is the missing information that `LEGACYFULL` is only executed on the host for the assembly, whereas `FULL` is compatible with device execution for the assembly?

fem/prestriction.cpp
YohannDudouit: Also, it is not mentioned in the description of the PR, but the `FactorizeBlocks` method has impact on Element Assembly, because "internal" face terms are added to elements matrices, and therefore don't require the extra step of applying them, therefore improving the performance of `Mult` for Element Assembly.

fem/restriction.hpp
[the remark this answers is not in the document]
pazner: Sounds good to me

fem/restriction.cpp
```diff
@@ -260,6 +211,312 @@ void ElementRestriction::BooleanMask(Vector& y) const
...
+      const int nnz = h_I[i];
+      h_I[i] = sum;
+      sum+=nnz;
+   }
+   h_I[nTdofs] = sum;
+   // We return the number of nnz
+   return h_I[nTdofs];
+}
+
+static MFEM_HOST_DEVICE int GetNnzInd(const int i_L, int* I)
```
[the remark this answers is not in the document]
YohannDudouit: Tell me if you like better the new one.

fem/prestriction.cpp
YohannDudouit: I forgot to say that one of the main change is that `ParL2FaceRestriction` now inherit from `L2FaceRestriction`, preventing from duplicating the new methods `FillI`, `FillJAndData`, `AddFaceMatricesToElementMatrices`.

fem/restriction.hpp
YohannDudouit: done.

examples/ex9.cpp
YohannDudouit: Yes.

fem/restriction.cpp
```diff
@@ -260,6 +211,312 @@ void ElementRestriction::BooleanMask(Vector& y) const
...
+   mat.GetMemoryJ().New(nnz, mat.GetMemoryJ().GetMemoryType());
+   mat.GetMemoryData().New(nnz, mat.GetMemoryData().GetMemoryType());
+   FillJandData(mat_ea, mat);
+}
+
+static MFEM_HOST_DEVICE int GetMinElt(const int *my_elts, const int nbElts,
+                                      const int *nbr_elts, const int nbrNbElts)
+{
+   //bulding the intersection
+   constexpr int MaxNbNbr = 16;//TODO remove magic number
```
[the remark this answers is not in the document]
YohannDudouit: Done, I made this an attribute of the class.

fem/restriction.cpp
barker29: Okay, I can't recommend merging into master until this is resolved.

fem/restriction.hpp
barker29: That's fine.

examples/ex9.cpp
YohannDudouit: The problem doesn't come from this PR.

fem/restriction.cpp
YohannDudouit: Okay, I'll remove it all.

fem/restriction.cpp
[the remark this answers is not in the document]
YohannDudouit: I removed them.

fem/bilinearform.hpp
YohannDudouit: I updated the documentation, tell me if it's okay with you.

fem/restriction.hpp
YohannDudouit: Updated, tell me if you like it.

fem/restriction.hpp
YohannDudouit: Updated, tell me if you like it.

YohannDudouit:
@pazner @barker29 I believe I have addressed all your concerns for this PR if you want to take a look.

examples/ex9.cpp
YohannDudouit: `DSmoother` is not yet GPU ready.

pazner: Good catch. I suppose we can always use `OperatorJacobiSmoother` (even in the matrix-based case). But that is not so relevant for this PR.

YohannDudouit: We would have to make `OperatorJacobiSmoother` work in the matrix case first though.

YohannDudouit:
@pazner @barker29 I modified `ex9` so that it works with `OperatorJacobiSmoother` only, like this `./ex9 -d cuda` works. I implemented `GetDiag()` in `SparseMatrix` on device for that.

fem/restriction.cpp
YohannDudouit: done.

linalg/sparsemat.hpp
YohannDudouit: done.

examples/ex1.cpp
YohannDudouit: They're gone.

examples/ex9.cpp
YohannDudouit: done.

YohannDudouit:
@artv3 I think I addressed all your concerns.

fem/restriction.cpp
```diff
@@ -281,6 +232,309 @@ void ElementRestriction::BooleanMask(Vector& y) const
...
+   auto d_offsets = offsets.Read();
+   auto d_indices = indices.Read();
+   auto d_gatherMap = gatherMap.Read();
+   MFEM_FORALL(i_L, vd*all_dofs+1,
+   {
+      I[i_L] = 0;
+   });
+   MFEM_FORALL(e, ne,
+   {
+      constexpr int Max = max;
```
pazner, line 300: Is this really necessary? I would have thought that `static constexpr int max = MaxNbNbr;` would be enough to make sure the lambda capture is OK (even for MSVC).

artv3:
Are you guys able to run ex9p with -fa? ``lrun -n 2 ./ex9p -fa`` segfaults for me
```
*** Error in `./ex9p': malloc(): memory corruption: 0x0000000012dff720 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x97118)[0x20000
```

pazner:
> Are you guys able to run ex9p with -fa? `lrun -n 2 ./ex9p -fa` segfaults for me
> 
> ```
> *** Error in `./ex9p': malloc(): memory corruption: 0x0000000012dff720 ***
> ======= Backtrace: =========
> /lib64/libc.so.6(+0x97118)[0x20000
> ```

Segfaults for me too (only when running with more than one MPI rank)

barker29:
Segfaults for me too, good catch @artv3 

Here's the first bit of what valgrind tells me:
```
==14654== Invalid read of size 4
==14654==    at 0x85B8C8: int mfem::AtomicAdd<int>(int&, int) (cuda.hpp:86)
==14654==    by 0x84CE10: mfem::AddNnz(int, int*, int) (restriction.cpp:512)
==14654==    by 0x851925: mfem::L2FaceRestriction::FillI(mfem::SparseMatrix&) const::{lambda(int)#2}::operator()(int) const (restriction.cpp:1304)
==14654==    by 0x85352B: void mfem::ForallWrap<1, mfem::L2FaceRestriction::FillI(mfem::SparseMatrix&) const::{lambda(int)#1}, mfem::L2FaceRestriction::FillI(mfem::SparseMatrix&) const::{lambda(int)#2}>(bool, int, mfem::L2FaceRestriction::FillI(mfem::SparseMatrix&) const::{lambda(int)#1}&&, mfem::L2FaceRestriction::Fi\
llI(mfem::SparseMatrix&) const::{lambda(int)#2}&&, int, int, int) (forall.hpp:378)
==14654==    by 0x8519F9: mfem::L2FaceRestriction::FillI(mfem::SparseMatrix&) const (restriction.cpp:1310)
==14654==    by 0x93EF22: mfem::FABilinearFormExtension::Assemble() (bilinearform_ext.cpp:640)
==14654==    by 0x7204D1: mfem::BilinearForm::Assemble(int) (bilinearform.cpp:381)
==14654==    by 0x6F2EAF: mfem::ParBilinearForm::Assemble(int) (pbilinearform.cpp:241)
==14654==    by 0x40C2A1: main (ex9p.cpp:353)
==14654==  Address 0x87b4cbc is 8 bytes after a block of size 6,020 alloc'd
==14654==    at 0x4C2A8E8: operator new[](unsigned long) (vg_replace_malloc.c:423)
==14654==    by 0x415E8A: mfem::Memory<int>::Alloc<16ul, true>::New(unsigned long) (mem_manager.hpp:468)
==14654==    by 0x415C61: mfem::Memory<int>::New(int, mfem::MemoryType) (mem_manager.hpp:699)
==14654==    by 0x93EEE7: mfem::FABilinearFormExtension::Assemble() (bilinearform_ext.cpp:636)
==14654==    by 0x7204D1: mfem::BilinearForm::Assemble(int) (bilinearform.cpp:381)
==14654==    by 0x6F2EAF: mfem::ParBilinearForm::Assemble(int) (pbilinearform.cpp:241)
==14654==    by 0x40C2A1: main (ex9p.cpp:353)
==14654==
```

fem/restriction.cpp
YohannDudouit: Let me try it, I tried a couple of different things, and I have to wait 1 hour each time to know if it worked...

fem/bilinearform_ext.cpp
```diff
@@ -651,20 +665,52 @@ void FABilinearFormExtension::Assemble()
...
+         {
+            I_face[i] = I_face[i-1];
+         }
+         I_face[0] = 0;
+      }
+
+      // std::cout << "mat:" << std::endl;
+      // std::cout << mat << std::endl;
+      // std::cout << "face_mat:" << std::endl;
+      // std::cout << face_mat << std::endl;
```
pazner, line 712:
```suggestion
```

fem/bilinearform_ext.cpp
```diff
@@ -677,11 +723,35 @@ void FABilinearFormExtension::Assemble()
 void FABilinearFormExtension::Mult(const Vector &x, Vector &y) const
 {
    mat.Mult(x, y);
+#ifdef MFEM_USE_MPI
+   if (const ParFiniteElementSpace *pfes =
+       dynamic_cast<const ParFiniteElementSpace*>(testFes))
+   {
+      ParGridFunction x_gf;
+      x_gf.MakeRef(const_cast<ParFiniteElementSpace*>(pfes),
+                   const_cast<Vector&>(x), 0);
```
pazner, line 732:
```suggestion
      x_gf.MakeRef(const_cast<ParFiniteElementSpace*>(pfes), x.GetData());
```

fem/bilinearform_ext.cpp
```diff
@@ -677,11 +723,35 @@ void FABilinearFormExtension::Assemble()
...
 void FABilinearFormExtension::MultTranspose(const Vector &x, Vector &y) const
 {
    mat.MultTranspose(x, y);
+#ifdef MFEM_USE_MPI
+   if (const ParFiniteElementSpace *pfes =
+       dynamic_cast<const ParFiniteElementSpace*>(testFes))
+   {
+      ParGridFunction x_gf;
+      x_gf.MakeRef(const_cast<ParFiniteElementSpace*>(pfes),
+                   const_cast<Vector&>(x), 0);
```
pazner, line 749:
```suggestion
      x_gf.MakeRef(const_cast<ParFiniteElementSpace*>(pfes), x.GetData());
```

YohannDudouit:
@artv3 @barker29 @pazner Sorry for this big bug, I believe it is fixed now.

general/cuda.hpp
```diff
@@ -77,6 +77,18 @@ void mfem_cuda_error(cudaError_t err, const char *expr, const char *func,
                      const char *file, int line);
 #endif
 
+template <typename T>
+MFEM_HOST_DEVICE T AtomicAdd(T &add, const T val)
+{
+#if defined(MFEM_USE_CUDA) && defined(__CUDA_ARCH__)
```
artv3, line 83: Minor comment, how far are we from supporting AMD GPUs? Might be good add the hip equivalent of this.

camierjs: We already do, there's this branch that will allow us to get right level of performances:
https://github.com/mfem/mfem/blob/hip-defs/general/hip.hpp

artv3:
The following led to an error  for me:
``lrun -n 2 ./ex9p -fa -d cuda``

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at
[rzansel19:157600] *** Process received signal ***
[rzansel19:157600] Signal: Aborted (6)
[rzansel19:157600] Signal code:  (-6)
<< Rank 1: Generating lwcore_cpu.440061_7.1 on rzansel19 Thu Jun 11 08:53:26 PDT 2020 (LLNL_COREDUMP_FORMAT_CPU=lwcore) >>
terminate called after throwing an instance of 'std::out_of_range
```

YohannDudouit:
Good catch @artv3 , the bug was introduced by simplifying:
```c++
x_gf.MakeRef(const_cast<ParFiniteElementSpace*>(pfes), const_cast<Vector&>(x), 0);
```
to
```c++
x_gf.MakeRef(const_cast<ParFiniteElementSpace*>(pfes), x.GetData());
```
I reverted that change.
However, I think it's dangerous to have methods with the same semantic but different effects... This bug might happen again in the future.

artv3:
> Good catch @artv3 , the bug was introduced by simplifying:
> 
> ```c++
> x_gf.MakeRef(const_cast<ParFiniteElementSpace*>(pfes), const_cast<Vector&>(x), 0);
> ```
> 
> to
> 
> ```c++
> x_gf.MakeRef(const_cast<ParFiniteElementSpace*>(pfes), x.GetData());
> ```
> 
> I reverted that change.
> However, I think it's dangerous to have methods with the same semantic but different effects... This bug might happen again in the future.

Should that be HostReadWrite()? 
On my platform rzansel (lassen) I'm still getting a segfault: 

``lrun -n 2 ./ex9p -fa -d cuda`` leads to: 

 *** Process received signal ***
Signal: Aborted (6)
Signal code:  (-6)
<< Rank 1: Generating lwcore_gpu.440190_3.1 on  Thu Jun 11 13:39:07 PDT 2020 (LLNL_COREDUMP_FORMAT_GPU=lwcore SLOW!) >>
GLVis visualization disabled.

YohannDudouit:
@artv3 Do you have more details because I can't reproduce your bug.

YohannDudouit:
```
[dudouit1@lassen738:examples]$ mpirun -n 2 ./ex9p -fa -d cuda
Options used:
   --mesh ../data/periodic-hexagon.mesh
   --problem 0
   --refine-serial 2
   --refine-parallel 0
   --order 3
   --no-partial-assembly
   --no-element-assembly
   --full-assembly
   --device cuda
   --ode-solver 4
   --t-final 10
   --time-step 0.01
   --visualization
   --no-visit-datafiles
   --no-paraview-datafiles
   --no-adios2-streams
   --ascii-datafiles
   --visualization-steps 5
Number of unknowns: 3072
Unable to connect to GLVis server at localhost:19916
GLVis visualization disabled.
time step: 5, time: 0.05
time step: 10, time: 0.1
time step: 15, time: 0.15
time step: 20, time: 0.2
time step: 25, time: 0.25
time step: 30, time: 0.3
time step: 35, time: 0.35
time step: 40, time: 0.4
time step: 45, time: 0.45
time step: 50, time: 0.5
time step: 55, time: 0.55
time step: 60, time: 0.6
time step: 65, time: 0.65
time step: 70, time: 0.7
time step: 75, time: 0.75
time step: 80, time: 0.8
time step: 85, time: 0.85
time step: 90, time: 0.9
time step: 95, time: 0.95
time step: 100, time: 1
time step: 105, time: 1.05
time step: 110, time: 1.1
time step: 115, time: 1.15
time step: 120, time: 1.2
time step: 125, time: 1.25
time step: 130, time: 1.3
time step: 135, time: 1.35
time step: 140, time: 1.4
time step: 145, time: 1.45
time step: 150, time: 1.5
time step: 155, time: 1.55
time step: 160, time: 1.6
time step: 165, time: 1.65
time step: 170, time: 1.7
time step: 175, time: 1.75
time step: 180, time: 1.8
time step: 185, time: 1.85
time step: 190, time: 1.9
time step: 195, time: 1.95
time step: 200, time: 2
time step: 205, time: 2.05
time step: 210, time: 2.1
time step: 215, time: 2.15
time step: 220, time: 2.2
time step: 225, time: 2.25
time step: 230, time: 2.3
time step: 235, time: 2.35
time step: 240, time: 2.4
time step: 245, time: 2.45
time step: 250, time: 2.5
time step: 255, time: 2.55
time step: 260, time: 2.6
time step: 265, time: 2.65
time step: 270, time: 2.7
time step: 275, time: 2.75
time step: 280, time: 2.8
time step: 285, time: 2.85
time step: 290, time: 2.9
time step: 295, time: 2.95
time step: 300, time: 3
time step: 305, time: 3.05
time step: 310, time: 3.1
time step: 315, time: 3.15
time step: 320, time: 3.2
time step: 325, time: 3.25
time step: 330, time: 3.3
time step: 335, time: 3.35
time step: 340, time: 3.4
time step: 345, time: 3.45
time step: 350, time: 3.5
time step: 355, time: 3.55
time step: 360, time: 3.6
time step: 365, time: 3.65
time step: 370, time: 3.7
time step: 375, time: 3.75
time step: 380, time: 3.8
time step: 385, time: 3.85
time step: 390, time: 3.9
time step: 395, time: 3.95
time step: 400, time: 4
time step: 405, time: 4.05
time step: 410, time: 4.1
time step: 415, time: 4.15
time step: 420, time: 4.2
time step: 425, time: 4.25
time step: 430, time: 4.3
time step: 435, time: 4.35
time step: 440, time: 4.4
time step: 445, time: 4.45
time step: 450, time: 4.5
time step: 455, time: 4.55
time step: 460, time: 4.6
time step: 465, time: 4.65
time step: 470, time: 4.7
time step: 475, time: 4.75
time step: 480, time: 4.8
time step: 485, time: 4.85
time step: 490, time: 4.9
time step: 495, time: 4.95
time step: 500, time: 5
time step: 505, time: 5.05
time step: 510, time: 5.1
time step: 515, time: 5.15
time step: 520, time: 5.2
time step: 525, time: 5.25
time step: 530, time: 5.3
time step: 535, time: 5.35
time step: 540, time: 5.4
time step: 545, time: 5.45
time step: 550, time: 5.5
time step: 555, time: 5.55
time step: 560, time: 5.6
time step: 565, time: 5.65
time step: 570, time: 5.7
time step: 575, time: 5.75
time step: 580, time: 5.8
time step: 585, time: 5.85
time step: 590, time: 5.9
time step: 595, time: 5.95
time step: 600, time: 6
time step: 605, time: 6.05
time step: 610, time: 6.1
time step: 615, time: 6.15
time step: 620, time: 6.2
time step: 625, time: 6.25
time step: 630, time: 6.3
time step: 635, time: 6.35
time step: 640, time: 6.4
time step: 645, time: 6.45
time step: 650, time: 6.5
time step: 655, time: 6.55
time step: 660, time: 6.6
time step: 665, time: 6.65
time step: 670, time: 6.7
time step: 675, time: 6.75
time step: 680, time: 6.8
time step: 685, time: 6.85
time step: 690, time: 6.9
time step: 695, time: 6.95
time step: 700, time: 7
time step: 705, time: 7.05
time step: 710, time: 7.1
time step: 715, time: 7.15
time step: 720, time: 7.2
time step: 725, time: 7.25
time step: 730, time: 7.3
time step: 735, time: 7.35
time step: 740, time: 7.4
time step: 745, time: 7.45
time step: 750, time: 7.5
time step: 755, time: 7.55
time step: 760, time: 7.6
time step: 765, time: 7.65
time step: 770, time: 7.7
time step: 775, time: 7.75
time step: 780, time: 7.8
time step: 785, time: 7.85
time step: 790, time: 7.9
time step: 795, time: 7.95
time step: 800, time: 8
time step: 805, time: 8.05
time step: 810, time: 8.1
time step: 815, time: 8.15
time step: 820, time: 8.2
time step: 825, time: 8.25
time step: 830, time: 8.3
time step: 835, time: 8.35
time step: 840, time: 8.4
time step: 845, time: 8.45
time step: 850, time: 8.5
time step: 855, time: 8.55
time step: 860, time: 8.6
time step: 865, time: 8.65
time step: 870, time: 8.7
time step: 875, time: 8.75
time step: 880, time: 8.8
time step: 885, time: 8.85
time step: 890, time: 8.9
time step: 895, time: 8.95
time step: 900, time: 9
time step: 905, time: 9.05
time step: 910, time: 9.1
time step: 915, time: 9.15
time step: 920, time: 9.2
time step: 925, time: 9.25
time step: 930, time: 9.3
time step: 935, time: 9.35
time step: 940, time: 9.4
time step: 945, time: 9.45
time step: 950, time: 9.5
time step: 955, time: 9.55
time step: 960, time: 9.6
time step: 965, time: 9.65
time step: 970, time: 9.7
time step: 975, time: 9.75
time step: 980, time: 9.8
time step: 985, time: 9.85
time step: 990, time: 9.9
time step: 995, time: 9.95
time step: 1000, time: 10
```

artv3:
Good news everyone! I got the code working. Additionally,  I tried it out with PR https://github.com/mfem/mfem/pull/1489

I used the following settings: 
lrun -n 2  ./ex9p -d cuda -fa -rs 7

Without PR  https://github.com/mfem/mfem/pull/1489 
Run time 10.980155 seconds

Profile looks like this: 
     Type  Time(%)      Time     Calls       Avg       Min       Max  Name
 GPU activities:   81.79%  8.46717s      8617  982.61us  38.815us  1.9714ms  _ZN4mfem10CuKernel1DIZNKS_12SparseMatrix7AddMultERKNS_6VectorERS2_dEUliE_EEviT_

With PR  https://github.com/mfem/mfem/pull/1489 
Run time 5.1951263

Profile looks like this: 
 Type  Time(%)      Time     Calls       Avg       Min       Max  Name
 GPU activities:  55.34%  2.56743s      4308  595.97us  455.55us  633.37us  void cusparse::csrmv_kernel

artv3:
Excited about this work! Some comments/suggestions I have is: 

1. To consider adding a -fa option to example 1p 
2. If ParBilinearForm::KeepNbrBlock() is not supported under -fa, throw an error  
3. As future work it looks like the memory for element matrices could be stored in a temporary pool? Since at the end we just want the Sparse Matrix.

YohannDudouit:
@artv3 
1. I think we want to keep `ex1` and `ex1p` simple, therefore we won't show `-ea` and `-fa` in these.
2. It wouldn't be too hard to implement, I just think we should do first what we talked about DG dofs, then we would have only on `SparseMatrix` and we could easily keep the rows you're interested in after that. Alternatively, we could store easily these rows in a different `SparseMatrix`.
3. It shouldn't be too hard to save some memory if it becomes an issue, that's a good point to keep in mind.

camierjs:
I'm getting 4 warnings:
```==38576== Conditional jump or move depends on uninitialised value(s)
==38576==    at 0x4C2ABCE: operator new[](unsigned long) (vg_replace_malloc.c:433)
==38576==    by 0x416E36: mfem::Memory<int>::Alloc<16ul, true>::New(unsigned long) (mem_manager.hpp:468)
==38576==    by 0x416C0D: mfem::Memory<int>::New(int, mfem::MemoryType) (mem_manager.hpp:698)
==38576==    by 0x9A7F71: mfem::FABilinearFormExtension::Assemble() (bilinearform_ext.cpp:686)
==38576==    by 0x7A1193: mfem::BilinearForm::Assemble(int) (bilinearform.cpp:381)
==38576==    by 0x76F39F: mfem::ParBilinearForm::Assemble(int) (pbilinearform.cpp:241)
==38576==    by 0x40D300: main (ex9p.cpp:355)
```
that seem to come from:
```
#if __cplusplus < 201703L
   template<bool dummy> struct Alloc<def_align_bytes,dummy>
   {
      static inline T *New(std::size_t size) { return new T[size]; }
   };
#endif
```
What was the logic for this `dummy`, @v-dobrev ?

camierjs:
It seems that `h_I_face` in `const int nnz = h_I_face[i];` is not initialized.

YohannDudouit:
@camierjs @v-dobrev My last commit fixed the Valgrind issue. However, I see an issue in `AssembleDiagonal` that I don't really understand:
```
==10725== Use of uninitialised value of size 8
==10725==    at 0x108C380C: mfem::L2ElementRestriction::MultTranspose(mfem::Vector const&, mfem::Vector&) const (mem_manager.hpp:235)
==10725==    by 0x10720143: mfem::PABilinearFormExtension::AssembleDiagonal(mfem::Vector&) const (bilinearform_ext.cpp:134)
==10725==    by 0x1039B563: mfem::BilinearForm::AssembleDiagonal(mfem::Vector&) const (bilinearform.cpp:638)
==10725==    by 0x100F423F: UnknownInlinedFun (solvers.cpp:130)
==10725==    by 0x100F423F: Memory (solvers.cpp:564)
==10725==    by 0x100F423F: mfem::OperatorJacobiSmoother::OperatorJacobiSmoother(mfem::BilinearForm const&, mfem::Array<int> const&, double) (vector.hpp:441)
==10725==    by 0x1002A56F: FE_Evolution::FE_Evolution(mfem::ParBilinearForm&, mfem::ParBilinearForm&, mfem::Vector const&) (ex9p.cpp:599)
==10725==    by 0x1002542B: CopyFrom (ex9p.cpp:478)
==10725==    by 0x1002542B: UnknownInlinedFun (array.hpp:637)
==10725==    by 0x1002542B: main (tmpxft_000028bd_00000000-5_ex9p.cudafe1.cpp:0)
==10725==  Uninitialised value was created by a stack allocation
==10725==    at 0x10BCDB10: mfem::MassIntegrator::AssembleDiagonalPA(mfem::Vector&) (bilininteg_mass_pa.cpp:444)
```

v-dobrev:
> However, I see an unrelated to this PR issue in `AssembleDiagonal` that I don't really understand:

Can you post the command line that reproduces this issue?

YohannDudouit:
@v-dobrev `valgrind ./ex9p -fa`

YohannDudouit:
@artv3 @pazner @camierjs I fixed the `uninitialized value` issue, thank you very much to @v-dobrev for his help, it was a nightmare to find the problem!

pazner:
I am getting a segfault in the unit tests when compiling with gcc 4.8 (but not gcc 4.9).

For example, on quartz after running `module load gcc/4.8-redhat`:
```
H1 pa_coeff
/usr/WS2/pazner1/branches-mfem/yohann/fa/tests/unit/fem/test_pa_coeff.cpp:48

/usr/WS2/pazner1/branches-mfem/yohann/fa/tests/unit/fem/test_pa_coeff.cpp:48: FAILED:
  {Unknown expression after the reported line}
```

It only shows up in release mode (not debug). Is this a bug in gcc 4.8?

pazner:
This segfault seems to happen on the master branch as well (thanks @YohannDudouit for the suggestion to check). Seems more likely that it is a bug in gcc? In any case, it does not appear related to this PR.

v-dobrev:
> This segfault seems to happen on the master branch as well (thanks @YohannDudouit for the suggestion to check). Seems more likely that it is a bug in gcc? In any case, it does not appear related to this PR.

I think this is the same (or similar) bug in gcc 4.8 that we've seen before -- it does not handle lambda functions correctly in some cases. This was the reason I had to switch to gcc 4.9 for testing on my desktop.

tzanio:

artv3:
> @artv3
> 
>     1. I think we want to keep `ex1` and `ex1p` simple, therefore we won't show `-ea` and `-fa` in these.
> 
>     2. It wouldn't be too hard to implement, I just think we should do first what we talked about DG dofs, then we would have only on `SparseMatrix` and we could easily keep the rows you're interested in after that. Alternatively, we could store easily these rows in a different `SparseMatrix`.
> 
>     3. It shouldn't be too hard to save some memory if it becomes an issue, that's a good point to keep in mind.

For the shared neighbors could we have the same behavior as the legacy version? 1 Sparse matrix.

YohannDudouit:
@artv3 I want to, but this requires a discussion first and quite some work that goes beyond this PR I think, we need to discuss about changing the `RAP` operator for DG spaces, because currently the local dofs "L-vector" are not enough to compute the "local" DG operator, therefore the 2 sparse matrices. I couldn't find an easy way around it, because the `FaceNeighborData` are not stored with the dofs.
If we change the `RAP` operator we can have one sparse matrix, and we can also get rid of the `ParL2FaceRestriction`.

examples/ex9.cpp
```diff
@@ -437,21 +448,10 @@ int main(int argc, char *argv[])
...
-   }
-   else
-   {
-      M_prec = new DSmoother(M.SpMat());
-      dg_solver = new DG_Solver(M.SpMat(), K.SpMat(), *M.FESpace());
-      M_solver.SetOperator(M.SpMat());
-   }
+   M_prec = new OperatorJacobiSmoother(M, ess_tdof_list);
+   M_solver.SetOperator(M);
+   dg_solver = NULL;
```
v-dobrev, line 454: This change breaks one of the sample runs, `ex9 -m ../data/amr-quad.mesh -p 1 -r 2 -dt 0.02 -s 13 -tf 9 -no-vis`.

YohannDudouit: Thank you, I'm investigating.

YohannDudouit: @v-dobrev It's fixed.

tzanio:

tzanio:
LGTM

PR #1528, opened 2020-06-08
Fix for superlu_dist 6.3.

pghysels:
Some SuperLU_Dist structures are now prefixed with 'd' (real double) or 'z' (complex double), etc. so that codes can link both real and complex versions of the SuperLU_Dist library.

See https://portal.nersc.gov/project/sparse/superlu/changes.html#slu_dist_change

### DIFF
--- a/linalg/superlu.cpp
+++ b/linalg/superlu.cpp
@@ -24,6 +24,18 @@
 #error "SuperLUDist has been built with 64bit integers. This is not supported"
 #endif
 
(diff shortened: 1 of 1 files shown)


pghysels:
> One question: Did you fix this for MFEM complex version? The current fix seems specific for the real version.

MFEM only uses the real double precision interface

tzanio:

tzanio:
LGTM

PR #1558, opened 2020-06-18
Fixing boundary attributes and 1D mesh support [bugfix/mesh-trimmer]

mlstowell:
The code now loops over boundary faces and interior faces separately so that it can easily copy the pre-existing boundary attributes to the new mesh.

Also added support for 1D meshes by creating new `Point` objects for the new boundary elements rather than duplicating interior face elements which don't exist in 1D.

I briefly looked at adding support for non-conforming meshes but I didn't see a way to build an NCMesh by copying selected elements, etc..  

| [#1558](https://github.com/mfem/mfem/pull/1558) | @mlstowell | @tzanio | @homijan + @delyank | 06/18/20 | 06/26/20 | 07/04/20 | |

### DIFF
--- a/miniapps/meshing/trimmer.cpp
+++ b/miniapps/meshing/trimmer.cpp
@@ -113,7 +113,7 @@ int main(int argc, char *argv[])
 
    // Count the number of boundary elements in the final mesh
    int num_bdr_elements = 0;
-   for (int f=0; f<mesh.GetNFaces(); f++)
+   for (int f=0; f<mesh.GetNumFaces(); f++)
    {
       int e1 = -1, e2 = -1;
       mesh.GetFaceElements(f, &e1, &e2);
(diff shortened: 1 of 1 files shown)


delyank:
It's interesting that this had a pushing need to be used in 1D.

It seems that the change is based on `GetBdrElementAdjacentElement()` and `GetBdrElement()` vs `GetFaceElements()` and `GetBdrElement()`. I didn't know that if the faces on the boundary were not explicitly extracted as boundary elements, this would confuse the boundary attributes, since after all `AddBdrElement()` is used to reintroduce the boundary faces. Kind of makes sense though, since the boundary attributes are separate from the mesh attributes and probably the `Duplicate()` procedure needs to know that the face being duplicated has a boundary "role".

Seems a legitimate change.

mlstowell:
@delyank , I was surprised to see this issue come up as well.  It's great that people are testing all the corner cases and giving us feedback.  It certainly helps to make MFEM more robust.  Thanks for the quick review!

tzanio:

PR #763, opened 2019-02-26
Holger/feature/gmshv4

jonesholger:
This updates gmsh reader for v4 Gmesh (.msh) files. This does not add functionality, i.e. the same element types are supported between v2 and v4, and any deficiencies wrt supporting files generated by Gmesh exist in both. I did remove some redundancies between the two readers, including some inlining vs macro. 

I added some sample .msh files into data/gmsh/v2 and data/gmsh/v4. 

I found some issues reading various simple tutorial meshes, and those files were not added into data, and should likely exist under a new branch supporting a reported issue.

This was originally based off of v3.4, and merged with master and astyled as last steps.

PR #1571, opened 2020-06-24
Adding high order Gmsh support [ho-gmsh-dev]

mlstowell:
This PR adds support for reading high order elements from Gmsh version 2.2 ".msh" files.

The Gmsh  [header](https://gitlab.onelab.info/gmsh/gmsh/-/blob/master/Common/GmshDefines.h) declares segments, triangles, quadrilaterals, and tetrahedra up to order 10 and wedges, pyramids, and hexahedra up to order 9.  This PR supports all of these types.

More testing is required but I believe it works for periodic as well as standard meshes.

Have modified the `periodic-annulus-sector.geo` and `periodic-torus-sector.geo` files to produce high order meshes.  With these scripts we can generate high order triangular, quadrilateral, or tetrahedral meshes which are periodic or not (it even works on 2D mixed meshes).  I was not able to build a high order periodic hexahedral mesh for testing but I did test the non-periodic case.  If anyone who is more adept with Gmsh than I would like to include a script for a high order periodic hex mesh I would be grateful.  High order prisms in non-periodic meshes have also been tested.  The pyramid support seems to produce the correct mappings up to order 4 (I didn't test beyond this) but MFEM cannot yet produce pyramid shaped elements so further testing was not possible.

| [#1571](https://github.com/mfem/mfem/pull/1571) | @mlstowell | @tzanio | @tzanio + @bslazarov | 06/27/20 | 07/30/20 | 08/18/20 | |


tzanio:
@mlstowell, which version of Gmsh will this work with? 

They just released 4.6.0. 

See also https://github.com/mfem/mfem/pull/763

mlstowell:
@tzanio , I have been working with Gmsh 4.5.6 and saving the `.msh` files as version 2.2 ASCII.

tzanio:
Given that the version 2 format is listed as [legacy](https://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format-version-2-_0028Legacy_0029) does it make sense to switch to the [new format](https://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format) for this functionality?

mlstowell:
@tzanio , the main difference between the two formats seems to be that 4.x provides partitioning and ghost element information.  Since this is a serial mesh reader we wouldn't be able to use the partitioning information in the newer format.  

I would suggest that we work towards supporting both formats.  Perhaps by revisiting PR #763 (which I was unaware of until you mentioned it).  However, I still think this PR makes sense on its own.

tzanio:
LGTM

mlstowell:
@tzanio , should I put this back as WIP until we sort out the higher order mappings?

tzanio:
> @tzanio , should I put this back as WIP until we sort out the higher order mappings?

I'm OK either way. If you do that, also remove the `in-review` label.

mlstowell:
@tzanio and @bslazarov , I just finished adding support for element orders up to 9 or 10, based on Gmsh limitations, so I have switched this PR back to `in-review` status.  Let me know if you have any concerns.

tzanio:
That's great, thank you @mlstowell!

bslazarov:
@mlstowell - I will add a periodic script for HO regular mesh. However it will be towards the end of this week.

mlstowell:
That would be great!  Thanks, @bslazarov.

tzanio:
LGTM

bslazarov:
@mlstowell  Could you please check the following gmsh input and let me know if you like it?

lc = 0.5;
Point(1) = {1, 0, 0, lc};
Point(2) = {0, 1, 0, lc};
Point(3) = {-1, 0, 0, lc};
Point(4) = {0, -1, 0, lc};
Point(5) = {0, 0, 0, lc};
Circle(1) = {1,5,2};
Circle(2) = {2,5,3};
Circle(3) = {3,5,4};
Circle(4) = {4,5,1};
Line Loop(100) = {1,2,3,4};
Plane Surface(200) = {100};
Recombine Surface {200} ;


//h = 1;

//Extrude {0,0,h} {
//  Surface{200}; Layers{ {8,2}, {0.5,1}}; Recombine;
//}

Extrude { {0,1,0} , {-1,0,0.5} , -Pi/4} {
  Surface{200}; Layers{7}; Recombine;
}

Periodic Surface{222} = {200} Rotate { {0,1,0} , {-1,0,0.5} , -Pi/4};

mlstowell:
@bslazarov, sure I like it! It's about the right size and it's an interesting shape.  The periodicity isn't working in my version of Gmsh but perhaps there's a workaround or newer version that I should be using.

bslazarov:
The geo file and 3rd order mesh are added to the repo - fell free to edit them. 

I tested the code using ex1p and everything seems to be fine. For some reason I could not generate mesh with element order higher than 8. If I push the order 9 or 10 several elements will have negative Jacobian.

mlstowell:
@bslazarov , thanks for the Gmsh script.  That was a big help!  You have no idea how long I struggled to make a high order periodic hex mesh.  You make it look so easy.

I incorporated your ideas into the `periodic-torus-sector.geo` file which is now capable of creating periodic or standard meshes of a portion of a torus with arbitrary order tetrahedra, wedges, or hexahedra.  Between this and the 2D `periodic-annulus-sector.geo` script I think we're demonstrating everything I had hoped to show.

I'm not sure what to do with your `periodic-sector.geo` script at this point.  I'll be happy to merge it into master if you want to do that (it is a cool shape!) but it's not strictly necessary now that `periodic-torus-sector.geo` does what I wanted.

bslazarov:
@mlstowell Just remove the geo file if it is redundant. I don't think that we should replicate the same functionality on two places. One can always modify the rotation to get a  skewed shape/volume for the torus sector.

mlstowell:
@tzanio, I believe this PR is ready whenever you get a chance.

tzanio:
LGTM

tzanio:
Please update `CHANGELOG`

mesh/mesh_readers.cpp
```diff
@@ -970,67 +984,298 @@ void Mesh::ReadGmshMesh(std::istream &input, int &curved, int &read_gf)
...
+            16,  /* 16-node third order quadrilateral (4 nodes associated with
+                       the vertices, 8 with the edges, 4 wth the face) */
+            25,  /* 25-node fourth order quadrilateral (4 nodes associated with
+                       the vertices, 12 with the edges, 9 wth the face) */
+            36,  /* 36-node fifth order quadrilateral (4 nodes associated with
+                       the vertices, 16 with the edges, 16 wth the face) */
+            -1,-1,-1, /* unsupported quadrilateral types */
+            28,  /* 28-node sixth order complete triangle (3 nodes associated
+                       with the vertices, 15 with the edges, 10 with the face)
+                 */
```
tzanio, line 1057: Comments on row like this (and maybe in general for this array) may look better with `//` instead of `/* ... */`:

```c++
            28,  // 28-node sixth order complete triangle (3 nodes associated
                 //    with the vertices, 15 with the edges, 10 with the face)
```

tzanio: I reformatted this section a bit -- let me know if you want me to revert it...

mesh/mesh_readers.cpp
```diff
@@ -1088,44 +1333,172 @@ void Mesh::ReadGmshMesh(std::istream &input, int &curved, int &read_gf)
...
+                           if (el_order > 1)
+                           {
+                              Array<int> * hov = new Array<int>;
+                              hov->Append(&vert_indices[0], n_elem_nodes);
+                              ho_verts_3D.push_back(hov);
+                              ho_el_order_3D.push_back(el_order);
+                           }
+                           break;
+                        }
+                     */
```
tzanio, line 1501: Do we want to keep the above commented code?

mesh/mesh_readers.cpp
```diff
@@ -1179,44 +1552,172 @@ void Mesh::ReadGmshMesh(std::istream &input, int &curved, int &read_gf)
...
+                        if (el_order > 1)
+                        {
+                           Array<int> * hov = new Array<int>;
+                           hov->Append(&vert_indices[0], n_elem_nodes);
+                           ho_verts_3D.push_back(hov);
+                           ho_el_order_3D.push_back(el_order);
+                        }
+                        break;
+                     }
+                  */
```
tzanio, line 1720: Do we want to keep the above commented code?

tzanio: If yes, can we add something like `// Gmsh pyramids not supported yet`?

mesh/mesh_readers.cpp
mlstowell: Yes, but we should indeed add a comment explaining why it's commented out.

mesh/mesh_readers.cpp
mlstowell: Either format is fine with me.  I'll be happy to change it if you like.

tzanio:
Should we include (maybe commented out) version of the Gmsh's high-order mesh optimization options from [tutorial/t5.geo](https://gitlab.onelab.info/gmsh/gmsh/-/blob/master/tutorial/t5.geo#L202)?

```
   // Mesh.ElementOrder = order;
   // Mesh.HighOrderOptimize = 2;
```

mesh/mesh_readers.cpp
```diff
@@ -917,6 +918,19 @@ void Mesh::ReadGmshMesh(std::istream &input, int &curved, int &read_gf)
    // (there may be gaps in the numbering, and also Gmsh enumerates vertices
    // starting from 1, not 0)
    map<int, int> vertices_map;
+
+   // Gmsh always outputs coordinates in 3D
+   spaceDim = 3;
```
tzanio, line 923: This causes 2D meshes to be read as surface meshes in MFEM, which is not what we want. Is there a way to check if the Gmsh file is actually a 2D mesh and set the `spaceDim` to 2 in that case?

mlstowell: We could scan the elements looking for 3D elements but that would not guarantee 2D geometry.  We would also need to scan the z coordinates and verify that they lie on a plane (within some tolerance).

mlstowell:
> Should we include (maybe commented out) version of the Gmsh's high-order mesh optimization options from [tutorial/t5.geo](https://gitlab.onelab.info/gmsh/gmsh/-/blob/master/tutorial/t5.geo#L202)?
> 
> ```
>    // Mesh.ElementOrder = order;
>    // Mesh.HighOrderOptimize = 2;
> ```

I'm not familiar with that but it's fine with me.

mesh/mesh_readers.cpp
tzanio: How about we just check that the z-coordinate if the nodes is (identically) zero? It is strange that there is no way to distinguish between 2D and 3D in the Gmsh file :(

tzanio:
> > Should we include (maybe commented out) version of the Gmsh's high-order mesh optimization options from [tutorial/t5.geo](https://gitlab.onelab.info/gmsh/gmsh/-/blob/master/tutorial/t5.geo#L202)?
> > ```
> >    // Mesh.ElementOrder = order;
> >    // Mesh.HighOrderOptimize = 2;
> > ```
> 
> I'm not familiar with that but it's fine with me.

They've done a lot of work on high-order mesh validity, so I assume this is implementation of some of that research. See https://gmsh.info/doc/texinfo/gmsh.html#index-Mesh_002eHighOrderOptimize. cc: @kmittal2, @vladotomov 

@mlstowell and @bslazarov -- do you know if there is a difference between `SetOrder order;` and `Mesh.ElementOrder = order;`?

vladotomov:
Yes we'd be interested comparing their optimized mesh with the result from the mesh-optimizer.

mesh/mesh_readers.cpp
mlstowell: I took a closer look at the `spaceDim` member data when reading Gmsh files.  Currently the master branch of MFEM reads in 3 coordinates and stores them in `Vertex` objects. However, the `spaceDim` is not set until `FinalizeTopology` is called where it is given a default value equal to `Dim`.  Technically this is a bug because the z coordinates could be non trivial but we are ignoring them.  This does, however, explain why 2D Gmsh meshes appear as 2D meshes rather than surface meshes.

mlstowell: I can add code to compute the bounding box of the mesh as the nodes are read.  We can then set the space dimension based primarily on the element types and secondarily on the relative sizes of the bounding box.

tzanio:
LGTM

tzanio:

tzanio:
Please address these compiler warnings from tonight's regression runs:

```
mesh/mesh_readers.cpp: In member function ‘void mfem::Mesh::ReadGmshMesh(std::istream&, int&, int&)’:
mesh/mesh_readers.cpp:1779:33: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             for (int el = 0; el < ho_el_order_3D.size(); el++)
                                 ^
mesh/mesh_readers.cpp:1808:33: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             for (int el = 0; el < ho_el_order_2D.size(); el++)
                                 ^
mesh/mesh_readers.cpp:1833:33: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             for (int el = 0; el < ho_el_order_1D.size(); el++)
                                 ^
```

tzanio:
Thanks Mark!

tzanio:

tzanio:

v-dobrev:
Here is an illustration of the small addition to the `mesh-explorer` miniapp that I just pushed:
torus-surf-variations

This shows the deviation of the surface of `data/periodic-torus-sector.msh` from the perfect torus.

v-dobrev:
As discussed on Slack, the large mesh `data/periodic-torus-sector.msh` should be replaced with a smaller version.

After that, I'd recommend squash-merging the PR (or rebasing the branch before merging) to remove the large file from the history.

tzanio:

mlstowell:
image

I managed to redefine the torus mesh using a butterfly configuration in the cross section.  This allows for a coarser mesh while maintaining well shaped elements.

v-dobrev:
Here is the surface variation plot for the new torus:
torus-surf-variations-2

tzanio:

tzanio:
Adding a few more days for testing since there were recent changes...

tzanio:

tzanio:
LGTM

tzanio:
@v-dobrev, I made an adjustment with the `level_function` plotting in Mesh Explorer 940ec88.

Can you approve if you like it? (We can discuss/revert otherwise)

v-dobrev:
> @v-dobrev, I made an adjustment with the `level_function` plotting in Mesh Explorer [940ec88](https://github.com/mfem/mfem/commit/940ec885ef6d39e5d5c548a35f1e91395c7e1a16).
> 
> Can you approve if you like it? (We can discuss/revert otherwise)

It's fine.

PR #1588, opened 2020-06-29
[HIP/AMD] #define fix & Diffusion kernel optimization

camierjs:
- [x] Fix `__HIP_DEVICE_COMPILE__`
- [x] Add top `backends.hpp` file for all backends
- [x] Add `MFEM_LAMBDA` vs `MFEM_DEVICE` for `hipcc` compiler
- [x] Use `MFEM_UNROLL` in `Diffusion` kernel
- [x] Fix unused variable warning

| [#1588](https://github.com/mfem/mfem/pull/1588) | @camierjs | @tzanio | @v-dobrev + @acfisher + @YohannDudouit | 07/04/20 | 07/14/20 | 07/19/20 | |


camierjs:
LGTM

general/cuda.hpp
```diff
@@ -15,17 +15,15 @@
...
-#include <cuda_runtime.h>
-#include <cuda.h>
-#endif
-
 // CUDA block size used by MFEM.
 #define MFEM_CUDA_BLOCKS 256
 
 #ifdef MFEM_USE_CUDA
 #define MFEM_DEVICE __device__
+#define MFEM_LAMBDA __host__
```
YohannDudouit, line 23: I would have found more consistent to call this `MFEM_HOST`, is there a reason to choose `MFEM_LAMBDA` over `MFEM_HOST`?

general/cuda.hpp
```diff
@@ -15,17 +15,15 @@
...
 #define MFEM_HOST_DEVICE __host__ __device__
+#define MFEM_DEVICE_SYNC MFEM_GPU_CHECK(cudaDeviceSynchronize())
+#define MFEM_STREAM_SYNC MFEM_GPU_CHECK(cudaStreamSynchronize(0))
```
YohannDudouit, line 26: Can you add some documentation to give the intuition when one should be used over the other?

general/hip.hpp
```diff
@@ -15,16 +15,15 @@
...
-#ifdef MFEM_USE_HIP
-#include <hip/hip_runtime.h>
-#endif
-
 // HIP block size used by MFEM.
 #define MFEM_HIP_BLOCKS 256
 
 #ifdef MFEM_USE_HIP
 #define MFEM_DEVICE __device__
+#define MFEM_LAMBDA __host__ __device__
```
YohannDudouit, line 23: It seems inconsistent that the hip `MFEM_LAMBDA` is defined as `__host__ __device__` whereas the cuda one is defined as `__host__`, is there a reason for that?

general/backends.hpp
```diff
@@ -0,0 +1,57 @@
...
+#ifdef MFEM_USE_CUDA
+#include <cuda_runtime.h>
+#include <cuda.h>
+#endif
+#include "cuda.hpp"
+
+#ifdef MFEM_USE_HIP
+#include <hip/hip_runtime.h>
+#endif
+#include "hip.hpp"
```
YohannDudouit, line 26: I would expect `cuda.hpp` and `hip.hpp` to not be included when Cuda and Hip are not used respectively, is it necessary to include them?

fem/bilininteg_mass_pa.cpp
```diff
@@ -41,6 +41,8 @@ void MassIntegrator::SetupPA(const FiniteElementSpace &fes, const bool force)
       InitCeedCoeff(Q, ptr);
       return CeedPAMassAssemble(fes, *ir, *ptr);
    }
+#else
+   MFEM_CONTRACT_VAR(force);
```
YohannDudouit, line 45: I don't understand this line, can you tell my why it is needed?

general/cuda.hpp
camierjs: `nvcc` only accepts `__host__` there, where `hipcc` requires `__host__ __device__` to work.
That's why I added this variant, to be able to pick the right one from the header.

general/hip.hpp
camierjs: the two compilers behave differently

fem/bilininteg_mass_pa.cpp
camierjs: it removes one compiler warning when we're not using `libCeed` and compiling with `nvcc`.

tzanio:

camierjs:
> Code looks good. How much performance was gained with these implementations?

There isn't that much of differences, but it allow to use less shared memory and to run the 8th order for the BP3 benchmarks.

[v100_1-7_bp3_v1.pdf](https://github.com/mfem/mfem/files/4925943/v100_1-7_diff_v1.pdf)

[v100_1-8_bp3_v2.pdf](https://github.com/mfem/mfem/files/4925944/v100_1-8_diff_v2.pdf)

general/cuda.hpp
camierjs: `done`

general/backends.hpp
camierjs: The Cuda/Hip API (`CuMemAlloc`) is made available through these two files for now.
It's been used in the `MemoryManager`.
We could be able to hide it - that's something we can think of while refactoring.

camierjs:
LGTM

tzanio:

PR #1602, opened 2020-07-02
Added Picard iteration to ex16p

ben-s-southworth:
I added a Picard iteration within ImplicitMult in Example 16p so that the nonlinearity is formally resolved. The code is not complicated, but to me there is a "right" way to do a Picard iteration within the MFEM ODE and Operator context that took me a little bit to figure out. I think this provides a better example of a simple way to handle a nonlinearity that is robust and provides high-order accuracy, which I do not believe taking one Picard iteration does.

### DIFF
--- a/examples/ex16p.cpp
+++ b/examples/ex16p.cpp
@@ -65,7 +65,7 @@ protected:
    HypreSmoother M_prec; // Preconditioner for the mass matrix M
 
    CGSolver T_solver;    // Implicit solver for T = M + dt K
-   HypreSmoother T_prec; // Preconditioner for the implicit solver
+   HypreBoomerAMG T_prec; // Preconditioner for the implicit solver
 
    double alpha, kappa;
 
@@ -353,7 +353,6 @@ int main(int argc, char *argv[])
          }
 #endif
       }
-      oper.SetParameters(u);
    }
 
 #ifdef MFEM_USE_ADIOS2
@@ -414,6 +413,7 @@ ConductionOperator::ConductionOperator(ParFiniteElementSpace &f, double al,
    T_solver.SetPrintLevel(0);
    T_solver.SetPreconditioner(T_prec);
 
+   T_prec.SetPrintLevel(0);
    SetParameters(u);
 }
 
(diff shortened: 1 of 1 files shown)


v-dobrev:
I think the idea for ex16p was to be very simple even if the method is not that accurate. Maybe we can put this extended version in a new example.

@jamiebramwell, what do you thing?

This discussion started here: #1597.

cc: @tzanio

tzanio:
I recall a discussion with @jamiebramwell and @acfisher about this. I'll leave the decision to Jamie, but maybe we can support both options?

ben-s-southworth:
Another option as Veselin said is to have another example where I can hook
up the Anderson acceleration and have that option or just a full Picard
solve - example to demonstrate non-newton Linearizations. I don’t know much
about nonlinear diffusion, but if anybody could contribute/suggest a more
difficult nonlinear diffusion or advection-diffusion like problem (that
takes more than 3-4 Picard iterations to solve) I’d be happy to add the
nonlinear solvers.

jamiebramwell:
As stated, the main concept behind the original ex16p was exploring a simple time dependent operator without the additional complication of a full nonlinear solve. If I recall at the time, there was a desire to also recompute the stiffness every cycle, so the simple single Picard step was added. I would be fully in favor of examples of other nonlinear algorithms as long as the original intent (simpler than ex10p) is maintained. Thanks for looking into this!

examples/ex16p.cpp
```diff
@@ -430,19 +430,53 @@ void ConductionOperator::Mult(const Vector &u, Vector &du_dt) const
...
                                        const Vector &u, Vector &du_dt)
 {
-   // Solve the equation:
-   //    du_dt = M^{-1}*[-K(u + dt*du_dt)]
-   // for du_dt
-   if (!T)
-   {
-      T = Add(1.0, Mmat, dt, Kmat);
-      current_dt = dt;
+   // Here we use Picard iterations to solve a nonlinear equation
```
barker29, line 433: I would put these comments above the function (kind of like a docstring).

examples/ex16p.cpp
```diff
@@ -430,19 +430,53 @@ void ConductionOperator::Mult(const Vector &u, Vector &du_dt) const
...
+   int maxiter = 100;
+
+   // Right-hand side for nonlinear iteration
+   Mmat.Mult(u, z);  // Add forcing function if one exists
+   du_dt = u;        // Set u as initial guess for x (2)
+   Vector temp(u);   // Vector to measure error
+   temp = u;
+   double error = 1;
+   int iter = 0;
+   while (error > tol) {
```
barker29, line 461: Brace style.

bensworth:
@jamiebramwell What are your thoughts on this? The code can be changed to support both options as well if you want to keep the single iteration option as a simple example. My confusion on the single Picard iteration is I don't know if it's really a formal time stepping scheme, not really implicit or explicit (kind of IMEX, but explicitly treating the nonlinearity rather than, say, a variable)?

PR #1610, opened 2020-07-08
Add option to derefine elements only if they have the same attribute [nc-attr-dev]

jakubcerveny:
This tries to address #1572 :

1. Mesh::SetAttribute is propagated into NCMesh elements.
2. ThresholdDerefiner and Mesh::DerefineByError now have the option to only derefine a group of elements only if all of them have the same attribute.

Test code in ex15, ex15p (to be removed):

derefine_attr

TODO:
- [ ] Parallel synchronization of NCMesh::SetAttribute (ghost elements)

### DIFF
--- a/examples/ex15.cpp
+++ b/examples/ex15.cpp
@@ -63,6 +63,15 @@ double rhs_func(const Vector &pt, double t);
 void UpdateProblem(Mesh &mesh, FiniteElementSpace &fespace,
                    GridFunction &x, BilinearForm &a, LinearForm &b);
 
+void UpdateAttributes(Mesh &mesh)
+{
+   Vector center;
+   for (int i = 0; i < mesh.GetNE(); i++)
+   {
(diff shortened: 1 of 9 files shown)


termi-official:
Looks good. I think this is the best approach for now, because the alternatives may require some wider changes in the API. 
I try to get my (mesh) unit test branch upstream asap, but this may take some days.

I guess we should move further discussion from he issue to this pr?

tzanio:
@termi-official can you confirm that this approach works in your application?

termi-official:
Hi @tzanio , generally speaking the appraoch works fine for me. I am still running into problems with the boundary attributes (which may to be not correctly propagated into the ncmesh) and load balancing (which messes up the attribute assignment in some cases). Sadly I could not find time to finish the implementation of the previously mentioned unit tests for the mesh classes yet. 

Best, Dennis

PR #1618, opened 2020-07-13
AbsMult and AbsMultTranspose

stefanhenneking:
Features:
- [x] `SparseMatrix`: Add `AbsMult` and `AbsMultTranspose`
- [x] `HypreParMatrix`: Add `AbsMult` and `AbsMultTranspose`
- [x] Unit tests for `AbsMult` and `AbsMultTranspose`
- [x] `AssembleDiagonal` (PA) should use `AbsMult` instead of `Mult` if the mesh in non-conforming

Previous discussion about assembling the diagonal for NCMesh with @tzanio, @barker29, @v-dobrev, and @psocratis:

_In order to obtain the correct diagonal with PA for a conforming mesh that is an `NCMesh` object, we need to compute `|P^T| d_e` with entry-wise absolute values from `P` because the `NCMesh`, when constructed from a conforming mesh, can have flipped orientations so that `P` is not boolean anymore but instead has `{-1, 0, +1}` values. We should therefore add a `Mult` and `MultTranspose` operator that applies the entry-wise absolute value of an operator, i.e. in particular for `SparseMatrix` and `HypreParMatrix`._

_Additional remark: We should test if this also gives a convergent diagonal preconditioner in the general AMR case._

| [#1618](https://github.com/mfem/mfem/pull/1618) | @stefanhenneking | @tzanio | @psocratis + @barker29 + @jakubcerveny | 07/15/20 | 08/04/20 | 08/14/20 | |


stefanhenneking:
> The code is all quite nice and clean. I recall that there were some ad-hoc tests you were running in `ex6p` that showed this difficulty - is that resolved by this PR? Should there be a unit test that shows this new feature's effect on matrix-free preconditioning with AMR?

@barker29 I started a separate PR to add the diagonal preconditioning for PA to ex6[p], see #1632. If you think it's a good idea to add that change in this PR instead, I can pull the changes from the other branch [ex6-pa-preconditioner](https://github.com/mfem/mfem/tree/ex6-pa-preconditioner).

With regard to an additional unit test: One test I can think of would be to ensure that a conforming mesh yields the same diagonal (except for permutations of the values) when using the standard (conforming) `Mesh` and the corresponding non-conforming `NCMesh`. Would this be something of interest?

barker29:
No, it's probably better for #1632 to be separate, I didn't know about it and its existence basically answers my first question.

The unit test you describe I think would be nice if it does not require too much effort.

stefanhenneking:
> No, it's probably better for #1632 to be separate, I didn't know about it and its existence basically answers my first question.
> 
> The unit test you describe I think would be nice if it does not require too much effort.

Sounds good. I will add a unit test for that.

stefanhenneking:
@barker29 I have added a unit test verifying that the conforming mesh yields the same diagonal (except for permutations of the values) when using the standard (conforming) Mesh and the corresponding non-conforming NCMesh. Beyond that, the NCMesh is currently not tested in unit tests, as far as I can see. Another, more sophisticated, unit test would be to somehow compare the actual diagonal from a non-conforming mesh (obtained by assembling the full matrix) to the one now obtained via PA. However, since the two _will_ be different, we would need to agree what exactly a unit test would be testing. I'm open to ideas for adding such a unit test if it's useful.

tests/unit/mesh/test_ncmesh.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+#include "mfem.hpp"
+#include "catch.hpp"
+
+namespace mfem
+{
+
+constexpr double EPS = 1.e-12;
+
+// Test case: Verify that a conforming mesh yields the same diagonal with PA
```
barker29, line 20: This is picky but: what you actually test is that the *norms* of the diagonals are the same. That's fine for a simple unit test, but the comments should reflect what it actually tests.

tests/unit/mesh/test_ncmesh.cpp
```diff
@@ -0,0 +1,239 @@
...
+                   "order: " << order << ", error: " << error << std::endl;
+         REQUIRE(error == Approx(EPS));
+      }
+   }
+
+} // test case
+
+#ifdef MFEM_USE_MPI
+
+// Test case: Verify that a conforming parallel mesh yields the same diagonal with PA
```
barker29, line 115: Same comment, basically.

barker29:
I should have also said thanks for adding the unit test! It is a good addition.

tests/unit/mesh/test_ncmesh.cpp
stefanhenneking: Thanks, @barker29! I made a minor to change to clarify this point.

tzanio:
There are no flipped orientations, but the `P` matrix can still have negative entries, no?

stefanhenneking:
> There are no flipped orientations, but the `P` matrix can still have negative entries, no?

Yes, we are now using |P| to assemble the diagonal (with PA) when the mesh in non-conforming.

jakubcerveny:
There are many negative entries potentially. I guess I don't understand what |P| actually represents... what do you gain by taking the absolute value? Aren't the interpolations meaningless then?

tzanio:
It gives you a convergent diagonal preconditioner (from a local convergent diagonal), see https://mfem.slack.com/archives/CB98FH3U4/p1594134273261000

stefanhenneking:
Thanks for linking to the thread where we discussed it, @tzanio. I should have been more specific when I said "sign-flips". @jakubcerveny I think our discussion in slack does give a better picture of the issue we try to address. Specifically, the issue I came accross was that:

> Applying P^T to the diagonals d_e yields some zeros (the same happens when applying to a test vector initialized with ‘1’). These zero entries correspond to dofs that are at the interface between MPI ranks, and they are associated with either edge dofs (when using an H(curl) fespace) or face dofs (when using H(div) fespace), so opposite orientations are probably the reason that the values are accumulated to ‘0’ instead of ‘2’.

linalg/hypre_parcsr.cpp
```diff
@@ -977,6 +977,196 @@ void hypre_ParCSRMatrixSplit(hypre_ParCSRMatrix *A,
...
+   /*-----------------------------------------------------------------
+    * y += abs(A)^T*x
+    *-----------------------------------------------------------------*/
+
+   for (i = 0; i < num_rows; i++)
+   {
+      for (jj = A_i[i]; jj < A_i[i+1]; jj++)
+      {
+         j = A_j[jj];
+         y_data[j] += fabs(A_data[jj]) * x_data[i];
```
jakubcerveny, line 1153: Maybe using `std::abs` is better these days?

linalg/sparsemat.cpp
```diff
@@ -784,6 +784,101 @@ void SparseMatrix::BooleanMultTranspose(const Array<int> &x,
...
+      double *yp = y.HostReadWrite();
+
+      //  The matrix is not finalized, but multiplication is still possible
+      for (int i = 0; i < height; i++)
+      {
+         RowNode *row = Rows[i];
+         double b = 0.0;
+         for ( ; row != NULL; row = row->Prev)
+         {
+            b += fabs(row->Value) * xp[row->Column];
```
jakubcerveny, line 809: `std::abs`?

linalg/hypre_parcsr.cpp
stefanhenneking: Thanks, Jakub. I've changed these to `std::abs`.

tzanio:

stefanhenneking:
> Nice work @stefanhenneking!
> 
> A few minor comments:
> 
>     * Did you test this in the general AMR case?
> 
>     * Should we mention it in the `README` (the capability to have convergent diagonals with AMR, not that much the AbsMult)
> 
>     * Should we add a comment to that effect in `BilinearForm::AssembleDiagonal` to explain why we use AbsMult there?

@tzanio Thanks for suggestions:

* I tried it out for the stix miniapp and example 6/p; it worked fine in both cases with AMR.
* Just to clarify: Did you mean the `README` or the `CHANGELOG`? The `README` is very high-level, so I'm not sure how it best fits in, i.e. is this not too specific there?
* Yes, I added a comment there to explain the usage.

stefanhenneking:
Also, thanks @psocratis for doing half the work on this PR.

tzanio:
> Just to clarify: Did you mean the README or the CHANGELOG? The README is very high-level, so I'm not sure how it best fits in, i.e. is this not too specific there?

Sorry, `CHANGELOG`

stefanhenneking:
> > Just to clarify: Did you mean the README or the CHANGELOG? The README is very high-level, so I'm not sure how it best fits in, i.e. is this not too specific there?
> 
> Sorry, `CHANGELOG`

Okay, I'll add that change in #1632. That way, I can refer to example 6.

Issue #1555, opened 2020-06-17
Trouble with compiling MFEM with CUDA


sshiraiwa:
Hi all,
I am trying to compile MFEM with GPU, and getting an error (below).  I am doing this on TRAVERSE cluster at Princeton Univ. (https://researchcomputing.princeton.edu/systems-and-services/available-systems/traverse). 
It seems like the error is coming from somewhere around autosimd, and in fact,  workaround I found so far is to use MFEM_USE_SIMD=OFF. I have two question.

1) My hope is to test GPU on this cluster. Does MFEM_USE_SIMD=OFF impact the performance?
If so, is there other option?
2) This cluster uses Power9. Does this relate to the error?

Thank you for your help

[shiraiwa@traverse cmbuild]$ cmake .. -DCMAKE_VERBOSE_MAKEFILE=1 -DCMAKE_INSTALL_PREFIX=${HOME}/sandbox/mfem/ser -DMFEM_USE_CUDA=1
-- (optional) USER_CONFIG = /home/shiraiwa/src/mfem/config/user.cmake
-- USE_XSDK_DEFAULTS = 'OFF'
-- The CXX compiler identification is GNU 4.8.5
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- The CUDA compiler identification is NVIDIA 10.2.89
-- Check for working CUDA compiler: /usr/local/cuda-10.2/bin/nvcc
-- Check for working CUDA compiler: /usr/local/cuda-10.2/bin/nvcc -- works
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Using CUDA architecture: sm_60
-- Looking for POSIXClocks ...
--    checking library: <standard c/c++>
-- Found POSIXClocks: TRUE  
-- MFEM: using package POSIXCLOCKS
-- MFEM build type: CMAKE_BUILD_TYPE = Release
-- MFEM version: v4.1.1
-- MFEM git string: heads/master-0-gb8ab8fce47d9081a3a7bf311b1025f64af9ce54d
-- Writing substitute header --> "mfem.hpp"
-- Writing substitute header --> "mfem-performance.hpp"
[...]

PR #1627, opened 2020-07-16
Set MFEM_USE_SIMD=NO by default

YohannDudouit:
`MFEM_USE_SIMD=YES` results in compilation errors on IBM machines (Lassen, Ray,...), I suggests that we change the default value of this option because the compilation bugs are very confusing, and I think MFEM master should compile on Lassen with its default configuration.

| [#1627](https://github.com/mfem/mfem/pull/1627) | @YohannDudouit | @tzanio | @camierjs + @v-dobrev | 07/19/20 | 07/27/20 | 08/01/20 | |

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -41,7 +41,7 @@ Performance improvements
     - x86 (SSE/AVX/AVX2/AVX512),
     - Power8 & Power9 (VSX),
     - BG/Q (QPX).
-  These are now enabled by default, and can be disabled with MFEM_USE_SIMD=NO.
+  These are disabled by default, and can be enabled with MFEM_USE_SIMD=YES.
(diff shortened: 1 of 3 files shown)


tzanio:
Please update also the corresponding line in `CHANGELOG`:

https://github.com/mfem/mfem/blob/yohann/remove-simd/CHANGELOG#L44

camierjs:
I'm not sure we want to disable the our best CPU code w/ intrinsics.

The main issue here is how `nvcc` handles these headers.

I'd prefer to disable the all-in-one `nvcc` + `intrinsics` that causes this issue and keep the best on both sides, like in PR #1642 which should also solve #1555.

tzanio:

PR #1649, opened 2020-07-23
On the fly NC mesh initialization [nc-init-dev]

jakubcerveny:
This improves the mesh initialization API (AddVertex, AddQuad, AddHex, ...) and adds the possibility to create meshes that are nonconforming from the beginning (AddVertexParents).

List of changes: 
* The user does not have to preallocate exactly the space for vertices, elements, etc., in the `Mesh` initialization constructor. This is useful if one doesn't know up front how many elements will be generated. If the space runs out when calling AddVertex and similar, the internal arrays are automatically enlarged. The original style of use still works of course. 
* Overloaded methods that take individual indices (like `AddTriangle(v1, v2, v3, attr)`) have been added. These can be more user friendly in many cases. All `Add*` methods now also return the index of the newly added object.
* A new method `AddVertexParents` can be used to mark vertices as hanging vertices. During mesh finalization, an NC mesh is automatically created if at least one non-conforming vertex exists.
* Added a new method `GridFunction::RestrictConforming` that multiplies the grid function by the R and P matrices, making the solution conforming. I wasn't sure if we had this somewhere already, but couldn't find anything.
* A new miniapp to create 2D and 3D polar meshes is added (`meshing/polar-nc`) that demonstrates the improved API. 

TODO: 
- [x] Improved mesh initialization interface
- [x] 2D polar mesh
- [x] Curved 2D mesh
- [x] SFC ordered 2D mesh
- [x] 3D polar mesh
- [x] Update CHANGELOG
- [x] Update website

GLVis_s01

| [#1649](https://github.com/mfem/mfem/pull/1649) | @jakubcerveny | @tzanio | @tzanio + @vladotomov | 08/13/20 | 08/30/20 | 09/06/20 | |


jakubcerveny:
The 2D mesh is now SFC ordered by default:
ordering

jakubcerveny:
3D version of the mesh works: 
polar-wedges

tzanio:
Very nice!

vladotomov:
These are very cool meshes, I'll definitely test with them at some point.

Here's one:
`./polar-nc --radius 1 --nsteps 50 -o 6 --aspect 0.1 -phi 320`

GLVis_s01

These extreme cases however hit an assert in `debug`. Is it possible to fix this?
`Assertion failed: (CheckElementOrientation(false) == 0) is false:
 --> 
 ... in function: virtual void mfem::Mesh::Finalize(bool, bool)
 ... in file: mesh/mesh.cpp:2544`

mesh/mesh.hpp
```diff
@@ -206,6 +206,9 @@ class Mesh
    Array<FaceGeometricFactors*>
    face_geom_factors; ///< Optional face geometric factors.
 
+   /// Used during initialization only.
+   Array<Triple<int, int, int> > tmp_vertex_parents;
```
vladotomov, line 210: Should this be `public`?

miniapps/meshing/polar-nc.cpp
```diff
@@ -0,0 +1,558 @@
...
+   {
+      args.PrintUsage(cout);
+      return EXIT_FAILURE;
+   }
+   args.PrintOptions(cout);
+
+   // validate options
+   MFEM_VERIFY(dim >= 2 && dim <= 3, "");
+   MFEM_VERIFY(angle > 0 && angle < 360, "");
+   MFEM_VERIFY(nsteps > 0, "");
```
vladotomov, line 524: Also `radius > 0` and `aspect > 0` (negative aspect ratio actually produces interesting meshes).

tzanio:
@jakubcerveny, I'll merge this for testing, but can you address @vladotomov's questions above?

tzanio:

mesh/mesh.hpp
jakubcerveny: No, this is an internal thing.

miniapps/meshing/polar-nc.cpp
jakubcerveny: Added, thanks.

jakubcerveny:
The extreme case, if not curved, may invert the center elements. I was actually finalizing the mesh before adding the curvature, so that's why the assert was triggered. I moved the finalization after making the mesh curved, hopefully it won't break anything. Still it may be possible that the elements get inverted in the extreme case (with --order 1), but I'm not sure we need to worry about it.

mesh/mesh.cpp
```diff
@@ -1302,19 +1380,60 @@ void Mesh::AddHexAsWedges(const int *vi, int attr)
    }
 }
 
-void Mesh::AddBdrSegment(const int *vi, int attr)
+int Mesh::AddElement(Element *elem)
+{
+   CheckEnlarge(boundary, NumOfBdrElements);
```
dylan-copeland, line 1385: @jakubcerveny Is this line intended? Should `CheckEnlarge` really be called in `AddElement`? This has broken a code I have, which uses a simple way to construct a mesh by just calling `AddVertex` and `AddElement` for a set of vertices and elements in another mesh. There are no boundary elements, so `FinalizeCheck()` fails due to `boundary.Size() == 1` not being equal to `NumOfBdrElements == 0`, which is caused by this `CheckEnlarge` (when `boundary` has size 0, it gets set to size 1).

jakubcerveny: Oops, this is a typo (or a copy/paste error), thanks for pointing this out. @tzanio can I push a one-line fix directly into master?

jakubcerveny: Maybe that's not such a good idea. I pushed the fix instead to this branch (nc-init-dev), but I'm not sure how to revive this PR...

dylan-copeland: I made PR #1764 for this one-line fix and a unit test that should prevent this (unusual) method of mesh construction from failing in the future.

PR #1650, opened 2020-07-23
Fix the const correctness of casting operators and make them explicit.

YohannDudouit:
This PR forces explicit casting from `DeviceTensor<T>` to `T*`, it also enforces a stronger const-correctness.


tzanio:
Can this be part of #1626?

YohannDudouit:
Done.

PR #1655, opened 2020-07-24
Support additional coefficient options in CurlCurlIntegrator

stefanhenneking:
Summary:

`CurlCurlIntegrator` should support `S`, `D`, `M` coefficients for both full assembly and partial assembly.

Features:

- [x] Add Diagonal Matrix Function support (using `VectorCoefficient`) in `CurlCurlIntegrator`
- [x] Example 25: Use Diagonal Matrix Function instead of General Matrix Function
- [x] Implement PA support for `VectorCoefficient` and `MatrixCoefficient` in `CurlCurlIntegrator`
- [x] Unit tests (PA)
- [x] Documentation (on merge in master): Add `D` coef. to `CurlCurlIntegrator` on [MFEM webpage](https://mfem.org/bilininteg/), see https://github.com/mfem/web/pull/74

Credit for implementation: @dylan-copeland and @psocratis 

Additional discussion:

- The [MFEM webpage](https://mfem.org/bilininteg/) indicates integrators' support for Diagonal Matrix Function coefficient by `D` (different from Vector Valued Function `V`) but the user has to use a `VectorCoefficient` (representing the diagonal) in the corresponding constructor (whereas `VectorCoefficient` is also used for `V`). This is not immediately clear from the description. Should there be a `DiagonalMatrixCoefficient` (could be typedef to `VectorCoefficient`)? If not, should it be pointed out expliclitly that the user needs to use `VectorCoefficient` in this case? Update: This is resolved in https://github.com/mfem/web/pull/74.

Depends on:
- #1620 

Related PRs:
- #1652  

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge| 
 | --- | --- | --- | --- | --- | --- | --- | 
| [#1655](https://github.com/mfem/mfem/pull/1655) | @stefanhenneking | @tzanio | @mlstowell + @psocratis + @dylan-copeland | 08/14/20 | 09/03/20 | 09/08/20 | |

PR #1620, opened 2020-07-15
Matrix coefficient support for H(curl) PA.

dylan-copeland:
Symmetric and general (asymmetric) matrix coefficients are supported for the PA mass integrator and diagonal assembly in H(curl). Unit tests verify the new capabilities. 

The implementation depends on whether the matrix coefficient is symmetric, to minimize the amount of PA data at quadrature points. This requires a way to determine whether a `MatrixCoefficient` is symmetric, and the implementation choice is debatable. The current choice is to have a flag for symmetry in `MatrixCoefficient` and a special `EvalSymmetric` function to call in the symmetric case. 

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge| 
 | --- | --- | --- | --- | --- | --- | --- | 
| [#1620](https://github.com/mfem/mfem/pull/1620) | @dylan-copeland | @tzanio | @mlstowell + @stefanhenneking | 07/15/20 | 07/27/20 | 08/13/20 | |

PR #1665, opened 2020-07-29
Matrix coefficients for diffusion PA

dylan-copeland:
Matrix coefficients (symmetric or general, as well as diagonal represented by `VectorCoefficient`) are supported in the standard and Smem diffusion PA kernels (operator application and diagonal assembly), with unit tests. The H(curl) PA setup now uses the diffusion setup functions, to reduce code.

| [#1665](https://github.com/mfem/mfem/pull/1665) | @dylan-copeland | @tzanio | @mlstowell + @stefanhenneking + @camierjs | 08/01/20 | 09/28/20 | 10/04/20 | |


sshiraiwa:
@dylan-copeland Thank you for quick update. This fixed the issue we had in our Fokker-Planck solver. Cuda and CPU solution agrees all digits except the last one.
sol_cuda

fem/bilininteg_diffusion_pa.cpp
```diff
@@ -80,28 +80,35 @@ static void OccaPADiffusionSetup3D(const int D1D,
...
                            Vector &d)
 {
    const int NQ = Q1D*Q1D;
+   const bool symmetric = (coeffDim != 4);
    const bool const_c = c.Size() == 1;
    auto W = w.Read();
    auto J = Reshape(j.Read(), NQ, 2, 2, NE);
-   auto C = const_c ? Reshape(c.Read(), 1, 1) : Reshape(c.Read(), NQ, NE);
-   auto D = Reshape(d.Write(), NQ, 3, NE);
+   auto C = const_c ? Reshape(c.Read(), 1, 1, 1) : Reshape(c.Read(), coeffDim,
```
YohannDudouit, line 109: I think it's `coeffDim, 1 , 1` here for the first `Reshape`.

dylan-copeland: If `const_c`, `c` is of size 1, so it only has entry `[0]`. This version passes the unit tests, so I think the dimensions are right.

YohannDudouit:
```suggestion
   auto C = const_c ? Reshape(c.Read(), coeffDim, 1, 1) : Reshape(c.Read(), coeffDim,
```

sshiraiwa:
It seems SetSymmetric is only for GPU? If so, can you make warning or error when it is not Partial Assembly?

dylan-copeland:
> It seems SetSymmetric is only for GPU? If so, can you make warning or error when it is not Partial Assembly?

@sshiraiwa Good question, and unfortunately the specification of symmetry in `MatrixCoefficient` is a little confusing. By default, the member variable `symmetric` is set to `false` by the constructors, except when specifying a symmetric `MatrixFunctionCoefficient` defined by a function. There is no logic within `MatrixCoefficient` to decide whether to call `Eval` (general) or `EvalSymmetric`. Instead, the user of the class is expected to call `EvalSymmetric` when `IsSymmetric` is `true`, which so far is typically only used in PA but could be used anywhere (not necessarily PA). The reason for this is that the output is either a `DenseMatrix` (`Eval`) or `Vector` (`EvalSymmetric`). `SetSymmetric` is not used anywhere and perhaps should be removed, since it should be set by the constructor only and not by the user.

sshiraiwa:
@dylan-copeland Thanks. Yes, this saves a user from running CPU version without paying attention that IsSymmetry is on (like I did ;D)

fem/bilininteg_diffusion_pa.cpp
YohannDudouit: What about a `ConstantMatrixCoefficient`?

fem/coefficient.cpp
```diff
@@ -319,6 +321,31 @@ void MatrixFunctionCoefficient::Eval(DenseMatrix &K, ElementTransformation &T,
    }
 }
 
+void MatrixFunctionCoefficient::EvalSymmetric(Vector &K,
```
YohannDudouit, line 324: Can you remind why you need this new method, and you can't just do a check inside `Eval`?
```c++
void MatrixFunctionCoefficient::Eval(DenseMatrix &K, ElementTransformation &T,
                                     const IntegrationPoint &ip)
{
  if(symmetric)
  {
  ...
  }
  else
  {
  ...
  }
}
```

dylan-copeland: @YohannDudouit In the symmetric case, it returns a `Vector` of 6 values in 3D instead of a `DenseMatrix` of 9 values.

fem/bilininteg_diffusion_pa.cpp
dylan-copeland: @YohannDudouit There are a few classes derived from `MatrixCoefficient` for which `EvalSymmetric` is not implemented, so the base class version will be called, resulting in an error. `MatrixConstantCoefficient` is one of those few classes. Support for those types can be added in the future as needed.

tzanio:
Please resolve the merge conflicts

YohannDudouit:
It would have been nice to synchronize this PR with #1656

YohannDudouit:
It would be very nice if we could merge #1656 before this PR, merging this PR in #1656 is really not trivial and it would also make this PR cleaner.

fem/bilininteg.hpp
```diff
@@ -2999,6 +3029,17 @@ class VectorInnerProductInterpolator : public DiscreteInterpolator
    VectorCoefficient *VQ;
 };
 
-}
 
+
+// PA Diffusion Assemble 2D kernel
+template<const int T_SDIM>
+void PADiffusionSetup2D(const int Q1D,
```
YohannDudouit, line 3036: I'm really not a fan of this method being in `bilininteg.hpp`, I also think that using templates is not useful here since you specialize each implementation and you don't have a generic implementation.

dylan-copeland: This was a template before this PR, but I agree it could be two separate functions named `PADiffusionSetup2D` and `PADiffusionSetup2Din3D`, for example. The declaration is in the header because it is being used in diffusion and H(curl) files. In bilininteg_vectorfe.cpp there is just a forward declaration of `PADiffusionSetup3D`. I'm not sure if that would work for the 2D template function, but I could change the template function to 2 functions with different names and put their declarations in bilininteg_vectorfe.cpp. Does that sound good?

fem/bilininteg_vectorfe.cpp
```diff
@@ -15,25 +15,18 @@
 namespace mfem
```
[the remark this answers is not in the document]
dylan-copeland: Why would we split it? It is only about 1000 lines.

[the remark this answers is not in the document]
camierjs: To put the `Assemble` & `MultAdd` of `div` or `curl` together in some other file to avoid the forward of the instantiation, to allow parallelization of the compilation of these files. 
We could also uniformise the name with the other existing one.

fem/bilininteg_vectorfe.cpp
```diff
@@ -143,57 +139,675 @@ void PAHdivMassApply3D(const int D1D,
...
+   const int i21 = transpose ? 1 : 3;
+   const int i22 = 4;
+   const int i23 = transpose ? 7 : 5;
+   const int i31 = transpose ? 2 : 6;
+   const int i32 = transpose ? 5 : 7;
+   const int i33 = 8;
+
+   MFEM_FORALL(e, NE,
+   {
+      for (int q = 0; q < NQ; ++q)
```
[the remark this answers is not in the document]
dylan-copeland: Done.

fem/bilininteg_vectorfe.cpp
```diff
@@ -143,57 +139,675 @@ void PAHdivMassApply3D(const int D1D,
...
+   auto y = Reshape(op.Write(), 4, NQ, NE);
+
+   const int i11 = 0;
+   const int i12 = transpose ? 2 : 1;
+   const int i21 = transpose ? 1 : 2;
+   const int i22 = 3;
+
+   MFEM_FORALL(e, NE,
+   {
+      for (int q = 0; q < NQ; ++q)
```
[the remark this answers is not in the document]
dylan-copeland: Done.

fem/bilininteg_vectorfe.cpp
```diff
@@ -143,57 +139,675 @@ void PAHdivMassApply3D(const int D1D,
...
+         const double A12 = (J32 * J13) - (J12 * J33);
+         const double A13 = (J12 * J23) - (J22 * J13);
+         const double A21 = (J31 * J23) - (J21 * J33);
+         const double A22 = (J11 * J33) - (J13 * J31);
+         const double A23 = (J21 * J13) - (J11 * J23);
+         const double A31 = (J21 * J32) - (J31 * J22);
+         const double A32 = (J31 * J12) - (J11 * J32);
+         const double A33 = (J11 * J22) - (J12 * J21);
+
+         if (coeffDim == 6 || coeffDim == 9) // Matrix coefficient version
```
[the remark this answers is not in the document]
dylan-copeland: An advantage of splitting this into specialized kernels would be avoiding the if-statement, and a disadvantage would be more lines of code. Do you think the if-statement has a significant impact on speed, or does it just make the code more difficult to read?

fem/bilininteg_vectorfe.cpp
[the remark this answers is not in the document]
dylan-copeland: The H(div) kernels are in bilininteg_hdiv.cpp, and the H(curl) kernels are in bilininteg_hcurl.cpp, and this file only contains `VectorFEMassIntegrator` and `MixedVectorGradientIntegator` member functions which call those kernels in other files, except for a few mixed H(curl)-H(div) kernels. I don't see a better way to organize things in separate files.

dylan-copeland:
Now that PR #1620 is merged, this PR shows much fewer changes (and should be easier to review).

dylan-copeland:
@tzanio It looks like the Travis CI build stalled. Does it just need to be restarted?

tzanio:
@mlstowell, @stefanhenneking and @camierjs -- can you please review. I have extended the deadline to this week.

stefanhenneking:
@dylan-copeland Is this PR still waiting for another PR to be merged, or did it get everything from the last merge from master?

dylan-copeland:
> @dylan-copeland Is this PR still waiting for another PR to be merged, or did it get everything from the last merge from master?

@stefanhenneking This PR is actually waiting for PR #1656 which was waiting on PR #1620 but now just needs one more approval. You can review this PR now, and some minor changes to the coefficient interface will be made after PR #1656 is merged, or you can wait until the merge is done before reviewing.

fem/coefficient.hpp
```diff
@@ -721,7 +721,6 @@ class MatrixCoefficient
    /// For backward compatibility get the width of the matrix.
    int GetVDim() const { return width; }
 
-   void SetSymmetric(bool s) { symmetric = s; }
```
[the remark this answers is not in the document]
dylan-copeland: Yes, it was recently added, but it is not used anywhere and can be confusing. The `symmetric` flag should be set in the constructor, and the user should not change it with this access method.

fem/bilininteg.hpp
```diff
@@ -1872,7 +1874,7 @@ class DiffusionIntegrator: public BilinearFormIntegrator
    const GeometricFactors *geom;  ///< Not owned
    int dim, ne, dofs1D, quad1D;
    Vector pa_data;
-
+   bool symmetric = true; ///< False if using a nonsymmetric matrix coefficient
```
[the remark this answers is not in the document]
dylan-copeland: This flag is used only in the PA setup and multiplication, and I think it is clear if the user understands the kernels; otherwise, the user should ignore it. Maybe `pa_symmetric` would be a more appropriate name.

fem/coefficient.cpp
```diff
@@ -294,6 +294,8 @@ void VectorRestrictedCoefficient::Eval(
 void MatrixFunctionCoefficient::Eval(DenseMatrix &K, ElementTransformation &T,
                                      const IntegrationPoint &ip)
 {
+   MFEM_VERIFY(!symmetric, "MatrixFunctionCoefficient is symmetric");
```
mlstowell, line 297: We should be able to support the symmetric case as well.  Users might have a reason for not calling EvalSymmetric.

mlstowell:
I know this hasn't been changed in this PR but in `MatrixFunctionCoefficient::EvalSymmetric` do we really need to restrict `height<4`?

fem/coefficient.cpp
dylan-copeland: @mlstowell `Eval` sets a `DenseMatrix`, whereas `EvalSymmetric` sets a `Vector` of the upper triangular entries. The flag `symmetric` does not necessarily mean that the matrix is symmetric; the real meaning is that the `MatrixCoefficient` is intended to be evaluated via `EvalSymmetric`, e.g. a function has been provided for computing the upper triangular entries. 

The user could call `Eval` for a symmetric `MatrixFunctionCoefficient` created with the constructor that takes a function defining all matrix entries. That constructor sets `MatrixCoefficient::symmetric` to `false`, by default, so that `Eval` can be called. It does not matter whether the matrix is actually symmetric mathematically, just whether it is specified by a function that computes all entries or just the upper triangular entries.

This summarizes my intended use of these functions, but your comment is still valid. We could decide to support the symmetric case in `MatrixFunctionCoefficient`, where `symmetric` being true means that `SymmFunction` should be called, not `Function` or `TDFunction`. In that case, `Eval` should call `SymmFunction` to get a `Vector` of upper triangular values, and then insert those into the full `DenseMatrix` `K`. This would be a more complete implementation, allowing the user to call `Eval`, but it would be less efficient than calling `EvalSymmetric`. It is a design question, of whether we want to allow the user to call the less efficient `Eval` for the sake of flexibility, or whether the user should copy those upper triangular entries to a full matrix in their code. I do not see a common case where the user should be allowed to call `Eval` when `EvalSymmetric` could be called. The typical usage is in integrators, where the matrix coefficient should be evaluated as efficiently as possible. Do you have other reasons to support the symmetric case in `Eval`?

dylan-copeland:
> I know this hasn't been changed in this PR but in `MatrixFunctionCoefficient::EvalSymmetric` do we really need to restrict `height<4`?

I removed that check. There should be no usage in MFEM so far where height > 3, but there is also not a reason to make this restriction. Maybe I put it there initially for debugging. Anyway, I agree it is unnecessary.

stefanhenneking (approved):

dylan-copeland:
@YohannDudouit @camierjs @mlstowell Do you still have any concerns about this PR?

fem/coefficient.cpp
mlstowell: I was mainly thinking of existing integrators.  This new matrix coefficient can only be used with integrators which are aware of the new scenario for handling symmetric coefficients.  It's possible that a user might need to send the same coefficient to two different integrators; a new one they have implemented, and a preexisting one that doesn't know how to handle symmetric coefficients.  This class compels the user to alter existing implementations.

dylan-copeland: Note that the interface changes (e.g. `EvalSymmetric`) were introduced in a recently merged PR, not this PR. This comment is on the `MFEM_VERIFY(!symmetric...` but seems to be about the interface change in general, not just this verify. That is fine, but I just want to clarify that we are discussing more than just this line of code.

@mlstowell For your concern about existing implementations, let's consider how an old integrator would have called `MatrixFunctionCoefficient`. Going back to the July 2, 2020 version of coefficient.hpp in branch `master`, `MatrixFunctionCoefficient` had `Function` and `TDFunction` which could be called by `Eval` to set a `DenseMatrix` (or it could be set to `DenseMatrix mat`, depending on the constructor used). There was no `SymmFunction` or `EvalSymmetric` to set a `Vector` of upper triangular entries. An old integrator would necessarily be using a constant `mat`, or calling `Function` or `TDFunction`. In branch `master` and in this PR, the flag `symmetric` is by default set to `false`. Therefore, `Eval` can still be called as before, and backwards compatibility is not broken.

The only case where `Eval` cannot be called is when the new `MatrixFunctionCoefficient` constructor is called to set `SymmFunction` and `symmetric` to `true`, i.e. the new usage. If this new kind of coefficient is passed to an integrator that calls `Eval`, then it will fail,	and I think this is the issue you are concerned about. In this case, `mat`, `Function`, and `TDFunction` are not set, just `SymmFunction`. At least it will fail with an `MFEM_VERIFY` rather than crashing, so the user has an error message and knows what happened.

This brings us back to the design question of whether we want to (1) force the user to update the integrator code to use `EvalSymmetric` (as in the current state of this PR), or (2) call `SymmFunction` in `Eval` and set the `DenseMatrix` by copying the upper triangular entries to the full matrix. This is a matter of opinion. There is	the choice between (1) which ensures efficient use of `SymmFunction` but is less flexible and user-friendly, and (2) which is flexible but inefficiently copies entries. If you think (2) is better, I can implement the copying.

@tzanio @stefanhenneking @camierjs Do you have a strong preference between (1) and (2)?

mlstowell: I do prefer option 2.  Option 1 seems like an "all or nothing" approach.  There may be a middle ground where a user wants to implement their `Function` or `TDFunction` by just computing the upper triangular portion but then use this with an older integrator which expects the full matrix coefficient.  Performing the copy to fill the full matrix may be less than optimal but it's not that expensive and it's easy enough or us to do.

I others have a strong opinion that we should use option 1 I'll accept that and approve.

YohannDudouit (commented):
@dylan-copeland I'm not a reviewer on this PR, I just had a few comments.
I reiterate my concern about the `EvalSymmetric` method and I agree with @mlstowell , I understand why it's useful, but I consider the current design more like a hack. We use `Vector` in many places with _implicit_ data layouts, and I don't think we should encourage this practice. In this case requesting from the user to be responsible of accessing the `Vector` like a symmetric matrix is cumbersome, bug prone, and makes the code hard to read. I think in some cases it was fine to use `Vector` with implicit layouts because we needed to run on GPU, and we don't have yet any better container. However, using `Vector` instead of an appropriate type will weaken the overall design and make MFEM harder to read and use properly. I think we ultimately want to replace in the code all the `Vector` that are used as a generic container (which it is not) and replace them with appropriate types (not yet existing).
In the present case, I would prefer to have a concept of symmetric matrix, meaning we should have a `SymmetricMatrix` class, with an associated `SymmetricMatrixCoefficient` class. It also means as @mlstowell pointed out that now, everywhere in the code we need to replace:
```c++
matCoeff.Eval(...);
```
with
```c++
if ( matCoeff.IsSymmetric() ) {
   matCoeff.Eval(...);
} else {
   matCoeff.EvalSymmetric(...);
}
```
Except that as you mentioned, one is returning a `DenseMatrix` and the other one a `Vector` making this process tricky. If we were to have a `SymmetricMatrixCoefficient` class we would avoid this issue, and get a compilation error when an `Integrator` does no support `SymmetricMatrixCoefficient`.
I also think a `SymmetricMatrix` should have the same accessor as a `DenseMatrix`, i.e. a row and column index accessor. I know that this is much beyond what is here, but that would be a cleaner way to do it I think. (I am not asking you to do it, this is just my opinion)

dylan-copeland:
@YohannDudouit You made excellent suggestions about introducing a `SymmetricMatrix` class and using it in `MatrixCoefficient`. I agree and will plan to do this in a future PR, after this one is merged.

fem/coefficient.cpp
dylan-copeland: @mlstowell I implemented the copying (option 2 above). There are no tests (unit or otherwise) that use this version (calling Eval for a symmetric function), but the code is short enough that it probably is bug-free.

stefanhenneking:
> @YohannDudouit You made excellent suggestions about introducing a `SymmetricMatrix` class and using it in `MatrixCoefficient`. I agree and will plan to do this in a future PR, after this one is merged.

Yes, I agree this could be a good idea to make things cleaner in a future PR. I think it would also apply to diagonal matrix coefficients, i.e. introducing`DiagonalMatrix` - something we briefly discussed previously (see description in https://github.com/mfem/mfem/pull/1655), as well as to anti-symmetric matrix coefficients.

fem/coefficient.cpp
mlstowell: Thanks, @dylan-copeland !

mlstowell (approved):

tzanio:
@camierjs, did you get a chance to review this?

camierjs:
LGTM

camierjs (approved):

tzanio:

v-dobrev:
LGTM

v-dobrev:
Re-merged in `next`.

tzanio (approved):
Thanks @dylan-copeland!

@mlstowell, @stefanhenneking and @camierjs -- Thanks for reviewing!

PR #1669, opened 2020-07-30
SparseMatrix Copy Constructor MemoryType

tomstitt:
Adds optional `MemoryType` parameter for the `SparseMatrix` Copy Constructor.

I've only tested the `if (mat.Finalized())` case since that's the only one we're using.

| [#1669](https://github.com/mfem/mfem/pull/1669) | @tomstitt | @tzanio | @artv3 + @v-dobrev + @YohannDudouit | 08/13/20 | ⌛due 08/27/20 | ⌛due 09/03/20 | |

### DIFF
--- a/linalg/sparsemat.cpp
+++ b/linalg/sparsemat.cpp
@@ -175,21 +175,9 @@ SparseMatrix::SparseMatrix(const SparseMatrix &mat, bool copy_graph,
       }
 
       // We probably do not need to set the ownership flags here.
-      if (mt != MemoryType::SIZE)
-      {
-         I.Reset(mt);
-         J.Reset(mt);
-         A.Reset(mt);
-      }
-      else
-      {
-         I.Reset();
-         J.Reset();
-         A.Reset();
-      }
-      I.SetHostPtrOwner(true);
-      J.SetHostPtrOwner(true);
-      A.SetHostPtrOwner(true);
+      I.Reset(); I.SetHostPtrOwner(true);
+      J.Reset(); J.SetHostPtrOwner(true);
+      A.Reset(); A.SetHostPtrOwner(true);
    }
 
    current_row = -1;
--- a/linalg/sparsemat.cpp
+++ b/linalg/sparsemat.cpp
@@ -121,7 +121,8 @@ SparseMatrix::SparseMatrix(int nrows, int ncols, int rowsize)
    }
 }
 
-SparseMatrix::SparseMatrix(const SparseMatrix &mat, bool copy_graph)
+SparseMatrix::SparseMatrix(const SparseMatrix &mat, bool copy_graph,
+                           MemoryType mt)
    : AbstractSparseMatrix(mat.Height(), mat.Width())
 {
    if (mat.Finalized())
@@ -129,8 +130,8 @@ SparseMatrix::SparseMatrix(const SparseMatrix &mat, bool copy_graph)
       const int nnz = mat.I[height];
       if (copy_graph)
       {
-         I.New(height+1, mat.I.GetMemoryType());
-         J.New(nnz, mat.J.GetMemoryType());
+         I.New(height+1, mt == MemoryType::SIZE ? mat.I.GetMemoryType() : mt);
+         J.New(nnz, mt == MemoryType::SIZE ? mat.J.GetMemoryType() : mt);
          I.CopyFrom(mat.I, height+1);
          J.CopyFrom(mat.J, nnz);
       }
@@ -141,7 +142,7 @@ SparseMatrix::SparseMatrix(const SparseMatrix &mat, bool copy_graph)
          I.ClearOwnerFlags();
          J.ClearOwnerFlags();
       }
-      A.New(nnz, mat.A.GetMemoryType());
+      A.New(nnz, mt == MemoryType::SIZE ? mat.A.GetMemoryType() : mt);
       A.CopyFrom(mat.A, nnz);
 
       Rows = NULL;
@@ -174,9 +175,12 @@ SparseMatrix::SparseMatrix(const SparseMatrix &mat, bool copy_graph)
       }
 
       // We probably do not need to set the ownership flags here.
-      I.Reset(); I.SetHostPtrOwner(true);
-      J.Reset(); J.SetHostPtrOwner(true);
-      A.Reset(); A.SetHostPtrOwner(true);
+      I.Reset(mt == MemoryType::SIZE ? I.GetMemoryType() : mt);
+      I.SetHostPtrOwner(true);
+      J.Reset(mt == MemoryType::SIZE ? J.GetMemoryType() : mt);
+      J.SetHostPtrOwner(true);
+      A.Reset(mt == MemoryType::SIZE ? A.GetMemoryType() : mt);
+      A.SetHostPtrOwner(true);
    }
 
    current_row = -1;
--- a/linalg/sparsemat.cpp
+++ b/linalg/sparsemat.cpp
@@ -175,11 +175,20 @@ SparseMatrix::SparseMatrix(const SparseMatrix &mat, bool copy_graph,
       }
 
       // We probably do not need to set the ownership flags here.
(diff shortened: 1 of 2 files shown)


artv3:
@tomstitt  does this depend on other changes? Travis reported an error: 

````
MFEM abort: internal error

 ... in function: static mfem::MemoryType mfem::MemoryManager::GetDeviceMemoryType_(void*)

 ... in file: general/mem_manager.cpp:913

```

tomstitt:
> @tomstitt does this depend on other changes? Travis reported an error:
> 
> ```
> MFEM abort: internal error
> 
>  ... in function: static mfem::MemoryType mfem::MemoryManager::GetDeviceMemoryType_(void*)
> 
>  ... in file: general/mem_manager.cpp:913
> 
> ```
> ```

@artv3 if `mt` isn't set it should be a exactly the same, I'll figure out why that isn't the case.

linalg/sparsemat.cpp
```diff
@@ -121,16 +121,17 @@ SparseMatrix::SparseMatrix(int nrows, int ncols, int rowsize)
...
 {
    if (mat.Finalized())
    {
       const int nnz = mat.I[height];
       if (copy_graph)
       {
-         I.New(height+1, mat.I.GetMemoryType());
-         J.New(nnz, mat.J.GetMemoryType());
+         I.New(height+1, mt == MemoryType::SIZE ? mat.I.GetMemoryType() : mt);
+         J.New(nnz, mt == MemoryType::SIZE ? mat.J.GetMemoryType() : mt);
```
YohannDudouit, line 134: If I understand correctly you're changing the semantic of `MemoryType::SIZE` to mean "use the same `MemoryType` as the given SparseMatrix", this should be documented.

linalg/sparsemat.cpp
```diff
@@ -174,9 +175,21 @@ SparseMatrix::SparseMatrix(const SparseMatrix &mat, bool copy_graph)
...
 
       // We probably do not need to set the ownership flags here.
-      I.Reset(); I.SetHostPtrOwner(true);
-      J.Reset(); J.SetHostPtrOwner(true);
-      A.Reset(); A.SetHostPtrOwner(true);
+      if (mt != MemoryType::SIZE)
+      {
+         I.Reset(mt);
+         J.Reset(mt);
+         A.Reset(mt);
```
YohannDudouit, line 182: The documentation of the `Reset` method says "Reset the memory and set the host memory type.", how do you know that `mt` is a host memory type?

linalg/sparsemat.cpp
```diff
@@ -174,9 +175,21 @@ SparseMatrix::SparseMatrix(const SparseMatrix &mat, bool copy_graph)
...
+      }
+      else
+      {
+         I.Reset();
+         J.Reset();
+         A.Reset();
+      }
```
YohannDudouit, line 189: I don't understand why you need to call the `Reset` method, this seems wrong to me, you just called `New` to get some memory, and you `Reset` the memory right after?

tomstitt:
> Could you give a bit more details why you need this?

@YohannDudouit, yes, we're creating a temporary `SparseMatrix` that is used on the device and since it is large we were allocating space for it once and storing it for the duration of the simulation to avoid `cudaMalloc` performance hits.

The SpMat's buffers aren't needed for much of our timestep loop so by updating the ctor we can pass in a `MemoryType` to make the SpMat use memory from a temporary device pool (the temp device pool is from a different PR that we're working on but that we're currently using in a custom branch); this way we both avoid the costs of `cudaMalloc`s and save on memory overhead since we can share the pool with other parts of the timestep loop that need temporary buffers.

linalg/sparsemat.cpp
tomstitt: Yes, good call, I'll add some documentation

linalg/sparsemat.cpp
tomstitt: That documentation is a bit out-of-date, in the `Memory` class there statements like [this](https://github.com/mfem/mfem/blob/master/general/mem_manager.hpp#L697) that will get the corresponding host type even when the argument is a device type. More fundamentally I don't think this is even needed since a call to `Finalize()`, which is what will actually modify `I`, `J`, and `A`, uses `Memory::New()`, which ignores the `Memory` object's current `MemoryType`, if that sounds right to you I can revert these changes.

linalg/sparsemat.cpp
tomstitt: It looks like in this `else` block (for `if (mat.Finalized())`) `I`, `J`, and `A` are not `New`'d in any ways, it's the `Rows` member that is allocated and used, does look true to you?

YohannDudouit: Thank you, all the `#ifdef` made me loose sight of what was happening.

YohannDudouit:
@tomstitt If you can just add the documentation, I'll approve your PR, I've already tested that it doesn't break tests.

tomstitt:
> @tomstitt If you can just add the documentation, I'll approve your PR, I've already tested that it doesn't break tests.

Thanks @YohannDudouit, I added some documentation and also remove the `Reset(mt)` from the `else` block since those `mt` values won't stick around

linalg/sparsemat.hpp
```diff
@@ -112,8 +112,13 @@ class SparseMatrix : public AbstractSparseMatrix
    /// Copy constructor (deep copy).
    /** If @a mat is finalized and @a copy_graph is false, the #I and #J arrays
        will use a shallow copy (copy the pointers only) without transferring
-       ownership. */
-   SparseMatrix(const SparseMatrix &mat, bool copy_graph = true);
+       ownership.
+       If @a mt is MemoryType::SIZE the memory type of the resulting
+       SparseMatrix's #I, #J, and #A arrays will be the same as @a mat,
+       otherwise the type with be @a mt for those arrays that are deep
```
YohannDudouit, line 118:
```suggestion
       otherwise the type will be @a mt for those arrays that are deep
```

YohannDudouit, line 118:
```suggestion
       otherwise the type will be @a mt for those arrays that are deep
```

linalg/sparsemat.cpp
```diff
@@ -121,16 +121,17 @@ SparseMatrix::SparseMatrix(int nrows, int ncols, int rowsize)
...
    : AbstractSparseMatrix(mat.Height(), mat.Width())
 {
    if (mat.Finalized())
    {
       const int nnz = mat.I[height];
       if (copy_graph)
       {
-         I.New(height+1, mat.I.GetMemoryType());
-         J.New(nnz, mat.J.GetMemoryType());
+         I.New(height+1, mt == MemoryType::SIZE ? mat.I.GetMemoryType() : mt);
```
camierjs, line 133: I'm not sure we should use `MemoryType::SIZE` here, it's confusing.
Do we need to add a `UNKNOWN` one?

linalg/sparsemat.hpp
tomstitt: Thanks @YohannDudouit !

linalg/sparsemat.cpp
tomstitt: I agree it would be good to have something else. We could call it `UNSET`, which would make the intention clear when it was used as the default arg but I also like `UNKNOWN` and `DEFAULT`. Would you like to hold this PR until we add one?

tzanio: Tom, does this need to be in the Copy constructor, e.g. can we add another constructor that explicitly provides the memory type?

tomstitt: It does need to be in the copy constructor. I think we should actually move these changes to GH-1644 since the interface might change and that's where we're working on it

tzanio: I'm Okay with that. Do you still want to keep this PR open?

tomstitt: No, let's close it

tzanio: thanks Tom!

PR #1680, opened 2020-08-03
Partial assembly gradient operator

barker29:
This PR is superseded by #1761, which includes all the useful code from this PR.

This PR implements a PA gradient operator, which is an `Interpolator` rather than an `Integrator` type object. As a result, it has some different behavior related to `Add` and `Set` values that as I understand it has not been implemented in a PA way before.

TODO:
- [x] clean up code (comments, todos)
- [ ] make an example (perhaps Hiptmair smoothing)
- [ ] more efficient kernels
- [ ] test with AMR (may be out of scope)


fem/bilininteg_hcurl.cpp
```diff
@@ -1802,6 +1802,536 @@ void PAHcurlH1Apply3D(const int D1D,
...
+                                   const Vector &_x,
+                                   Vector &_y)
+{
+   auto B = Reshape(_B.Read(), c_dofs1D, c_dofs1D);
+   auto G = Reshape(_G.Read(), o_dofs1D, c_dofs1D);
+
+   auto x = Reshape(_x.Read(), c_dofs1D, c_dofs1D, c_dofs1D, NE);
+   auto y = Reshape(_y.ReadWrite(), (3 * c_dofs1D * c_dofs1D * o_dofs1D), NE);
+
+   Vector pxwork1(c_dofs1D * c_dofs1D * c_dofs1D);
```
dylan-copeland, line 1968: In existing PA kernels in MFEM, usually arrays like this are defined with something like `double a[m][n][p]` inside the `MFEM_FORALL`. Is there a difference between that and the `ReadWrite` used here, in terms of how the memory is allocated and used?

barker29: Probably? The Nedelec dofs come in three separate groups (x, y, z directions), and each one of those groups has size `c_dofs * c_dofs * o_dofs`. I wasn't sure the best way to organize that, I chose this and got it to work, but probably there's more efficient ways for memory access etc.

barker29: Oh, I misunderstood your comment (thought you were talking about line 1966). You are right and I should probably change the way `pxwork1` is allocated.

dylan-copeland: I am not familiar enough with the memory management to know the details of the difference, so it is unclear to me whether a change is needed. The GPU experts should have more insight (@camierjs @YohannDudouit ).

YohannDudouit: I agree with @dylan-copeland you should declare these inside the `MFEM_FORALL`, currently this generates a lot a memory allocations/deallocations which are very expensive on GPU. I'm also suspicious that this code works on GPU, because all the threads compute in the same memory space. If you declare the the `Vector` inside the `MFEM_FORALL` using arrays, then the memory space is private to each thread.

fem/bilininteg_hcurl.cpp
```diff
@@ -3208,4 +3717,90 @@ void MixedVectorWeakCurlIntegrator::AddMultPA(const Vector &x, Vector &y) const
...
+   const int dims = trial_el->GetDim();
+   MFEM_VERIFY(dims == 2 || dims == 3, "");
+
+   dim = mesh->Dimension();
+   MFEM_VERIFY(dim == 2 || dim == 3, "");
+
+   MFEM_VERIFY(trial_el->GetOrder() == test_el->GetOrder(), "");
+
+   ne = trial_fes.GetNE();
+   geom = mesh->GetGeometricFactors(*old_ir, GeometricFactors::JACOBIANS);
```
dylan-copeland, line 3748: It looks like `geom` is not used and can be removed.

fem/bilininteg_hcurl.cpp
```diff
@@ -1802,6 +1802,515 @@ void PAHcurlH1Apply3D(const int D1D,
...
+{
+   auto B = Reshape(_B.Read(), c_dofs1D, c_dofs1D);
+   auto G = Reshape(_G.Read(), o_dofs1D, c_dofs1D);
+
+   auto x = Reshape(_x.Read(), c_dofs1D, c_dofs1D, c_dofs1D, NE);
+   auto y = Reshape(_y.ReadWrite(), (3 * c_dofs1D * c_dofs1D * o_dofs1D), NE);
+
+   MFEM_FORALL(e, NE,
+   {
+      double pxw1[c_dofs1D][c_dofs1D][c_dofs1D];
```
dylan-copeland, line 1965: Although these are "small" arrays, they can cause GPU stack memory failures for high order. Instead of 6 arrays, you can just use 2 arrays `w1` and `w2` of size `[c_dofs1D][c_dofs1D][c_dofs1D]`

fem/bilininteg_hcurl.cpp
```diff
@@ -1802,6 +1802,515 @@ void PAHcurlH1Apply3D(const int D1D,
...
+      for (int ez = 0; ez < c_dofs1D; ++ez)
+      {
+         for (int ey = 0; ey < c_dofs1D; ++ey)
+         {
+            for (int ex = 0; ex < o_dofs1D; ++ex)
+            {
+               for (int dx = 0; dx < c_dofs1D; ++dx)
+               {
+                  const int local_index = ez*c_dofs1D*o_dofs1D + ey*o_dofs1D + ex;
+                  y(local_index, e) += G(ex, dx) * pxw2[dx][ey][ez];
```
dylan-copeland, line 2018: In summations like this, I think it is more efficient to use a local variable `double sum = 0.0` and then set `y(local_index, e) = sum` after the `dx` loop.

barker29:
Superseded by #1761.

tzanio:
Can we also delete the branch, or do you want to keep it?

Issue #1678, opened 2020-08-03
Improve error message in DenseMatrix::Eigensystem


benzwick:
Attempting to compute the eigensystem of a dense matrix without LAPACK results in the following error:
```
DenseMatrix::Eigensystem
Aborted
```
which is not very informative.

Could we add some extra information to this line in [densemat.cpp](https://github.com/mfem/mfem/blob/ba51fe2c53862699eadf465537a6a66a290ea909/linalg/densemat.cpp#L1086):
```
   mfem_error("DenseMatrix::Eigensystem");
```
e.g. like this:
```
   mfem_error("DenseMatrix::Eigensystem: requires MFEM built with LAPACK");
```
to make it immediately obvious why this failed?

Maybe the instructions for using LAPACK could also be added to the instructions on the [website](https://mfem.org/building/).

benzwick:
On [line 1167](https://github.com/mfem/mfem/blob/ba51fe2c53862699eadf465537a6a66a290ea909/linalg/densemat.cpp#L1167) there is also this one:
```
   mfem_error("DenseMatrix::Eigensystem for generalized eigenvalues");
```

On [line 3372](https://github.com/mfem/mfem/blob/ba51fe2c53862699eadf465537a6a66a290ea909/linalg/densemat.cpp#L3372) (and a few other places) there is this message which could be used for the similar errors above:
```
   mfem_error("DenseMatrixEigensystem::Eval(): Compiled without LAPACK");
```
There are also some other errors in `densemat.cpp` that could output more information like "incorrect matrix dimensions" or "dimension mismatch", for example [here](https://github.com/mfem/mfem/blob/ba51fe2c53862699eadf465537a6a66a290ea909/linalg/densemat.cpp#L2399):
```
   if (A.Height() != ABt.Height() || B.Height() != ABt.Width() ||
       A.Width() != B.Width())
   {
      mfem_error("MultABt(...)");
   }
```

barker29:
In my view these suggestions are very reasonable. Could you collect them in a pull request and submit it?

benzwick:
Yes, I will create a PR for this.

Somewhat related, can someone please explain to me what the `MFEM_CONTRACT_VAR` macro is used for. It is defined [here](https://github.com/mfem/mfem/blob/b9a40daf5b3b6b5338907c3936941415abd5e4b4/general/error.hpp#L113) as:
```
#define MFEM_CONTRACT_VAR(x) if (false && (&x)+1){}
```
[...]

PR #1685, opened 2020-08-05
Improve DenseMatrix error messages

benzwick:
See #1678

| [#1685](https://github.com/mfem/mfem/pull/1685) | @benzwick | @tzanio | @barker29 + @mlstowell | 08/05/20 | 08/23/20 | 08/30/20 | |

### DIFF
--- a/linalg/densemat.cpp
+++ b/linalg/densemat.cpp
@@ -401,7 +401,7 @@ void DenseMatrix::InvSymmetricScaling(const Vector & s)
 {
    if (height != width || s.Size() != width)
    {
-      mfem_error("DenseMatrix::SymmetricScaling: dimension mismatch");
+      mfem_error("DenseMatrix::InvSymmetricScaling: dimension mismatch");
    }
 
    double * ss = new double[width];
@@ -528,7 +528,7 @@ double DenseMatrix::Weight() const
       double F = d[0] * d[3] + d[1] * d[4] + d[2] * d[5];
       return sqrt(E * G - F * F);
    }
-   mfem_error("DenseMatrix::Weight(): dimension mismatch");
+   mfem_error("DenseMatrix::Weight(): mismatched or unsupported dimensions");
    return 0.0;
 }
 
@@ -1164,7 +1164,7 @@ void DenseMatrix::Eigensystem(DenseMatrix &b, Vector &ev,
    MFEM_CONTRACT_VAR(b);
    MFEM_CONTRACT_VAR(ev);
    MFEM_CONTRACT_VAR(evect);
-   mfem_error("DenseMatrix::Eigensystem: Compiled without LAPACK");
+   mfem_error("DenseMatrix::Eigensystem(generalized): Compiled without LAPACK");
 #endif
 }
 
@@ -2093,7 +2093,7 @@ void CalcAdjugate(const DenseMatrix &a, DenseMatrix &adja)
 #ifdef MFEM_DEBUG
    if (a.Width() > a.Height() || a.Width() < 1 || a.Height() > 3)
    {
-      mfem_error("CalcAdjugate(...): dimension mismatch");
+      mfem_error("CalcAdjugate(...): unsupported dimensions");
    }
    if (a.Width() != adja.Height() || a.Height() != adja.Width())
    {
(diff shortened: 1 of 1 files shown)


linalg/densemat.cpp
```diff
@@ -401,7 +401,7 @@ void DenseMatrix::InvSymmetricScaling(const Vector & s)
 {
    if (height != width || s.Size() != width)
    {
-      mfem_error("DenseMatrix::SymmetricScaling");
+      mfem_error("DenseMatrix::SymmetricScaling: dimension mismatch");
```
mlstowell, line 404: Should probably read `DenseMatrix::InvSymmetricScaling: dimension mismatch`

linalg/densemat.cpp
```diff
@@ -528,7 +528,7 @@ double DenseMatrix::Weight() const
       double F = d[0] * d[3] + d[1] * d[4] + d[2] * d[5];
       return sqrt(E * G - F * F);
    }
-   mfem_error("DenseMatrix::Weight()");
+   mfem_error("DenseMatrix::Weight(): dimension mismatch");
```
mlstowell, line 531: "dimension mismatch" is not the best description of this error.  Perhaps something like "unsupported dimensions".

linalg/densemat.cpp
```diff
@@ -1164,7 +1164,7 @@ void DenseMatrix::Eigensystem(DenseMatrix &b, Vector &ev,
    MFEM_CONTRACT_VAR(b);
    MFEM_CONTRACT_VAR(ev);
    MFEM_CONTRACT_VAR(evect);
-   mfem_error("DenseMatrix::Eigensystem for generalized eigenvalues");
+   mfem_error("DenseMatrix::Eigensystem: Compiled without LAPACK");
```
mlstowell, line 1167: It would be best to distinguish this message from that on line 1086.  Try to incorporate the fact that this is for the generalized eigenvalue case.

linalg/densemat.cpp
```diff
@@ -2093,11 +2093,11 @@ void CalcAdjugate(const DenseMatrix &a, DenseMatrix &adja)
 #ifdef MFEM_DEBUG
    if (a.Width() > a.Height() || a.Width() < 1 || a.Height() > 3)
    {
-      mfem_error("CalcAdjugate(...)");
+      mfem_error("CalcAdjugate(...): dimension mismatch");
    }
    if (a.Width() != adja.Height() || a.Height() != adja.Width())
    {
-      mfem_error("CalcAdjugate(...)");
+      mfem_error("CalcAdjugate(...): dimension mismatch");
```
mlstowell, line 2100: We should distinguish these messages.  Perhaps simply by adding integers similar to lines 1679 and 1709.

linalg/densemat.cpp
benzwick: I agree, but the first conditional here is
```
if (Height() == Width())
```
then there are some specializations for smaller dimensions where height and width are different so maybe "mismatched or unsupported dimensions" would be better.

linalg/densemat.cpp
benzwick: It looks like the first one checks that both matrices are between 1x1 and 3x3, whereas the second checks that they are the same size and symmetric so I'll change the first error message to "unsupported dimensions".

linalg/densemat.cpp
benzwick: I agree but adding to much information will exceed the character limit for the line. I can fit "(generalized)" in there otherwise we'll need to split the line.

linalg/densemat.cpp
benzwick: Done.

benzwick:
Would it be better to use the `MFEM_ASSERT` [macro](https://github.com/mfem/mfem/blob/b9a40daf5b3b6b5338907c3936941415abd5e4b4/general/error.hpp#L118) for the dimension checks as done e.g. [here](https://github.com/mfem/mfem/blob/b9a40daf5b3b6b5338907c3936941415abd5e4b4/linalg/densemat.cpp#L182) instead?

tzanio:
@mlstowell, were you comments addressed?

tzanio:

PR #1701, opened 2020-08-12
libCEED - add new qfunction context data object, mild perf improvement

jeremylt:
This new interface is ~~in the [final review in libCEED](https://github.com/CEED/libCEED/pull/596)~~ merged in libCEED. This will prevent extra copies of the context data from host -> device and device -> host on every operator application, which should offer a mild performance improvement.

ToDo:
- [x] Update libCEED git hash

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge| 
 | --- | --- | --- | --- | --- | --- | --- | 
| [#1701](https://github.com/mfem/mfem/pull/1701) | @jeremylt | @tzanio | @camierjs + @YohannDudouit | 08/14/20 | ⌛due 08/28/20 | ⌛due 09/04/20 | |

PR #1705, opened 2020-08-13
libCEED - prevent duplicate Ceed objects from MFEM FES

jeremylt:
This PR slightly factors the libCEED integration, using a hash table to look up libCEED Basis and ElemRestriction objects corresponding to a particular MFEM FES.

This prevents excessive copies of identical data on the GPU from repeated libCEED objects associated with a single FES across multiple integrators (mass and diffusion). This will also improve the future libCEED grid transfer operators.

Note, this PR must be merged after #1701 

| [#1705](https://github.com/mfem/mfem/pull/1705) | @jeremylt | @tzanio | @YohannDudouit + @tzanio | 08/16/20 | 08/23/20 | 08/30/20 | |

### DIFF
--- a/fem/libceed/ceed.cpp
+++ b/fem/libceed/ceed.cpp
@@ -35,8 +35,8 @@ extern Ceed ceed;
 
 std::string ceed_path;
 
-extern std::unordered_map<std::string, CeedBasis> ceed_basis_table;
-extern std::unordered_map<std::string, CeedElemRestriction> ceed_restr_table;
+extern CeedBasisMap ceed_basis_map;
+extern CeedRestrMap ceed_restr_map;
 
 }
 
@@ -368,16 +368,16 @@ void InitCeedBasisAndRestriction(const FiniteElementSpace &fes,
    const int P = fe->GetDof();
    const int Q = irm.GetNPoints();
    const int nelem = mesh->GetNE();
-   std::string basis_key = std::to_string(reinterpret_cast<intptr_t>(&fes)) +
-                           std::to_string(reinterpret_cast<intptr_t>(&irm)) +
-                           std::to_string(P) + std::to_string(Q);
-   auto basis_itr = internal::ceed_basis_table.find(basis_key);
-   std::string restr_key = std::to_string(reinterpret_cast<intptr_t>(&fes)) +
-                           std::to_string(P) + std::to_string(nelem);
-   auto restr_itr = internal::ceed_restr_table.find(restr_key);
+   CeedBasisKey basis_key (const_cast<FiniteElementSpace *>(&fes),
+                           const_cast<IntegrationRule *>(&irm),
+                           P, Q);
+   auto basis_itr = internal::ceed_basis_map.find(basis_key);
+   CeedRestrKey restr_key (const_cast<FiniteElementSpace *>(&fes),
+                           P, nelem);
+   auto restr_itr = internal::ceed_restr_map.find(restr_key);
 
    // Init or retreive key values
-   if (basis_itr == internal::ceed_basis_table.end())
+   if (basis_itr == internal::ceed_basis_map.end())
    {
       if (UsesTensorBasis(fes))
       {
@@ -388,13 +388,13 @@ void InitCeedBasisAndRestriction(const FiniteElementSpace &fes,
       {
          InitCeedNonTensorBasis(fes, irm, ceed, basis);
       }
-      internal::ceed_basis_table[basis_key] = *basis;
+      internal::ceed_basis_map[basis_key] = *basis;
    }
    else
    {
       basis = &basis_itr->second;
    }
-   if (restr_itr == internal::ceed_restr_table.end())
+   if (restr_itr == internal::ceed_restr_map.end())
    {
       if (UsesTensorBasis(fes))
       {
@@ -405,7 +405,7 @@ void InitCeedBasisAndRestriction(const FiniteElementSpace &fes,
       {
          InitCeedNonTensorRestriction(fes, irm, ceed, restr);
       }
-      internal::ceed_restr_table[restr_key] = *restr;
+      internal::ceed_restr_map[restr_key] = *restr;
    }
    else
    {
--- a/fem/libceed/ceed.cpp
+++ b/fem/libceed/ceed.cpp
@@ -392,7 +392,7 @@ void InitCeedBasisAndRestriction(const FiniteElementSpace &fes,
    }
    else
    {
-      basis = &basis_itr->second;
+      *basis = basis_itr->second;
    }
    if (restr_itr == internal::ceed_restr_map.end())
    {
@@ -409,7 +409,7 @@ void InitCeedBasisAndRestriction(const FiniteElementSpace &fes,
    }
    else
    {
-      restr = &restr_itr->second;
+      *restr = restr_itr->second;
    }
 }
 
--- a/fem/libceed/ceed.cpp
+++ b/fem/libceed/ceed.cpp
@@ -99,7 +99,6 @@ static void InitCeedNonTensorBasis(const FiniteElementSpace &fes,
    Vector qweight(Q);
    Vector shape_i(P);
    DenseMatrix grad_i(P, dim);
-   CeedInt compstride = fes.GetOrdering()==Ordering::byVDIM ? 1 : fes.GetNDofs();
    const Table &el_dof = fes.GetElementToDofTable();
    Array<int> tp_el_dof(el_dof.Size_of_connections());
    const TensorBasisElement * tfe =
@@ -261,7 +260,6 @@ static void InitCeedTensorBasis(const FiniteElementSpace &fes,
    const TensorBasisElement * tfe =
       dynamic_cast<const TensorBasisElement *>(fe);
    MFEM_VERIFY(tfe, "invalid FE");
-   const Array<int>& dof_map = tfe->GetDofMap();
    const FiniteElement *fe1d =
       fes.FEColl()->FiniteElementForGeometry(Geometry::SEGMENT);
    DenseMatrix shape1d(fe1d->GetDof(), ir.GetNPoints());
@@ -298,7 +296,6 @@ static void InitCeedTensorRestriction(const FiniteElementSpace &fes,
 {
    Mesh *mesh = fes.GetMesh();
    const FiniteElement *fe = fes.GetFE(0);
-   const int order = fes.GetOrder(0);
    const TensorBasisElement * tfe =
       dynamic_cast<const TensorBasisElement *>(fe);
    MFEM_VERIFY(tfe, "invalid FE");
--- a/fem/libceed/ceed.cpp
+++ b/fem/libceed/ceed.cpp
@@ -97,7 +97,6 @@ static void InitCeedNonTensorBasisAndRestriction(const FiniteElementSpace &fes,
    Vector qweight(Q);
    Vector shape_i(P);
    DenseMatrix grad_i(P, dim);
-
    CeedInt compstride = fes.GetOrdering()==Ordering::byVDIM ? 1 : fes.GetNDofs();
    const Table &el_dof = fes.GetElementToDofTable();
    Array<int> tp_el_dof(el_dof.Size_of_connections());
@@ -178,13 +177,16 @@ static void InitCeedNonTensorBasisAndRestriction(const FiniteElementSpace &fes,
          }
       }
    }
-   CeedBasisCreateH1(ceed, GetCeedTopology(fe->GetGeomType()), fes.GetVDim(),
-                     fe->GetDof(), ir.GetNPoints(), shape.GetData(),
-                     grad.GetData(), qref.GetData(), qweight.GetData(), basis);
-   CeedElemRestrictionCreate(ceed, mesh->GetNE(), fe->GetDof(), fes.GetVDim(),
-                             compstride, (fes.GetVDim())*(fes.GetNDofs()),
-                             CEED_MEM_HOST, CEED_COPY_VALUES,
-                             tp_el_dof.GetData(), restr);
+   if (basis)
+      CeedBasisCreateH1(ceed, GetCeedTopology(fe->GetGeomType()), fes.GetVDim(),
+                        fe->GetDof(), ir.GetNPoints(), shape.GetData(),
+                        grad.GetData(), qref.GetData(), qweight.GetData(), basis);
+
+   if (restr)
+      CeedElemRestrictionCreate(ceed, mesh->GetNE(), fe->GetDof(), fes.GetVDim(),
+                                compstride, (fes.GetVDim())*(fes.GetNDofs()),
+                                CEED_MEM_HOST, CEED_COPY_VALUES,
+                                tp_el_dof.GetData(), restr);
 }
 
 static void InitCeedTensorBasisAndRestriction(const FiniteElementSpace &fes,
@@ -223,11 +225,14 @@ static void InitCeedTensorBasisAndRestriction(const FiniteElementSpace &fes,
          grad1d(j, i) = grad_i(dof_map_1d[j], 0);
       }
    }
-   CeedBasisCreateTensorH1(ceed, mesh->Dimension(), fes.GetVDim(), order + 1,
-                           ir.GetNPoints(), shape1d.GetData(),
-                           grad1d.GetData(), qref1d.GetData(),
-                           qweight1d.GetData(), basis);
+   if (basis)
+      CeedBasisCreateTensorH1(ceed, mesh->Dimension(), fes.GetVDim(), order + 1,
+                              ir.GetNPoints(), shape1d.GetData(),
+                              grad1d.GetData(), qref1d.GetData(),
+                              qweight1d.GetData(), basis);
 
+   if (!restr)
+      return;
    CeedInt compstride = fes.GetOrdering()==Ordering::byVDIM ? 1 : fes.GetNDofs();
    const Table &el_dof = fes.GetElementToDofTable();
    Array<int> tp_el_dof(el_dof.Size_of_connections());
--- a/fem/libceed/ceed.cpp
+++ b/fem/libceed/ceed.cpp
@@ -251,15 +251,6 @@ static void InitCeedNonTensorRestriction(const FiniteElementSpace &fes,
                              tp_el_dof.GetData(), restr);
 }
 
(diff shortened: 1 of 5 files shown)


jeremylt:
Note - this PR will require some minor coordination with libCEED. We will need expand our avaliable key tuple sizes.

jeremylt:
I'll be following up this PR with a PR to add libCEED grid transfer operators. Without this PR, we'd have double the number of libCEED objects holding copies of FES data with the libCEED grid transfer operators.

As it is currently, any single FES used in multiple libCEED integrators has duplicate libCEED objects.

fem/libceed/ceed.cpp
```diff
@@ -35,6 +35,9 @@ extern Ceed ceed;
 
 std::string ceed_path;
 
+extern khash_t(basis) *ceed_basis_from_fes;
+extern khash_t(restr) *ceed_restr_from_fes;
```
tzanio, line 39: Hmmm... what these? Where are they defined?

general/device.cpp
```diff
@@ -154,6 +158,17 @@ Device::~Device()
    {
       free(device_option);
 #ifdef MFEM_USE_CEED
+      // Destroy FES -> CeedBasis, CeedElemRestriction hash table contents
+      CeedBasis basis;
+      kh_foreach_value(internal::ceed_basis_from_fes,
+                       basis, CeedBasisDestroy(&basis));
+      kh_destroy(basis, internal::ceed_basis_from_fes);
```
tzanio, line 165: Same here -- if these are internal libCEED functions they need to be namespaced.

More importantly, why should a libCEED application even know about them?

fem/libceed/ceed.cpp
jeremylt: These functions are part of Khash. Khash is a header only hashing library that we have been using in libCEED. These functions are defined in `khash.h` and `ceed-hash.h` in libCEED's include dir.

If MFEM's native hash tables can map to arbitrary types instead of only integers (this did not look to be the case), I can use that instead in this PR.

fem/libceed/ceed.cpp
```diff
@@ -223,11 +228,16 @@ static void InitCeedTensorBasisAndRestriction(const FiniteElementSpace &fes,
...
    }
-   CeedBasisCreateTensorH1(ceed, mesh->Dimension(), fes.GetVDim(), order + 1,
-                           ir.GetNPoints(), shape1d.GetData(),
-                           grad1d.GetData(), qref1d.GetData(),
-                           qweight1d.GetData(), basis);
+   if (basis)
+      CeedBasisCreateTensorH1(ceed, mesh->Dimension(), fes.GetVDim(), order + 1,
+                              ir.GetNPoints(), shape1d.GetData(),
+                              grad1d.GetData(), qref1d.GetData(),
+                              qweight1d.GetData(), basis);
```
YohannDudouit, line 235: This `if` statement could start line 209.

fem/libceed/ceed.cpp
```diff
@@ -223,11 +228,16 @@ static void InitCeedTensorBasisAndRestriction(const FiniteElementSpace &fes,
...
 
+   if (!restr)
+   {
+      return;
+   }
```
YohannDudouit, line 240: I don't really like this very C style programming.

fem/libceed/ceed.cpp
```diff
@@ -178,13 +180,16 @@ static void InitCeedNonTensorBasisAndRestriction(const FiniteElementSpace &fes,
...
+   if (basis)
+      CeedBasisCreateH1(ceed, GetCeedTopology(fe->GetGeomType()), fes.GetVDim(),
+                        fe->GetDof(), ir.GetNPoints(), shape.GetData(),
+                        grad.GetData(), qref.GetData(), qweight.GetData(), basis);
+
+   if (restr)
+      CeedElemRestrictionCreate(ceed, mesh->GetNE(), fe->GetDof(), fes.GetVDim(),
+                                compstride, (fes.GetVDim())*(fes.GetNDofs()),
+                                CEED_MEM_HOST, CEED_COPY_VALUES,
+                                tp_el_dof.GetData(), restr);
```
YohannDudouit, line 192: This is really not clean to put the `if` at this level, you created everything to call these functions right before, you shouldn't do the computation if you already know that you're not going to use the result.

fem/libceed/ceed.cpp
```diff
@@ -258,14 +268,56 @@ void InitCeedBasisAndRestriction(const FiniteElementSpace &fes,
                                  Ceed ceed, CeedBasis *basis,
                                  CeedElemRestriction *restr)
 {
+   // Check for FES -> basis, restriction in hash tables
+   const Mesh *mesh = fes.GetMesh();
+   const FiniteElement *fe = fes.GetFE(0);
+   const int P = fe->GetDof();
+   const int Q = irm.GetNPoints();
+   const int nelem = mesh->GetNE();
+   CeedHashIJKLKey basis_key = {(int)(long)&fes, (int)(long)&irm, P, Q};
```
YohannDudouit, line 277: That's not how you cast in a C++ code.

jeremylt:
I've made a major refactor to address the comments so far. 

One point of note - now `InitCeed*TensorBasisAndRestriction`warns as unused (which they are), but I'm hesitant to just remove these two functions.

fem/libceed/ceed.cpp
```diff
@@ -253,19 +348,68 @@ static void InitCeedTensorBasisAndRestriction(const FiniteElementSpace &fes,
...
+   const int P = fe->GetDof();
+   const int Q = irm.GetNPoints();
+   const int nelem = mesh->GetNE();
+   std::string basis_key = std::to_string(reinterpret_cast<intptr_t>(&fes)) +
+                           std::to_string(reinterpret_cast<intptr_t>(&irm)) +
+                           std::to_string(P) + std::to_string(Q);
+   auto basis_itr = internal::ceed_basis_table.find(basis_key);
+   std::string restr_key = std::to_string(reinterpret_cast<intptr_t>(&fes)) +
+                           std::to_string(P) + std::to_string(nelem);
+   auto restr_itr = internal::ceed_restr_table.find(restr_key);
```
YohannDudouit, line 377: I think it would be cleaner to use a tuple rather than doing the hash function manually each time you want to create a key, that's bug prone:
```c++
// For CeedBasis
using BasisKey = std::tuple<FiniteElementSpace*, IntegrationRule*, int, int>;
struct BasisHash
{
   std::size_t operator()(const BasisKey& k) const
   {
      return ...; // use std::get<n>(k) for computing the hash
   }
};
using BasisMap = std::unordered_map<const BasisKey, CeedBasis, BasisHash>;
auto basis_key = std::make_tuple(&fes,&irm,P,Q);

// For CeedRestriction
using RestrKey = std::tuple<FiniteElementSpace*, int, int>;
struct RestrHash
{
   std::size_t operator()(const RestrKey& k) const
   {
      return ...; // use std::get<n>(k) for computing the hash
   }
};
using RestrMap = std::unordered_map<const RestrKey, CeedRestriction, RestrHash>;
auto restr_key = std::make_tuple(&fes,P,nelem);
```

jeremylt: I thought about that and its what I had at first. I just wasn't sure that we wanted to also build out the hashing and comparison operators.

This will, amusingly enough, result in me duplicating the code already included libCEED for hashing keys tuples.

general/device.hpp
```diff
@@ -15,6 +15,12 @@
 #include "globals.hpp"
 #include "mem_manager.hpp"
 
+
+#ifdef MFEM_USE_CEED
+// FES -> CeedBasis, CeedElemRestriction hash tables
+#include <unordered_map>
+#endif
```
YohannDudouit, line 22: You don't use `unordered_map` in this file, this include can go directly in the cpp file.

jeremylt: I put it here so that it's included in both `device.cpp` and `ceed.cpp`? Is including separately in those two files preferable?

jeremylt: I might also need it to get the new function signatures in `device.cpp` and `ceed.cpp` for the tuple->hash function I'm going to re-implement here, right?

YohannDudouit: In general it's better to include files in the `cpp` files, rather than the `hpp`, if you can. This prevents including unwanted libraries when including a `hpp` file, which can result in various forms of trouble.

fem/libceed/ceed.cpp
YohannDudouit: You don't need to build the comparison, only the hash doesn't have a default implementation for `tuple`.

fem/libceed/ceed.cpp
```diff
@@ -35,6 +35,9 @@ extern Ceed ceed;
 
 std::string ceed_path;
 
+extern std::unordered_map<std::string, CeedBasis> ceed_basis_table;
+extern std::unordered_map<std::string, CeedElemRestriction> ceed_restr_table;
```
YohannDudouit, line 39: I think these should be in the `hpp` file, that's why you need to include `unordered_map` in device. It's strange to put a declaration in a `cpp` file.

general/device.cpp
```diff
@@ -33,6 +33,10 @@ occa::device occaDevice;
 
 #ifdef MFEM_USE_CEED
 Ceed ceed = NULL;
+
+// Initalize FES -> CeedBasis, CeedElemRestriction hash tables
+std::unordered_map<std::string, CeedBasis> ceed_basis_table;
+std::unordered_map<std::string, CeedElemRestriction> ceed_restr_table;
```
YohannDudouit, line 39: Use `BasisMap` and `RestrMap` here.

fem/libceed/ceed.cpp
YohannDudouit: Not completely sure of what I'm saying though.

jeremylt:
What about this version?

jeremylt:
Huh, somethings off with ex26.

Investigating

fem/libceed/ceed.cpp
```diff
@@ -253,19 +348,68 @@ static void InitCeedTensorBasisAndRestriction(const FiniteElementSpace &fes,
...
+   const Mesh *mesh = fes.GetMesh();
+   const FiniteElement *fe = fes.GetFE(0);
+   const int P = fe->GetDof();
+   const int Q = irm.GetNPoints();
+   const int nelem = mesh->GetNE();
+   CeedBasisKey basis_key (const_cast<FiniteElementSpace *>(&fes),
+                           const_cast<IntegrationRule *>(&irm),
+                           P, Q);
+   auto basis_itr = internal::ceed_basis_map.find(basis_key);
+   CeedRestrKey restr_key (const_cast<FiniteElementSpace *>(&fes),
```
YohannDudouit, line 375: Instead of casting you should just change the type of your keys to `const FiniteElementSpace*` and `const IntegrationRule*`

fem/libceed/ceed.hpp
```diff
@@ -27,7 +30,45 @@ class GridFunction;
 class IntegrationRule;
 class Coefficient;
 
-namespace internal { extern Ceed ceed; } // defined in device.cpp
+// Hash table for CeedBasis
+using CeedBasisKey =
+   std::tuple<FiniteElementSpace*, IntegrationRule*, int, int>;
```
YohannDudouit, line 35: here use `const FiniteElementSpace*` and `const IntegrationRule*`

fem/libceed/ceed.hpp
```diff
@@ -27,7 +30,45 @@ class GridFunction;
...
+                                                reinterpret_cast<CeedHash64_t>(std::get<1>(k)))),
+                             CeedHashCombine(CeedHashInt(std::get<2>(k)),
+                                             CeedHashInt(std::get<3>(k))));
+   }
+};
+using CeedBasisMap =
+   std::unordered_map<const CeedBasisKey, CeedBasis, CeedBasisHash>;
+
+// Hash table for CeedElemRestriction
+using CeedRestrKey = std::tuple<FiniteElementSpace*, int, int>;
```
YohannDudouit, line 52: same here, use `const FiniteElementSpace*`.

general/device.cpp
```diff
@@ -169,7 +183,7 @@ Device::~Device()
 
 void Device::Configure(const std::string &device, const int dev)
 {
-   // If a device was configured via the environment, skip the configuration,
+   // If device a was configured via the environment, skip the configuration,
```
YohannDudouit, line 186: I think this change was unintentional.

fem/libceed/ceed.cpp
jeremylt: Yeah that's better, and an obvious change I should have seen.

jeremylt:
This leaves 
```
fem/libceed/ceed.cpp(254): warning: function "mfem::InitCeedNonTensorBasisAndRestriction" was declared but never referenced

fem/libceed/ceed.cpp(351): warning: function "mfem::InitCeedTensorBasisAndRestriction" was declared but never referenced
```
These functions are not in `ceed.hpp`, which should mean that these functions are not used outside of `ceed.cpp` and its fine to remove them, right?

fem/libceed/ceed.cpp
```diff
@@ -253,19 +348,65 @@ static void InitCeedTensorBasisAndRestriction(const FiniteElementSpace &fes,
...
 }
 
+static void InitCeedTensorBasisAndRestriction(const FiniteElementSpace &fes,
+                                              const IntegrationRule &ir,
+                                              Ceed ceed, CeedBasis *basis,
+                                              CeedElemRestriction *restr)
+{
+   InitCeedTensorBasis(fes, ir, ceed, basis);
+   InitCeedTensorRestriction(fes, ir, ceed, restr);
+}
```
YohannDudouit, line 358: You can remove it, it's a static function so it's not going to create backward compatibility issues, and it also generates warnings at compilation.

fem/libceed/ceed.cpp
```diff
@@ -178,19 +245,24 @@ static void InitCeedNonTensorBasisAndRestriction(const FiniteElementSpace &fes,
...
-                                              Ceed ceed, CeedBasis *basis,
-                                              CeedElemRestriction *restr)
+static void InitCeedNonTensorBasisAndRestriction(const FiniteElementSpace &fes,
+                                                 const IntegrationRule &ir,
+                                                 Ceed ceed, CeedBasis *basis,
+                                                 CeedElemRestriction *restr)
+{
+   InitCeedNonTensorBasis(fes, ir, ceed, basis);
+   InitCeedNonTensorRestriction(fes, ir, ceed, restr);
+}
```
YohannDudouit, line 261: Remove this one too, to avoid warnings.

tzanio:

fem/libceed/ceed.cpp
```diff
@@ -124,7 +125,74 @@ static void InitCeedNonTensorBasisAndRestriction(const FiniteElementSpace &fes,
...
+   const int dim = mesh->Dimension();
+   const int P = fe->GetDof();
+   const int Q = ir.GetNPoints();
+   DenseMatrix shape(P, Q);
+   Vector grad(P*dim*Q);
+   DenseMatrix qref(dim, Q);
+   Vector qweight(Q);
+   Vector shape_i(P);
+   DenseMatrix grad_i(P, dim);
+   CeedInt compstride = fes.GetOrdering()==Ordering::byVDIM ? 1 : fes.GetNDofs();
```
YohannDudouit, line 170: unused.

fem/libceed/ceed.cpp
```diff
@@ -178,19 +245,15 @@ static void InitCeedNonTensorBasisAndRestriction(const FiniteElementSpace &fes,
...
-static void InitCeedTensorBasisAndRestriction(const FiniteElementSpace &fes,
-                                              const IntegrationRule &ir,
-                                              Ceed ceed, CeedBasis *basis,
-                                              CeedElemRestriction *restr)
+static void InitCeedTensorBasis(const FiniteElementSpace &fes,
+                                const IntegrationRule &ir,
+                                Ceed ceed, CeedBasis *basis)
 {
    Mesh *mesh = fes.GetMesh();
    const FiniteElement *fe = fes.GetFE(0);
```
YohannDudouit, line 259: dof_map is unused

fem/libceed/ceed.cpp
```diff
@@ -227,6 +290,29 @@ static void InitCeedTensorBasisAndRestriction(const FiniteElementSpace &fes,
...
                            qweight1d.GetData(), basis);
+}
+
+static void InitCeedTensorRestriction(const FiniteElementSpace &fes,
+                                      const IntegrationRule &ir,
+                                      Ceed ceed, CeedElemRestriction *restr)
+{
+   Mesh *mesh = fes.GetMesh();
+   const FiniteElement *fe = fes.GetFE(0);
+   const int order = fes.GetOrder(0);
```
YohannDudouit, line 301: unused.

fem/libceed/ceed.cpp
```diff
@@ -258,14 +344,51 @@ void InitCeedBasisAndRestriction(const FiniteElementSpace &fes,
...
+   // Check for FES -> basis, restriction in hash tables
+   const Mesh *mesh = fes.GetMesh();
+   const FiniteElement *fe = fes.GetFE(0);
+   const int P = fe->GetDof();
+   const int Q = irm.GetNPoints();
+   const int nelem = mesh->GetNE();
+   CeedBasisKey basis_key(&fes, &irm, P, Q);
+   auto basis_itr = internal::ceed_basis_map.find(basis_key);
+   CeedRestrKey restr_key(&fes, P, nelem);
+   auto restr_itr = internal::ceed_restr_map.find(restr_key);
```
YohannDudouit, line 353: These keys don't seem to uniquely define a basis and restriction and result into a bug.

jeremylt: The key was ok (I think) but a stronger key doesn't hurt.

It was a good old 'double free' type issue.

tzanio:
I am a little confused -- what exactly happened here?

- Should this be re-reviewed?
- Should @camierjs be added as a reviewer from #1701?

YohannDudouit:
@tzanio I tested my test suit for libCEED (#1719) on this PR, which made me realize there was a bug with `GridFunctionCoefficient`, I thought it was an issue with the hash-keys, but it was a double free issue that @jeremylt found, now this PR works with #1719.

YohannDudouit:
I just tested #1719 on Lassen too, all the CEED tests are passing, meaning that this PR works as far as I can tell.

PR #1714, opened 2020-08-20
Enable HIP backend in libCEED

nbeams:
This PR allows MFEM to use libCEED + HIP. 

_Minimum working example to test:_

- Install libCEED.  To keep the build/linking process as simple as possible, I built libCEED with the `clang` and `clang++` in `/opt/rocm/llvm/bin` (set through `CC` and `CXX` when building libCEED). 

- Build MFEM from this branch with the following options: 
```
MFEM_USE_HIP=YES
MFEM_USE_CEED=YES
CEED_DIR=/path/to/libCEED
```
You can also set `HIP_ARCH` accordingly for your system. (I was using gfx906.)

The previous error with 3D problems has been resolved; all meshes should work with MFEM `ex1` and libCEED + HIP.
The MFEM device string is `ceed-hip`.

| [#1714](https://github.com/mfem/mfem/pull/1714) | @nbeams | @tzanio | @YohannDudouit + @camierjs + @barker29 + @artv3 | 09/17/20 | 10/13/20 | 10/18/20 | |


nbeams:
MFEM team, please feel free to tag anyone else you think would be interested or could help.

general/device.cpp
```diff
@@ -504,6 +508,18 @@ void Device::Setup(const int device)
          CeedDeviceSetup(device_option);
       }
    }
+   if (Allows(Backend::CEED_HIP))
+   {
+      if (!device_option)
+      {
+         // NOTE: libCEED's /gpu/hip/ref backend is non-deterministic!
+         CeedDeviceSetup("/gpu/hip/ref");
```
jeremylt, line 516: BTW, my PR in libCEED earlier today was so we can use `/gpu/hip` here and get the fastest (lowest priorityvvalue) HIP backend to date as we make new ones like is node with the CPU impl

nbeams: Yes, I was just following what CUDA had, but since HIP support still has a long way to go in terms of development, it would probably make sense to not have to remember to change this all the time (though we'd still have to update the non-deterministic comment, if it's important to have it).

jeremylt: Yeah, it comes into play for the MFEM test suite, I believe, and perhaps some applications

I'm not sure why the GPU impl doesn't do the same, but it probably should

general/error.hpp
```diff
@@ -151,6 +151,11 @@ void mfem_warning(const char *msg = NULL);
       printf(msg);             \
       asm("trap;");            \
```
tzanio, line 152: The above `asm("trap;");` line didn't work at all with, HIP, correct?

tzanio: cc: @dylan-copeland, @artv3

nbeams: Yes, that's correct.  I tried some variations on the compiler's suggested replacement (`s_trap`, I believe it was) but I couldn't get one that worked, so I just removed the line.

artv3: For hip abort should work with device code:
``` 
#if defined(__HIP_DEVICE_COMPILE__)
  abort();
#endif
```

nbeams: Thanks @artv3. Is `__HIP_DEVICE_COMPILE__` something that should be detected by my HIP configuration, or something I should set when building?  When I tried adding the above code, I got the same error as before I added anything related to HIP (error because it's trying to call the standard MFEM abort, so it says it's trying to call a host function from the device) which leads me to believe it didn't think `__HIP_DEVICE_COMPILE__` was defined

artv3: ``__HIP_DEVICE_COMPILE__`` should be automatically defined when the compiler is compiling for device code. 

See: Identifying Current Compilation Pass: Host or Device in https://rocmdocs.amd.com/en/latest/Programming_Guides/HIP-porting-guide.html

nbeams: Hmm, I'm not sure what's going on.  I keep getting errors like 
`mfem/linalg/sparsemat.cpp:2370:16: error: reference to __host__ function 'mfem_warning' in __device__ function`

Looks like it's coming from an `MFEM_ABORT` call inside `MFEM_FORALL`

artv3: Was this on Corona?

nbeams: No, on a local machine here at ICL.

moes1: What's going on here is that when hip-clang does its host and compile passes all of the code is ALWAYS visible. So when it is doing the host compilation pass __HIP_DEVICE_COMPILE__ is not defined, but it still checks that the __device__ functions are sane in that situation. This makes using macros much more difficult. 

In this case when doing the host pass but looking at the device code, the check in the macro for #defined(__HIP_DEVICE_COMPILE__) says no, and it goes on to try and place the following chain of macros into __device__ code:

#define MFEM_ABORT_KERNEL(msg) MFEM_ABORT(msg) 
#define MFEM_ABORT(msg) _MFEM_MESSAGE("MFEM abort: " << msg, 0)
#define _MFEM_MESSAGE(msg, warn)                                        \
   {                                                                    \
      std::ostringstream mfemMsgStream;                                 \
      mfemMsgStream << std::setprecision(16);                           \
      mfemMsgStream << std::setiosflags(std::ios_base::scientific);     \
      mfemMsgStream << msg << MFEM_LOCATION;                            \
      if (!(warn))                                                      \
         mfem::mfem_error(mfemMsgStream.str().c_str());                 \
      else                                                              \
         mfem::mfem_warning(mfemMsgStream.str().c_str());               \
   }


and obviously the MFEM_MESSAGE macro contains lots of host code the GPU doesn't understand. 

A way to get this to work with hip clang would be something like this:


#if defined(MFEM_USE_HIP)
template <typename T>
__host__ void abort_msg(T & msg)
{
   MFEM_ABORT(msg);
}

template <typename T>
__device__ void abort_msg(T & msg)
{
   abort();
}

#else
template <typename T>
__host__ __device__ void abort_msg(T & msg)
{
#if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE)
#else
   MFEM_ABORT(msg);
#endif
}
#endif
#if defined(__CUDA_ARCH__)
#define MFEM_ABORT_KERNEL(msg) \
   {                           \
      printf(msg);             \
      asm("trap;");            \
   }
#else
#define MFEM_ABORT_KERNEL(msg) \
   {                           \
      abort_msg(msg);          \
   }
#endif
#endif


You have to separate the nvcc and hip-clang cases because with nvcc you cannot have __host__ and __device__ functions with the same function signature (with clang the __host__ and __device__ tags are part of the function signature but they aren't with nvcc).

nbeams: @moes1, thanks for the explanation.   In f2d5e1a1d3768864aef09c0778f324207b1b8a26 I tried incorporating this in a way that had fewer changes to what was in MFEM currently -- but I'm not entirely sure it is functionally the same as what you suggested.  And, if I understand correctly, if MFEM-HIP aborts from inside a device kernel, it still won't print the error message, unlike MFEM-CUDA?  Or no?

general/device.cpp
nbeams: This has been changed in 28d366b77c726963ac69d930685b0b664d62fdfb 
I also went ahead and removed the comment about non-deterministic behavior in anticipation of libCEED's PR 623 being merged before this one.

nbeams:
Good news!  After updating this branch with the recent changes in master, the reported error with `ex1` and 3D meshes seems to be fixed.  I have confirmed with `rocprof` that it is, indeed, using the `hip-ref` backend of libCEED.

Though, full disclosure: the AMD GPU I was using for testing was moved to a different machine, so I am now working on a different CPU.  The new machine also has ROCm 3.7 instead of 3.5.  However, @moes1 mentioned that he was still seeing errors with ROCm 3.7 before, so it could be that one of the changes in the master branch fixed the issue?  If someone else could test this updated branch locally to reproduce, that would be much appreciated.

general/error.hpp
moes1: Yes I removed the printf from the __device__ abort_msg implementation. You could add it and then it would print from within the kernel when it aborts. However I think the recommended practice then would be to call it from the abort_msg function decorated with "__attribute__((noinline))" because if the compiler attempts to inline the printf calls it can cause problems. 

So:

__attribute__((noinline)) __device__ void abort_msg(T & msg)
{
   printf(msg);
   abort();
}

moes1:
@nbeams I have verified that with this update on the machine I had been testing on there is now no longer an issue, all the tests listed above pass.

general/error.hpp
moes1: Actually after further thought, I would recommend not using printf with hip unless debugging if possible right now...

nbeams:
> @nbeams I have verified that with this update on the machine I had been testing on there is now no longer an issue, all the tests listed above pass.

Great, thanks! 

MFEM team, are there any other tests I should check?  Was the goal to review this with the intent of merging, or not?

YohannDudouit:
@nbeams If this works we should review and merge it.

nbeams:
@YohannDudouit Yes, it appears to be working now. I tested both the hexahedron `beam-hex` mesh and tetrahedron `escher` mesh, and `ceed-hip` gives the same output as `ceed-cpu`.

I am not sure which of the recent changes in the master branch should be credited with this happy turn of events...

YohannDudouit:
@nbeams Can you change from `WIP` label to `ready-for-review` if you don't want to add anything else.

barker29:
I'm happy to review this but I don't think I have access to a HIP-enabled machine. Is there one I can use here at LLNL?

tzanio:
> Is there one I can use here at LLNL?

Corona

artv3:
I might be missing something, but I am hitting the following error: 
``fem/libceed/ceed.hpp:21:10: fatal error: 'ceed-hash.h' file not found
#include <ceed-hash.h>``

I installed libceed like this: 
```
make install prefix=/g/g17/vargas45/MFEM_REVIEW/libCEED-install
```
Then I tried to build mfem as follows: 
```
make config MFEM_USE_HIP=YES MFEM_USE_CEED=YES CEED_DIR=/g/g17/vargas45/MFEM_REVIEW/libCEED-install 
```

jeremylt:
What version of libCEED did you install?

artv3:
> What version of libCEED did you install? The hash file was not in v0.6, so you'll need the latest libCEED main branch (which will be released as v0.7 as soon as we make new Python wheels).

I tried using libCEED main, hash: 8bbde58485a37954da92a5c228699254b35dc377 . Perhaps I made a mistake? I can try again.

nbeams:
Does `ceed-hash.h` need to be added to the copy list for installing out-of-source in libCEED?  Right now, it seems the Makefile only installs `ceed.h` and `ceedf.h` in `install-dir/include`.

@artv3 , it should work if you install libCEED in its source directory (or just do a plain `make` since that will build the library, which is what we need).

artv3:
> Does `ceed-hash.h` need to be added to the copy list for installing out-of-source in libCEED? Right now, it seems the Makefile only installs `ceed.h` and `ceedf.h` in `install-dir/include`.
> 
> @artv3 , it should work if you install libCEED in its source directory (or just do a plain `make` since that will build the library, which is what we need).

Thanks @nbeams , it does look like ``ceed-hash.h`` pointing to the source after typing make worked for me.

artv3 (approved):
This is cool. Beyond this PR it would be super nice if there were timers in MFEM examples to compare run-time of different backends.

jeremylt:
Install for the ceed-hash header is fixed now.

fem/libceed/ceed.cpp
```diff
@@ -547,7 +547,8 @@ void CeedAddMultPA(const CeedData *ceedDataPtr,
    CeedScalar *y_ptr;
    CeedMemType mem;
    CeedGetPreferredMemType(internal::ceed, &mem);
-   if ( Device::Allows(Backend::CUDA) && mem==CEED_MEM_DEVICE )
+   if ( (Device::Allows(Backend::CUDA) || Device::Allows(Backend::HIP))
+        && mem==CEED_MEM_DEVICE )
```
YohannDudouit, line 551:
```suggestion
   if ( Device::Allows(Backend::DEVICE_MASK) && mem==CEED_MEM_DEVICE )
```

fem/libceed/ceed.cpp
```diff
@@ -575,7 +576,8 @@ void CeedAssembleDiagonalPA(const CeedData *ceedDataPtr,
    CeedScalar *d_ptr;
    CeedMemType mem;
    CeedGetPreferredMemType(internal::ceed, &mem);
-   if ( Device::Allows(Backend::CUDA) && mem==CEED_MEM_DEVICE )
+   if ( (Device::Allows(Backend::CUDA) || Device::Allows(Backend::HIP))
+        && mem==CEED_MEM_DEVICE )
```
YohannDudouit, line 580:
```suggestion
   if ( Device::Allows(Backend::DEVICE_MASK) && mem==CEED_MEM_DEVICE )
```

fem/libceed/ceed.hpp
```diff
@@ -197,6 +197,7 @@ void CeedAssembleDiagonalPA(const CeedData *ceedDataPtr,
 inline bool DeviceCanUseCeed()
 {
    return Device::Allows(Backend::CEED_CUDA) ||
+          Device::Allows(Backend::CEED_HIP) ||
```
YohannDudouit, line 200:
```suggestion
   return Device::Allows(Backend::DEVICE_MASK)
```

YohannDudouit (changes requested):
This looks good to me, I'll test it and approve after these small changes.

fem/libceed/ceed.hpp
nbeams: In this one, do I really want to change it to `DEVICE_MASK`?  Isn't it specifically checking for `CEED_CUDA` or `CEED_HIP`, not any device usage?  Should I replace all three checks with `CEED_MASK` instead?

nbeams: Or wait, I see, the `CEED_CPU` check goes with the check on not allowing `DEVICE` or `OMP`.  So should there be a `CEED_DEVICE_MASK` to check for `CEED_CUDA` or `CEED_HIP` together?

YohannDudouit: You're right in this one, `CEED_MASK` is the correct one. I don't really understand the logic for `(Device::Allows(Backend::CEED_CPU) && !Device::Allows(Backend::DEVICE_MASK|Backend::OMP_MASK));`, I think it's a bit overkill, we could probably replace everything with:
```c++
return Device::Allows(Backend::CEED_MASK);
```

nbeams: I did this in b267750431f09086aefa0494e2ab366ba2ef2620

fem/libceed/ceed.cpp
nbeams: These have been fixed in b267750431f09086aefa0494e2ab366ba2ef2620

general/device.hpp
```diff
@@ -61,31 +61,35 @@ struct Backend
           backend. Enabled when MFEM_USE_CEED = YES and MFEM_USE_CUDA = YES.
           NOTE: The current default libCEED GPU backend is non-deterministic! */
```
barker29, line 62: `GPU` -> `CUDA` in this comment.

nbeams: Good catch, I updated it in aa278f92a3912fd5a648ed470c84f70668b708f8. 
I also removed the non-deterministic comment for the libCEED HIP backend, since due to recent changes in libCEED, the default HIP backend *is* deterministic now.

tzanio:
@YohannDudouit and @camierjs -- can you please review?

camierjs (approved):
Works great on Corona: thank you for the integration @nbeams !
The 3D mesh with libCEED HIP results are good, we can update the table.

YohannDudouit (approved):
It works, thank you for doing this @nbeams .
MFEM is about 7sec faster than libCEED, I suspect hiprtc since it's a constant time slower.
I merged this branch into #1719 to run the tests, most of them pass, but I have issues with `VectorDiffusion` in 3D. Since this PR is about enabling the hip backends of libCEED, this shouldn't stop the merge, but this should be investigated.
The tests can be run with after merging #1719 :
```
cd tests/unit
make ceed_tests
./ceed_tests --device ceed-hip
```
With a constant coefficient:
```
:0:rocdevice.cpp            :2159: 3010156718459 us: Device::callbackQueue aborting with status: 0x100f
CEED
/g/g91/dudouit1/Projects/mfem/tests/unit/ceed/test_ceed.cpp:165

/g/g91/dudouit1/Projects/mfem/tests/unit/ceed/test_ceed.cpp:175: FAILED:
  assembly: PARTIAL
  coeff_type: Const
  pb: VectorDiffusion
  order: 1
  mesh: ../../data/inline-hex.mesh
  SIGABRT - Abort (abnormal termination) signal
```
With a `GridFunctionCoefficient`:
```
:0:rocdevice.cpp            :2159: 3010928889699 us: Device::callbackQueue aborting with status: 0x100f
CEED
/g/g91/dudouit1/Projects/mfem/tests/unit/ceed/test_ceed.cpp:165

/g/g91/dudouit1/Projects/mfem/tests/unit/ceed/test_ceed.cpp:176: FAILED:
  assembly: PARTIAL
  coeff_type: Grid
  pb: VectorDiffusion
  order: 1
  mesh: ../../data/inline-hex.mesh
  SIGABRT - Abort (abnormal termination) signal
```
With a coefficient evaluated at quadrature points:
```
:0:rocdevice.cpp            :2159: 3011470916617 us: Device::callbackQueue aborting with status: 0x100f
CEED
/g/g91/dudouit1/Projects/mfem/tests/unit/ceed/test_ceed.cpp:165

/g/g91/dudouit1/Projects/mfem/tests/unit/ceed/test_ceed.cpp:176: FAILED:
  assembly: PARTIAL
  coeff_type: Quad
  pb: VectorDiffusion
  order: 1
  mesh: ../../data/inline-hex.mesh
  SIGABRT - Abort (abnormal termination) signal
```

tzanio:

tzanio (approved):
Thank you so much for this useful contribution @nbeams!

@YohannDudouit, @camierjs, @barker29, @artv3, @moes1 -- thank you for the help and the review!

PR #1380, opened 2020-03-27
CVODES Adjoint-Sensitivity Support

jonwong12:
Adds basic support for CVODES adjoint-sensitivity analysis to MFEM's existing SUNDIALS support.

- Added methods to perform a backwards adjoint problem for a given forward problem
- Added some helper methods and refactor `sundials.cpp` to facilitate allocating and using NVectors in mfem
- Added support for root-finding
- Includes MFEM-ified SUNDIALS CVODES examples (serial and parallel)

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge| 
 | --- | --- | --- | --- | --- | --- | --- | 
| [#1380](https://github.com/mfem/mfem/pull/1380) | @jonwong12 | @tzanio | @jandrej + @bslazarov + @jamiebramwell | 03/29/20 | 06/23/20 | 07/21/20 | |

PR #1726, opened 2020-08-26
Add support for SUNDIALS GPU features [sundials-gpu-dev]

balos1:
This PR adds support for SUNDIALS CUDA features. These changes will be easily extendable to support future GPU support provided by SUNDIALS, e.g. for HIP. To keep things clean, this creates a `SundialsNVector` class to interface with the SUNDIALS `N_Vector`. This class inherits from the `mfem::Vector` class, and follows the pattern used in MFEM interfaces to other libraries such as PETSc and hypre. The PR also synchronizes the Examples 9/9p and the SUNDIALS modifications.

Building MFEM with `MFEM_USE_CUDA=ON` and `MFEM_USE_SUNDIALS=ON` will require the upcoming SUNDIALS 5.4.0 release. _With `MFEM_USE_CUDA=OFF` and `MFEM_USE_SUNDIALS=ON`,  SUNDIALS version 5.0.0 remains the minimum required version_.

~~The SUNDIALS GPU features added in this PR can be built and tested against this branch in my SUNDIALS fork: https://github.com/balos1/sundials/tree/mfem-gpu-dev.~~

The SUNDIALS GPU features added in this PR have been tested against [SUNDIALS 5.4.0](https://github.com/LLNL/sundials/releases/tag/v5.4.0).

CC: @camierjs,  @acfisher 

**TODO:**
- [x] Resolve `ToNVector` comments
- [x] Coordinate with SUNDIALS 5.4.0 release (release is 9/24)

| [#1726](https://github.com/mfem/mfem/pull/1726) | @balos1 | @tzanio | @camierjs + @cjvogl + @acfisher | 08/30/20 | 09/23/20 | 10/04/20 | |

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,9 +242,12 @@ endif()
 if (MFEM_USE_SUNDIALS)
   if (NOT MFEM_USE_MPI)
     find_package(SUNDIALS REQUIRED NVector_Serial CVODES ARKODE KINSOL)
-  else()
+  elseif(NOT MFEM_USE_CUDA)
     find_package(SUNDIALS REQUIRED
       NVector_Serial NVector_Parallel NVector_MPIPlusX CVODES ARKODE KINSOL)
+  else()
+    find_package(SUNDIALS REQUIRED
+      NVector_Serial NVector_Cuda NVector_Parallel NVector_MPIPlusX CVODES ARKODE KINSOL)
   endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -244,7 +244,7 @@ if (MFEM_USE_SUNDIALS)
     find_package(SUNDIALS REQUIRED NVector_Serial CVODES ARKODE KINSOL)
   else()
     find_package(SUNDIALS REQUIRED
-      NVector_Serial NVector_Parallel NVector_ParHyp CVODES ARKODE KINSOL)
+      NVector_Serial NVector_Parallel NVector_MPIPlusX CVODES ARKODE KINSOL)
   endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,15 +240,15 @@ endif()
 
 # SUNDIALS
 if (MFEM_USE_SUNDIALS)
-  if (NOT MFEM_USE_MPI)
-    find_package(SUNDIALS REQUIRED NVector_Serial CVODES ARKODE KINSOL)
-  elseif(NOT MFEM_USE_CUDA)
-    find_package(SUNDIALS REQUIRED
-      NVector_Serial NVector_Parallel NVector_MPIPlusX CVODES ARKODE KINSOL)
-  else()
-    find_package(SUNDIALS REQUIRED
-      NVector_Serial NVector_Cuda NVector_Parallel NVector_MPIPlusX CVODES ARKODE KINSOL)
+  set(SUNDIALS_COMPONENTS "CVODES ARKODE KINSOL NVector_Serial")
+  if (MFEM_USE_MPI)
+    set(SUNDIALS_COMPONENTS
+      "${SUNDIALS_COMPONENTS} NVector_Parallel NVector_MPIPlusX")
   endif()
+  if (MFEM_USE_CUDA)
+    set(SUNDIALS_COMPONENTS "${SUNDIALS_COMPONENTS} NVector_Cuda")
+  endif()
+  find_package(SUNDIALS REQUIRED ${SUNDIALS_COMPONENTS})
 endif()
 
 # Mesquite
--- a/INSTALL
+++ b/INSTALL
@@ -535,11 +535,14 @@ The specific libraries and their options are:
   Beginning with MFEM v3.3, SUNDIALS v2.7.0 is supported.
   Beginning with MFEM v3.3.2, SUNDIALS v3.0.0 is also supported.
   Beginning with MFEM v4.1, only SUNDIALS v5.0.0+ is supported.
+  When MFEM_USE_CUDA is enabled, only SUNDIALS v5.4.0+ is supported.
   If MFEM_USE_MPI is enabled, we expect that SUNDIALS is built with support for
   both MPI and hypre.
+  If MFEM_USE_CUDA is enabled, we expect that SUNDIALS is built with support
+  for CUDA.
   URL: http://computation.llnl.gov/projects/sundials/sundials-software
   Options: SUNDIALS_OPT, SUNDIALS_LIB.
-  Versions: SUNDIALS >= 5.0.0.
+  Versions: SUNDIALS >= 5.0.0, SUNDIALS >= 5.4.0 for CUDA suppport.
 
 - Mesquite (optional), used when MFEM_USE_MESQUITE = YES.
   URL: http://trilinos.org/oldsite/packages/mesquite
--- a/config/cmake/modules/FindSUNDIALS.cmake
+++ b/config/cmake/modules/FindSUNDIALS.cmake
@@ -22,8 +22,8 @@ mfem_find_package(SUNDIALS SUNDIALS SUNDIALS_DIR
     "include" nvector/nvector_serial.h "lib" sundials_nvecserial
   ADD_COMPONENT NVector_Parallel
     "include" nvector/nvector_parallel.h "lib" sundials_nvecparallel
-  ADD_COMPONENT NVector_ParHyp
-    "include" nvector/nvector_parhyp.h "lib" sundials_nvecparhyp
+  ADD_COMPONENT NVector_MPIPlusX
+    "include" nvector/nvector_mpiplusx.h "lib" sundials_nvecmpiplusx
   ADD_COMPONENT CVODE "include" cvode/cvode.h "lib" sundials_cvode
   ADD_COMPONENT CVODES "include" cvodes/cvodes.h "lib" sundials_cvodes
   ADD_COMPONENT ARKODE "include" arkode/arkode.h "lib" sundials_arkode
--- a/config/cmake/modules/FindSUNDIALS.cmake
+++ b/config/cmake/modules/FindSUNDIALS.cmake
@@ -20,6 +20,10 @@ mfem_find_package(SUNDIALS SUNDIALS SUNDIALS_DIR
   "Paths to headers required by SUNDIALS." "Libraries required by SUNDIALS."
   ADD_COMPONENT NVector_Serial
     "include" nvector/nvector_serial.h "lib" sundials_nvecserial
+  ADD_COMPONENT NVector_Cuda
+    "include" nvector/nvector_cuda.h "lib" sundials_nveccuda
+  ADD_COMPONENT NVector_ParHyp
+    "include" nvector/nvector_parhyp.h "lib" sundials_nvecparhyp
   ADD_COMPONENT NVector_Parallel
     "include" nvector/nvector_parallel.h "lib" sundials_nvecparallel
   ADD_COMPONENT NVector_MPIPlusX
--- a/config/defaults.cmake
+++ b/config/defaults.cmake
@@ -88,7 +88,7 @@ set(METIS_DIR "${MFEM_DIR}/../metis-4.0" CACHE PATH "Path to the METIS library."
 
 set(LIBUNWIND_DIR "" CACHE PATH "Path to Libunwind.")
 
-# For sundials_nvecparhyp and nvecparallel remember to build with MPI_ENABLED=ON
+# For sundials_nvecmpiplusx and nvecparallel remember to build with MPI_ENABLED=ON
 # and modify cmake variables for hypre for sundials
 set(SUNDIALS_DIR "${MFEM_DIR}/../sundials-5.0.0/instdir" CACHE PATH
     "Path to the SUNDIALS library.")
--- a/config/defaults.cmake
+++ b/config/defaults.cmake
@@ -88,7 +88,7 @@ set(METIS_DIR "${MFEM_DIR}/../metis-4.0" CACHE PATH "Path to the METIS library."
 
 set(LIBUNWIND_DIR "" CACHE PATH "Path to Libunwind.")
 
-# For sundials_nvecmpiplusx and nvecparallel remember to build with MPI_ENABLED=ON
+# For sundials_nvecmpiplusx and nvecparallel remember to build with MPI_ENABLE=ON
 # and modify cmake variables for hypre for sundials
 set(SUNDIALS_DIR "${MFEM_DIR}/../sundials-5.0.0/instdir" CACHE PATH
     "Path to the SUNDIALS library.")
--- a/config/defaults.mk
+++ b/config/defaults.mk
@@ -189,15 +189,18 @@ OPENMP_LIB =
 POSIX_CLOCKS_LIB = -lrt
 
 # SUNDIALS library configuration
-# For sundials_nvecparhyp and nvecparallel remember to build with MPI_ENABLED=ON
+# For sundials_nvecmpiplusx and nvecparallel remember to build with MPI_ENABLE=ON
 # and modify cmake variables for hypre for sundials
 SUNDIALS_DIR = @MFEM_DIR@/../sundials-5.0.0/instdir
 SUNDIALS_OPT = -I$(SUNDIALS_DIR)/include
-SUNDIALS_LIB = -Wl,-rpath,$(SUNDIALS_DIR)/lib64 -L$(SUNDIALS_DIR)/lib64\
+SUNDIALS_LIB = $(XLINKER)-rpath,$(SUNDIALS_DIR)/lib64 -L$(SUNDIALS_DIR)/lib64\
  -lsundials_arkode -lsundials_cvodes -lsundials_nvecserial -lsundials_kinsol
 
 ifeq ($(MFEM_USE_MPI),YES)
-   SUNDIALS_LIB += -lsundials_nvecparhyp -lsundials_nvecparallel
+   SUNDIALS_LIB += -lsundials_nvecparallel -lsundials_nvecmpiplusx
+endif
+ifeq ($(MFEM_USE_CUDA),YES)
+   SUNDIALS_LIB += -lsundials_nveccuda
 endif
 # If SUNDIALS was built with KLU:
 # MFEM_USE_SUITESPARSE = YES
--- a/config/defaults.mk
+++ b/config/defaults.mk
@@ -193,7 +193,7 @@ POSIX_CLOCKS_LIB = -lrt
 # and modify cmake variables for hypre for sundials
 SUNDIALS_DIR = @MFEM_DIR@/../sundials-5.0.0/instdir
 SUNDIALS_OPT = -I$(SUNDIALS_DIR)/include
-SUNDIALS_LIB = $(XLINKER)-rpath,$(SUNDIALS_DIR)/lib64 -L$(SUNDIALS_DIR)/lib64\
+SUNDIALS_LIB = $(XLINKER)-rpath,$(SUNDIALS_DIR)/lib -L$(SUNDIALS_DIR)/lib\
  -lsundials_arkode -lsundials_cvodes -lsundials_nvecserial -lsundials_kinsol
 
 ifeq ($(MFEM_USE_MPI),YES)
--- a/config/defaults.mk
+++ b/config/defaults.mk
@@ -191,9 +191,10 @@ POSIX_CLOCKS_LIB = -lrt
 # SUNDIALS library configuration
 # For sundials_nvecmpiplusx and nvecparallel remember to build with MPI_ENABLE=ON
 # and modify cmake variables for hypre for sundials
-SUNDIALS_DIR = @MFEM_DIR@/../sundials-5.0.0/instdir
-SUNDIALS_OPT = -I$(SUNDIALS_DIR)/include
-SUNDIALS_LIB = $(XLINKER)-rpath,$(SUNDIALS_DIR)/lib -L$(SUNDIALS_DIR)/lib\
+SUNDIALS_DIR    = @MFEM_DIR@/../sundials-5.0.0/instdir
+SUNDIALS_OPT    = -I$(SUNDIALS_DIR)/include
+SUNDIALS_LIBDIR = $(wildcard $(SUNDIALS_DIR)/lib*)
+SUNDIALS_LIB    = $(XLINKER)-rpath,$(SUNDIALS_LIBDIR) -L$(SUNDIALS_LIBDIR)\
  -lsundials_arkode -lsundials_cvodes -lsundials_nvecserial -lsundials_kinsol
 
 ifeq ($(MFEM_USE_MPI),YES)
--- a/examples/ex9.cpp
+++ b/examples/ex9.cpp
@@ -135,6 +135,8 @@ public:
    virtual void ImplicitSolve(const double dt, const Vector &x, Vector &k);
 
    virtual ~FE_Evolution();
+
+   MemoryClass GetMemoryClass() const { return Device::GetMemoryClass(); }
 };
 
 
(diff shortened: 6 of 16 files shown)


linalg/sundials.hpp
```diff
@@ -43,25 +47,159 @@ namespace mfem
...
+   /** The Vector does not assume ownership of the new data. The new size is
+       also used as the new Capacity().
+       @warning This method should be called only when OwnsData() is false. */
+   void SetDataAndSize(double *d, int s, long glob_size = 0);
+
+   /// Reset the Vector to be a reference to a sub-vector of @a base.
+   inline void MakeRef(Vector &base, int offset, int s)
+   {
+      // Ensure that the base is registered/initialized before making an alias
+      base.Read();
```
balos1, line 131: This is not optimal, but see https://github.com/mfem/mfem/issues/1578#issuecomment-649962962.

linalg/vector.hpp
```diff
@@ -393,6 +393,9 @@ class Vector
    { return mfem::ReadWrite(data, size, false); }
 
 #ifdef MFEM_USE_SUNDIALS
+   // TODO: I think this should be removed, but not sure if it might break old code.
```
balos1, line 396: Highlighting this code comment for discussion. With the addition of the `SundialsNVector` class, I don't think this is needed, at least for the interface to SUNDIALS.

jandrej: I agree but don't think it is necessary to do in this PR. It would be good enough to have in a shortly-after follow up.

v-dobrev: Unless this constructor is now broken, we should keep it for now and just mark it as deprecated. For an example how we do that see https://github.com/mfem/mfem/blob/4f521a1be6624826ff25924977b43e5667b9ac1c/fem/coefficient.hpp#L149-L152

linalg/sundials.hpp
```diff
@@ -547,6 +686,10 @@ class ARKStepSolver : public ODESolver, public SundialsSolver
 
    /// Destroy the associated ARKode memory and SUNDIALS objects.
    virtual ~ARKStepSolver();
+
+   virtual MemoryClass GetMemoryClass() const
+   { return Device::GetMemoryClass(); }
```
gardner48, line 691: Should this be in the ` SundialsSolver` class?

linalg/sundials.hpp
```diff
@@ -43,25 +47,159 @@ namespace mfem
...
-   SUNLinearSolver    LSM; ///< Linear solver for M.
-   SUNNonlinearSolver NLS; ///< Nonlinear solver.
+   void *sundials_mem;        ///< SUNDIALS mem structure.
+   mutable int flag;          ///< Last flag returned from a call to SUNDIALS.
+   bool reinit;               ///< Flag to signal memory reinitialization is need.
+   long saved_global_size;    ///< Global vector length on last initialization.
+
+   SundialsNVector*   Y;      ///< State vector.
+   SUNMatrix
+   A;      ///< Linear system A = I - gamma J, M - gamma J, or J.
```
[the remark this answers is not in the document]
balos1: This is how the `make style`command formats it unfortunately.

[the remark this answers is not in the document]
v-dobrev: You can help `make style` here. You can either turn the `///<` comment into a multi-line comment as here https://github.com/mfem/mfem/blob/b2983cb965544c09e8c7dcc793dea13d6478fad6/fem/fe.hpp#L41-L42

Or you can move the doxygen comment before the declaration as a `///` doxygen comment.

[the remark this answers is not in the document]
balos1: Fixed, thanks @v-dobrev.

linalg/sundials.hpp
balos1: This code is unused, so I removed it entirely.

cjvogl:
@tzanio or @v-dobrev, can one of you restart the Travis Mac Parallel test... looks like it may have stalled out.

tzanio:
Restarted

cjvogl (changes requested):
Having a problem when building using the following steps:
```bash
make config MFEM_USE_MPI=YES HYPRE_DIR=<local-hypre-dir> MFEM_USE_SUNDIALS=YES SUNDIALS_DIR=<local-sundials-5.3.0-dir> METIS_DIR=<local-metis-dir>
make -j 4
cd examples
make
```
The problem seems to be the `sundials/lib64` directory is being specified for libraries that exist in the `sundials/lib` directory:
``` bash
mpicxx  -O3 -std=c++11 -I.. -I/mnt/c/Users/vogl2/workspace/mhd-mfem/local/Release/hypre/include  -I/mnt/c/Users/vogl2/workspace/mhd-mfem/local/Release/sundials/include ex1p.cpp -o ex1p -L.. -lmfem -L/mnt/c/Users/vogl2/workspace/mhd-mfem/local/Release/hypre/lib -lHYPRE -L/mnt/c/Users/vogl2/workspace/mhd-mfem/local/metis -lmetis -Wl,-rpath,/mnt/c/Users/vogl2/workspace/mhd-mfem/local/Release/sundials/lib64 -L/mnt/c/Users/vogl2/workspace/mhd-mfem/local/Release/sundials/lib64 -lsundials_arkode -lsundials_cvodes -lsundials_nvecserial -lsundials_kinsol -lsundials_nvecparallel -lsundials_nvecmpiplusx -lrt
/usr/bin/ld: cannot find -lmetis
/usr/bin/ld: cannot find -lsundials_arkode
/usr/bin/ld: cannot find -lsundials_cvodes
/usr/bin/ld: cannot find -lsundials_nvecserial
/usr/bin/ld: cannot find -lsundials_kinsol
/usr/bin/ld: cannot find -lsundials_nvecparallel
/usr/bin/ld: cannot find -lsundials_nvecmpiplusx
collect2: error: ld returned 1 exit status
makefile:66: recipe for target 'ex1p' failed
make: *** [ex1p] Error 1
```
I remember @gardner48 mentioning this can be an issue while we were working on another project... not sure what the fix is here though.

gardner48:
> I remember @gardner48 mentioning this can be an issue while we were working on another project... not sure what the fix is here though.

Sundials automatically chooses between installing in `lib` or `lib64` depending on the system it's built on but you can override this choice by setting the CMake variable `CMAKE_INSTALL_LIBDIR` to what you want you want to use for the library directory name.

cjvogl:
@gardner48, thanks for the quick response!  I suppose the trick now is make sure the MFEM looks in the right place: it appears that MFEM will look in `sundials/lib` if using cmake and in `sundials/lib64`.  

Seems to me like the best response is to have MFEM look in a single place so that we can tell the user what to specify for `CMAKE_INSTALL_LIBDIR` when building Sundials.  @tzanio or @v-dobrev, who would be the best person to ping for that (*sheepishly looking around, hoping the answer isn't me*)

examples/ex9.cpp
```diff
@@ -135,6 +135,8 @@ class FE_Evolution : public TimeDependentOperator
    virtual void ImplicitSolve(const double dt, const Vector &x, Vector &k);
 
    virtual ~FE_Evolution();
+
+   MemoryClass GetMemoryClass() const { return Device::GetMemoryClass(); }
```
cjvogl, line 139: Is this function used anywhere?

cjvogl:
I don't know that I have the experience to test the GPU/CUDA versions of the Sundials example 9 code, but I was able to test the CPU version of Sundials example 9.  @acfisher, @camierjs, can I punt on testing the GPU/CUDA functionality to one of you guys?

examples/ex9.cpp
balos1: This function is not used anymore (when I started this branch months ago it was), and [I have now removed it](94e07ba735dd370d3dbd89d4bc54b2872c6793c9).

balos1:
> @gardner48, thanks for the quick response! I suppose the trick now is to make sure the MFEM looks in the right place: it appears that MFEM will look in `sundials/lib` if using cmake and in `sundials/lib64` if using GNU make.
> 
> Seems to me like the best response is to have MFEM look in a single place so that we can tell the user what to specify for `CMAKE_INSTALL_LIBDIR` when building Sundials. @tzanio or @v-dobrev, who would be the best person to ping for that (_sheepishly looking around, suspecting the answer will be me_)

I have updated the GNU make to match the cmake and look in `lib` instead of `lib64`.

cjvogl (approved):
I am not really in a position to test the GPU/CUDA features, but I did test the updated CPU Sundials examples: LGTM

v-dobrev:
@cjvogl, @gardner48, @balos1, I think the `"lib"` subdirectory used in `config/cmake/modules/FindSUNDIALS.cmake` is just a hint and CMake actually searches other suitable subdirectories too, so this is probably not an issue there.

For the GNU make build system, we can easily add some auto-detection logic (using `wildcard` command) for `lib` vs `lib64` in the file `config/defaults.mk` for the variable `SUNDIALS_LIB`. Of course, this variable can always be set on the `make` command line by the user to override the defaults -- e.g. this is what we do in the mfem spack package.

@balos1, I suspect you were testing on Lassen? Is this where Sundials libraries are installed under `lib64` and for that reason you originally switched `lib` to `lib64` in `config/defaults.mk`? Or was there a different reason to switch to `lib64`?

balos1:
@v-dobrev I was testing on my Linux workstation and also Lassen. In both of those places lib64 is the default detected by GNUInstallDirs.

linalg/sundials.cpp
```diff
@@ -35,12 +40,392 @@ using namespace std;
...
+      /* Set the ops */
+      h->ops->alloc     = SundialsMemHelper_Alloc;
+      h->ops->dealloc   = SundialsMemHelper_Dealloc;
+#ifdef MFEM_USE_CUDA
+      h->ops->copy      = SUNMemoryHelper_Copy_Cuda;
+      h->ops->copyasync = SUNMemoryHelper_CopyAsync_Cuda;
+#endif
+
+      /* Set content (user data) to NULL */
+      h->content = NULL;
```
gardner48, line 71: The content is`NULL` by default so this could be removed

linalg/sundials.cpp
```diff
@@ -35,12 +40,392 @@ using namespace std;
...
+   }
+
+   ~SundialsMemHelper()
+   {
+      SUNMemoryHelper_Destroy(h);
+   }
+
+   /// Typecasting to SUNDIALS' N_Vector type
```
gardner48, line 79: N_Vector to SUNMemoryHelper

linalg/sundials.cpp
```diff
@@ -35,12 +40,392 @@ using namespace std;
...
+         if (known) { data.ClearOwnerFlags(); }
+         break;
+      }
+#ifdef MFEM_USE_CUDA
+      case SUNDIALS_NVEC_CUDA:
+      {
+         double *h_ptr = N_VGetHostArrayPointer_Cuda(local_x);
+         double *d_ptr = N_VGetDeviceArrayPointer_Cuda(local_x);
+         const bool known = mm.IsKnown(h_ptr);
+         size = N_VGetLength_Cuda(local_x);
```
gardner48, line 256: Is it possible for `h_ptr` to be known but not `d_ptr` (or vice versa)?

linalg/sundials.cpp
```diff
@@ -35,12 +40,392 @@ using namespace std;
...
+#ifdef MFEM_USE_MPI
+SundialsNVector::SundialsNVector(MPI_Comm comm)
+   : Vector()
+{
+   UseDevice(Device::IsAvailable());
+   x = MakeNVector(comm, UseDevice());
+   own_NVector = 1;
+}
+
+SundialsNVector::SundialsNVector(MPI_Comm comm, int s, long glob_size)
```
gardner48, line 313: `s` is unused?

linalg/sundials.cpp
```diff
@@ -230,15 +586,23 @@ void CVODESolver::Init(TimeDependentOperator &f_)
...
+      else
+      {
+         Y->SetSize(local_size, global_size);
+         saved_global_size = global_size;
+      }
+#endif
 
       // Create CVODE
       sundials_mem = CVodeCreate(lmm_type);
-      MFEM_VERIFY(sundials_mem, "error in CVodeCreate()");
```
gardner48, line 238: Why was this removed?

linalg/sundials.cpp
balos1: I believe only h_ptr can be unknown, but @camierjs can answer definitively.

gardner48 (commented):
A couple of minor comments but otherwise this looks good to me.

linalg/sundials.cpp
balos1: Accidental. I have re-added.

linalg/sundials.cpp
balos1: Fixed. `s` should have been passed to the `Vector` constructor.

linalg/sundials.cpp
balos1: `IsKnown` can only be called with a host pointer according to the documentation, so this must be the case.

gardner48 (approved):

examples/sundials/ex9p.cpp
```diff
@@ -273,13 +384,51 @@ int main(int argc, char *argv[])
...
+   {
+      std::string postfix(mesh_file);
+      postfix.erase(0, std::string("../data/").size() );
+      postfix += "_o" + std::to_string(order);
+      const std::string collection_name = "ex9-p-" + postfix + ".bp";
+
+      adios2_dc = new ADIOS2DataCollection(MPI_COMM_WORLD, collection_name, pmesh);
+      // output data substreams are half the number of mpi processes
+      adios2_dc->SetParameter("SubStreams", std::to_string(num_procs/2) );
+      // adios2_dc->SetLevelsOfDetail(2);
```
acfisher, line 424: Dead code?

linalg/sundials.cpp
```diff
@@ -259,32 +621,21 @@ void CVODESolver::Init(TimeDependentOperator &f_)
...
-   }
-   else
-   {
-#ifdef MFEM_USE_MPI
-      NV_DATA_P(y) = x.GetData();
-      MFEM_VERIFY(NV_LOCLENGTH_P(y) == x.Size(), "");
-#endif
-   }
+   Y->MakeRef(x, 0, x.Size());
+   MFEM_VERIFY(Y->Size() == x.Size(), "");
```
acfisher, line 625: Maybe a size mismatch comment instead of an empty string here?

linalg/sundials.cpp
```diff
@@ -1082,34 +1377,24 @@ void ARKStepSolver::Init(TimeDependentOperator &f_)
...
```
acfisher, line 1381: Error string...

acfisher (approved):

config/defaults.mk
```diff
@@ -189,15 +189,18 @@ OPENMP_LIB =
 POSIX_CLOCKS_LIB = -lrt
 
 # SUNDIALS library configuration
-# For sundials_nvecparhyp and nvecparallel remember to build with MPI_ENABLED=ON
+# For sundials_nvecmpiplusx and nvecparallel remember to build with MPI_ENABLE=ON
 # and modify cmake variables for hypre for sundials
 SUNDIALS_DIR = @MFEM_DIR@/../sundials-5.0.0/instdir
 SUNDIALS_OPT = -I$(SUNDIALS_DIR)/include
-SUNDIALS_LIB = -Wl,-rpath,$(SUNDIALS_DIR)/lib64 -L$(SUNDIALS_DIR)/lib64\
+SUNDIALS_LIB = $(XLINKER)-rpath,$(SUNDIALS_DIR)/lib -L$(SUNDIALS_DIR)/lib\
```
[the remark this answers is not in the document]
balos1: You would have to tell sundials to put the libraries in `lib` with the cmake option `CMAKE_INSTALL_LIBDIR=lib`, because we still default to wherever GNUInstallDirs decides. However, I will change this to use the `wildcard` command like @v-dobrev suggested.

linalg/sundials.cpp
balos1: Fixed.

linalg/sundials.cpp
balos1: Fixed.

config/defaults.mk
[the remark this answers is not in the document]
balos1: Fixed.

camierjs:
LGTM

camierjs (approved):

camierjs:
Looks good @balos1 ! Thank you for adding this MFEM+SUNDIALS GPU feature!

tzanio:

balos1:
The SUNDIALS 5.4.0 release (needed for the CUDA features in this PR) is now available: https://github.com/LLNL/sundials/releases/tag/v5.4.0.

config/defaults.cmake
```diff
@@ -88,7 +88,7 @@ set(METIS_DIR "${MFEM_DIR}/../metis-4.0" CACHE PATH "Path to the METIS library."
 
 set(LIBUNWIND_DIR "" CACHE PATH "Path to Libunwind.")
 
-# For sundials_nvecparhyp and nvecparallel remember to build with MPI_ENABLED=ON
+# For sundials_nvecmpiplusx and nvecparallel remember to build with MPI_ENABLED=ON
```
v-dobrev, line 91: For consistency with the fix in `config/defaults.mk`, let's fix the typo here too: the correct option is `MPI_ENABLE=ON` and not `MPI_ENABLED=ON`.

INSTALL
```diff
@@ -535,11 +535,14 @@ The specific libraries and their options are:
   Beginning with MFEM v3.3, SUNDIALS v2.7.0 is supported.
   Beginning with MFEM v3.3.2, SUNDIALS v3.0.0 is also supported.
   Beginning with MFEM v4.1, only SUNDIALS v5.0.0+ is supported.
+  When MFEM_USE_CUDA is enabled, only SUNDIALS v5.4.0+ is supported.
```
v-dobrev, line 538: It looks like SUNDIALS v5.0.0 is not sufficient for a non-CUDA build. The internal tests on `toss3` give the following error with v5.0.0:
```console
../../libmfem.so: undefined reference to `N_VGetLocalVector_MPIPlusX'
```
When was `N_VGetLocalVector_MPIPlusX` introduced in SUNDIALS? If it was added before v5.4.0, we can build without CUDA with that older version but it is probably simpler to just say that we require v5.4.0+ now.

In this case, the above text and the version requirement below should be updated appropriately.

It is also a good idea to mention the new version requirement in the `CHANGELOG` in the new paragraph about SUNDIALS.

CMakeLists.txt
```diff
@@ -244,7 +244,7 @@ if (MFEM_USE_SUNDIALS)
     find_package(SUNDIALS REQUIRED NVector_Serial CVODES ARKODE KINSOL)
   else()
     find_package(SUNDIALS REQUIRED
-      NVector_Serial NVector_Parallel NVector_ParHyp CVODES ARKODE KINSOL)
+      NVector_Serial NVector_Parallel NVector_MPIPlusX CVODES ARKODE KINSOL)
```
v-dobrev, line 247: What about the CUDA dependency in the CMake build system?

config/cmake/modules/FindSUNDIALS.cmake
```diff
@@ -22,8 +22,8 @@ mfem_find_package(SUNDIALS SUNDIALS SUNDIALS_DIR
     "include" nvector/nvector_serial.h "lib" sundials_nvecserial
   ADD_COMPONENT NVector_Parallel
     "include" nvector/nvector_parallel.h "lib" sundials_nvecparallel
-  ADD_COMPONENT NVector_ParHyp
-    "include" nvector/nvector_parhyp.h "lib" sundials_nvecparhyp
+  ADD_COMPONENT NVector_MPIPlusX
+    "include" nvector/nvector_mpiplusx.h "lib" sundials_nvecmpiplusx
```
v-dobrev, line 26: There is really no need to remove the `NVector_ParHyp` component since these are just definitions of components -- the actual requirements are in the `find_package(SUNDIALS ...)` command.

The CUDA NVector component can be added here too.

INSTALL
balos1: `N_VGetLocalVector_MPIPlusX` is in SUNDIALS `v5.0.0` so I am not sure why that error is occurring. Can you share details about how SUNDIALS and mfem were built in this case? I am able to do a non CUDA build on my box with v5.0.0.

config/cmake/modules/FindSUNDIALS.cmake
balos1: I have added `NVector_ParHyp` back and added the CUDA NVector component.

CMakeLists.txt
balos1: I have added a link to handle this.

config/defaults.cmake
balos1: Fixed.

config/cmake/modules/FindSUNDIALS.cmake
gardner48: `NVector_ParHyp` was used in the MFEM SUNDIALS interface some time in the past but is no longer used anywhere so I think that can be safely removed.

INSTALL
v-dobrev: Okay, I think you are right -- the problem is on our side. In the internal tests we explicitly set the SUNDIALS libraries with which to link, overriding the defaults, and we need to update our scripts to include `-lsundials_nvecmpiplusx`.

CMakeLists.txt
```diff
@@ -242,9 +242,12 @@ endif()
...
   if (NOT MFEM_USE_MPI)
     find_package(SUNDIALS REQUIRED NVector_Serial CVODES ARKODE KINSOL)
+  elseif(NOT MFEM_USE_CUDA)
+    find_package(SUNDIALS REQUIRED
+      NVector_Serial NVector_Parallel NVector_MPIPlusX CVODES ARKODE KINSOL)
   else()
     find_package(SUNDIALS REQUIRED
-      NVector_Serial NVector_Parallel NVector_ParHyp CVODES ARKODE KINSOL)
+      NVector_Serial NVector_Cuda NVector_Parallel NVector_MPIPlusX CVODES ARKODE KINSOL)
   endif()
```
v-dobrev, line 251: The case of `MFEM_USE_CUDA=ON MFEM_USE_MPI=OFF` is not handled correctly. Maybe defining a variable for which `NVector_*` components to require can simplify the logic.

INSTALL
v-dobrev: This reminds me: we'll need to update the mfem spack package accordingly here: https://github.com/spack/spack/blob/59720e0ef9b28a6bc1b7f2deee7e079ce378da02/var/spack/repos/builtin/packages/mfem/package.py#L676-L682. This will need to include `cvode` --> `cvodes` (from #1380), `nvecparhyp` --> `nvecmpiplusx`, and adding `nveccuda` when CUDA is enabled. On top of that, maintaining older mfem+sundials versions will complicate the logic even more. 🙁

balos1: Since v5.0.0 does work without cuda I think all you need to do in the spack package is add something like

```
depends_on('sundials@5.4.0:+mpi+hypre+cuda', when='@4.2.0:+sundials+mpi+cuda')
```

and then 

```
@property
    def sundials_components(self):
        """Return the SUNDIALS components needed by MFEM."""
        sun_comps = 'arkode,cvode,nvecserial,kinsol'
        if '+mpi' in self.spec:
            sun_comps += ',nvecparhyp,nvecparallel,nvecmpiplusx'
        if '+cuda' in self.spec:
            sun_comps += ',nveccuda'
        return sun_comps
```

CMakeLists.txt
balos1: Oops... fix incoming.

v-dobrev:
Re-merged in `next`.

tzanio (approved):
Thank you so much for this really useful contribution @balos1! I'm excited to see the fruits of this GPU integration of MFEM and SUNDIALS 🚀 

@camierjs, @cjvogl, @acfisher, @v-dobrev, @gardner48 and @jandrej -- thank you all for your help and reviews!

PR #1738, opened 2020-09-02
Implement "BR2" method for DG diffusion

pazner:
This PR implements the "BR2" method for DG diffusion, which is a popular method for diffusion problems with a compact stencil that does not require the user to specify a penalty parameter (kappa in ex14) in order to ensure stability.

The integrator is an example of using DG "lifting operators," which could be useful for other DG methods (LDG, etc.).

Sample runs (note all of these use 0.0 for the interior penalty stabilization parameter):
```
./ex14 -m ../data/star-mixed.mesh -r 2 -o 3 -k 0.0 -e 1
./ex14 -m ../data/amr-quad.mesh -r 1 -o 4 -k 0.0 -e 1
mpirun -np 4 ./ex14p -m ../data/fichera-mixed.mesh -rs 1 -rp 1 -o 1 -k 0.0 -e 1
mpirun -np 4 ./ex14p -m ../data/amr-hex.mesh -rs 1 -rp 1 -o 1 -k 0.0 -e 1
```

| [#1738](https://github.com/mfem/mfem/pull/1738) | @pazner | @tzanio | @tzanio + @YohannDudouit | 09/02/20 | 11/30/20 | 12/02/20 | |


fem/bilininteg.cpp
```diff
@@ -2904,7 +2905,6 @@ void DGDiffusionIntegrator::AssembleFaceMatrix(
    }
 }
 
-
 // static method
```
tzanio, line 2908: Revert (two empty lines between different classes)

examples/ex14.cpp
```diff
@@ -44,6 +44,7 @@ int main(int argc, char *argv[])
    int order = 1;
    double sigma = -1.0;
    double kappa = -1.0;
+   double eta = 1.0;
```
tzanio, line 47: Are these the default values we want?

tzanio (approved):
Thanks @pazner!

Can you update the `CHANGELOG`?

examples/ex14.cpp
pazner: I changed eta to be zero by default, so the previous behavior remains the same. This is probably preferable, since the BR2 terms can make the system harder to solve.

tzanio:
@YohannDudouit, can you please review?

fem/bilininteg_br2.cpp
```diff
@@ -0,0 +1,241 @@
...
+         {
+            Re(ndof1 + i, ndof1 + j) = R22(i, j);
+            MinvRe(ndof1 + i, ndof1 + j) = MinvR22(i, j);
+         }
+      }
+   }
+
+   // Compute the matrix associated with (r_e([u]), r_e([u])).
+   // The matrix for r_e([u]) is `MinvRe`, and so we need to form the product
+   // `MinvRe^T M MinvRe`. Canceling Minv and M, we obtain `Re^T M MinvRe`.
```
YohannDudouit, line 237: I think the following correspond to the code bellow:
```suggestion
   // `MinvRe^T M MinvRe`. Canceling Minv and M, we obtain `Re^T MinvRe`.
```
However, why is it not:
```suggestion
   // `MinvRe^T M MinvRe`. Canceling Minv and M, we obtain `MinvRe^T Re`.
   MultAtB(MinvRe, Re, elmat);
```
?

YohannDudouit (changes requested):
Just a small question, otherwise this looks good, the code is clean and clear and it works.

fem/bilininteg_br2.cpp
pazner: The notation is a bit confusing, `MinvRe^T` is supposed to mean `(MinvRe)^T`, so we have `(MinvRe)^T M MinvRe = Re^T Minv^T M Minv Re`, and `M` is symmetric, so we obtain `Re^T Minv Re`.

fem/bilininteg_br2.cpp
```diff
@@ -0,0 +1,241 @@
...
+         {
+            Re(ndof1 + i, ndof1 + j) = R22(i, j);
+            MinvRe(ndof1 + i, ndof1 + j) = MinvR22(i, j);
+         }
+      }
+   }
+
+   // Compute the matrix associated with (r_e([u]), r_e([u])).
+   // The matrix for r_e([u]) is `MinvRe`, and so we need to form the product
+   // `(MinvRe)^T M MinvRe`. Canceling Minv and M, we obtain `Re^T MinvRe`.
```
YohannDudouit, line 237: If you stick with this instead of `MultAtB(MinvRe, Re, elmat);`, then maybe that the following would be more clear:
```suggestion
   // `(MinvRe)^T M MinvRe`. Using  `Minv^T M = Minv M = I`, we obtain `Re^T MinvRe`.
```

YohannDudouit (approved):

tzanio:

tzanio (approved):
Thanks again @pazner !

PR #1761, opened 2020-09-14
PA for GradientInterpolator and IdentityInterpolator

dylan-copeland:
This PR is a continuation of PR #1680 by @barker29 which implemented a PA gradient operator, which is an Interpolator rather than an Integrator type object. As a result, it has some different behavior related to Add and Set values that as I understand it has not been implemented in a PA way before.

PA for IdentityInterpolator is now also implemented. Both the gradient and identity operators are implemented in 2D and 3D, with unit tests, for H(curl) and H1 spaces. Other choices of spaces can be added in the future as needed.

| [#1761](https://github.com/mfem/mfem/pull/1761) | @dylan-copeland | @tzanio | @barker29 + @pazner + @tzanio | 09/17/20 | 01/11/21 | 01/16/21 | |


barker29:
The identity interpolator fails for me for the `star.mesh` mesh, try putting
```
mesh = new Mesh("../../data/star.mesh", order);
```
at the start of the `compare_pa_id_assembly()` unit test. I suspect something is wrong with the handling of element orientation (which is much more difficult for the Pi operator than the G operator). If you want I can share my 2D template-based implementation, which is terrible (and only 2D) but does work for the star mesh.

dylan-copeland:
Good catch @barker29 . So far I had only tested on meshes aligned with the coordinate axes in 2D and 3D, and I realize now that the problem is in the PA kernels, which neglected the coupling of some components. Those couplings are zero in the cases tested so far, but not in general. I fixed it in 2D with the last commit and will fix the 3D case next (on Monday). In 3D, I also need to set up a unit test not aligned with the axes, which should fail with the current version. Note that a higher relative tolerance (1e-5 vs 1e-14) is needed for the star mesh test.

dylan-copeland:
With the latest commit, the 2D tests are succeeding with low tolerance (not 1e-5 as in previous commit). Also, the 3D case is corrected with a unit test using a hexahedral mesh not aligned with the coordinate axes.

dylan-copeland:
I ran the IdentityInterpolator PA kernel unit tests up to order 11 without any memory failures, so it seems we will not need shared memory versions for moderately high order.

tzanio:
@barker29, @pazner and @tzanio -- the reviews for this PR are due

dylan-copeland:
@tzanio @barker29 @pazner The review deadline should be extended. @barker29 is looking into a bug in the prolongation on device, in the parallel case. Either that should be fixed, or prolongation should just be done on host.

tzanio:
Thanks for the update Dylan, I've extended the deadline

fem/restriction.hpp
```diff
@@ -57,6 +57,10 @@ class ElementRestriction : public Operator
    /// Compute MultTranspose without applying signs based on DOF orientations.
    void MultTransposeUnsigned(const Vector &x, Vector &y) const;
 
+   /// Compute MultTranspose by setting (rather than adding) element
+   /// contributions; this is a left inverse of the Mult() operation
+   void MultLeftInverse(const Vector &x, Vector &y) const;
```
pazner, line 62: I don't see how the R operator has a left inverse — it has fewer rows than columns, and so it cannot have full column rank.

tests/unit/fem/test_pa_idinterp.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+   pa_y -= assembled_y;
+   double error = pa_y.Norml2() / assembled_y.Norml2();
+   std::cout << "dim " << dim << " ne " << num_elements << " order "
+             << order;
+   if (transpose)
+   {
+      std::cout << " T";
+   }
+   std::cout << ": error in PA identity: " << error << std::endl;
```
pazner, line 109: Personally, I don't like all the output in tests (I know a lot of other existing unit tests do this). It really pollutes the terminal, and usually isn't very informative.

I suggest using an `INFO` message here, and if you want to see all of the messages, then run the tests with flag `-s`.

tests/unit/fem/test_pa_idinterp.cpp
```diff
@@ -0,0 +1,134 @@
...
+   for (bool transpose : {false, true})
+   {
+      for (int dim = 2; dim < 4; ++dim)
+      {
+         for (int num_elements = 0; num_elements < 5; ++num_elements)
+         {
+            for (int order = 1; order < 5; ++order)
+            {
+               double error = compare_pa_id_assembly(dim, num_elements, order, transpose);
+               REQUIRE(error < 1.0e-14);
```
pazner, line 129: Why not use `REQUIRE(error == MFEM_Approx(0.0))`?

This communicates the intent better, avoids the arbitrary choice of tolerance (and would also fail if the error is a large negative number for any reason...)

tests/unit/fem/test_pa_idinterp.cpp
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) 2010-2020, Lawrence Livermore National Security, LLC. Produced
```
pazner, line 1: `tests/unit/CMakeLists.txt` should be updated.

tests/unit/fem/test_pa_grad.cpp
```diff
@@ -0,0 +1,254 @@
+// Copyright (c) 2010-2020, Lawrence Livermore National Security, LLC. Produced
```
pazner, line 1: `tests/unit/CMakeLists.txt` should be updated.

tests/unit/fem/test_pa_grad.cpp
```diff
@@ -0,0 +1,254 @@
...
+   }
+
+   if (false)
+   {
+      std::cout << "true   \tpa\n";
+      for (int i = 0; i < assembled_y.Size(); ++i)
+      {
+         std::cout << i << " : " << assembled_y(i) << "\t" << pa_y(i) << std::endl;
+      }
+   }
```
pazner, line 96: This should be removed?

tests/unit/fem/test_pa_grad.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+   pa_y -= assembled_y;
+   double error = pa_y.Norml2() / assembled_y.Norml2();
+   std::cout << "dim " << dim << " ne " << num_elements << " order "
+             << order;
+   if (transpose)
+   {
+      std::cout << " T";
+   }
+   std::cout << ": error in PA gradient: " << error << std::endl;
```
pazner, line 106: See my other comment in `test_pa_idinterp.cpp` about writing to stdout in tests.

tests/unit/fem/test_pa_grad.cpp
```diff
@@ -0,0 +1,254 @@
...
+   // in the test
+   double error = error_vec.Norml2() / assembled_y.Norml2();
+
+   for (int p = 0; p < size; ++p)
+   {
+      if (rank == p)
+      {
+         std::cout << "[" << rank << "]";
+         // std::cout << "pa_y.Norml2() = " << pa_y.Norml2() << std::endl;
+         // std::cout << "assembled_y.Norml2() = " << assembled_y.Norml2() << std::endl;
```
pazner, line 215: Delete?
```suggestion
         // std::cout << "pa_y.Norml2() = " << pa_y.Norml2() << std::endl;
         // std::cout << "assembled_y.Norml2() = " << assembled_y.Norml2() << std::endl;
```

tests/unit/fem/test_pa_grad.cpp
```diff
@@ -0,0 +1,254 @@
...
+   delete nd_fec;
+   delete assembled_grad_mat;
+   delete mesh;
+
+   return error;
+}
+
+TEST_CASE("ParallelPAGradient", "[Parallel], [ParallelPAGradient]")
+{
+   for (bool transpose : {false, true})
```
pazner, line 238: I suggest using `GENERATE` here instead of for loops (see my comment in `test_pa_idinterp.cpp`)

tests/unit/fem/test_pa_grad.cpp
```diff
@@ -0,0 +1,254 @@
...
+   {
+      for (int dim = 2; dim < 4; ++dim)
+      {
+         for (int num_elements = 4; num_elements < 6; ++num_elements)
+         {
+            for (int order = 1; order < 5; ++order)
+            {
+               double error = par_compare_pa_assembly(dim, num_elements, order, transpose);
+               REQUIRE(error < 1.e-14);
```
pazner, line 247: I suggest using `MFEM_Approx` here (see my comment in `test_pa_idinterp.cpp`)

fem/fespace.hpp
```diff
@@ -330,6 +330,11 @@ class FiniteElementSpace
    virtual const Operator *GetProlongationMatrix() const
    { return GetConformingProlongation(); }
 
+   /// Return a prolongation operator that only acts on local dofs
+   /** The returned operator is owned by the FiniteElementSpace. */
+   virtual const Operator *GetLocalProlongationMatrix() const
+   { return GetConformingProlongation(); }
```
pazner, line 336: It's not clear what this means in the context of a serial FiniteElementSpace. I suppose it's just the AMR P matrix, but perhaps the documentation should make clear that there is only a distinction in the parallel case?

fem/pfespace.hpp
```diff
@@ -335,6 +338,7 @@ class ParFiniteElementSpace : public FiniteElementSpace
    HYPRE_Int GetMyTDofOffset() const;
 
    virtual const Operator *GetProlongationMatrix() const;
+   virtual const Operator *GetLocalProlongationMatrix() const;
```
pazner, line 341: I'm not sure that `GetLocalProlongationMatrix` is a good name. There is a nice symmetry with `GetProlongationMatrix` and `GetRestrictionMatrix`, but this (just like `GetProlongationMatrix`) actually doesn't return a matrix object, and instead returns a specialized matrix-free operator.

It seems like `GetProlongationMatrix` is probably a legacy name from a time when a `HypreParMatrix` was returned, and so it is not changed for backwards compatibility reasons.

But that is not a concern here, so it doesn't seem to me like we should make a function `GetLocalProlongationMatrix` that does not return a matrix.

fem/pfespace.hpp
```diff
@@ -72,6 +72,9 @@ class ParFiniteElementSpace : public FiniteElementSpace
    mutable HypreParMatrix *P;
    /// Optimized action-only prolongation operator for conforming meshes. Owned.
    mutable Operator *Pconf;
+   /// Optimized action-only diagonal prolongation operator on ldofs for
+   /// conforming meshes. Owned.
+   mutable Operator *Pconf_local;
```
pazner, line 77: If I understand correctly, this converts T-dofs to L-dofs, setting any not-owned DOF to zero. Is this an abstraction that makes sense more generally? Is this something that would be useful for users to have access to? It does not seem very intuitive to me.

fem/pbilinearform.cpp
```diff
@@ -565,6 +565,53 @@ HypreParMatrix* ParDiscreteLinearOperator::ParallelAssemble() const
...
+      Finalize();
+      ParallelAssemble(p_mat);
+      delete mat;
+      mat = NULL;
+      delete mat_e;
+      mat_e = NULL;
+      p_mat_e = NULL;
+   }
+   A = p_mat;
+   */
```
pazner, line 610: Delete?

tests/unit/fem/test_pa_idinterp.cpp
barker29: You don't see the output when you run `make test`, only when you run `./unit_tests` or similar. I thought that was intentional so users could decide how much to see, but then again I didn't know about the `INFO` option or the `-s` flag.

fem/pbilinearform.cpp
```diff
@@ -565,6 +565,53 @@ HypreParMatrix* ParDiscreteLinearOperator::ParallelAssemble() const
...
+   OperatorHandle P_test(A.Type()), P_trial(A.Type());
+
+   // TODO - construct the Dof_TrueDof_Matrix directly in the required format.
+   P_test.ConvertFrom(range_fes->Dof_TrueDof_Matrix());
+   P_trial.ConvertFrom(domain_fes->Dof_TrueDof_Matrix());
+
+   A.MakeRAP(P_test, dA, P_trial);
+}
+
+//// @todo copied from ParMixedBilinearForm, should be some inheritance?
```
dylan-copeland, line 588: @barker29 Should this be addressed in this PR?

tests/unit/fem/test_pa_idinterp.cpp
pazner: Good point — I almost always run the unit tests directly rather than with `make test` (usually so I can specify a subset of tests to run so it completes faster).

fem/pbilinearform.cpp
barker29: Probably, but it's not immediately clear to me how to do this. I'll try to think about it a bit.

tests/unit/fem/test_pa_idinterp.cpp
dylan-copeland: This has been changed to an `INFO` message. I was also unaware of `INFO`. I agree that using `INFO` is an improvement, but a disadvantage is that people might not know about the `-s` flag. A lot of these `catch` features are not well documented or advertised, so it can be tricky for people to follow, whereas `cout` is straightforward.

tests/unit/fem/test_pa_idinterp.cpp
dylan-copeland: I changed this to `MFEM_Approx(0.0, 1.0e-14)`. I agree it is an improvement, but a tolerance still needs to be given (otherwise, the default 1e-12 is used, which is arbitrary). Typically we set tolerances manually, based on the error magnitude observed in individual tests, so it makes sense to set the tolerance rather than use the default.

tests/unit/fem/test_pa_idinterp.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+TEST_CASE("PAIdentityInterp", "[CUDA]")
+{
+   for (bool transpose : {false, true})
+   {
+      for (int dim = 2; dim < 4; ++dim)
+      {
+         for (int num_elements = 0; num_elements < 5; ++num_elements)
+         {
+            for (int order = 1; order < 5; ++order)
```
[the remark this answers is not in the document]
dylan-copeland: This is done. `GENERATE` is a nice convenience, which perhaps should be documented somewhere. It seems the best documentation is the usage in the kernels that use it already.

tests/unit/fem/test_pa_idinterp.cpp
pazner: Agreed, we could edit our README or similar to better document these features. (There is also the flag `-r compact` for less verbose `-s` output). The available options are also shown with the `-h` or `--help` flags.

The point of `INFO` rather than outputting to stdout is that `INFO` will by default only output the message on test failure. I don't think these messages are very useful otherwise.

tests/unit/fem/test_pa_idinterp.cpp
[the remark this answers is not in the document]
pazner: The relevant Catch documentation is here: https://github.com/catchorg/Catch2/blob/devel/docs/generators.md#top

We could link to this page or add a short description somewhere in our own docs.

barker29:
Update: I'm having real trouble figuring out the parallel cuda gradient bug. It only shows up in 3D, only for certain polynomial orders, and is very painful to debug. Hopefully eventually I will get there but the reviewers (and editors) should not hold their breath.

v-dobrev:
> Update: I'm having real trouble figuring out the parallel cuda gradient bug. It only shows up in 3D, only for certain polynomial orders, and is very painful to debug. Hopefully eventually I will get there but the reviewers (and editors) should not hold their breath.

Can the issue be reproduced with the debug device: `-d debug`? If that is the case, debugging should be simpler -- everything is on host.

barker29:
Good idea @v-dobrev , but I can't reproduce the bug on the `debug` device, everything looks fine. Does that fact by itself tell me anything?

v-dobrev:
The debug device emulates the separate memory spaces, so maybe that part of the code is okay. If the debug device is okay, it looks the issue needs to be debugged on GPU.

You can try adding `HostRead()` + `printf` at some key places and compare the results from a host run with the GPU run.

dylan-copeland:
@v-dobrev Thanks for the suggestions. Also note that `HostRead()` sometimes fixes a bug, as I have found will debugging.

fem/restriction.hpp
barker29: Despite its name, `ElementRestriction` is more of a P matrix than a R matrix, it takes L-vectors to E-vectors and has more rows than columns. Is there a way to make the documentation clearer?

fem/pfespace.hpp
barker29: Changing to `GetLocalProlongationOperator`.

fem/pfespace.hpp
barker29: This is how we implement an MFEM `Interpolator` (as opposed to an `Integrator`) with partial assembly. Where in an assembled matrix you use `SetValues` instead of `AddValues`, for matrix-free you have to take care of this in the element restriction (Ceed's `G` operator). I think it makes sense as generally as the `Interpolator` idea.

fem/pbilinearform.cpp
barker29: Done.

fem/fespace.hpp
barker29: Done.

fem/restriction.hpp
pazner: Ok, that makes sense (I should have looked more closely at the rest of `restriction.hpp`) — I guess it's using the libCEED "ElemRestriction" terminology. There is some potential for confusion with the restriction that is given by `FiniteElementSpace::GetRestrictionMatrix`.

fem/pfespace.hpp
pazner: Ok, that makes sense to me. My only concern would be that between `GetConformingProlongation`, `GetProlongationMatrix` and `GetLocalProlongationOperator`, it may be a little bit confusing for users. I think this could be solved with documentation/comments.

tzanio:
Added more time for debugging.

barker29:
I think the bug is actually fixed, so this is (again) ready for review.

fem/bilinearform_ext.hpp
```diff
@@ -251,6 +252,33 @@ class PAMixedBilinearFormExtension : public MixedBilinearFormExtension
    void Update();
 };
 
+
+/**
```
barker29, line 256: This comment block could be formatted better (with `@brief`) and maybe clarified.

fem/bilininteg.hpp
```diff
@@ -2878,11 +2880,33 @@ class DiscreteInterpolator : public BilinearFormIntegrator { };
...
+   virtual ~GradientInterpolator() { delete fake_fe; }
+
    virtual void AssembleElementMatrix2(const FiniteElement &h1_fe,
                                        const FiniteElement &nd_fe,
                                        ElementTransformation &Trans,
                                        DenseMatrix &elmat)
    { nd_fe.ProjectGrad(h1_fe, Trans, elmat); }
+
+   using BilinearFormIntegrator::AssemblePA;
+   /**
```
barker29, line 2893: Reformat and expand, using `@param`.

fem/bilininteg.hpp
```diff
@@ -2878,11 +2880,33 @@ class DiscreteInterpolator : public BilinearFormIntegrator { };
...
+   */
+   virtual void AssemblePA(const FiniteElementSpace &trial_fes,
+                           const FiniteElementSpace &test_fes);
+
+   virtual void AddMultPA(const Vector &x, Vector &y) const;
+   virtual void AddMultTransposePA(const Vector &x, Vector &y) const;
+
+private:
+   /// 1D finit element that generates and owns the 1D DofToQuad maps below
+   FiniteElement * fake_fe;
```
barker29, line 2905: Don't really like the name `fake_fe`, maybe `dofquad_fe` or something?

fem/bilininteg.hpp
```diff
@@ -2897,6 +2921,24 @@ class IdentityInterpolator : public DiscreteInterpolator
...
+
+   virtual void AssemblePA(const FiniteElementSpace &trial_fes,
+                           const FiniteElementSpace &test_fes);
+
+   virtual void AddMultPA(const Vector &x, Vector &y) const;
+   virtual void AddMultTransposePA(const Vector &x, Vector &y) const;
+
+private:
+   /// 1D finit element that generates and owns the 1D DofToQuad maps below
+   FiniteElement * fake_fe;
```
barker29, line 2935: Ditto.

fem/bilininteg_hcurl.cpp
```diff
@@ -4843,6 +4844,1415 @@ void MixedVectorWeakCurlIntegrator::AddMultPA(const Vector &x, Vector &y) const
...
+   auto B = Reshape(_B.Read(), c_dofs1D, c_dofs1D);
+   auto G = Reshape(_G.Read(), o_dofs1D, c_dofs1D);
+
+   auto x = Reshape(_x.Read(), c_dofs1D, c_dofs1D, c_dofs1D, NE);
+   auto y = Reshape(_y.ReadWrite(), (3 * c_dofs1D * c_dofs1D * o_dofs1D), NE);
+
+   constexpr static int MAX_D1D = HCURL_MAX_D1D;
+   MFEM_VERIFY(c_dofs1D <= MAX_D1D && o_dofs1D <= c_dofs1D, "");
+
+   MFEM_FORALL(e, NE,
```
barker29, line 5018: For this (and all the other kernels), I wonder if there's a way to use eg `MFEM_FORALL_2/3D` in some places or `MFEM_FOREACH_THREAD` in some of the nested for loops inside this. Would that make it faster? Has anyone who knows anything about GPUs actually looked at these kernels?

fem/bilininteg_hcurl.cpp
```diff
@@ -4843,6 +4844,1415 @@ void MixedVectorWeakCurlIntegrator::AddMultPA(const Vector &x, Vector &y) const
...
+   auto Bo = Reshape(Bopen.Read(), o_dofs1D, c_dofs1D);
+
+   auto x = Reshape(_x.Read(), (3 * c_dofs1D * c_dofs1D * o_dofs1D), NE);
+   auto y = Reshape(_y.ReadWrite(), c_dofs1D, c_dofs1D, c_dofs1D, 3, NE);
+
+   auto vk = Reshape(pa_data.Read(), 3, (3 * c_dofs1D * c_dofs1D * o_dofs1D),
+                     NE);
+
+   constexpr static int MAX_D1D = HCURL_MAX_D1D;
+   //constexpr static int MAX_Q1D = HCURL_MAX_Q1D;
```
barker29, line 5686: Remove?

fem/bilininteg_hcurl.cpp
```diff
@@ -4843,6 +4844,1415 @@ void MixedVectorWeakCurlIntegrator::AddMultPA(const Vector &x, Vector &y) const
...
+   auto Bc = Reshape(Bclosed.Read(), c_dofs1D, c_dofs1D);
+   auto Bo = Reshape(Bopen.Read(), o_dofs1D, c_dofs1D);
+
+   auto x = Reshape(_x.Read(), c_dofs1D, c_dofs1D, 2, NE);
+   auto y = Reshape(_y.ReadWrite(), (2 * c_dofs1D * o_dofs1D), NE);
+
+   auto vk = Reshape(pa_data.Read(), 2, (2 * c_dofs1D * o_dofs1D), NE);
+
+   constexpr static int MAX_D1D = HCURL_MAX_D1D;
+   //constexpr static int MAX_Q1D = HCURL_MAX_Q1D;
```
barker29, line 5912: Remove?

fem/pbilinearform.cpp
```diff
@@ -565,6 +565,38 @@ HypreParMatrix* ParDiscreteLinearOperator::ParallelAssemble() const
...
+   dA.MakeRectangularBlockDiag(domain_fes->GetComm(),
+                               range_fes->GlobalVSize(),
+                               domain_fes->GlobalVSize(),
+                               range_fes->GetDofOffsets(),
+                               domain_fes->GetDofOffsets(),
+                               mat);
+
+   OperatorHandle P_test(A.Type()), P_trial(A.Type());
+
+   // TODO - construct the Dof_TrueDof_Matrix directly in the required format.
```
barker29, line 581: This comment is copied (by me!) from `ParBilinearForm` and `ParMixedBilinearForm` but I don't know what it means or why it's here. But I figured if it was important to be in those places it should also be here?

fem/pbilinearform.hpp
```diff
@@ -160,6 +160,9 @@ class ParBilinearForm : public BilinearForm
    /// Get the parallel finite element space prolongation matrix
    virtual const Operator *GetProlongation() const
    { return pfes->GetProlongationMatrix(); }
+   /// Get the diagonal part of the parallel finite element space prolongation matrix
```
barker29, line 163:
```suggestion
   /// Get the local part of the parallel finite element space prolongation matrix
```

fem/pfespace.cpp
```diff
@@ -3147,9 +3199,9 @@ static void SetSubVector(const int N,
                          const Array<int> &indices,
                          const Vector &in, Vector &out)
 {
-   auto y = out.Write();
-   const auto x = in.Read();
-   const auto I = indices.Read();
+   auto y = out.HostWrite();
```
barker29, line 3202: I realize this was the bug fix we were looking for for a long time, but it bothers me a little to do `SetSubVector` on the host in an object called `DeviceConformingProlongationOperator`. Shouldn't all the memory be on the device? Does this do a bunch of unnecessary copies?

fem/pfespace.hpp
```diff
@@ -335,6 +338,7 @@ class ParFiniteElementSpace : public FiniteElementSpace
    HYPRE_Int GetMyTDofOffset() const;
 
    virtual const Operator *GetProlongationMatrix() const;
+   virtual const Operator *GetLocalProlongationOperator() const;
```
barker29, line 341: As mentioned by Will above, we need some comments to distinguish/explain `GetProlongationMatrix()` and `GetLocalProlongationOperator()` here.

fem/restriction.cpp
```diff
@@ -195,6 +195,32 @@ void ElementRestriction::MultTransposeUnsigned(const Vector& x, Vector& y) const
...
+   auto d_y = Reshape(y.Write(), t?vd:ndofs, t?ndofs:vd);
+   MFEM_FORALL(i, ndofs,
+   {
+      const int nextOffset = d_offsets[i + 1];
+      for (int c = 0; c < vd; ++c)
+      {
+         double dofValue = 0;
+         const int j = nextOffset - 1;
+         const int idx_j = (d_indices[j] >= 0) ? d_indices[j] : -1 - d_indices[j];
+         dofValue =
```
barker29, line 216: This passes `make style` but it's still terrible. Is there a way to fix the indentation on the next few lines?

barker29 (changes requested):
(I wrote a chunk of this PR, but I'm also assigned as an official reviewer, so here goes.)

This PR is very close to being ready - I've marked a few minor things, mostly documentation, that should be corrected (I will do some of them myself at some point). There are also a couple questions about kernel `FORALL` and `HostRead` etc that are probably premature optimization - if we are happy with the device performance overall we can probably ignore those questions.

fem/pfespace.cpp
dylan-copeland: @barker29 Good question, and I think the answer is that the memory is already on host. `SetSubVector` is called only by `DeviceConformingProlongationOperator::BcastLocalCopy` and `DeviceConformingProlongationOperator::BcastEndCopy`, which are called only by `DeviceConformingProlongationOperator::Mult`, and when they are called, it is after some `HostRead` and `HostWrite` calls which are necessitated by `MPI_Isend` and `MPI_Irecv` calls. Thus the data is on host because of some `MPI` calls. An extra copy to host might hurt performance, but if the data is already on host, then setting a subvector on host seems fine to me.

fem/bilininteg_hcurl.cpp
dylan-copeland: @YohannDudouit @camierjs Please correct me if I am wrong, but my understanding is that `MFEM_FORALL_2/3D` sets up blocks of GPU threads indexed with respect to quadrature indices (`qx`, `qy`, `qz`), which is beneficial for shared memory usage. I don't see a benefit for just doing some computations on each element. The usage of `MFEM_FORALL` in this kernel assigns one GPU thread per element which should be fine. I could be wrong, or maybe my understanding needs to be updated.

fem/pfespace.cpp
camierjs: We should look at it as I remember that there can be direct `mpi` calls with data on the `device`.
I'm not sure that the `data` is always on the `host` there.

fem/bilininteg_hcurl.cpp
v-dobrev: If there is fine-grain parallelism within elements then there will be benefit in assigning multiple threads (a thread block) to one element.

fem/pfespace.cpp
v-dobrev: It does not make sense to call `Host{Read,Write,ReadWrite}` and then use these host pointers inside `MFEM_FORALL` which expects device pointers. If this copy has to happen on host, then a normal `for` loop is what you need.

However, I do think we want to do these copies on device.

v-dobrev: One thing that is a bit tricky here is that only part of `y` is overwritten, so using `Write` may be incorrect if `y` is actually on host.

dylan-copeland: @camierjs @v-dobrev Thanks for your comments, and I agree with them. First, I was not aware of `mpi_gpu_aware` (I guess @barker29 wrote this part?). This is interesting, and I did not know MPI calls could be made with data on device. It must be doing something more efficient than copying from device to host in this case. Is there some documentation on this? I only see a very concise declaration in `device.hpp`. For a few tests (unit tests and ex3p) with `pcuda` builds on lassen, I printed out `mpi_gpu_aware` and found it is always `false`. How can it be enabled? Assuming it can be `false` or `true`, I suppose the implementation of `SetSubVector` should depend on `mpi_gpu_aware` and `local` (if `local`, there is no MPI communication in any case). At least we can determine cases where data need not be on host.

fem/bilininteg_hcurl.cpp
barker29: Looking at this and some of the other kernel implementations more carefully, I don't think there's much to be gained here.

v-dobrev: For, example each entry of `w1`, `w1[dx][dy][ez]`, can be computed independently, right? I think this is the same as (or at least similar to) what all the shared memory kernels do. It is not crucial to do it right now, just something to keep in mind for the future as an optimization opportunity.

v-dobrev: One other observation regarding the gradient computation: in the typical use case, where `Q_k` H1 function is mapped to `Nedelec_k` H(curl) function, the 1D `B` matrix is the identity. Therefore, from a performance point of view, a specialized kernel for that case is worth implementing.

barker29: This is a good point! I just pushed a commit that includes these specialized kernels. It works fine, but it is four new (long, lots of copied code) kernels and if we want to roll back that commit because we think the maintenance cost is higher than what we gain in performance, I would not object.

barker29 (approved):
I think this is ready - the last thing we were worrying about is the host/device copies in `DeviceConformingProlongationOperator`, but the current code works fine and I don't think this is a very serious issue to prevent merging.

tzanio:
@pazner, can you please take a look?

fem/pfespace.cpp
camierjs: It can be enabled with `Device::SetGPUAwareMPI()`. The `MPI` implementation _should_ do something efficient, but we haven't seen yet one avoiding some internal copies (that was ~one year ago with `Laghos`) and could be revisited to see if the expected gain is reachable.

fem/bilininteg_hcurl.cpp
```diff
@@ -4843,6 +4844,1923 @@ void MixedVectorWeakCurlIntegrator::AddMultPA(const Vector &x, Vector &y) const
    }
 }
 
+// Apply to x corresponding to DOFs in H^1 (domain) the (topological) gradient
+// to get a dof in H(curl) (range). You can think of the range as the "test" space
+// and the domain as the "trial" space, but there's no integration.
+static void PAHcurlApplyGradient2D(const int c_dofs1D,
+                                   const int o_dofs1D,
+                                   const int NE,
+                                   const Array<double> &_B,
```
camierjs, line 4853: We should remove all the `_*` variable names to avoid errors with preprocessors and `cygwin`.

dylan-copeland: @camierjs I will fix this, but I don't understand why this is an issue if there are so many occurrences beyond this PR. If you do `grep "&\_" *` in `mfem/fem` you will find many occurrences in many files, in the master branch. Does the master branch not give errors? Should I fix all of them or just the ones in this PR?

camierjs: I was just looking at #1828 and thinking that we should not add some more. I agree that we will have to go all over the code to sanitize this names. Just the ones in this PR would be great.

dylan-copeland: Thanks for clarifying; now it makes sense. I think I fixed all instances introduced in this PR, with the latest commit.

barker29: To be clear, the issue is with the *leading* underscore `_something`. Trailing underscores `something_` are fine.

dylan-copeland: @barker29 Thanks for underscoring that.

pazner (approved):
Looks good to me!

fem/pfespace.cpp
dylan-copeland: @camierjs I tried adding `Device::SetGPUAwareMPI()` to ex3p, just before the declaration of `Device device`, but there is a segmentation fault, and TotalView shows the crash somewhere in the internals of `MPI_Isend`. Maybe I did something wrong (building configuration?).

tzanio:
I'm extending the deadline because I need a bit more time to review this.

fem/fespace.hpp
```diff
@@ -330,6 +330,12 @@ class FiniteElementSpace
    virtual const Operator *GetProlongationMatrix() const
    { return GetConformingProlongation(); }
 
+   /// Return a prolongation operator that only acts on local dofs
+   /** The returned operator is owned by the FiniteElementSpace. This
+       differs from GetProlongationMatrix only in parallel. */
+   virtual const Operator *GetLocalProlongationOperator() const
+   { return GetConformingProlongation(); }
```
tzanio, line 337: This is confusing and needs to be clarified, see the comment in `operator.hpp`

fem/bilinearform.hpp
```diff
@@ -375,6 +375,9 @@ class BilinearForm : public Matrix
    /// Get the output finite element space prolongation matrix
    virtual const Operator *GetOutputProlongation() const
    { return GetProlongation(); }
+   /// Get the output finite element space prolongation matrix (local diagonal)
+   virtual const Operator *GetLocalOutputProlongation() const
+   { return GetOutputProlongation(); }
```
tzanio, line 380: This is confusing and needs to be clarified, see the comment in `operator.hpp`

fem/bilinearform.hpp
```diff
@@ -984,9 +987,17 @@ class DiscreteLinearOperator : public MixedBilinearForm
...
+   /** This method must be called before assembly. */
+   void SetAssemblyLevel(AssemblyLevel assembly_level);
+
    /** @brief Construct the internal matrix representation of the discrete
        linear operator. */
    virtual void Assemble(int skip_zeros = 1);
+
+   /// Get the output finite element space prolongation matrix (local diagonal)
+   virtual const Operator *GetLocalOutputProlongation() const
+   { return test_fes->GetLocalProlongationOperator(); }
```
tzanio, line 1000: This is confusing and needs to be clarified, see the comment in `operator.hpp`

fem/pbilinearform.hpp
```diff
@@ -301,10 +304,18 @@ class ParDiscreteLinearOperator : public DiscreteLinearOperator
    /// Returns the matrix "assembled" on the true dofs
    HypreParMatrix *ParallelAssemble() const;
 
+   /** @brief Returns the matrix assembled on the true dofs, i.e.
+       @a A = P_test^t A_local P_trial, in the format (type id) specified by
+       @a A. */
+   void ParallelAssemble(OperatorHandle &A);
```
tzanio, line 310: Maybe I'm missing something, but this seems incorrect -- if we want to have a `ParallelAssemble` for `ParDiscreteLinearOperator` it should be `R_test A_local P_trial` not `P_test^t A_local P_trial`...

fem/pbilinearform.cpp
```diff
@@ -565,6 +565,38 @@ HypreParMatrix* ParDiscreteLinearOperator::ParallelAssemble() const
...
+   P_test.ConvertFrom(range_fes->Dof_TrueDof_Matrix());
+   P_trial.ConvertFrom(domain_fes->Dof_TrueDof_Matrix());
+
+   A.MakeRAP(P_test, dA, P_trial);
```
tzanio, line 585: This doesn't look right, we should use `R_test` here.

fem/pfespace.cpp
```diff
@@ -3147,9 +3199,9 @@ static void SetSubVector(const int N,
...
+   const auto x = in.HostRead();
+   const auto I = indices.HostRead();
    MFEM_FORALL(i, N, y[I[i]] = x[i];);
 }
```
tzanio, line 3206: I am confused by this too -- do we want the above assignment on the host or on the device? In the former case, `MFEM_FORALL` may look confusing, in the latter case the `HostRead` is confusing...

tzanio (changes requested):
I apologize for the significant delay of my review, I should have commented on this earlier.

I have several concerns described below, and I wonder if I'm missing something...

On a general level the difference between a bilinear form and a discrete interpolator, despite the computation of the local element matrix `A`, is that the global matrix is `R_test A P_trial` instead of `R_test^T A P_trial` where `R_test` is the left inverse of `P_test`.

Maybe I've missed it but I don't quite see that in the code. Instead there are several references to a "LocalProlongationOperator" which is confusing for me.

I am sorry for not bringing this up earlier, but can we maybe have technical meeting to discuss this after the break?

linalg/operator.hpp
```diff
@@ -112,6 +113,10 @@ class Operator
    {
       return GetProlongation(); // Assume square unless specialized
    }
+   /** @brief Prolongation operator from linear algebra (linear system) vectors,
+       to output vectors for the operator, including only processor-local portion.
+       `NULL` means identity. */
+   virtual const Operator *GetLocalOutputProlongation() const { return NULL; }
```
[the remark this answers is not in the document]
barker29: The way I think about this operator is that it assembles with `SetValues` instead of `AddValues`. You need to enforce that in two places, once at the element-assembly level (by using `SetValues`) and once at the parallel assembly level (by simply not writing on dofs you do not own, which means using a "local" matrix/operator so you only write on your own dofs). This is probably equivalent to your `R_test` left-inverse way of thinking, but I'm trying to explain where these names/documentation comes from. In particular, the `Operator` this returns is in fact processor-local.

I recall having this discussion with Veselin either on Slack or GitHub, and he suggested these "local" names, but I can't find it now and I could be remembering wrong.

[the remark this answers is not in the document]
tzanio: Thanks Andrew, I think we are on the same page.

Assembling is just the action of `P_test^T`, while "setting" is the action of `R_test`. This is easy to see in the Boolean conforming case, but also holds in more general cases, e.g. nonconforming AMR.

Are you saying that `LocalOutputProlongation` above is just the local portion of `R_test`? In that case I'm confused by the name "prolongation", this is really a restriction.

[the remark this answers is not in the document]
barker29: `LocalOutputProlongation` is the local portion of `R_test^T`, it's shaped like a prolongation but I guess it's constructed more like `R_test`. In addition to clarifying the documentation (which I will do), do you have a suggestion for the name of the method itself? `GetOutputProlongation` already exists, and adding `Local` seemed like a reasonable way to distinguish this method from that one.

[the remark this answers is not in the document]
tzanio: Ok, I understand now.

I personally prefer a name like `GetOutputRestrictionTranspose`, but I'd like to understand something first -- why do we need another object, can't we just use `ParFiniteElementSpace::GetRestrictionMatrix`?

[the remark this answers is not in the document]
barker29: Well, for one thing `ParFiniteElementSpace::GetRestrictionMatrix` returns a `SparseMatrix*`, and the whole point of this PR is to do things matrix-free. You could view this as another reason for the name, we build on `(Device)ConformingProlongationOperator` for the matrix-free action of these operators. I guess one option is to change the interface for `GetRestrictionMatrix()` to return an `Operator`, but that seems invasive.

It might be best to schedule a technical meeting about all this. I don't object to changing names etc., but there are reasons for what @dylan-copeland  and I did.

fem/pbilinearform.cpp
```diff
@@ -565,6 +565,38 @@ HypreParMatrix* ParDiscreteLinearOperator::ParallelAssemble() const
...
+   // construct the rectangular block-diagonal matrix dA
+   OperatorHandle dA(A.Type());
+   dA.MakeRectangularBlockDiag(domain_fes->GetComm(),
+                               range_fes->GlobalVSize(),
+                               domain_fes->GlobalVSize(),
+                               range_fes->GetDofOffsets(),
+                               domain_fes->GetDofOffsets(),
+                               mat);
+
+   OperatorHandle P_test(A.Type()), P_trial(A.Type());
```
barker29, line 579: rename to `R_test_transpose` (locally)

barker29:
Based on discussion on 5 January 2021, we have decided that `OutputProlongation`, `LocalOutputProlongation` etc should be renamed to be some kind of `Restriction` or `RestrictionTranspose`.

fem/pfespace.cpp
dylan-copeland: It has been a while since I tested this, and I cannot reproduce the bug that these `HostRead` calls fixed. I think the bug was only revealed in `MatrixFreeSolver` in branch `barker29/matrix-free-ams`, which seems to be working fine without `HostRead`. So I am reverting this change, and any issues can be fixed in a future PR for `barker29/matrix-free-ams`. It looks like there is a case that may need to be fixed, where `mpi_gpu_aware` and `local` are false in `DeviceConformingProlongationOperator::Mult` and the data is on host for the MPI communication. If so, the fix should not be in `SetSubVector` but in `DeviceConformingProlongationOperator::Mult`, with copying on host just in that case.

barker29:
I think all the review comments have been addressed.

fem/fespace.hpp
```diff
@@ -330,6 +330,12 @@ class FiniteElementSpace
    virtual const Operator *GetProlongationMatrix() const
    { return GetConformingProlongation(); }
 
+   /// Return a prolongation operator that only acts on local dofs
+   /** The returned operator is owned by the FiniteElementSpace. This
+       differs from GetProlongationMatrix only in parallel. */
```
pazner, line 335: Comment should be updated given the new name.

fem/fespace.hpp
```diff
@@ -330,6 +330,12 @@ class FiniteElementSpace
...
+   virtual const Operator *GetRestrictionTransposeOperator() const
+   { return GetConformingProlongation(); }
```
pazner, line 337: Perhaps it would be more intuitive to provide `FiniteElementSpace::GetRestrictionOperator`, which returns an `Operator` that performs the exact same action as `FiniteElementSpace::GetRestrictionMatrix`, but potentially matrix-free. The caller can transpose using `TransposeOperator` (or `GetRestrictionTransposeOperator` can also be provided to avoid stacking transposes).

I think that would be easier to understand conceptually, and the fact that the transpose is implemented using `ConformingProlongationOperator` is (if I understand correctly) mostly done out of the convenience, and is an implementation detail that the user does not need to be aware of.

barker29: I can certainly do this. Here's my hesitation.

`FiniteElementSpace` and `ParFiniteElementSpace` currently have `Operator * GetProlongationMatrix()` and `SparseMatrix* GetRestrictionMatrix()`. At some point in the transition to partial assembly, the return type of `GetProlongationMatrix()` was changed from `SparseMatrix*` to `Operator*`.

One option is what you propose, which results in having `Operator* GetRestrictionOperator()`, `SparseMatrix* GetRestrictionMatrix()`, and `Operator* GetProlongationMatrix()`, which seems confusing.

Another option is to change the name of `GetProlongationMatrix()` to `GetProlongationOperator()`, which is sensible but changes interfaces and potentially breaks user code (though I suspect this is almost exclusively used internally, I doubt users touch it much).

What I think is the "correct" thing to do (this is only becoming apparent to me now because of our discussions) is to change the return type of `GetRestrictionMatrix()` to `Operator*`. This will often be a `SparseMatrix`, and when it is not it could be implemented with `(Device)ConformingProlongationOperator` with a flag or with `TransposeOperator`. The problem here is that `GetRestrictionMatrix` is called ~40 times in the code base, more than a few of these calls seem to require an explicit assembled matrix, and I would need significant help and time to sort it out.

pazner: I actually think that `Operator* GetRestrictionOperator()` and `SparseMatrix* GetRestrictionMatrix()` seem quite reasonable. The intent is conveyed pretty well by the name of each function.

What I agree is confusing is that `GetProlongationMatrix` may return something that isn't a matrix. That is definitely unfortunate, but it isn't related to this PR.

I would prefer to avoid a situation where there are two different, but similarly named functions that differ in some way that is not necessarily clear up-front (just as an example, `Mesh::GetNFaces` returns the 2D entities in a 3D mesh, `Mesh::GetNumFaces` returns the codim-1 entities in a general mesh — the user cannot infer this distinction from the difference between `N` and `Num`). Similarly, it might not be clear that the really important distinction between `GetRestrictionTransposeOperator` and `GetRestrictionMatrix` is not that one is transposed (which is really just an artifact of this particular implementation), but that one is matrix-free (the confusion is mitigated by the distinction between `Operator` and `Matrix`, but I think it's still present).

I also agree that the "best" interface would require breaking backwards compatibility (and so this is one instance where it would be useful to have some clear guidance on what the process is for introducing breaking API changes, cc: @tzanio), but I think there is still some room for improvement here without breaking changes.

rcarson3: Just saw your comment pop up on my feed @barker29. I figured briefly mention that I'd probably be one of the few people where changing the name `GetProlongationMatrix` would break parts of my application code that deal with partial and element assembly methods. Although, I will say I only use that function like once, so it would be fairly simple fix.

barker29: Thanks Will, that's very clear and you have convinced me. I will adopt your suggestion and we can deal with `Operator* GetProlongationMatrix` later.

barker29: I believe this is resolved.

linalg/operator.hpp
```diff
@@ -112,6 +113,11 @@ class Operator
    {
       return GetProlongation(); // Assume square unless specialized
    }
+   /** @brief Transpose of GetOutputRestriction, directly available in this
+       form to facilitate matrix-free RAP-type operators.
+
+       `NULL` means identity. */
```
tzanio, line 119: Thanks for adding these comments

tzanio (commented):
Please mention in `CHANGELOG`

tzanio:
@pazner, can you confirm that you still approve this (after the recent changes)?

pazner:
Yes, looks good to me. I think the latest changes help to clarify matters.

tzanio (approved):
Thanks for clarifying and reworking this @dylan-copeland and @barker29 !

tzanio:

PR #1765, opened 2020-09-17
Default implementation of AssembleElementVector and AssembleFaceVector for bilinear integrators [blfi-default-assemble-vector-dev]

tuckerbabcock:
As stated in the title, this adds a default implementation of `BilinearFormIntegrator::AssembleElementVector` and `BilinearFormIntegrator::AssembleFaceVector`that simply multiply the state vector by the element/face stiffness matrix to compute the action of the integrator.

I realize this isn't necessarily a very efficient implementation, but it should allow someone to easily use a bilinear form integrator in a `NonlinearForm`, and the necessary functions can be overridden if need be for performance. Thoughts?

Thanks,

Tucker

| [#1765](https://github.com/mfem/mfem/pull/1765) | @tuckerbabcock | @tzanio | @rcarson3 + @jandrej | 09/22/20 | 09/27/20 | 10/04/20 | |

### DIFF
--- a/fem/bilininteg.cpp
+++ b/fem/bilininteg.cpp
@@ -133,6 +133,7 @@ void BilinearFormIntegrator::AssembleElementVector(
    const FiniteElement &el, ElementTransformation &Tr, const Vector &elfun,
    Vector &elvect)
 {
+   // Note: This default implementation is general but not efficient
    DenseMatrix elmat;
    AssembleElementMatrix(el, Tr, elmat);
    elvect.SetSize(elmat.Height());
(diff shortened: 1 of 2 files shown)


jandrej:
Good approach and helpful in general.

jandrej (approved):

rcarson3 (approved):
So, I don't see the harm of approving this as is. Since, it just performs the local action of the element matrix. 

I do worry about potentially new users incorrectly using this within a NonlinearForm class, but I guess most of those would go for the BilinearForm/LinearForm combo instead. So, the likely hood of that occurring are pretty low.

tzanio:

tzanio (approved):
Thanks for the contribution @tuckerbabcock !

PR #1766, opened 2020-09-17
Refactor the libCEED integration and add support for Convection.

YohannDudouit:
This PR adds support using libCEED for:
- [x] `ConvectionIntegrator`,
- [x] `VectorConvectionNLFIntegrator`,
- [x] `VectorCoefficient`,
~~This should be reviewed after #1719~~

| [#1766](https://github.com/mfem/mfem/pull/1766) | @YohannDudouit | @v-dobrev | @barker29 + @pazner | 10/29/20 | 02/08/21 | 02/16/21 | |


YohannDudouit:
@barker29 This PR adds support for `ConvectionIntegrator` if you want to try your libCEED based solver on it.

YohannDudouit:
@jandrej This PR adds support for `VectorConvectionNLFIntegrator` if you want to try it in your Navier-Stokes miniapp.

YohannDudouit:
I've caught some bugs, I was making a mistake for the tests of non-linear forms, I put this back in WIP.

barker29 (commented):
Overall this looks very good, but I'm seeing differences in `ex9p`. Running `./ex9p -pa` is fine, but `./ex9p -pa -d ceed-cpu` fails with an infinite norm in CG after some timesteps. I tried also with a non-periodic mesh and got the same result, and also adding a `ConvectionIntegrator` to `ex1p` (and switching to GMRES) gives a similar problem.

YohannDudouit:
@barker29 I think I have addressed your remarks.

barker29:
I still have the problem with `ex9p`.

tzanio:
@YohannDudouit can merge `master` and resolve conflicts?

pazner (commented):
The changes and additions look good to me.

Has the issue with ex9 been resolved? Looks like it was an error in the coefficient computation in the Q-function... I would have thought the unit tests would catch this, do you know what happened?

YohannDudouit:
@tzanio I will merge master, I found the bugs but I didn't fix it in a clean way, I want to clean this, and I'll merge master.

tests/unit/ceed/test_ceed.cpp
```diff
@@ -20,11 +20,25 @@ using namespace mfem;
...
+// Velocity coefficient
+void velocity_function(const Vector &x, Vector &v)
+{
+   int dim = x.Size();
+   switch (dim)
+   {
+      case 1: v(0) = 1.0; break;
+      case 2: v(0) = 1.0; v(1) = 1.0; break;
+      case 3: v(0) = 1.0; v(1) = 1.0; v(2) = 1.0; break;
+   }
```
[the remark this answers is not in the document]
YohannDudouit: I think this is why the tests didn't find the bug, this was too simple of a vector coefficient.

CHANGELOG
```diff
@@ -25,6 +25,11 @@ Version 4.2.1 (development)
 - Implemented a filter method for the Navier miniapp to stabilize highly
   turbulent flows in direct numerical simulation.
 
+libCEED integration improvements
+--------------------------------
+- Add support for  VectorCoefficient.
+
+- Add support for ConvectionIntegrator, and VectorConvectionNLFIntegrator.
```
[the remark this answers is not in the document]
YohannDudouit: It's under:
```
libCEED integration improvements
--------------------------------
```
Are you sure it's not redundant?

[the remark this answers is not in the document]
pazner: Yeah, I noticed that, but personally I think this is clearer. Up to you though.

tests/unit/ceed/test_ceed.cpp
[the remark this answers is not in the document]
pazner: Aha, yeah, I think it would be good to double check that a variable coefficient would have really caught the problem so we can be confident that our unit tests are robust.

YohannDudouit:
@barker29 @pazner ex9 is working now, my fix is not the cleanest, currently `Convection` only works with `alpha = -1` (this value is hardcoded on the libCEED side). I have to redesign the way `BuildContext` are handled to make this cleaner but this is not a small change so I prefer to postpone this refactor.

pazner:
> @barker29 @pazner ex9 is working now, my fix is not the cleanest, currently `Convection` only works with `alpha = -1` (this value is hardcoded on the libCEED side). I have to redesign the way `BuildContext` are handled to make this cleaner but this is not a small change so I prefer to postpone this refactor.

Maybe for now a compromise would be to have `CeedScalar coeff[4]` in the `BuildContext`, and use `coeff[3]` as the `alpha` for convection?

I agree this is an ugly fix, but this could be revamped when you do a redesign of `BuildContext`.

I'd prefer not to have `alpha=-1` hard-coded. It makes sense for ex9, since the convection term is moved to the RHS for time integration, but wouldn't make sense for steady problems in residual formulation.

tests/unit/ceed/test_ceed.cpp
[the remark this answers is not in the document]
pazner: Did you have the chance to see if the unit tests with the new velocity would have caught the bugs from before?

fem/libceed/ceed.cpp
```diff
@@ -110,21 +110,97 @@ void InitCeedCoeff(Coefficient *Q, Mesh &mesh,
...
+   else
+   {
+      CeedVecQuadCoeff *ceedCoeff = new CeedVecQuadCoeff;
+      const int ne = mesh.GetNE();
+      const int dim = mesh.Dimension();
+      const int nq = ir.GetNPoints();
+      ceedCoeff->coeff.SetSize(dim * nq * ne);
+      auto C = Reshape(ceedCoeff->coeff.HostWrite(), dim, nq, ne);
+      DenseMatrix Q_ir;
+      for (int e = 0; e < ne; ++e)
```
barker29, line 168: How expensive is this loop in general? To me it looks like linear cost in the number of quad points, so I almost want to just do this all the time regardless of the type of `VQ`. That would make the code simpler (remove all the `select...case` blocks), but I don't know if it is worth the performance penalty.

fem/libceed/convection.h
```diff
@@ -0,0 +1,523 @@
...
+      case 22:
+         for (CeedInt i = 0; i < Q; i++)
+         {
+            // J: 0 2   qd: 0 1   adj(J):  J22 -J12
+            //    1 3       1 2           -J21  J11
+            const CeedScalar J11 = J[i + Q * 0];
+            const CeedScalar J21 = J[i + Q * 1];
+            const CeedScalar J12 = J[i + Q * 2];
+            const CeedScalar J22 = J[i + Q * 3];
+            const CeedScalar w = - qw[i]; // TODO remove - and use alpha
```
barker29, line 49: Will's suggestion is better, but if you don't do that I recommend putting `w = alpha * wq[i];` here (and in all similar lines) and `constexpr double alpha = -1.0` at the top of the function or something similar. That way future extension / maintainability is easier even if we don't fix the alpha issue in this PR.

barker29 (approved):
I have a couple comments/suggestions, but honestly I think we're reaching the point where potential improvements are more like restructuring of the whole MFEM/Ceed interface that should go in a different PR. Since I'm confident @YohannDudouit will continue maintaining/refining that interface, I don't think this is our last chance to get this right and so I'm ready to approve.

pazner (approved):
> I have a couple comments/suggestions, but honestly I think we're reaching the point where potential improvements are more like restructuring of the whole MFEM/Ceed interface that should go in a different PR. Since I'm confident @YohannDudouit will continue maintaining/refining that interface, I don't think this is our last chance to get this right and so I'm ready to approve.

I agree — I think this PR provides nice functionality, and it would be great to see a version of the Navier-Stokes miniapp using libCEED for all of the integrators (does this exist in a branch somewhere?). It's also a good step in the direction of getting ex9 to work using libCEED (waiting only on DG support in libCEED). The limitations regarding `alpha` and the general structure of the `BuildContext` will need to be addressed, because I believe the current implementation is too limited. That being said, it's already a lot of code to review, and restructuring the `BuildContext` will be significant work, so I think it makes sense to approve this PR with the understanding that those issues will be addressed in a follow-up PR.

fem/libceed/ceed.cpp
YohannDudouit: I think this loop is much more expensive than the other cases in general (it is never computed on GPU for instance).

YohannDudouit:
@barker29 @pazner I merge the libCEED refactor branch, besides this I only moved the `MFEMCeedOperator` into `NonlinearFormIntegrator` to avoid having to duplicate code inside each `Integrator` that supports libCEED.

@tzanio I lowered the order of test to only order 1, now the libCEED tests take about 6 sec in debug an 1 sec in release version.

fem/libceed/convection.cpp
```diff
@@ -0,0 +1,83 @@
...
+#endif
+
+namespace mfem
+{
+
+namespace ceed
+{
+
+#ifdef MFEM_USE_CEED
+struct ConvectionInfo
```
barker29, line 26: I would prefer a name that describes the difference with `ConvectionContext`, maybe `ConvectionKernelInfo`.

fem/libceed/coefficient.hpp
```diff
@@ -0,0 +1,216 @@
...
+};
+
+struct QuadCoefficient : VariableCoefficient
+{
+   mfem::Vector coeff;
+   CeedElemRestriction restr;
+   QuadCoefficient(int ncomp_) : VariableCoefficient(ncomp_, CEED_EVAL_NONE) { }
+};
+
+/** @brief Identifies the type of coefficient of the Integrator to initialize
```
barker29, line 77: This documentation could be improved. The reference to `Integrator` is confusing because that type isn't used at all here. The documentation should specify which of `Q` and `coeff_ptr` is input and which is output, and probably emphasize that `coeff_ptr` is `mfem::ceed::Coefficient`, a different type from `Q`.

fem/libceed/coefficient.hpp
```diff
@@ -0,0 +1,216 @@
...
+         {
+            C(q,e) = Q->Eval(T, ir.IntPoint(q));
+         }
+      }
+      InitVector(ceedCoeff->coeff, ceedCoeff->coeffVector);
+      coeff_ptr = ceedCoeff;
+   }
+}
+
+/** @brief Identifies the type of vector coefficient of the Integrator to
```
barker29, line 140: Same comment as above.

fem/libceed/diffusion.cpp
```diff
@@ -11,53 +11,69 @@
...
-   CeedPAOperator diffOp = {fes, irm,
-                            dim * (dim + 1) / 2, "/diffusion.h",
-                            ":f_build_diff_const", f_build_diff_const,
-                            ":f_build_diff_quad", f_build_diff_quad,
-                            ":f_apply_diff", f_apply_diff,
-                            CEED_EVAL_GRAD,
-                            CEED_EVAL_GRAD
-                           };
-   CeedPAAssemble(diffOp, ceedData);
+struct DiffusionInfo
```
barker29, line 26: Maybe `DiffusionKernelInfo`

fem/libceed/mass.cpp
```diff
@@ -11,52 +11,66 @@
...
 namespace mfem
 {
 
-void CeedPAMassAssemble(const FiniteElementSpace &fes,
-                        const mfem::IntegrationRule &irm, CeedData& ceedData)
+namespace ceed
+{
+
+#ifdef MFEM_USE_CEED
+struct MassInfo
```
barker29, line 26: Suggest `MassKernelInfo`

fem/libceed/nlconvection.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace mfem
+{
+
+namespace ceed
+{
+
+/** Represent a VectorConvectionNLFIntegrator with AssemblyLevel::Partial
+    using libCEED. */
+class PANLConvectionIntegrator : public PAIntegrator
```
barker29, line 26: Should be named `PAVectorConvectionNLFIntegrator` for consistency.

fem/libceed/nlconvection.hpp
```diff
@@ -0,0 +1,48 @@
...
+{
+public:
+   PANLConvectionIntegrator(const mfem::FiniteElementSpace &fes,
+                            const mfem::IntegrationRule &irm,
+                            mfem::Coefficient *coeff);
+};
+
+/** Represent a VectorConvectionNLFIntegrator with AssemblyLevel::None
+    using libCEED. */
+class MFNLConvectionIntegrator : public MFIntegrator
```
barker29, line 36: Should be `MFVectorConvectionNLFIntegrator`

fem/libceed/util.hpp
```diff
@@ -0,0 +1,119 @@
...
+/** @brief Remove from ceed_basis_map and ceed_restr_map the entries associated
+    with the given @a fes. */
+void RemoveBasisAndRestriction(const mfem::FiniteElementSpace *fes);
+
+#ifdef MFEM_USE_CEED
+
+/// Initialize a CeedVector from a Vector
+void InitVector(const mfem::Vector &v, CeedVector &cv);
+
+/// Initialize a strided CeedElemRestriction
```
barker29, line 47: This comment doesn't tell you any more than a name. Maybe describe the inputs, or at least refer to the libCEED documentation.

fem/libceed/util.hpp
```diff
@@ -0,0 +1,119 @@
...
+void InitStridedRestriction(const mfem::FiniteElementSpace &fes,
+                            CeedInt nelem, CeedInt nqpts, CeedInt qdatasize,
+                            const CeedInt *strides,
+                            CeedElemRestriction *restr);
+
+/// Initialize a CeedBasis and a CeedElemRestriction
```
barker29, line 53: Something like "Based on MFEM element description in fes and integration rules, construct Ceed basis and elemrestriction objects."

fem/libceed/nlconvection.cpp
```diff
@@ -0,0 +1,79 @@
...
+#endif
+
+namespace mfem
+{
+
+namespace ceed
+{
+
+#ifdef MFEM_USE_CEED
+struct NLConvectionInfo
```
barker29, line 26: suggest `NLConvectionKernelInfo`

fem/libceed/mf_integrator.hpp
```diff
@@ -0,0 +1,253 @@
...
+#include "../gridfunc.hpp"
+
+namespace mfem
+{
+
+namespace ceed
+{
+
+#ifdef MFEM_USE_CEED
+/** This structure contains the data to assemble a MF operator with libCEED.
```
barker29, line 28:
```suggestion
/** This structure contains the data to assemble a matrix-free operator with libCEED.
```

fem/libceed/pa_integrator.hpp
```diff
@@ -0,0 +1,275 @@
...
+#include "../gridfunc.hpp"
+
+namespace mfem
+{
+
+namespace ceed
+{
+
+#ifdef MFEM_USE_CEED
+/** This structure contains the data to assemble a PA operator with libCEED.
```
barker29, line 27: PA -> partially assembled

fem/libceed/pa_integrator.hpp
```diff
@@ -0,0 +1,275 @@
...
+protected:
+   CeedOperator build_oper;
+   CeedQFunction build_qfunc;
+
+public:
+   PAIntegrator()
+      : MFIntegrator(),  build_oper(nullptr), build_qfunc(nullptr) { }
+
+   template <typename CeedInfo, typename CoeffType>
+   PAOperator InitPA(CeedInfo &info,
```
barker29, line 68: This could use a few lines of documentation; also using the template parameter `CeedKernelInfo` IMO would make it more readable.

fem/libceed/mf_integrator.hpp
```diff
@@ -0,0 +1,253 @@
...
+   /** The evaluation mode to apply to the trial function (CEED_EVAL_INTERP,
+       CEED_EVAL_GRAD, etc.) */
+   EvalMode trial_op;
+   /** The evaluation mode to apply to the test function ( CEED_EVAL_INTERP,
+       CEED_EVAL_GRAD, etc.) */
+   EvalMode test_op;
+};
+#endif
+
+class MFIntegrator : public Operator
```
barker29, line 51: I think this class needs to be documented, especially since `PAIntegrator` inherits from it.

tests/unit/ceed/test_ceed.cpp
```diff
@@ -13,19 +13,36 @@
 #include "mfem.hpp"
```
barker29, line 13: I would like to see at least one test with multiple integrators, ie mass + diffusion. The machinery here is complex enough that it's not so obvious multiple integrators will work out of the box (I've tried it and they do, but I think it's worth having a test).

barker29 (commented):
I have a bunch of comments mostly about documentation and naming, but overall I think this is going to be a great simplification for the CEED integration.

fem/libceed/convection.cpp
YohannDudouit: I went for `ConvectionOperatorInfo`.

fem/libceed/coefficient.hpp
YohannDudouit: I improved the documentation, tell me if you like it.

fem/libceed/coefficient.hpp
YohannDudouit: I improved the documentation, tell me if you like it.

fem/libceed/diffusion.cpp
YohannDudouit: Renamed to `DiffusionOperatorInfo`.

fem/libceed/mass.cpp
YohannDudouit: Renamed to `MassOperatorInfo`.

fem/libceed/nlconvection.hpp
YohannDudouit: done.

fem/libceed/nlconvection.hpp
YohannDudouit: done.

fem/libceed/util.hpp
YohannDudouit: I improved the documentation, tell me if you like it.

fem/libceed/util.hpp
YohannDudouit: I improved the documentation, tell me if you like it.

fem/libceed/nlconvection.cpp
YohannDudouit: Renamed to `NLConvectionOperatorInfo`.

fem/libceed/mf_integrator.hpp
YohannDudouit: done.

fem/libceed/pa_integrator.hpp
YohannDudouit: done.

fem/libceed/mf_integrator.hpp
YohannDudouit: I added a short documentation, and `PAIntegrator` no longer inherit from `MFIntegrator`.

tests/unit/ceed/test_ceed.cpp
YohannDudouit: I added a Mass+Diffusion test.

fem/libceed/pa_integrator.hpp
YohannDudouit: I think I could (should?) get rid of this method and the `PAOperator` struct (and `MFOperator` struct) and change the `Assemble` method's signature.

YohannDudouit: I went ahead and removed `InitPA` and `InitMF` and fused them with `Assemble`, it's indeed more clean and clear. Thank you for the suggestion.

v-dobrev:
This PR has grown quite a bit -- it changes a lot of files (45). The scope has probably also grown far beyond what the title says. Should we consider breaking it into smaller pieces?

YohannDudouit:
@v-dobrev This is correct that the title is not relevant any more, different branches have been merged and already reviewed, because this PR is a refactor of the libCEED integration. I think it would only be disruptive to restart the review process since I believe we're seeing the end of it.

fem/libceed/pa_integrator.hpp
YohannDudouit: In order to make the documentation more clear I introduced an `OperatorInfo` struct from which the template parameter `CeedInfo` is supposed to inherit (see `mass.cpp` for instance), I think the design is even more clean and clear now.

fem/libceed/mass.cpp
```diff
@@ -11,52 +11,66 @@
...
+                     ":f_apply_mass_mf_const",
+                     ":f_apply_mass_mf_quad",
+                     &f_build_mass_const,
+                     &f_build_mass_quad,
+                     &f_apply_mass,
+                     &f_apply_mass_mf_const,
+                     &f_apply_mass_mf_quad,
+                     EvalMode::Interp,
+                     EvalMode::Interp,
+                     1}
```
pazner, line 43: My only concern with this is that it seems very easy to mix up the correct order of the parameters.

pazner (commented):
The new refactor is a big improvement! I think the libCEED integration is looking really good.

One very minor comment: perhaps we should rename the `libceed` subdirectory to `ceed` in order to match the name of the namespace.

tzanio:
> One very minor comment: perhaps we should rename the `libceed` subdirectory to `ceed` in order to match the name of the namespace.

I will also prefer `ceed`

fem/libceed/mass.cpp
YohannDudouit: An alternative is to write the following (I think it should work):
```c++
{
   header = "/mass_qf.h";
   build_func_const = ":f_build_mass_const";
   build_func_quad = ":f_build_mass_quad";
   apply_func = ":f_apply_mass";
   apply_func_mf_const = ":f_apply_mass_mf_const";
   apply_func_mf_quad = ":f_apply_mass_mf_quad";
   build_qf_const = &f_build_mass_const;
   build_qf_quad = &f_build_mass_quad;
   apply_qf = &f_apply_mass;
   apply_qf_mf_const = &f_apply_mass_mf_const;
   apply_qf_mf_quad = &f_apply_mass_mf_quad;
   trial_op = EvalMode::Interp;
   test_op = EvalMode::Interp;
   qdatasize = 1;
}
```

YohannDudouit: I can also reorganize to pair together the corresponding `const char*` and `CeedQFunctionUser`, e.g. `":f_build_mass_const"` and `&f_build_mass_const`.

pazner: I think the more explicit one is clearer, personally.

pazner (approved):
Looks great!

barker29 (approved):
Since a few changes have made, I'll go ahead and re-approve in case there's any doubt. This PR is ready.

fem/nonlinearform.cpp
```diff
@@ -24,7 +24,7 @@ void NonlinearForm::SetAssemblyLevel(AssemblyLevel assembly_level)
    switch (assembly)
    {
       case AssemblyLevel::NONE:
-         // This is the default behavior.
+         ext = new MFNonlinearFormExtension(this);
```
v-dobrev, line 27: Currently, in the default constructor `AssemblyLevel::NONE` is used as the default. We need to change that to something different since now we want `AssemblyLevel::NONE` to mean the new "matrix-free" (`MF` is not the best prefix for nonlinear operator since there is no "matrix" version anyway) assembly level.

We can either introduce `AssemblyLevel::LEGACYNONE` or rename `AssemblyLevel::LEGACYFULL` to `AssemblyLevel::LEGACY` and use that to denote the "legacy" assembly level for both `BilinearForm` and `NonlinearForm`.

What do you think?

fem/nonlinearform_ext.hpp
```diff
@@ -25,7 +25,7 @@ class NonlinearFormExtension : public Operator
    NonlinearForm *n; ///< Not owned
 public:
    NonlinearFormExtension(NonlinearForm *form);
-   virtual void AssemblePA() = 0;
+   virtual void Setup() = 0;
```
v-dobrev, line 28: In #1731, this method is renamed to `Assemble`. We need to unify these.

fem/nonlinearform_ext.hpp
```diff
@@ -37,8 +37,22 @@ class PANonlinearFormExtension : public NonlinearFormExtension
    const Operator *elem_restrict_lex; // Not owned
 public:
    PANonlinearFormExtension(NonlinearForm*);
-   void AssemblePA();
+   void Setup();
    void Mult(const Vector &x, Vector &y) const;
 };
+
+/// Data and methods for unassembled nonlinear forms
+class MFNonlinearFormExtension : public NonlinearFormExtension
```
v-dobrev, line 45: Since all nonlinear form representations are "matrix-free", it maybe best to replace `MF` with a different prefix. One option is `UA` (for UnAssembled) -- I've used that abbreviation before, e.g. here: https://ceed.exascaleproject.org/bps/#terminology-and-notation.

What do you think?

fem/nonlinearform.cpp
YohannDudouit: I was proposing both solutions to @tzanio , I'm fine with either, with a slight preference for `AssemblyLevel::LEGACY`.

fem/nonlinearform_ext.hpp
YohannDudouit: I will rename this method to `Assemble` too then.

fem/nonlinearform_ext.hpp
YohannDudouit: I think it could be confusing to introduce another acronym, but I don't have a strong opinion.

fem/nonlinearform.cpp
tzanio: `AssemblyLevel::LEGACY` is ok with me

fem/nonlinearform_ext.hpp
v-dobrev: Any other thoughts? @pazner, @barker29, @tzanio?

I see that @YohannDudouit pushed the change `MF` -> `UA` but I'm not sure myself if I like it, so I want to see what everyone thinks.

fem/nonlinearform.cpp
v-dobrev: For backward compatibility, I think it maybe good to keep the constant `LEGACYFULL = 0` in addition to `LEGACY = 0`.

fem/nonlinearform_ext.hpp
pazner: I am probably missing some context here. Would it not be confusing to use `MF` for linear forms and `UA` for nonlinear forms, where they both represent essentially the same "assembly level"?

tzanio: I do not like `UA` (it is not standard and make me think of United Airlines).

I think it is fine to use `MF`. In MFEM we have already said that we are using this for action without storage.

fem/ceed/nlconvection_qf.h
```diff
@@ -0,0 +1,481 @@
...
+CEED_QFUNCTION(f_build_conv_const)(void *ctx, CeedInt Q,
+                                   const CeedScalar *const *in,
+                                   CeedScalar *const *out)
+{
+   NLConvectionContext *bc = (NLConvectionContext*)ctx;
+   // in[0] is Jacobians with shape [dim, nc=dim, Q]
+   // in[1] is quadrature weights, size (Q)
+   //
+   // At every quadrature point, compute qw/det(J).adj(J).adj(J)^T and store
+   // the symmetric part of the result.
```
v-dobrev, line 26: This comment and other comments below seem to be copy-pasted from another place and not updated -- they are incorrect for this integrator I think.

fem/ceed/convection_qf.h
```diff
@@ -0,0 +1,531 @@
...
+CEED_QFUNCTION(f_build_conv_const)(void *ctx, CeedInt Q,
+                                   const CeedScalar *const *in,
+                                   CeedScalar *const *out)
+{
+   ConvectionContext *bc = (ConvectionContext*)ctx;
+   // in[0] is Jacobians with shape [dim, nc=dim, Q]
+   // in[1] is quadrature weights, size (Q)
+   //
+   // At every quadrature point, compute qw/det(J).adj(J).adj(J)^T and store
+   // the symmetric part of the result.
```
v-dobrev, line 30: Another copy-pasted comment that is incorrect I think. Check other comments below too.

v-dobrev:

v-dobrev:
Re-merged in `next`.

v-dobrev:
LGTM

PR #1781, opened 2020-09-24
Matrix-free auxiliary Maxwell solver [barker29/matrix-free-ams]

barker29:
This implements a matrix-free (or at least "matrix-light") auxiliary space Maxwell solver for high-order Nedelec elements. It uses an (assembled) AMG V-cycle for low-order refined approximations as a replacement for the auxiliary space solves in the auxiliary space framework.

The ideas and algorithm are described here: https://doi.org/10.1002/nla.2348

Run
```
./ex3p -pa -o 8 --mesh ../data/star.mesh
```
to see the new solver in action.

| [#1781](https://github.com/mfem/mfem/pull/1781) | @barker29 | @tzanio | @tzanio + @dylan-copeland + @psocratis | 01/22/21 | 03/07/21 | 03/11/21 | |


linalg/auxiliary.cpp
```diff
@@ -0,0 +1,458 @@
...
+   smoother_.Mult(residual, temp);
+   y += temp;
+   chrono.Stop();
+   smooth_time_ += chrono.RealTime();
+}
+
+// Pi-space constructor
+MatrixFreeAuxiliarySpace::MatrixFreeAuxiliarySpace(
+   mfem::ParMesh& mesh_lor,
+   mfem::Coefficient* alpha_coeff,
```
dylan-copeland, line 169: It may be convenient to treat `NULL` coefficients as 1.

psocratis:
> ./ex3p -pa -o 8 --mesh ../data/star.mesh

This crashes when I tried the branch out of the box. I think `HCURL_MAX_D1D` in `biliniteg.hpp` needs to be increased.

dylan-copeland:
> > ./ex3p -pa -o 8 --mesh ../data/star.mesh
> 
> This crashes when I tried the branch out of the box. I think `HCURL_MAX_D1D` in `biliniteg.hpp` needs to be increased.

I think `HCURL_MAX_D1D` should not be too high, because otherwise MFEM will take longer to build. This can be changed for a user who wants to run higher order, but should not be in the master branch.

psocratis:
> I think `HCURL_MAX_D1D` should not be too high, because otherwise MFEM will take longer to build. This can be changed for a user who wants to run higher order, but should not be in the master branch.

Sure, I agree with this. At least in debug mode can we add an abort message? As it is now it just segfaults

barker29:
@psocratis Are you suggesting we change that in this PR? In my view it is a separate issue. Are you saying that since this command line runs (but does not converge) in master it should also work in this branch?

linalg/auxiliary.cpp
```diff
@@ -0,0 +1,542 @@
...
+   bool useAmgX_,
+#endif
+   int cg_iterations)
+   :
+   Solver(curlcurl_oper.Height()),
+   comm(mesh_lor.GetComm()),
+   matfree(NULL),
+   cg(NULL),
+#ifdef MFEM_USE_AMGX
+   useAmgX_(useAmgX),
```
dylan-copeland, line 250: It looks like the underscores got mixed up.

psocratis:
> @psocratis Are you suggesting we change that in this PR? In my view it is a separate issue. Are you saying that since this command line runs (but does not converge) in master it should also work in this branch?

It is indeed a separate issue, it does not run on master either.  However, since this new matrix-free solver is optimal for high orders, then an example run demonstrating this could be added, and therefore (unavoidably) the MAX_D1D numbers have to be increased. Just so I know ... how much impact in compilation time are we talking about?

dylan-copeland:
@psocratis I'm not sure how much the build time will be affected. We should just experiment and find out. Another issue is memory. If I remember correctly, the original H(curl) PA kernels ran out of memory on the GPU quickly (order 4 or 5) at run time, and then I wrote the shared memory kernels which can go to higher orders, but I think the shared memory kernels can run out of memory at build time if the order is too high. So there are multiple challenges to high order, which may require changes to the kernels other than just increasing the max order allowed. We can experiment with this and see if the order can be increased (probably not much if at all), but for this PR I would be satisfied with order 5 or so, considering the PA kernel issues.

linalg/auxiliary.cpp
barker29: Good catch, fixed.

dylan-copeland:
I thought there may be a broken case, from PR #1761 , depending on `mpi_gpu_aware` in `DeviceConformingProlongationOperator::Mult`, but it seems to work in all cases. If you set `Device::SetGPUAwareMPI(true);` in `ex3p` and then run with `lrun -M "-gpu", it works fine. I also tested with and without AMGx, and everything is working.

linalg/auxiliary.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+   ~MatrixFreeAuxiliarySpace();
+
+   void Mult(const Vector& x, Vector& y) const;
+
+   void SetOperator(const Operator& op) {}
+
+private:
+   void SetupAMG(int
+                 system_dimension);  // TODO: document meaning of system_dimension
```
dylan-copeland, line 100: TODO?

dylan-copeland (approved):
Thanks @barker29 for this major advance!

CHANGELOG
```diff
@@ -11,6 +11,9 @@
 Version 4.2.1 (development)
 ===========================
 
+- Added high-order matrix-free auxiliary Maxwell solver for H(curl) problems,
+  as described in Barker and Kolev 2021 (https://doi.org/10.1002/nla.2348).
```
psocratis, line 15: small thing: The publication date is in 2020 (not 2021)

examples/ex3p.cpp
```diff
@@ -159,9 +167,9 @@ int main(int argc, char *argv[])
    //    by marking all the boundary attributes from the mesh as essential
    //    (Dirichlet) and converting them to a list of true dofs.
    Array<int> ess_tdof_list;
+   Array<int> ess_bdr(pmesh->bdr_attributes.Max());
```
psocratis, line 170: I think this can cause an issue if a processor does not own any bdr element, i.e., the array on the processor is empty. So it's better to do 

````cpp
   Array<int> ess_tdof_list;
   Array<int> ess_bdr;
   if (pmesh->bdr_attributes.Size())
   {
      ess_bdr.SetSize(pmesh->bdr_attributes.Max());
      ess_bdr = 1;
 

````

File: examples/ex3p.cpp [hunk 6]
```diff
@@ -208,17 +216,20 @@ int main(int argc, char *argv[])
    // 13. Solve the system AX=B using PCG with the AMS preconditioner from hypre
    //     (in the full assembly case) or CG with Jacobi preconditioner (in the
```
psocratis, line 217: The comment for Jacobi needs to change here

linalg/auxiliary.cpp
```diff
@@ -0,0 +1,542 @@
...
+   pispacesolver(pispacesolver_),
+   gspacesolver(gspacesolver_),
+   smoother(smoother_),
+   ess_tdof_list(ess_tdof_list_),
+   residual_time(0.0),
+   smooth_time(0.0),
+   gspacesolver_time(0.0),
+   pispacesolver_time(0.0)
+{
+   // could assert a bunch of sizes...
```
psocratis, line 44: Is this comment (and some others of the same kind) intended to stay?

linalg/auxiliary.cpp
```diff
@@ -0,0 +1,542 @@
...
+   y += residual;
+
+   // smooth (don't need the residual if smoother_ has iterative_mode ?)
+   FormResidual(x, y, residual);
+   m_chrono.Clear();
+   m_chrono.Start();
+   smoother.Mult(residual, temp);
+   y += temp;
+   m_chrono.Stop();
+   smooth_time += m_chrono.RealTime();
```
psocratis, line 150: Are these timers supposed to stay here?

examples/ex3p.cpp
barker29: Nice catch, thanks.

linalg/auxiliary.cpp
```diff
@@ -0,0 +1,542 @@
...
+{
+   int order = nd_fespace.GetFE(0)->GetOrder();
+   ParMesh *mesh = nd_fespace.GetParMesh();
+   int dim = mesh->Dimension();
+
+   // smoother
+   const double scale = 0.25; // not so clear what exactly to put here...
+   Array<int> ess_tdof_list;
+   nd_fespace.GetEssentialTrueDofs(ess_bdr, ess_tdof_list);
+   smoother = new OperatorJacobiSmoother(aform, ess_tdof_list, scale);
```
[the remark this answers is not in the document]
barker29: I added an optional parameter - I can remove it or structure it differently if think the interface is getting too ugly here.

linalg/auxiliary.cpp
barker29: I have gone over the comments in this file, removing some and rewriting others to be more professional as hints to future maintainers. There could still probably be improvements, feel free to make more suggestions.

linalg/auxiliary.cpp
```diff
@@ -0,0 +1,513 @@
...
+   FormResidual(x, y, residual);
+   Vector gspacetemp(gradient.Width());
+   gradient.MultTranspose(residual, gspacetemp);
+   Vector gspacecorrection(gradient.Width());
+   gspacecorrection = 0.0;
+   gspacesolver.Mult(gspacetemp, gspacecorrection);
+   gradient.Mult(gspacecorrection, residual);
+   y += residual;
+
+   Vector temp(x.Size());
```
psocratis, line 92: Maybe move this just before line 116, where it's actually used?

psocratis (approved):
This looks really good. Results from some sample runs look awesome!

tzanio:
This looks great overall, but I'd like to take a closer look and will need a bit more time for the review. Sorry for the delay.

linalg/auxiliary.cpp
```diff
@@ -0,0 +1,512 @@
...
+   H1_FECollection * fec_lor = new H1_FECollection(1, mesh_lor.Dimension());
+   ParFiniteElementSpace fespace_lor_d(&mesh_lor, fec_lor, mesh_lor.Dimension(),
+                                       Ordering::byVDIM);
+
+   // build LOR AMG v-cycle
+   if (ess_bdr.Size())
+   {
+      fespace_lor_d.GetEssentialTrueDofs(ess_bdr, ess_tdof_list);
+   }
+   ParBilinearForm a_space(&fespace_lor_d);
```
dylan-copeland, line 146: A more descriptive name for `a_space` might be `a_lor`.

linalg/auxiliary.hpp
```diff
@@ -0,0 +1,233 @@
...
+   void SetupAMG(int system_dimension);
+   void SetupVCycle();
+
+   /// inner_cg_iterations > 99 applies an exact solve here
+   void SetupCG(Operator& curlcurl_oper, Operator& conn,
+                int inner_cg_iterations);
+
+   MPI_Comm comm;
+   Array<int> ess_tdof_list;
+   HypreParMatrix * aspacematrix;
```
dylan-copeland, line 112: A more descriptive name for `aspacematrix` might be `LORmatrix`.

linalg/auxiliary.cpp
```diff
@@ -0,0 +1,512 @@
...
+
+   aspacewrapper = cg;
+}
+
+void MatrixFreeAuxiliarySpace::SetupVCycle()
+{
+   aspacewrapper = aspacepc;
+}
+
+class ZeroWrap : public Solver
```
dylan-copeland, line 311: It would be more accurate to name this something like `ZeroWrapAMG`.

linalg/auxiliary.cpp
barker29: Done.

linalg/auxiliary.hpp
barker29: Done for `aspacematrix` and `aspacepc`, for `aspacewrapper` I am being picky but it sometimes wraps a CG solver for the high-order operator (which is not LOR) so I didn't change it. I am open to other names but `aspacewrapper` wraps the actual auxiliary space solver (approximation) so I think its name is appropriate.

linalg/auxiliary.cpp
barker29: Done.

tzanio:
LGTM

tzanio (approved):
Very cool thank you Andrew! 🚀 

Some small suggestions below

tzanio:

CHANGELOG
```diff
@@ -11,6 +11,9 @@
 Version 4.2.1 (development)
 ===========================
 
+- Added high-order matrix-free auxiliary Maxwell solver for H(curl) problems,
+  as described in Barker and Kolev 2020 (https://doi.org/10.1002/nla.2348).
+
```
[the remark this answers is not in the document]
tzanio: thank you

tzanio:

PR #1789, opened 2020-09-29
[WIP] Coefficient Factory class for mini apps [coef-fact-dev]

mlstowell:
This PR adds a class which reads a file (or any input stream) and allocates (and returns) `Coefficient`, `VectorCoefficient`, or `MatrixCoefficient` classes described in that file.  The current implementation recognizes most of the basic coefficient types appearing in `coefficients.hpp` and is extendable.  There is a small mini app which demonstrates its usage in `miniapps/tools/coef-fact.cpp`.

The goal of this object is to provide a way to allow an application user to easily specify, in an input file, various coefficients for boundary conditions, initial conditions, material coefficients, etc.. This scheme is not as flexible as converting user input into compilable code but I think it could be useful and much easier to implement.

At this point I'm mainly interested in feedback.  The file format attempts to closely follow the argument lists needed by the various coefficient constructors.  Examining a header file entry for a given coefficient should give you most of the information you need to figure out what data is needed in the file but clearly more documentation will be needed before making this public (unless we choose a different path).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -199,9 +199,12 @@ miniapps/performance/sol.*
 
 miniapps/tools/display-basis
 miniapps/tools/load-dc
+miniapps/tools/coef-fact
 miniapps/tools/convert-dc
 miniapps/tools/lor-transfer
 miniapps/tools/get-values
+
+miniapps/tools/coef-fact.inp
 
 miniapps/toys/automata
 miniapps/toys/life
--- a/miniapps/common/fem_extras.cpp
+++ b/miniapps/common/fem_extras.cpp
@@ -10,6 +10,7 @@
 // CONTRIBUTING.md for details.
 
 #include "fem_extras.hpp"
+#include "../../general/text.hpp"
 
 using namespace std;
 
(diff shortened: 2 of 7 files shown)


goxberry:
Thanks for writing this PR @mlstowell! I like the idea and the conversation you've started. One cool thing about this idea is the possibility of enabling people who don't write much C++ (say they use ANSYS a lot) to tweak inputs to mfem-linked executables as part of running parameter studies or as part of doing design work. From the perspective of someone who uses mfem to do design work, this capability would be a big force multiplier for us, because it would be possible for our customers to do those sorts of design studies with minimal intervention from the design optimization team. It would free us up to do more feature development.

As currently executed, I agree that it would be possible, but difficult to use this file format for that purpose. As you've alluded to in the PR description, the numbers in the input file format look kind of magic to me, and I have to read the corresponding mfem classes and the source code to figure out what's going on. My customers wouldn't have that option, and I don't know that they'd look at public headers to understand what the inputs in the current input file format mean.

I think using a more semantically rich file format -- one that has tags and doesn't rely so much on positional inputs -- would make it easier for non-developers to understand what's going on, and for developers and non-developers alike to spot questionable or syntactically-invalid inputs.

For example, I could write an input file with the snippet:
```console
##### snip #####
scalar_coef
DeltaCoefficient
2 3 4 5 6
##### snip #####
```
and not really know that the file parser is going to silently ignore the 6 in that input. Something that tagged those inputs with argument names would be really helpful. For example:
```yaml
scalar-coefficients:
    - delta-coefficient:
          - name: "my-point-load" # maybe this name could be used to pass the coefficient in as a particular argument
          - dimension: 2
          - location: [3, 4, 5]    # oops! the dimension could also possibly be inferred from the location array
          - amplitude: 6
```
I don't think the input needs to be YAML-like -- I think lots of options are available that could add useful semantic information, and this semantic information makes it much easier for a non-developer to see what's going on. 

I like the factory logic, too. I think refactoring `CoefFactory` into three abstract factories, one for each of the distinct coefficient class hierarchies (i.e., `Coefficient`, `VectorCoefficient`, `MatrixCoefficient`), and then having those abstract factories delegate to concrete factories would make the resulting design easier to test, debug, and extend.

re: the delegation logic inside your current factories, if you refactored them to the pattern above, I think the delegation logic inside the abstract factories could be simplified to creating a hash table keyed by the `std::string name` associated with concrete factories. (Similar sorts of constructions, like a factory function mapping keys to concrete factories, would also work.) The factory-specific logic would be in separate classes that could be tested independently. Delegating creation to concrete factories with hash tables would make it easier to change the keys corresponding to each factory if the input file format changes.

I'm curious to see how this effort evolves and hope something like it lands in MFEM eventually. Thanks again!

PR #1046, opened 2019-08-27
Removing need for ReorientTetMesh [no-tet-reorder-dev]

mlstowell:
This PR adds a `DofTransformation` class which performs a linear mapping of degrees of freedom on triangular faces to enable Nedelec basis functions to align properly regardless of the relative orientation of the elements.

With the addition of `DofTransformation`, the methods `[Par]Mesh::ReorientTetMesh` are no longer needed.  This PR currently comments out these methods rather than removing them.  Unless someone makes an argument for keeping them in the library I think we should remove these methods before merging this PR.

The `test_eigs.cpp` and `test_eigs_nd.cpp` programs were used to validate the new DoF ordering scheme for all triangle face orientations at a variety of basis function orders.  These programs don't need to be part of this PR but they may be useful as part of the test suite.  Edit: a version of `test_eigs.cpp` has been added as a unit test in `tests/unit/fem/test_eigs.cpp`.

The `ex_nd_amr_p.cpp` program can be used to test the new code in an AMR context.  This code implements an exact solution which is only valid on square or cube-shaped domains.  It can be run on any 2D or 3D mesh but the solution errors will only be correct on squares or cubes.

This PR has the potential to interfere with any code that uses high order Nedelec basis functions on tetrahedral meshes.  Unfortunately, we don't have that many relevant test cases in our examples/miniapps/unit tests.  I've invested a lot of time testing this branch with the unit tests, autotest, and custom examples but I know there are pieces of code that have not been covered by these tests.  I would really appreciate it if the reviewers could test this code further if they know of any suitable test cases.

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge| 
 | --- | --- | --- | --- | --- | --- | --- | 
| [#1046](https://github.com/mfem/mfem/pull/1046) | @mlstowell | @tzanio | @jakubcerveny + @psocratis | 02/09/21 | 08/01/21 | 08/20/21 | |

PR #975, opened 2019-06-24
Mesh reader support for VTK high order elements [feature/vtk_high_order]

jonesholger:
# Updated PR information (Dec. 2020 – Will)

This PR adds two new features to MFEM's VTK support:

1. Reading XML VTK mesh files in addition to the "legacy" ASCII format
2. Reading high-order VTK meshes. High-order support works for **both** XML and legacy VTK formats.

XML parsing is done using the TinyXML2 library.

To facilitate systematic conversion between VTK's high-order node ordering and MFEM's node ordering, mappings are created that use an intermediate lexicographic ordering. Some remarks about the lexicographic ordering:

* All NodalFiniteElement subclasses now support GetLexicographicOrdering, which plays the same role as TensorBasisElement::GetDofMap, but with support for all non-tensor geometries
* The "refined geometry vertices" for tetrahedral elements are now lexicographically ordered in the standard reference element rather than in an "auxiliary element"
* As a consequence, the lexicographic DOF ordering given by GetLexicographicOrdering always agrees with the refined geometry vertex ordering.
* These orderings are tested for all geometries, for orders 1 through 6 in the unit tests.
* The VTK mesh reader supports arbitrary-order Lagrange elements. The node orderings are handled using the same mappings defined for high-order ParaView output, so pre-generated data tables are not required.

As part of this PR, the handling of the VTK format for input and output are both made more systematic. The files `vtk.hpp` and `vtk.cpp` now have a lot more features that should make dealing with VTK more convenient, for instance converting from a VTK element type to MFEM `Geometry::Type` can now be performed using `VTKGeometry:: GetMFEMGeometry` instead of manual switch statements. This simplifies the code in `Mesh::PrintVTK` and `Mesh::PrintVTU`.

For testing purposes, VTK meshes of Cartesian grids with low-order and high-order elements of all types can be created in ParaView with the source "Unstructured Cell Types".

# Previous information (out of date)

> This mesh reader update subtask for vtk is related to MFEM issue https://github.com/mfem/mfem/issues/678  & https://blog.kitware.com/modeling-arbitrary-order-lagrange-finite-elements-in-the-visualization-toolkit/
> 
> Quads to arbitrary order vtk ascii
> Hexes to arbitrary order vtk ascii
> Triangles to order 6 (vtk limit) vtk ascii
> Tets to order 6 (vtk limit) vtk ascii
> Wedge (or prism) to order 6 (vtk limit) vtk ascii
[...]

PR #1817, opened 2020-10-16
Variable order finite element collections and spaces [var-order-space-dev]

jakubcerveny:
I'm separating the basis of `p-refinement-dev` (#1423) into its own pull request, so it can be merged in more manageable chunks.

The `FiniteElementCollection` base class now contains a mechanism to turn any collection into one that supports arbitrary orders. A collection is constructed as usual with a "default" order. However, if one of the new generalized methods is called (`GetFE`, `GetNumDof`, `GetDofOrdering`) with an order different than the default, a new collection of that order is constructed internally
and the call is rerouted to it. This is taken care of in the base class and all that a derived collection needs to do is implement the protected virtual method `CloneOrder` to construct a new instance of itself with a different order. Note that it is not mandatory to override this method: only the H1 and L2 collections currently support variable orders.                            

The `FiniteElementSpace` also works as usual, with all elements assumed to be of the default order of its FE collection, until the new method SetElementOrder is called. At that moment the space starts tracking element orders in the `elem_order` array and becomes a variable order space. There is a lot of internal changes: edges and faces in a variable order space may hold more than one set of DOFs (called DOF 'variants', held in the members `var_edge_dofs` and `var_face_dofs`) and there 
are new constraints in the P matrix construction algorithm to handle hp interpolation. But on the outside the space looks the same, it all should 'just work'.

The hp interpolation should work in both 2D and 3D on arbitrarily refined NC meshes, in serial. The parallel space is not handled yet in this PR. Also serialization of variable order spaces is not implemented yet. In the demo example we thus need to interpolate the solutions to the maximum order to visualize them in GLVis. 

Will Pazner contributed variable order support in `Mesh::CreateRefinedMesh`: one can now low-order refine non-uniformly. I'm not sure how serious are the FIXMEs in the ParMesh, but I assume we'll sort these out in the upcoming parallel version.

Example of running the demo: `./hpdemo -m ../data/fichera-mixed.mesh -o 2 -u 1 -r 3`

mixed-hp-2
mixed-hp-1

TODO: 
- [x] Checks that orders changed
- [x] Rename ConformingRestrictionInterpolation?
- [x] Fix 3D hpdemo
- [x] Fix mixed mesh support
- [x] Add variable-order space unit tests
- [x] Update CHANGELOG

| [#1817](https://github.com/mfem/mfem/pull/1817) | @jakubcerveny | @tzanio | @v-dobrev + @pazner + @jamiebramwell | 12/07/20 | 02/17/21 | 03/11/21 | |


jakubcerveny:
@tzanio, I'm trying to finalize this variable-order spaces branch, but I have a problem: 

For RT/ND spaces, there seems to be an inconsistence between the 'order' the FE collection is constructed with and the order of the actual finite elements contained in it, which pretty much sabotages the handling of variable orders in my new code.

Example: I create an order zero `RT_Trace_FECollection`, but the contained `FiniteElement::GetOrder` reports order 1. 

The problem is that my variable order `FiniteElementSpace` thinks it contains order zero elements, so `GetElementOrder` returns 0, which screws up code like HypreAMS and `ex8p` and probably other things.

How do we fix this?

tzanio:
This is a special case just for RT. 😁 

Unfortunately in the literature the lowest (1st) order Raviart-Thomas space is denoted with RT_0, not RT_1 as one may expect. That is not the case for Nedelec elements, where the lowest order space is ND_1 and there is no ND_0.

We chose to conform to the literature in the code. Can't we handle RT as a special case?

v-dobrev:
We can simply add a new method that returns the number that is given to the constructor, e.g. something like `GetNotationOrder` or `GetLogicalOrder`.

jakubcerveny:
This should be finished. I'm still thinking of adding solve tests in `test_var_order.cpp` but that's a minor thing. I think this is ready for review.

tzanio:
More details on the hp-refinement algorithm (internal LLNL link):

https://myconfluence.llnl.gov/pages/viewpage.action?spaceKey=MFEM&title=hp-refinement+in+MFEM

tzanio:
@v-dobrev, @pazner and @jamiebramwell -- Reminder that the reviews for this PR are due this week.

fem/fe.cpp
```diff
@@ -40,22 +40,19 @@ FiniteElement::FiniteElement(int D, Geometry::Type G, int Do, int O, int F)
 void FiniteElement::CalcVShape (
    const IntegrationPoint &ip, DenseMatrix &shape) const
 {
-   mfem_error ("FiniteElement::CalcVShape (ip, ...)\n"
-               "   is not implemented for this class!");
+   MFEM_ABORT("method is not implemented for this class");
```
pazner, line 43: Is there a reason for these cases (and below)?

fem/convergence.cpp
```diff
@@ -29,7 +29,7 @@ double ConvergenceStudy::GetNorm(GridFunction *gf, Coefficient *scalar_u,
 {
    bool norm_set = false;
    double norm=0.0;
-   int order = gf->FESpace()->GetOrder(0);
+   int order = gf->FESpace()->GetElementDegree(0);
```
pazner, line 32:
```suggestion
   int order = gf->FESpace()->GetMaxElementOrder();
```
Would this be better?

fem/fespace.hpp
```diff
@@ -169,31 +187,80 @@ class FiniteElementSpace
        boundary. */
    void BuildNURBSFaceToDofTable() const;
 
+   typedef std::uint64_t VarOrderBits;
+   enum { MaxVarOrder = 8*sizeof(VarOrderBits) - 1 };
```
pazner, line 191: Rather than an `enum` maybe this could be a `static constexpr`?

fem/fespace.hpp
```diff
@@ -320,12 +390,36 @@ class FiniteElementSpace
...
 
+   /// Sets the order of the i'th finite element.
+   /** By default, all elements are assumed to be of fec->GetOrder(). Once
+       SetElementOrder is called, the space becomes a variable order space. */
+   void SetElementOrder(int i, int p);
+
+   /// Returns the order of the i'th finite element.
+   /** @note This is the 'order' of the FiniteElementCollection. The actual
+       polynomial degree may differ, see @a GetElementDegree(). */
+   int GetElementOrder(int i) const;
```
pazner, line 401: Is this only the case for RT spaces? If so, that should be spelled out in this comment. Otherwise I find it a bit confusing.

fem/fespace.hpp
```diff
@@ -388,10 +486,20 @@ class FiniteElementSpace
...
-   /// Returns the order of the i'th finite element
-   int GetOrder(int i) const;
-   /// Returns the order of the i'th face finite element
-   int GetFaceOrder(int i) const;
+   /// Returns the polynomial degree of the i'th element.
+   /** This is normally the same as GetElementOrder(i), except for RT finite
+       element spaces, where the actual degree of the RTp collection is (p+1).*/
+   int GetElementDegree(int i) const { return GetFE(i)->GetOrder(); }
+
+   /// Returns the polynomial degree of the i'th finite element (deprecated).
```
pazner, line 494:
```suggestion
   /// Returns the polynomial degree of the i'th finite element (deprecated).
   /// Users should use @a GetElementDegree instead.
```

fem/pfespace.hpp
```diff
@@ -262,7 +262,7 @@ class ParFiniteElementSpace : public FiniteElementSpace
 
    /** Returns the indexes of the degrees of freedom for i'th face
        including the dofs for the edges and the vertices of the face. */
-   virtual void GetFaceDofs(int i, Array<int> &dofs) const;
+   virtual int GetFaceDofs(int i, Array<int> &dofs, int variant = 0) const;
```
pazner, line 265: Document return value here?

mesh/mesh.cpp
```diff
@@ -3738,28 +3738,58 @@ Mesh::Mesh(Mesh *mesh_array[], int num_pieces)
...
 {
    Dim = orig_mesh->Dimension();
-   MFEM_VERIFY(ref_factor >= 1, "the refinement factor must be >= 1");
-   MFEM_VERIFY(ref_type == BasisType::ClosedUniform ||
-               ref_type == BasisType::GaussLobatto, "invalid refinement type");
-   MFEM_VERIFY(Dim == 1 || Dim == 2 || Dim == 3,
-               "only implemented for Segment, Quadrilateral and Hexahedron "
-               "elements in 1D/2D/3D");
-   MFEM_VERIFY(orig_mesh->GetNumGeometries(Dim) <= 1,
-               "meshes with mixed elements are not supported");
```
pazner, line 3750: Some of these checks should probably be brought back (I think I was the one who deleted them in the first place...)

mesh/mesh.hpp
```diff
@@ -706,6 +710,9 @@ class Mesh
        @note The constructed Mesh is linear, i.e. it does not have nodes. */
    Mesh(Mesh *orig_mesh, int ref_factor, int ref_type);
 
+   /// A version of the above constructor for non-uniform refinement.
+   Mesh(Mesh *orig_mesh, Array<int> &ref_factors, int ref_type);
```
pazner, line 714:
```suggestion
   /// A version of the above constructor for non-uniform refinement.
   /// The input array @a ref_factors contains one refinement factor per element
   /// of the input mesh.
   Mesh(Mesh *orig_mesh, const Array<int> &ref_factors, int ref_type);
```
```suggestion
   /// A version of the above constructor for non-uniform refinement.
   Mesh(Mesh *orig_mesh, Array<int> &ref_factors, int ref_type);
```

mesh/mesh.hpp
```diff
@@ -830,21 +837,31 @@ class Mesh
...
 
+   // deprecated: "base geometry" no longer means anything
+   Geometry::Type GetFaceBaseGeometry(int i) const
+   { return GetFaceGeometry(i); }
+
+   Geometry::Type GetElementBaseGeometry(int i) const
+   { return GetElementGeometry(i); }
+
+   Geometry::Type GetBdrElementBaseGeometry(int i) const
+   { return GetBdrElementGeometry(i); }
```
pazner, line 863: Should these be marked deprecated? I am guessing you didn't so as to not generate tons of warnings.

I think this is a good change since I never knew what the distinction between "base geometry" and "geometry" was.

fem/fe.cpp
jakubcerveny: I'm not sure I understand. The point of these changes is to use the macro MFEM_ABORT, which should be standard. The "old" way with `mfem_error` is pretty weird: you need to manually enter the function name, you cannot use `<<` to format the message, etc. It's true these changes are unrelated to the PR, but I tend to fix these things when stumble upon them.

fem/convergence.cpp
jakubcerveny: This is a good idea. I'm just not sure I want to risk breaking something since "order" and "degree" may be different.

fem/fespace.hpp
jakubcerveny: OK

fem/fespace.hpp
jakubcerveny: Done

fem/fespace.hpp
jakubcerveny: Done

fem/pfespace.hpp
jakubcerveny: Done.

mesh/mesh.cpp
jakubcerveny: OK, which ones do you want brought back? (I don't know this part of the code so I'm really not sure)

mesh/mesh.hpp
jakubcerveny: There used to be special refined Elements (e.g., `BisectedElement`) and the base geometry was the original geometry. This is long gone. I would like to use MFEM_DEPRECATED and change all occurrences to the simpler names, but it's a lot of changes that I felt would better be done in a separate PR.

mesh/mesh.hpp
jakubcerveny: Done, also made the array `const &`

tzanio:
Thank you for the review @pazner.

@jamiebramwell and @v-dobrev -- can you please take a look?

fem/fe.cpp
pazner: Yes, you're right — I hadn't noticed that difference between `mfem_error` and `MFEM_ABORT`.

mesh/mesh.cpp
pazner: For now, all of them (except maybe the `Dim` test, which seems unnecessary)

mesh/mesh.hpp
pazner: That sounds good to me 👍

fem/convergence.cpp
pazner: It looks like the order is just used to determine the quadrature rule, so this should be OK, right?

fem/fespace.hpp
```diff
@@ -388,10 +486,21 @@ class FiniteElementSpace
...
+       element spaces, where the actual degree of the RTp collection is (p+1).*/
+   int GetElementDegree(int i) const { return GetFE(i)->GetOrder(); }
+
+   /// Returns the polynomial degree of the i'th finite element.
+   /** Deprecated: please use @a GetElementDegree instead. */
+   MFEM_DEPRECATED int GetOrder(int i) const { return GetElementDegree(i); }
+
+   /** Returns the order of the specified edge. In a variable order space, order
+       of the given variant is returned, or -1 if there are no more variants. */
+   int GetEdgeOrder(int edge, int variant = 0) const;
```
jamiebramwell, line 500: Some additional comments on what `variant` means would be appreciated.

fem/fespace.cpp
```diff
@@ -1632,167 +1879,404 @@ void FiniteElementSpace::Construct()
...
-      dim = mesh->Dimension();
-      nv = fec->DofForGeometry(Geometry::POINT);
-      ne = (dim > 1) ? ( fec->DofForGeometry(Geometry::SEGMENT) ) : ( 0 );
-      nb = (dim > 0) ? fec->DofForGeometry(mesh->GetElementBaseGeometry(i)) : 0;
-      if (nv > 0)
+      mesh->GetElementEdges(i, E, ori);
+      for (int j = 0; j < E.Size(); j++)
       {
-         mesh->GetElementVertices(i, V);
+         edge_orders[E[j]] |= mask;
```
jamiebramwell, line 2025: Some comments describing the purpose of this bitmask operation would be appreciated.

fem/fespace.hpp
```diff
@@ -169,31 +187,80 @@ class FiniteElementSpace
        boundary. */
    void BuildNURBSFaceToDofTable() const;
 
+   typedef std::uint64_t VarOrderBits;
+   static constexpr int MaxVarOrder = 8*sizeof(VarOrderBits) - 1;
```
jamiebramwell, line 191: A description of how this data type is used would be nice.

jamiebramwell (commented):
I still need to pull this and try it out a bit, but this is a really great addition to MFEM. I look forward to trying it out! My comments are mostly about documentation (particularly with regards to the new bitwise data structure).

fem/convergence.cpp
jakubcerveny: OK, this case is probably safe. Done.

mesh/mesh.cpp
jakubcerveny: Done.

fem/fespace.hpp
jakubcerveny: Added.

fem/fespace.cpp
jakubcerveny: Added.

fem/fespace.hpp
jakubcerveny: Added.

jamiebramwell (approved):
I'm super excited to use this!

fem/fe_coll.hpp
```diff
@@ -153,14 +153,54 @@ class FiniteElementCollection
...
+   int GetOrder() const { return base_p; }
+
+protected:
+   const int base_p;
+
+   FiniteElementCollection() : base_p(0) {}
+   FiniteElementCollection(int p) : base_p(p) {}
+
+   /// Instantiate a new collection of the same type with a different order.
+   virtual FiniteElementCollection* CloneOrder(int p) const;
```
pazner, line 191: Maybe `CloneWithOrder` would be clearer?

fem/fe_coll.hpp
```diff
@@ -177,15 +217,23 @@ class H1_FECollection : public FiniteElementCollection
...
    virtual int GetContType() const { return CONTINUOUS; }
+   int GetBasisType() const { return b_type; }
+
    FiniteElementCollection *GetTraceCollection() const;
 
-   int GetBasisType() const { return b_type; }
    /// Get the Cartesian to local H1 dof map
    const int *GetDofMap(Geometry::Type GeomType) const;
+   /// Variable order version of GetDofMap
+   const int *GetDofMap(Geometry::Type GeomType, int p) const;
```
pazner, line 230: The way `GetDofMap` is implemented will change once #975 is merged (which should be very soon). I believe that won't affect this PR, but would be good to double-check.

pazner (approved):
I did another pass, and this PR is looking really good to me!

I've used this functionality extensively (when it was still in the `p-refinement-dev` branch), and it worked basically flawlessly and was super useful for my purposes.

This is a big effort and will be a wonderful feature to have in MFEM. Thanks @jakubcerveny!

fem/fe_coll.hpp
jakubcerveny: I'd prefer the simpler name...

YohannDudouit: I also prefer the more self documented name `CloneWithOrder`

fem/fespace.hpp
```diff
@@ -473,28 +586,37 @@ class FiniteElementSpace
...
+   virtual int GetFaceDofs(int face, Array<int> &dofs, int variant = 0) const;
+
+   /** @brief Returns the indices of the degrees of freedom for the specified
+       edge, including the DOFs for the vertices of the edge. */
+   /** In variable order spaces, multiple sets of DOFs may exist on an edge,
+       corresponding to the different polynomial orders of incident elements.
+       The 'variant' parameter is the zero-based index of the desired DOF set.
+       The variants are ordered from lowest polynomial degree to the highest.
+       @return Order of the selected variant, or -1 if there are no more
+       variants. */
```
YohannDudouit, line 610: I don't think this is the right place to define the concept of `variant` which is central in all this new code, moreover after reading the documentation I still don't understand fully what it is.

fem/pfespace.cpp
```diff
@@ -1549,16 +1553,19 @@ void ParFiniteElementSpace::UnpackDof(int dof,
...
-         else // uniform faces
+         if (uni_fdof >= 0) // uniform faces
          {
             int nf = fec->DofForGeometry(pncmesh->GetFaceGeometry(0));
             index = dof / nf, edof = dof % nf;
          }
+         else // mixed faces or var-order space
+         {
+            const Table &table = var_face_dofs;
+            MFEM_ASSERT(table.Size(), "");
```
YohannDudouit, line 1564: Should we give a more meaningful message?

YohannDudouit (commented):
Overall variable order finite elements are introduced with minimal changes, I just think we should better define the `variant` concept or find a more explicit name.

fem/fespace.hpp
jakubcerveny: OK, I'm thinking of adding more documentation to the tables `var_edge_dofs` and `var_face_dofs`, where the "variant" concept could be better defined, maybe with an example too.

fem/pfespace.cpp
jakubcerveny: If you insist... I just can't think of anything better than "internal error" or similar. This is just a routine check that the table is not empty, which the following code assumes.

fem/fe_coll.hpp
jakubcerveny: I didn't expect this non-public method to be controversial. I guess both names are acceptable, but at the same time both are not optimal: in the short version, "order" is supposed to mean the whole sub-collection (this might be confusing) while in the longer version "order" refers to the integer parameter, which I find a bit clumsy. The short version has the benefit of at least being short, which is why I prefer it.

fem/fespace.hpp
jakubcerveny: Done, there is now a Doxygen comment describing the content of `var_edge_dofs` and `var_face_dofs`.

dylan-copeland:
In the master version with h-refinement, a `GridFunction` can be updated (with correct size) by calling its `Update()` function. In `hpdemo` in this branch, `Update()` does not change the size of the `GridFunction`. Doing `SetSpace(&fespace)` after `fespace.Update(false)` resizes the `GridFunction`. Is this intended to be the only way to update a `GridFunction`?

dylan-copeland:
The result for running `hpdemo` looks fine, but the result of `hpdemo -o 2` is this:
Screen Shot 2021-01-29 at 1 14 17 PM

fem/fespace.cpp
```diff
@@ -93,16 +92,68 @@ FiniteElementSpace::FiniteElementSpace(const FiniteElementSpace &orig,
...
 
-int FiniteElementSpace::GetOrder(int i) const
+void FiniteElementSpace::SetElementOrder(int i, int p)
 {
-   Geometry::Type GeomType = mesh->GetElementBaseGeometry(i);
-   return fec->FiniteElementForGeometry(GeomType)->GetOrder();
+   MFEM_VERIFY(sequence == mesh->GetSequence(), "space has not been Updated()");
+   MFEM_VERIFY(i >= 0 && i < GetNE(), "invalid element index");
+   MFEM_VERIFY(p >= 0 && p <= MaxVarOrder, "order ouf of range");
+   MFEM_ASSERT(!elem_order.Size() || elem_order.Size() == GetNE(),
```
YohannDudouit, line 100: Using `!IsVariableOrder()` would make all the `!elem_order.Size()` more self-documented.

fem/fespace.cpp
```diff
@@ -93,16 +92,68 @@ FiniteElementSpace::FiniteElementSpace(const FiniteElementSpace &orig,
...
+               "internal error");
+
+   if (elem_order.Size()) // already a variable-order space
```
YohannDudouit, line 103: ditto.

fem/fespace.cpp
YohannDudouit: I also don't understand why there is this `MFEM_ASSERT`, `!elem_order.Size()` is treated just below.

YohannDudouit: Also, isn't it expected that `elem_order.Size() == GetNE()`?

fem/fespace.cpp
```diff
@@ -93,16 +92,68 @@ FiniteElementSpace::FiniteElementSpace(const FiniteElementSpace &orig,
...
+   return elem_order.Size() ? elem_order[i] : fec->GetOrder();
+}
+
+int FiniteElementSpace::GetEdgeOrder(int edge, int variant) const
+{
+   if (!IsVariableOrder()) { return fec->GetOrder(); }
+
+   const int* beg = var_edge_dofs.GetRow(edge);
+   const int* end = var_edge_dofs.GetRow(edge + 1);
+   if (variant >= end - beg) { return -1; } // past last variant
```
YohannDudouit, line 143: Wouldn't it make more sense to throw an error?

fem/fespace.cpp
```diff
@@ -615,6 +667,62 @@ FiniteElementSpace::AddDependencies(SparseMatrix& deps, Array<int>& master_dofs,
...
+         edge_pm(j, 1) = (*pm)(j, b);
+         mid[j] = 0.5*((*pm)(j, a) + (*pm)(j, b));
+      }
+
+      // check that the edge does not coincide with the master face's edge
+      const double eps = 1e-14;
+      if (mid[0] > eps && mid[0] < 1-eps &&
+          mid[1] > eps && mid[1] < 1-eps)
+      {
+         int order = GetEdgeDofs(E[i], slave_dofs, 0);
```
YohannDudouit, line 716: I don't think you need the magic number `0`.

fem/fespace.cpp
```diff
@@ -629,26 +737,29 @@ bool FiniteElementSpace::DofFinalizable(int dof, const Array<bool>& finalized,
...
-   GetEdgeDofs(-1 - index, edof);
+   int order = GetEdgeDofs(-1 - index, edof, variant);
 
    int nv = fec->DofForGeometry(Geometry::POINT);
    int ne = fec->DofForGeometry(Geometry::SEGMENT);
    int nn = 2*nv + ne;
 
    dofs.SetSize(nn*nn);
-   if (!dofs.Size()) { return; }
+   if (!dofs.Size()) { return 0; }
```
YohannDudouit, line 762: Wouldn't an error be more appropriate?

fem/fespace.cpp
```diff
@@ -728,19 +866,100 @@ void FiniteElementSpace::BuildConformingInterpolation() const
...
+               (entity == 1) ? Geometry::SEGMENT : mesh->GetFaceGeometry(i);
+
+            if (geom != last_geom)
+            {
+               T.SetIdentityTransformation(geom);
+               last_geom = geom;
+            }
+
+            // get lowest order variant DOFs and FE
+            int p = GetEntityDofs(entity, i, master_dofs, geom, 0);
```
YohannDudouit, line 943: Why `GetEntityDofs` is returning the order `p`?

fem/fespace.cpp
```diff
@@ -728,19 +866,100 @@ void FiniteElementSpace::BuildConformingInterpolation() const
...
+            const auto *master_fe = fec->GetFE(geom, p);
+
+            // constrain all higher order DOFs: interpolate lowest order function
+            for (int variant = 1; ; variant++)
```
YohannDudouit, line 947: `while(q >= 0)`?

fem/fespace.cpp
```diff
@@ -1632,167 +1879,406 @@ void FiniteElementSpace::Construct()
...
+   cR_hp = NULL;
    cP_is_set = false;
    // 'Th' is initialized/destroyed before this method is called.
 
-   nvdofs = mesh->GetNV() * fec->DofForGeometry(Geometry::POINT);
+   int dim = mesh->Dimension();
+   int order = fec->GetOrder();
+
+   MFEM_ASSERT(mesh->GetNumGeometries(dim),
+               "Mesh was not correctly finalized.");
```
YohannDudouit, line 1905: I think it can be `MFEM_VERIFY` it can give valuable information without impacting the performance.

fem/fespace.cpp
```diff
@@ -1632,167 +1879,406 @@ void FiniteElementSpace::Construct()
...
+            {
+               // degenerate face (i.e., edge-face constraint)
+               slave_orders |= edge_orders[-1 - slave.index];
+            }
+         }
+
+         int min_order = MinOrder(slave_orders);
+         if (min_order < MinOrder(face_orders[master.index]))
+         {
+            face_orders[master.index] |= (VarOrderBits(1) << min_order);
```
YohannDudouit, line 2102: A helper function for this kind of operation would make the code more readable.

fem/fespace.hpp
```diff
@@ -320,12 +400,36 @@ class FiniteElementSpace
...
 
    /// The returned SparseMatrix is owned by the FiniteElementSpace.
    const SparseMatrix *GetConformingRestriction() const;
 
+   /** Return a version of the conforming restriction matrix for variable-order
+       spaces with complex hp interfaces, where some true DOFs are not owned by
+       any elements and need to be interpolated from higher order edge/face
+       variants (see also @a SetRelaxedHpConformity()). */
+   /// The returned SparseMatrix is owned by the FiniteElementSpace.
+   const SparseMatrix *GetHpConformingRestriction() const;
```
YohannDudouit, line 431: Is it correct that this method doesn't make sense in most cases?

fem/fespace.hpp
```diff
@@ -712,6 +844,20 @@ class FiniteElementSpace
...
+       whether strict conformity is enforced in cases where master edges/faces
+       have higher polynomial order than their refined slaves. In the default
+       (strict) case, the master polynomial order is reduced to that of the
+       lowest order slave, so the slave can interpolate the master side exactly.
+       If relaxed == true, discontinuities in the solution in such cases are
+       allowed and the master side is not restricted. */
+   void SetRelaxedHpConformity(bool relaxed = true)
+   {
+      relaxed_hp = relaxed;
+      orders_changed = true; // force update
```
YohannDudouit, line 857: Wouldn't `needs_update` be more clear than `orders_changed`?

mesh/mesh.cpp
```diff
@@ -3776,19 +3812,21 @@ Mesh::Mesh(Mesh *orig_mesh, int ref_factor, int ref_type)
...
       max_nv = std::max(max_nv, nvert);
       rfes.GetElementDofs(el, rdofs);
       MFEM_ASSERT(rdofs.Size() == RG.RefPts.Size(), "");
       const FiniteElement *rfe = rfes.GetFE(el);
       orig_mesh->GetElementTransformation(el)->Transform(rfe->GetNodes(),
                                                          phys_pts);
-      const int *c2h_map = rfec.GetDofMap(geom);
-      const int *vertex_map = vertex_fec.GetDofMap(geom);
+      const int *c2h_map = rfec.GetDofMap(geom, ref);
+      const int *vertex_map = vertex_fec.GetDofMap(geom, 1);
```
YohannDudouit, line 3829: magic number.

YohannDudouit (commented):
A bit too many magic numbers to my taste, and some esoteric bitwise operations could be made self-documented behind a function with a well chosen name. Otherwise the user interface is really simple and intuitive.

mesh/mesh.cpp
pazner: I think this should work too, would need to double-check:
```suggestion
      const int *vertex_map = vertex_fec.GetDofMap(geom);
```

tzanio:
@YohannDudouit and @v-dobrev can we finish the review of this PR this week?

jakubcerveny:
@dylan-copeland `FiniteElementSpace::Update` should eventually work just like for h-refinement, i.e., you request an interpolation matrix/operator (`want_transform == true`) and then call `Update` on your GridFunctions. However, the update operator is not implemented yet for p-refinement in this PR. I'm not sure where exactly Update doesn't work for you in `hpdemo`, could you describe your use case?

jakubcerveny:
@dylan-copeland Yeah, sorry, the saved solution in `hpdemo` is incorrect (this was forgotten from `ex1`, removed). Serialization of variable-order spaces and solutions is not implemented yet (this will require a new serialization format for the space).

fem/fe_coll.hpp
jakubcerveny: `Clone` is fine with me.

fem/fespace.cpp
```diff
@@ -93,16 +92,68 @@ FiniteElementSpace::FiniteElementSpace(const FiniteElementSpace &orig,
...
+      orders_changed = true;
+   }
+}
+
+int FiniteElementSpace::GetElementOrder(int i) const
+{
+   MFEM_VERIFY(sequence == mesh->GetSequence(), "space has not been Updated()");
+   MFEM_VERIFY(i >= 0 && i < GetNE(), "invalid element index");
+   MFEM_ASSERT(!elem_order.Size() || elem_order.Size() == GetNE(),
+               "internal error");
```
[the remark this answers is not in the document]
jakubcerveny: The idea is that all user-facing APIs are thoroughly checked with MFEM_VERIFY for valid input. When calling `GetElementOrder` internally we don't need to be so strict so the actual implementation (`GetElementOrderImpl`) is called directly.

fem/fespace.hpp
```diff
@@ -107,25 +107,49 @@ class FiniteElementSpace
...
+       and faces, each of which in a variable order space may host several DOF
+       sets, called DOF set variants. Example: an edge 'i' shared by 4 hexes of
+       orders 2, 3, 4, 5 will hold four DOF sets starting at indices e.g. 100,
+       101, 103, 106. These numbers are stored in row 'i' of \a var_edge_dofs.
+       Variant zero is always the lowest order DOF set, which is followed by
+       consecutive ranges of higher order DOFs. Variable order faces are handled
+       similarly by \a var_face_dofs, which holds at most two DOF set variants
+       per face. The tables are empty for constant-order spaces. */
+   Table var_edge_dofs; // each row contains beginnings of DOF sets for an edge
+   Table var_face_dofs; // var-order space or mixed mesh: similar for faces
```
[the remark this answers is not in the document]
jakubcerveny: You can see this e.g. in `GetElementDofs`: we know the order of the element and need edge and faces DOFs that correspond to the element order. It's also clear how many DOFs the edges/faces should have for that order (`ne`, `nfd`). The number of edge/face DOFs is used to search for the right variant (whatever it's index is) -- the (non-public) method `FindDofs` (also wrapped in `FindEdgeDofs` and `FindFaceDofs`) is used for this.

fem/fespace.cpp
```diff
@@ -269,13 +321,13 @@ void FiniteElementSpace::BuildFaceToDofTable() const
    fc_dof->MakeI(mesh->GetNumFaces());
    for (int i = 0; i < fc_dof->Size(); i++)
    {
-      GetFaceDofs(i, dofs);
+      GetFaceDofs(i, dofs, 0);
```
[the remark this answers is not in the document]
jakubcerveny: I don't think zero is a magic number. Also this is deliberate, we explicitly need the first (lowest-order) variant here. Not using the default is defensive programming: the default really means "don't care", but we do care about the particular variant here.

fem/fespace.cpp
```diff
@@ -269,13 +321,13 @@ void FiniteElementSpace::BuildFaceToDofTable() const
...
       fc_dof->AddColumnsInRow(i, dofs.Size());
    }
    fc_dof->MakeJ();
    for (int i = 0; i < fc_dof->Size(); i++)
    {
-      GetFaceDofs(i, dofs);
+      GetFaceDofs(i, dofs, 0);
```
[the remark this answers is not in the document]
jakubcerveny: Same as above.

fem/fespace.cpp
jakubcerveny: Yes it is expected, the expression `!elem_order.Size() || elem_order.Size() == GetNE()` is an invariant that needs to hold, and we check it in debug with an ASSERT, isn't that what asserts are for?

jakubcerveny: To break the expression down, the array `elem_order` is either empty (regular constant-order space) or it is exactly of length `GetNE()` (in case of variable-order space) -- nothing else, which is exactly what the ASSERT checks.

jakubcerveny: As for using `IsVariableOrder`, I tend to use it everywhere else, but this particular method is one of the few whose very job is to manipulate `elem_order` directly so it makes sense IMO that it's using `elem_order.Size()` directly.

fem/fespace.cpp
jakubcerveny: See above.

fem/fespace.cpp
jakubcerveny: This is for consistency with `GetEdgeDofs` where the -1 is returned at the end when looping over all variants. Here it's strictly not necessary, but since we have no method to return the total number of variants for an edge/face this is "correct" return value at the moment. One may ask if looping until you get -1 is the best design, but that's a different question.

fem/fespace.cpp
jakubcerveny: This is intentional, same as above.

fem/fespace.cpp
jakubcerveny: This is not an error.

fem/fespace.cpp
jakubcerveny: It's the actual order of the variant. It's needed for `GetFE` in subsequent code. This is very internal, why does it bother you?

fem/fespace.cpp
jakubcerveny: Agreed.

fem/fespace.cpp
jakubcerveny: I disagree. These are some of the most elementary bit operations (why would they be "esoteric"?) Wrapping this in a function would be counterproductive to readability IMO.

fem/fespace.hpp
jakubcerveny: What do you mean in "most cases"? If you don't care about hp interpolation then yes, but if you need to do nodal projections on hp spaces then you need this version of the matrix.

fem/fespace.hpp
jakubcerveny: No, `needs_update` would be too broad. This means exactly what it says, `orders_changed` is `true` when any of the element order changed. There may be other reasons the space needs updating (e.g. the mesh changed).

dylan-copeland:
> @dylan-copeland `FiniteElementSpace::Update` should eventually work just like for h-refinement, i.e., you request an interpolation matrix/operator (`want_transform == true`) and then call `Update` on your GridFunctions. However, the update operator is not implemented yet for p-refinement in this PR. I'm not sure where exactly Update doesn't work for you in `hpdemo`, could you describe your use case?

@jakubcerveny It is fine if this PR does not implement this yet. This PR is a great accomplishment already. The issues I have run into, which could be fixed in the future, are that in `hpdemo`, changing to `fespace.Update(true);` results in 

`MFEM abort: Error in update sequence. Space needs to be updated after each mesh modification.`

With `Update(false)`, if a `GridFunction` calls its `Update` function, its size does not change.

jakubcerveny:
@dylan-copeland You are right, this is a bug in the checks in `FiniteElementSpace::Update` (should be fixed now). Now if you change element orders and call `Update(true)`, it still doesn't work, but at least you get the correct error message :-) ("not implemented yet")

GridFunctions will not be updated at the moment (`Update(false)` leaves the update transformation `Th` empty). I think I'll implement the p-refinement update as the first thing in a follow-up PR.

fem/fespace.cpp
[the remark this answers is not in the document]
YohannDudouit: Shouldn't you use `MFEM_ASSERT` in the unsafe version then?

fem/fespace.hpp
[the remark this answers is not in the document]
YohannDudouit: So, if I understand what you wrote: with the order you get the number of edge/face DOFs (is it really that obvious?), and this number of edge/face DOFs is used to search (how?) the variant index?

fem/fespace.cpp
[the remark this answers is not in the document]
YohannDudouit: I think it is a magic number because I need to read the documentation of `GetFaceDofs` to understand the meaning of 0, but this is not a big deal.

fem/fespace.cpp
YohannDudouit: Not convinced, especially with the comment...

fem/fespace.cpp
YohannDudouit: I'm just trying to understand the code...

fem/fespace.cpp
YohannDudouit: Is something like the following really less clear to you?
```suggestion
            AddOrder(face_orders, master.index, min_order);
```

fem/fespace.hpp
YohannDudouit: Usually a method should make sense in all cases, not only in one case. I would have expected `GetConformingRestriction` to internally check for variable p, and to return the meaningful `SparseMatrix` that is currently returned by `GetHpConformingRestriction` when needed.

jakubcerveny: That was my hope as well, but it turns out in the hp case you need both the standard `R` matrix (for things like `GetEssentialTrueDofs` which does `BooleanMult` where the extra dependencies would screw up the boundary DOFs) *and* the `R_hp` version (e.g. to fix the solution after nodal projection).

fem/fespace.cpp
jakubcerveny: Yes, because it's excessive nesting. The reader has to go and check what `AddOrder` does, only to find out that it's a one-liner. Do you really go and wrap every line that looks like e.g. 
```array[index] += constant * variable?```

File: fem/fespace.hpp (hunk 30 above)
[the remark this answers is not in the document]
Comment (jakubcerveny) [reply fem/fespace.hpp:128]: Please keep in mind that:
1. Everything that has to do with these tables is internal to `FiniteElementSpace`
2. The variant index itself is not that important. The supporting methods around `var_edge/face_dofs` are designed to help the main methods which are `GetElementDofs`, `GetFaceDofs`, etc., and their way of operation.
3. The design goal for the data structure of variable order edge and faces was to minimize memory footprint. That's why the information is tightly packed in Tables. 

You're right: based on the number of desired DOFs for an edge/face, the corresponding row of the Table is scanned. The difference between J[i+1] and J[i] is the number of DOFs of a particular variant. When the right variant is found, its first DOF is returned.

File: fem/fespace.cpp (hunk 29 above)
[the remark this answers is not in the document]
Comment (jakubcerveny) [reply fem/fespace.cpp:126]: No. The the public method is completely checked, with some checks hard (to catch user mistakes) and some checks soft (to catch our mistakes). The internal method is free of any checks for performance reasons.

File: fem/fespace.cpp (hunk 25 above)
Comment (YohannDudouit) [reply fem/fespace.cpp:2101]: Yes, I usually try to separate high level algorithms to implementation details. I don't think `AddOrder` requires to be checked, I find it self documented, how it's implemented is not really relevant. But, I'm okay to keep the code as is, it's not a big deal.

**Comment by jakubcerveny** [comment]
@pazner, there seems to be a problem in `mesh.cpp:3859` (in this branch) where a FiniteElementSpace gets created over a non-finalized mesh (we now check this because we need the number of geometries to be known). I'm not sure if I can move the finalization before this point. Do you think you could take look or advise. Miniapps that are using refined meshes (`navier_*`, `performance/ex1p`) are now failing because of this problem.

**Comment by pazner** [comment]
> @pazner, there seems to be a problem in `mesh.cpp:3859` (in this branch) where a FiniteElementSpace gets created over a non-finalized mesh (we now check this because we need the number of geometries to be known). I'm not sure if I can move the finalization before this point. Do you think you could take look or advise. Miniapps that are using refined meshes (`navier_*`, `performance/ex1p`) are now failing because of this problem.

It should be fine to just move that block of code down until after the mesh topology has been finalized. I pushed a commit to try that approach…

There will also be some interactions with #2014 and #2016 for the LOR mesh creation.

**Comment by dylan-copeland** [comment]
Another issue for future work, which is not critical for this PR, is support for `H1_FECollection(order, dim, BasisType::Positive)`. For example, with `-o 2` and `Positive` type, there is the error `--> Error creating cP matrix: n_finalized = 587, ndofs = 863`. If only 1 element has its order increased by 1 (from 2 to 3), then there is no error message, and CG gets nan. In that case, `slave_fe->GetTransferMatrix(*master_fe, T, I);` on fespace.cpp:953 gets `inf` in `I` as the computation involves the inverse of a singular matrix. Again, I don't think this needs to be fixed right away, but hopefully this info will help you debug.

**Comment by jakubcerveny** [comment]
Thank you @pazner, the miniapps are now working. There is one last occurrence (probaly unrelated to LOR) of the problem in the unit tests and I'm really confused becase it only happens in Travis CI and not on my machine:
```
Verification failed: (mesh->GetNumGeometries(dim)) is false:

 --> Mesh was not correctly finalized.

 ... in function: void mfem::FiniteElementSpace::Construct()

 ... in file: fem/fespace.cpp:1904

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

unit_tests is a Catch v2.13.2 host application.

Run with -? for options

Identity Linear Interpolators

  Operators on H1 for element type 1

/home/travis/build/mfem/mfem/tests/unit/fem/test_lin_interp.cpp:177

/home/travis/build/mfem/mfem/tests/unit/fem/test_lin_interp.cpp:177: FAILED:

  SIGABRT - Abort (abnormal termination) signal

test cases:    105 |    104 passed | 1 failed

assertions: 223347 | 223346 passed | 1 failed
```
However in `test_lin_interp.cpp` the Mesh is just a regular hex mesh that should (and is) finalized. I can't reproduce the problem no matter how I run the unit test (serial/parallel build, debug/release, valgrind or not). Is the .gitignore run of Travis doing something special (I tried building exactly as in the log, but no luck). Does anyone have some idea?

**Comment by pazner** [comment]
@jakubcerveny, I think you should be able to reproduce locally if you merge the latest origin/master into your branch

**Comment by jakubcerveny** [comment]
Supporting `BasisType::Positive` is an interesting question. I don't see why it shouldn't work out of the box, at least in `master` it just works. The difference now is that we are using `FiniteElement::GetTransferMatrix` instead of `GetLocalInterpolation`. I debugged this on the simple case of a conforming edge where order 1 and order 2 elements meet. Here the quadratic `SEGMENT` FE is constrained and for nodal elements the matrix looks like this:
```
  master_dofs: 1 4
  slave_dofs: 1 4 7
  I: [row +0] +1.000000e+00 +0.000000e+00
     [row +1] +0.000000e+00 +1.000000e+00
     [row +2] +5.000000e-01 +5.000000e-01
```
This is clear, the vertex DOFs coincide, hence the identity. The quadratic internal DOF is constrained with weights 0.5, 0.5, that's clear too.

Now just switching to `BasisType::Positive` gives:
```
  master_dofs: 1 4
  slave_dofs: 1 4 7
  I: [row +0] +3.125000e-01 -7.500000e-01
     [row +1] -6.875000e-01 +2.500000e-01
     [row +2] +1.875000e+00 +2.000000e+00
```
which I don't really get. The matrix now comes from `ScalarFiniteElement::ScalarLocalInterpolation` which should be able to support different coarse and fine FEs. I can see that unlike for nodal elements, this time a linear system has to be solved to get the coefficients, but it should just work too. Do you see what could be the problem, @v-dobrev?

**Comment by jakubcerveny** [comment]
@pazner thank you, again :-) I didn't realize the Travis runs are actually done on the "merged" state, now I can reproduce the error.

**Comment by jakubcerveny** [comment]
The wrong version of the Mesh constructor was likely called for the 1D cases in `test_lin_interp.cpp` (I think this is a good case in support of #2016). @mlstowell, is https://github.com/mfem/mfem/pull/1817/commits/d3730cb887d24ebd67000dcf31a4d9276ff9e616 the intended meaning?

File: fem/fespace.cpp (hunk 24 above)
Comment (jakubcerveny) [reply fem/fespace.cpp]: This was a good suggestion that revealed several cases where an unfinalized Mesh was being used... :+1:

**Comment by jakubcerveny** [comment]
@dylan-copeland, I have slightly reworked the way the Space and GridFunctions detect if they are up to date. In this branch it is possible for the space to change while the mesh is left untouched (only the element orders can change) so the GridFunction needs to update too. The Space now has its own `sequence` as well as `mesh_sequence` that are both independent. This should fix the problem you had with `hptest`, which should now work in this branch too.

**Comment by dylan-copeland** [comment]
Last week visualization was working for hpdemo (if I remember correctly), but now I get a `glvis` error:
MFEM abort: Unknown input mesh format: MFEM NC mesh v1.0
 ... in function: void mfem::Mesh::Loader(std::istream &, int, std::string)
 ... in file: mesh/mesh.cpp:3057

**Comment by tzanio** [comment]
> Last week visualization was working for hpdemo (if I remember correctly), but now I get a glvis error:
MFEM abort: Unknown input mesh format: MFEM NC mesh v1.0
... in function: void mfem::Mesh::Loader(std::istream &, int, std::string)
... in file: mesh/mesh.cpp:3057

Rebuild `glvis` with the latest `master`

File: linalg/hypre.cpp [hunk 33]
```diff
@@ -3878,7 +3878,7 @@ void HypreAMS::Init(ParFiniteElementSpace *edge_fespace)
       }
       else
       {
-         p = edge_fespace->GetOrder(0);
+         p = edge_fespace->GetElementDegree(0);
```
dylan-copeland, line 3881: Should there be a `MFEM_VERIFY` that the space is not of variable order?

mesh/pmesh.cpp
```diff
@@ -1208,7 +1208,7 @@ ParMesh::ParMesh(ParMesh *orig_mesh, int ref_factor, int ref_type)
          const Geometry::Type geom = Geometry::SEGMENT;
          const int nvert = Geometry::NumVerts[geom];
          RefinedGeometry &RG = *GlobGeometryRefiner.Refine(geom, ref_factor);
-         const int *c2h_map = rfec.GetDofMap(geom);
+         const int *c2h_map = rfec.GetDofMap(geom, ref_factor); // FIXME hp
```
dylan-copeland, line 1211: What is the `FIXME` for?

examples/hpdemo.cpp
```diff
@@ -0,0 +1,318 @@
+//                     hp refinement demo (NOT TO BE MERGED)
```
dylan-copeland, line 1: Why not merge `hpdemo` into master? Also the `hpfem` miniapp from branch `p-refinement` works (in my branch `hpfem-var-order-space` from this branch). Why not include it? The unit test for variable order is good, but users might find an example or miniapp more illustrative.

fem/fe_coll.cpp
```diff
@@ -289,6 +290,31 @@ FiniteElementCollection *FiniteElementCollection::New(const char *name)
    return fec;
 }
 
+FiniteElementCollection *FiniteElementCollection::Clone(int p) const
+{
+   // default implementation for collections that don't care about variable p
+   MFEM_ABORT("Collection " << Name() << " does not support variable orders.");
+   (void) p;
```
dylan-copeland, line 297: What does `(void) p;` do?

fem/fe_coll.cpp
```diff
@@ -1504,6 +1530,8 @@ const int *RT1_3DFECollection::DofOrderForOrientation(Geometry::Type GeomType,
 
 
 H1_FECollection::H1_FECollection(const int p, const int dim, const int btype)
+   : FiniteElementCollection(p)
+   , dim(dim)
```
dylan-copeland, line 1534: Not a big deal, but this is unusual style. Was it intentional, or an artifact of `make style`?

fem/fespace.hpp
```diff
@@ -107,25 +107,49 @@ class FiniteElementSpace
...
+   int nvdofs, nedofs, nfdofs, nbdofs;
+   int uni_fdof; ///< # of single face DOFs if all faces uniform; -1 otherwise
+   int *bdofs; ///< internal DOFs of elements if mixed/var-order; NULL otherwise
+
+   /** Variable-order spaces only: these tables hold DOF assignments for edges
+       and faces, each of which in a variable order space may host several DOF
+       sets, called DOF set variants. Example: an edge 'i' shared by 4 hexes of
+       orders 2, 3, 4, 5 will hold four DOF sets starting at indices e.g. 100,
+       101, 103, 106. These numbers are stored in row 'i' of \a var_edge_dofs.
+       Variant zero is always the lowest order DOF set, which is followed by
```
dylan-copeland, line 123: I am more confused with this example than without it. Are the indices 100, 101, 103, 106 determined by the orders, or are they just randomly chosen numbers that might be the indices?

fem/fespace.hpp
```diff
@@ -169,31 +204,81 @@ class FiniteElementSpace
        boundary. */
    void BuildNURBSFaceToDofTable() const;
 
+   /// Bit-mask representing a set of orders needed by an edge/face.
+   typedef std::uint64_t VarOrderBits;
+   static constexpr int MaxVarOrder = 8*sizeof(VarOrderBits) - 1;
+
```
dylan-copeland, line 210: Is `MaxVarOrder = 63` adequate for all use cases. If it is not large enough, will there be an informative error or a seg-fault?

fem/fespace.hpp
```diff
@@ -320,12 +408,36 @@ class FiniteElementSpace
    bool Conforming() const { return mesh->Conforming(); }
    bool Nonconforming() const { return mesh->Nonconforming(); }
 
+   /// Sets the order of the i'th finite element.
+   /** By default, all elements are assumed to be of fec->GetOrder(). Once
+       SetElementOrder is called, the space becomes a variable order space. */
+   void SetElementOrder(int i, int p);
+
+   /// Returns the order of the i'th finite element.
+   /** @note This is the 'order' of the FiniteElementCollection. The actual
```
dylan-copeland, line 417: Would it be clearer to say the `order` of the element rather than `FiniteElementCollection`? The collection can have multiple (variable) orders.

fem/fespace.hpp
```diff
@@ -724,6 +865,20 @@ class FiniteElementSpace
...
    }
 
+   /** In variable order spaces on nonconforming meshes, this function controls
+       whether strict conformity is enforced in cases where master edges/faces
+       have higher polynomial order than their refined slaves. In the default
+       (strict) case, the master polynomial order is reduced to that of the
+       lowest order slave, so the slave can interpolate the master side exactly.
+       If relaxed == true, discontinuities in the solution in such cases are
+       allowed and the master side is not restricted. */
+   void SetRelaxedHpConformity(bool relaxed = true)
```
dylan-copeland, line 875: This is a function that general users need to understand on a high level, even if they do not understand anything about the implementation. The case described here may not be entirely clear to users. It may help to explain that "master edges/faces" are coarser than their slaves in h-refinement (if I understand correctly). The reader might confuse h- and p-refinement, so "refined slaves" might allow them to think of "p-refined slaves" with higher order. This could be clarified by explicitly saying "h-refined slaves." It could also be mentioned that the case of equal or lower order on the master always has strict conformity. Also, it is not clear why `relaxed == false` would ever be desirable, and what it means exactly. It is not the same as complete discontinuity (as in L2 spaces), as only a subset of DOFs are allowed to be discontinuous. Is there a use case for `false` for which this is designed?

fem/fespace.cpp
```diff
@@ -90,19 +89,74 @@ FiniteElementSpace::FiniteElementSpace(const FiniteElementSpace &orig,
...
    Constructor(mesh, NURBSext, fec, orig.vdim, orig.ordering);
 }
 
-int FiniteElementSpace::GetOrder(int i) const
+void FiniteElementSpace::SetElementOrder(int i, int p)
+{
+   MFEM_VERIFY(mesh_sequence == mesh->GetSequence(),
+               "Space has not been Updated() after a Mesh change.");
+   MFEM_VERIFY(i >= 0 && i < GetNE(), "Invalid element index");
+   MFEM_VERIFY(p >= 0 && p <= MaxVarOrder, "Order ouf of range");
```
dylan-copeland, line 101: ouf -> out

fem/fespace.cpp
```diff
@@ -615,6 +670,62 @@ FiniteElementSpace::AddDependencies(SparseMatrix& deps, Array<int>& master_dofs,
...
+{
+   // In variable-order spaces in 3D, we need to only constrain interior face
+   // DOFs (this is done one level up), since edge dependencies can be more
+   // complex and are primarily handled by edge-edge dependencies. The one
+   // exception is edges of slave faces that lie in the interior of the master
+   // face, which are not covered by edge-edge relations. This function finds
+   // such edges and makes them constrained by the master face.
+   // See also https://github.com/mfem/mfem/pull/1423#issuecomment-633916643
+
+   Array<int> V, E, Eo; // TODO: LocalArray
```
dylan-copeland, line 687: What is `LocalArray` and why `TODO`?

dylan-copeland:
@jakubcerveny Thanks again for this great PR. I have tested it successfully with the hpfem miniapp from the old p-refinement-dev branch, and everything seems to be working fine. For such a large change (with new pointers and ownership issues), it is a good idea to run valgrind on the new tests (unit and hpdemo). If you could report the results of valgrind here it would help the reviewers know there are no memory leaks.

fem/fespace.cpp
jakubcerveny: Fixed.

mesh/pmesh.cpp
jakubcerveny: I'm not sure, this was added by Will @pazner

linalg/hypre.cpp
jakubcerveny: Added.

examples/hpdemo.cpp
jakubcerveny: The demo is too simple, not really useful and has to use `ProlongToMaxOrder` at the moment, which would probably confuse users. There will be at least two more PRs, eventually we will probably have a proper example too. The hp miniapp is IMO not ready either, we should first get rid of the oscillations due to the relaxed mode (if you are using the miniapp and wouldn't mind helping experimenting with DG integrators tu suppress the oscillations, that would be a great help).

fem/fe_coll.cpp
jakubcerveny: This is a common technique to suppress an "unused variable" compiler warning.

jakubcerveny: Also it explicitly documents that we are not using the parameter and are aware of it.

fem/fe_coll.cpp
jakubcerveny: Yeah this is a style that I brought from another project, I can remove it if there are objections.

fem/fespace.hpp
jakubcerveny: These are examples of DOF numbers, as in "this 5th-order edge has internal DOFs 100-103". I didn't feel that users need to study this in detail but Yohann requested a documentation for these tables, so I added one. If it's more confusing than useful I can remove the comment, what do you think @tzanio?

fem/fespace.hpp
jakubcerveny: Hopefully this is more than enough for most uses. There is an `MFEM_VERIFY` in `SetElementOrder` so there shouldn't be a mysterious failure if the user tries to set an order that is too high. If this ever is not enough, we can come up with a solution. Right now I think this is sufficient *and* many times more efficient than e.g. having an `std::set` for each edge and face.

fem/fespace.hpp
jakubcerveny: This refers to the `order` parameter the user constructed the `FiniteElementCollection` with. The point is that this may not be the actual polynomial degree (there is an unfortunate quirk in the naming of the RT collections).

fem/fespace.cpp
jakubcerveny: I envision a descendant of `Array` for internal use like here, that would use a small buffer local to the class to hold its elements. Using `LocalArray` would avoid heap allocations. It's a common use case that we need small temporary arrays in methods like this, and a large number of heap allocations may hurt performance.

fem/fespace.hpp
jakubcerveny: I have updated this comment to avoid the internal terminology, added explicit mentions that these are coarse/fine NC neighbors (not p-refinements), and added a link to a github comment with pictures.

The relaxed case was initially considered for special uses only, but it turns out it will be commonly needed, as seen e.g. in the hptest miniapp, where the strict conformity is overconstraining the solution in some cases (see discussions on the L-shaped problem in the `p-refinement-dev` PR) and disturbing the calculation hp refinement candidates.

jakubcerveny:
@dylan-copeland, thanks for testing and reviewing this in detail. As for valgrind, I can report that the unit test and hpdemo run cleanly (confirmed again now), but I believe the reviewers shouldn't take my word for it a make a couple of valgrind runs themselves, just to be sure.

fem/fe_coll.cpp
tzanio: I am flexible on this. I'll put the comma on the previous line (or fit both on one line) if it was me.

fem/fespace.hpp
tzanio: This is definitely for developers, not users. Could there be a better place, maybe somewhere in the implementation, where this can be moved and expanded? I believe @YohannDudouit's request was for information that can help MFEM developers (e.g. folks at LLNL) that wants to understand the algorithm and what basic data structures are involved...

fem/fespace.hpp
dylan-copeland: In the variable order case, this returns `elem_order[i]` which is the order for the element. Isn't this different from the `FiniteElementCollection` order?

linalg/hypre.cpp
dylan-copeland: Thanks, this could also be added in `HypreADS`.

fem/fespace.hpp
jakubcerveny: There are two things: `order` as passed at the very beginning to `FiniteElementCollection`, this is the same kind of order as `elem_order[i]`, and then there's the real polynomial degree as returned by `FiniteElement::GetOrder()`. For example, `RT_FiniteElementCollection` of order 0 contains `FiniteElements` whose `GetOrder` returns 1.

linalg/hypre.cpp
jakubcerveny: Added.

linalg/hypre.cpp
```diff
@@ -4143,13 +4144,14 @@ void HypreADS::Init(ParFiniteElementSpace *face_fespace)
    int p = 1;
    if (face_fespace->GetNE() > 0)
    {
+      MFEM_VERIFY(!edge_fespace->IsVariableOrder(), "");
```
dylan-copeland, line 4147: `face_fespace`

jakubcerveny: Oops, my bad, I didn't build `parallel`. Fixed

jakubcerveny:
@dylan-copeland it seems `BasisType::Positive` is working now, thanks to Veselin's fix in `ScalarLocalInterpolation`.

fem/fespace.hpp
dylan-copeland: It seems the relaxed case is useful for studying convergence and understanding the hp algorithm, which justifies having this feature, but in practice I don't see how it would be useful for solving PDEs, in which case one would want full discontinuity (L2) or full continuity (H1), not something in between.

tzanio: Maybe it could be a kind of non-conforming method for H1 like Crouzeix-Raviart?

dylan-copeland (approved):

fem/fespace.hpp
v-dobrev: I think both options can be useful in practice. The users have the option to try both and decide for themselves.

tzanio:
@jakubcerveny and @pazner can you help merge `master` here (after the merge of #2014)?

jakubcerveny:
I looked at the conflicts but it's not obvious/trivial to resolve them, I think it will be best handled by Will since it's his code.

pazner:
> I looked at the conflicts but it's not obvious/trivial to resolve them, I think it will be best handled by Will since it's his code.

I'll take a crack at resolving this.

pazner:
Hopefully it should work now…


GLVis_s02

tzanio:
Travis is failing now, can you please take a look?

jakubcerveny:
This seems similar to the finalization problem that we resolved by moving the Nodes construction block, maybe the fix got lost in the merge?

pazner:
> This seems similar to the finalization problem that we resolved by moving the Nodes construction block, maybe the fix got lost in the merge?

Good catch, I think you're right (the construction of nodes is quite different now). I moved it to be after finalizing the topology, let's see if that works…

pazner:
CI looks ok now 👍  thanks for the pointer @jakubcerveny

tzanio:

v-dobrev:
Please take a look at the comment I left here: https://github.com/mfem/mfem/pull/1046#issuecomment-781022119

I think we should try to coordinate and unify, if possible, some of the changes these two PRs.

tzanio:
There are some autotest errors due to this PR, see the [#autotest channel](https://mfem.slack.com/archives/C0106S7MNTS) on Slack

jakubcerveny:
I had to remove the new methods `FiniteElementSpace::GetEdgeOrder` and `GetFaceOrder`, because the old `GetFaceOrder` is needed by `HypreADS` (in case static condensation is on) and behaves differently for RT spaces. `GetFaceOrder` is now deprecated and replaced by `GetFaceDegree`, similar to `GetElementDegree`. Examples `ex4p` and `ex8p` should now be working as before.

tzanio:

jakubcerveny:
Oops, @tzanio did you merge with the latest commit (e328861), that was a fix of d8bf00f

tzanio:
Yes, https://github.com/mfem/mfem/commits/next

tzanio:
Another error from the `build-test-tpls` test on LC:

https://mybitbucket.llnl.gov/projects/MFEM/repos/autotest/browse/toss_3_x86_64_ib/2021-02-19-llnl-gh-next/build-test-tpls.err

jakubcerveny:
The LC error should hopefully be fixed too.

tzanio:

jakubcerveny:
Please hold merging until next week, I'm trying a new approach to fix the order/degree ambiguity: 2949fa9563815d6b1c6a7667b43aad3faa94b550

jakubcerveny:
There is a last-minute modification in #2075, please take a look...

tzanio:

tzanio:
@jakubcerveny, you saw the autotest errors with the latest changes in https://mfem.slack.com/archives/C0106S7MNTS/p1614875750033300?

jakubcerveny:
The problems in autotest are caused by line 4150 in hypre.cpp:
```
     p = face_fespace->GetFaceOrder(0) + 1;
```
In master, when one runs `mpirun -np 4 ex4p -m ../data/escher.mesh -o 2 -sc` the result is `p = 2`, because `GetFaceOrder` returns 1. This is strange because the collection is `RT_Trace_3D_P1`, i.e., quadratic elements. How come the faces are of order 1?

trace_rt

In this branch `GetFaceOrder` currently returns 0 for the trace collection, which is a bug, but when I fix it to return 2 (the polynomial degree of the collection), it's still not right. I may need some help to understand this.

tzanio:
Hi Jakub,

I'm writing this quickly while listening to a talk, so take it with a grain of salt, but for RT_0 the order (polynomial degree) in the element is 1, but on the face it is zero (which is why you need just single degree of freedom per face).

In general `RT_k-1` has basis functions `(Q_k,k-1,k-1 ; Q_k-1,k,k-1 ; Q_k-1,k-1,k)` so e.g. on the x-faces the normal component is a `Q_k,k-1,k-1` function but for a fixed `x` so it is really a `Q_0,k-1,k-1` function.

(There are similar considerations for Nedelec elements, the basis for `ND_k` is `(Q_k-1,k,k ; Q_k,k-1,k ; Q_k,k,k-1)`)

v-dobrev:
Just to clarify: for the RT_k spaces the _normal_ trace on the faces is a scalar polynomial of degree k, P_k/Q_k.

For ND_k spaces the _tangential_ trace on the faces is the same order, ND_k.

jakubcerveny:
OK, so now `GetFaceOrder` does the same thing as before for constant-order spaces (asks the actual face FE), hopefully this will fix the problem.

tzanio:

v-dobrev:
Re-merged in `next`.

tzanio:
Error in tonight's regressions tests:

```
> mpirun -np 4 ex6p -m ../data/inline-segment.mesh -o 1 -md 200 -no-vis

MFEM abort: Unsupported boundary element geometry.
 ... in function: mfem::NCMesh::NCMesh(const mfem::Mesh*)
 ... in file: mesh/ncmesh.cpp:172
```

tzanio:
There is also an issue on LC. I haven't dug into it, but some of the parallel unit tests fail there:

```
    Parallel unit tests [ srun -n 4 punit_tests ... ]: FAILED  (4.45s 6580kB)

Verification failed: (mesh->GetNumGeometries(dim)) is false:
 --> Mesh was not correctly finalized.
 ... in function: void mfem::FiniteElementSpace::Construct()
 ... in file: fem/fespace.cpp:1888

Backtrace:
0) [0x2aaaaae9385e]: mfem::mfem_error(char const*)
1) [0x2aaaaba81425]: mfem::FiniteElementSpace::Construct()
2) [0x2aaaaba72f15]: mfem::FiniteElementSpace::Constructor(mfem::Mesh*, mfem::NURBSExtension*, mfem::FiniteElementCollection const*, int, int)
3) [0x2aaaab39d963]: mfem::ParFiniteElementSpace::ParFiniteElementSpace(mfem::ParMesh*, mfem::FiniteElementCollection const*, int, int)
4) [0x6e6a79]: _INTERNAL1dab4fe6::____C_A_T_C_H____T_E_S_T____2()
5) [0x43ffec]: Catch::RunContext::runCurrentTest(std::string&, std::string&)
6) [0x43f102]: Catch::RunContext::runTest(Catch::TestCase const&)
7) [0x43e965]: Catch::(anonymous namespace)::TestGroup::execute()
8) [0x435823]: Catch::Session::runInternal()
9) [0x426261]: main
10) [0x2aaabceb7554]: __libc_start_main
11) [0x425918]: _start

```

jakubcerveny:
> mpirun -np 4 ex6p -m ../data/inline-segment.mesh -o 1 -md 200 -no-vis

> MFEM abort: Unsupported boundary element geometry.

I think this comes from #2038

jakubcerveny:
I can't reproduce the "LC" error in this branch, could it be from some other PR too?

tzanio:
>I can't reproduce the "LC" error in this branch (also not in #2038), could it be from some other PR too?

@adrienbernede, can you help us investigate this?

adrienbernede:
LGTM

v-dobrev:
The gitlab test uncovered another issue when building with PUMI -- this is not tested elsewhere. The error is:
```console
mesh/pumi.cpp: In constructor 'mfem::GridFunctionPumi::GridFunctionPumi(mfem::Mesh*, apf::Mesh2*, apf::Numbering*, int)':
mesh/pumi.cpp:747:4: error: 'sequence' was not declared in this scope
    sequence = 0;
    ^~~~~~~~
```

fem/fespace.cpp
```diff
@@ -1632,434 +1862,772 @@ void FiniteElementSpace::Construct()
...
+   cR_hp = NULL;
    cP_is_set = false;
    // 'Th' is initialized/destroyed before this method is called.
 
-   nvdofs = mesh->GetNV() * fec->DofForGeometry(Geometry::POINT);
+   int dim = mesh->Dimension();
+   int order = fec->GetOrder();
+
+   MFEM_VERIFY(mesh->GetNumGeometries(dim),
+               "Mesh was not correctly finalized.");
```
v-dobrev, line 1888: This check caused autotest failures. It does not work with empty processors.

jakubcerveny: Thank you, should hopefully be fixed now.

v-dobrev: By the way, this was triggered only when building with direct parallel solvers, like SuperLU_DIST -- from here: https://github.com/mfem/mfem/blob/33827fe5ccaccc5a6ae322c198ffd945ba0001dd/tests/unit/linalg/test_direct_solvers.cpp#L165
at least in 1D the mesh has 2 elements and we run the test on 4 ranks.

v-dobrev:
LGTM

v-dobrev:
Re-merged in `next`.

brendankeith (approved):

tzanio (approved):
Amazing work 🚀 

Thank you so much @jakubcerveny!

adam-sim-dev:
Dear developers, any progress of the parallel version variable order space? @jakubcerveny @tzanio

PR #1818, opened 2020-10-16
Adaptive linear rtol for NewtonSolver

jandrej:
This PR implements adaptive relative tolerances for the linear solves in `NewtonSolver` using a Eisenstat-Walker type method. Additionally it introduces a "Matrix-Vector-Product" callback inside `NewtonSolver` for using Finite Difference Jacobian actions.

| [#1818](https://github.com/mfem/mfem/pull/1818) | @jandrej | @tzanio | @kmittal2 + @vladotomov | 12/02/20 | 01/04/21 | 01/08/21 | |


vladotomov:
I'd like to review this when it's ready.

tzanio:
The documentation test on Travis is broken: https://travis-ci.com/github/mfem/mfem/jobs/452723012

examples/ex10p.cpp
```diff
@@ -564,7 +568,7 @@ HyperelasticOperator::HyperelasticOperator(ParFiniteElementSpace &f,
    J_minres->SetRelTol(rel_tol);
    J_minres->SetAbsTol(0.0);
    J_minres->SetMaxIter(300);
-   J_minres->SetPrintLevel(-1);
+   J_minres->SetPrintLevel(2);
```
vladotomov, line 571: revert this?

linalg/solvers.hpp
```diff
@@ -434,6 +465,43 @@ class NewtonSolver : public IterativeSolver
...
+                                const double fnorm) const;
+
+   /** @brief Method for the adaptive linear solver rtol invoked after the
+       linear solve. */
+   void AdaptiveLinRtolPostSolve(const Vector &x,
+                                 const Vector &b,
+                                 const int it,
+                                 const double fnorm) const;
+
+   static void DQOpTimes(const Vector &v, Vector &Jv, void *ctx);
```
vladotomov, line 502: These function pointers seem too intrusive (or ignore) the existing linear solvers' interface. Why don't you create a custom `Operator` that knows about the `NewtonSolver` and give it to the linear solver? This would avoid changes in `IterativeSolver`. For example, the `NewtonSolver` can pass this operator [here](https://github.com/mfem/mfem/pull/1818/files#diff-3d145a6c87b1ab35fc52ca1fe9793356cbde85403f3b872b0c7eca7c1a554a6eR1622), if approximate Hessians are enabled.

linalg/solvers.hpp
```diff
@@ -406,7 +415,29 @@ void MINRES(const Operator &A, Solver &B, const Vector &b, Vector &x,
...
+   // Linear residual norm of the previous iterate
+   mutable double lnorm_last = 0.0;
+   // Forcing term (linear residual rtol) from the previous iterate
+   mutable double eta_last = 0.0;
+   // Eisenstat-Walker factor gamma
+   double gamma;
+   // Eisenstat-Walker factor alpha
+   double alpha;
+   // Safeguard threshold
+   double sg_threshold = 0.1;
```
vladotomov, line 440: Can be a local variable.

linalg/solvers.hpp
```diff
@@ -434,6 +465,43 @@ class NewtonSolver : public IterativeSolver
...
+    iteration, based on the algorithm presented in [1].
+
+    The maximum linear solver relative tolerance @a rtol_max should be < 1.
+    For @a version 1 the parameters @a alpha and @a gamma are ignore. For @a
+    version 2 @a alpha has to be between 0 and 1 and @a alpha between 1 and 2.
+
+    [1] Eisenstat, Stanley C., and Homer F. Walker. "Choosing the forcing terms
+    in an inexact Newton method."
+    */
+   void SetAdaptiveLinRtol(const int version = 2,
```
vladotomov, line 483: Maybe `type` instead of `version`.

linalg/solvers.hpp
```diff
@@ -434,6 +465,43 @@ class NewtonSolver : public IterativeSolver
...
+                           const double rtol0 = 0.5,
+                           const double rtol_max = 0.9,
+                           const double alpha = 0.5 * (1.0 + sqrt(5.0)),
+                           const double gamma = 1.0);
+
+   /** @brief Method for the adaptive linear solver rtol invoked before the
+       linear solve. */
+   void AdaptiveLinRtolPreSolve(const Vector &x,
```
vladotomov, line 491: `PreSolve`, `PostSolve` can be `protected`.

vladotomov (changes requested):

linalg/solvers.hpp
jandrej: My intention here was to circumvent the linear solver interface, otherwise we can't keep the preconditioning operators from the application perspective. If I overwrite the operator before the linear solve I'd have to memorize the original operator, no?

In a discussion with @v-dobrev we concluded that it might be a good addition to have a "callback" that resembles a custom Jacobian-Vector-Product in the IterativeSolver interface.

I'm all for a simpler, or more clear, approach, but I don't understand the proposed alternative.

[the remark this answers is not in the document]
jandrej: Looking at this method, the fixed `eps` was not intended. I have to input the correct formula there.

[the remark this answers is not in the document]
kmittal2: Yep, I was wondering about that. I was also wondering if one-sided difference is sufficient but I didn't see a big difference in the final result with two-sided difference as compared to one-sided for the experiments I tried.

vladotomov: I must be missing something. I'll discuss in the context of `NewtonSolver` as it is the main interest of the PR.

You have an approximate `Mult` (which doesn't need the matrix assembly) for the gradient operator , and you give it to the linear solver, and the linear solver uses it. But then you always compute the original operator (assemble the matrix), as a preconditioner might need it.

Then what is the point of doing the approximate `Mult` if the exact action is available / computed anyway?

linalg/solvers.cpp
```diff
@@ -1641,6 +1656,110 @@ void NewtonSolver::Mult(const Vector &b, Vector &x) const
...
+
+   constexpr double epsmach = 2.0 * DBL_EPSILON;
+   const double eps = sqrt((1.0 + c->Norm(x)) * epsmach)
+                      / max(sqrt(epsmach), c->Norm(v));
+
+   for (int i = 0; i < Jv.Size(); i++)
+   {
+      xpev(i) = x(i) + eps * v(i);
+   }
+
```
vladotomov, line 1754: Maybe you could replace the loops here with function calls to `Vector` to reuse its kernels.

linalg/solvers.cpp
```diff
@@ -1641,6 +1656,110 @@ void NewtonSolver::Mult(const Vector &b, Vector &x) const
...
+}
+
+void NewtonSolver::DQOpTimes(const Vector &v, Vector &Jv, void *ctx)
+{
+   NewtonSolver *c = static_cast<NewtonSolver *>(ctx);
+   Vector &xpev = c->z;
+   const Vector &x = c->GetCurrentIterate();
+   const Vector &fx = c->GetCurrentResidual();
+
+   constexpr double epsmach = 2.0 * DBL_EPSILON;
```
vladotomov, line 1746: Why the `2.0`?

linalg/solvers.hpp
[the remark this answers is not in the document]
vladotomov: I second Ketan's request for adding comments; also the function name is somewhat cryptic.

jandrej: Ok. If the exact action is available, it could definitely be used. However, in practice it shows often disadvantages since the "correctness" (or rather incorrectness) of the gradient and the relative tolerance algorithms can be out of agreement. It could be made optional to use DQ.

linalg/solvers.cpp
jandrej: Habit introduced by old fortran books :-/

linalg/solvers.hpp
```diff
@@ -434,6 +476,30 @@ class NewtonSolver : public IterativeSolver
...
+    */
+   void SetAdaptiveLinRtol(const int type = 2,
+                           const double rtol0 = 0.5,
+                           const double rtol_max = 0.9,
+                           const double alpha = 0.5 * (1.0 + sqrt(5.0)),
+                           const double gamma = 1.0);
+
+   static void DQOpTimes(const Vector &v, Vector &Jv, void *ctx);
+
+   mutable Vector z;
```
vladotomov, line 502: This is recomputed by `DQOpTimes`, but I don't see how the stored values are used for anything.

vladotomov (changes requested):
The adaptive relative tolerance looks very good and it definitely leads to improvements in the `mesh-optimizer`. It should be mentioned in the changelog.

I don't think the approximate Jacobian-vector-product code is ready. We won't use it in this form in the `mesh-optimizer`, as it leads to slowdowns, with unclear benefits. I commented out the usage in `mesh-optimizer`, so that capability is not tested/used by anything now. I think it should be moved to another PR with some clear use cases. There should be a way to skip the matrix assembly. I can see how it could combine with action-based preconditioners.

tzanio:
There have been recent comments and conflicts to address... extending the deadline

linalg/solvers.hpp
```diff
@@ -74,6 +74,9 @@ class IterativeSolver : public Solver
    Solver *prec;
    IterativeSolverMonitor *monitor = nullptr;
 
+   void (*OpTimes)(const Vector &v, Vector &Jv, void *ctx) = nullptr;
+   void *OpTimesCtx = nullptr;
```
vladotomov, line 78: Remove?

miniapps/meshing/pmesh-optimizer.cpp
```diff
@@ -889,12 +895,17 @@ int main (int argc, char *argv[])
    {
       // Specify linear solver when we use a Newton-based solver.
       solver.SetPreconditioner(*S);
+      //static_cast<IterativeSolver *>(S)->SetOpTimes(&solver, NewtonSolver::DQOpTimes);
```
vladotomov, line 898: Remove.

vladotomov (approved):

tzanio:
@kmittal2, a gentle reminder that you review is due 😄

kmittal2 (approved):

kmittal2:
> @kmittal2, a gentle reminder that you review is due 😄

Apologies @tzanio, I was on travel until recently so missed the deadline.

tzanio:

tzanio:
@jandrej, can you please mention this in `CHANGELOG`?

CHANGELOG
```diff
@@ -49,6 +49,9 @@ Version 4.2.1 (development)
   deprecated EvalSymmetric in MatrixCoefficient. Added DiagonalMatrixCoefficient
   for clarity, which is a typedef of VectorCoefficient.
 
+- Implemented an adaptive linear solver tolerance option for NewtonSolver based
+  on the algorithm of Eisenstat and Walker.
+
```
tzanio, line 54: Thanks!

examples/ex10p.cpp
```diff
@@ -178,6 +178,7 @@ int main(int argc, char *argv[])
    double visc = 1e-2;
    double mu = 0.25;
    double K = 5.0;
+   bool adaptive_lin_rtol = true;
```
v-dobrev, line 181: It looks like this parameter is not used. Also, why not leave it `false` by default?

v-dobrev: Also, should we add this option in the serial `ex10`?

tzanio: Good points

PR #1825, opened 2020-10-20
ParaView output of boundary elements and attributes

pazner:
Add `Mesh::PrintBdrVTU` ﻿to output the boundary elements and attributes of the mesh for visualization using ParaView (useful for checking boundary conditions).

Also allows for output of parallel meshes in the PVTU format (so now both `ParMesh::PrintVTU` and `ParMesh::PrintBdrVTU` should work properly).

2d_bdr
3d_bdr

| [#1825](https://github.com/mfem/mfem/pull/1825) | @pazner | @tzanio | @tzanio + @jandrej + @rcarson3 | 10/20/20 | 10/22/20 | 10/23/20 | |

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -267,6 +267,9 @@ Miscellaneous
   ParaView v5.8.1. In addition, ADIOS2 allows for setting a user-defined number
   of data substreams/subfiles at scale. See examples 5, 9, 12, 16.
 
+- Added VTU output of boundary elements and attributes and parallel VTU (PVTU)
+  output of parallel meshes for visualization using ParaView.
+
 - The integration order used in the ComputeLpError and ComputeElementLpError
   methods of class GridFunction has been increased.
 
--- a/fem/datacollection.cpp
+++ b/fem/datacollection.cpp
@@ -908,7 +908,7 @@ void ParaViewDataCollection::Save()
 
       // CELL DATA
       out << "<PCellData>\n";
-      out << "\t<PDataArray type=\"Int32\" Name=\"" << "material"
+      out << "\t<PDataArray type=\"Int32\" Name=\"" << "attribute"
           << "\" NumberOfComponents=\"1\""
           << " format=\"" << GetDataFormatString() << "\"/>\n";
       out << "</PCellData>\n";
--- a/fem/datacollection.hpp
+++ b/fem/datacollection.hpp
@@ -382,6 +382,10 @@ public:
    int Error() const { return error; }
    /// Reset the error state
    void ResetError(int err = NO_ERROR) { error = err; }
+
+#ifdef MFEM_USE_MPI
+   friend class ParMesh;
+#endif
 };
 
 
--- a/mesh/mesh.cpp
+++ b/mesh/mesh.cpp
@@ -9041,7 +9041,8 @@ void Mesh::PrintVTK(std::ostream &out)
 void Mesh::PrintVTU(std::string fname,
                     VTKFormat format,
                     bool high_order_output,
-                    int compression_level)
+                    int compression_level,
+                    bool bdr)
 {
    int ref = (high_order_output && Nodes) ? Nodes->FESpace()->GetOrder(0) : 1;
    fname = fname + ".vtu";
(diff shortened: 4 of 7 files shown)


rcarson3:
@pazner could you maybe provide a small example problem just for the PR that we can test this on?

pazner:
> @pazner could you maybe provide a small example problem just for the PR that we can test this on?

You can add
```c++
pmesh.PrintVTU("pmesh");
pmesh.PrintBdrVTU("pmesh_bdr");
```
to ex1p (or really any of the examples). It would be good to try this with meshes that have different boundary attributes (for example, inline-quad, inline-hex, the various "beam" meshes, etc.).

tzanio:
If you are interested, @benzwick and @tangqi can you give this PR a quick look? We'd like to merge it as part of the mfem-4.2 release this Monday.

tzanio (approved):
Looks great, thanks @pazner!

Should we mention it in `CHANGELOG`?

pazner:
> Should we mention it in `CHANGELOG`?

Sure, I updated it.

Another small change to note in this PR is that the element attributes used to be output as a field called "material" in ParaView. I changed this field name to "attribute" to be more consistent with MFEM terminology.

tangqi (approved):

tangqi:
This looks pretty good to me. I also like changing the name to "attribute". I was confused when I first saw "material id" in ParaView.

benzwick (commented):
I tested this with a few different examples (serial and parallel) and the output looks good to me.

My only question is whether the subdirectory for the parallel output is really necessary? I would prefer to have the files named as e.g:
```
ex1p_bdr.pvtu
ex1p_bdr.proc000000.vtu
ex1p_bdr.proc000001.vtu
```
in the top level directory to match more closely the output from serial runs, e.g:
```
ex1_bdr.vtu
```
As far as I know the output from `pmesh->Print(mesh_ofs)` is similar to this (it does not create a new directory).

tangqi:
I personally would actually prefer to have a directory for ParaView. I am routinely running my applications with thousands of cores and many time steps. I use ParPrint to output checkpoints for pmesh and gridfunctions, which is already in the current directory. 

Ideally, I would like to have a folder for checkpoint, a folder for ParaView or Visit files. But at least the ParaView files are all collected in one folder now.

benzwick:
> I personally would actually prefer to have a directory for ParaView

Is it not possible to do this manually by creating the directory as part of the code and adding the prefix to the filename? This would be more flexible than a predefined folder that cannot be changed. I tried this approach and it appears to work for the serial `.vtu` file and the parallel `proc...vtu` files but not the `.pvtu` file.

I agree that it would probably be preferable to have the `proc...vtu` files in a directory but it might also be a good idea to prefix them with the file name (e.g. `myfilename_proc000001.vtu` with matching `myfilename.pvtu` instead of just `proc000001.vtu`) otherwise there will be many files with the same name (albeit in different directories).

tzanio:
Thanks @benzwick and @tangqi!

jandrej (approved):

jandrej:
LGTM

pazner:
Thanks for the feedback. I'm open to changing the way the files and directories are handled.

Personally, I prefer creating a directory in the parallel case since otherwise things get quite cluttered, especially when running with many MPI ranks. This is also similar to how data collections work in MFEM.

All in all, I think it's better not to require the user to explicitly create directories, since filesystem handling in C++ can be a bit obscure (less so with `std::filesystem` library, but that is only C++17 and above).

One possible option would be to allow something like `pmesh.PrintVTU("pmesh_dir/pmesh_file");` that would automatically create the directory `pmsh_dir`, and then use `pmesh_file` as the filename prefix. This currently doesn't work, but I can fix it.

rcarson3 (approved):
Nice addition @pazner!

rcarson3:
So, I definitely prefer to keep things saved off in directories pretty much in line with what @tangqi has already said above. Lots of processors and cycles definitely causes a ton of clutter if they weren't in directories.

pazner:
> I agree that it would probably be preferable to have the `proc...vtu` files in a directory but it might also be a good idea to prefix them with the file name (e.g. `myfilename_proc000001.vtu` with matching `myfilename.pvtu` instead of just `proc000001.vtu`) otherwise there will be many files with the same name (albeit in different directories).

Good suggestion, I implemented this.

After discussions with the MFEM team, we decided to keep the output files in a directory. The relevant code and Doxygen comments have been improved/clarified.

tzanio:

benzwick:
This works perfectly now with the following code to output the mesh and boundary in the single directory `ex14output` (with subdirectories and file extensions created automatically):
```
pmesh->PrintVTU("ex14output/ex14p_mesh");
pmesh->PrintBdrVTU("ex14output/ex14p_bdr");
```
Thanks @pazner for this very useful addition!

benzwick (approved):

PR #1826, opened 2020-10-20
Absolute value function 'abs' when argument is of floating point type [tbnlf]

camierjs:
Removes `using integer absolute value function 'abs' when argument is of floating point type` warning.

| [#1826](https://github.com/mfem/mfem/pull/1826) | @camierjs | @tzanio | @tzanio + @jandrej | 10/20/20 | 10/20/20 | 10/21/20 | |

### DIFF
--- a/tests/unit/fem/test_blocknonlinearform.cpp
+++ b/tests/unit/fem/test_blocknonlinearform.cpp
@@ -154,7 +154,7 @@ TEST_CASE("ParBlockNonlinearForm",
                 << " Expected" << M_PI / 6.0 << " diff=" << (A4 - M_PI / 6.0)
                 << std::endl;
 
-      REQUIRE(abs(A4 - M_PI / 6.0) < 1e-2);
+      REQUIRE(fabs(A4 - M_PI / 6.0) < 1e-2);
 
       delete nf;
       delete pmesh;


tzanio:
Is this the only place in the code where you get a warning about this?

Seems like there are other instances in the code, e.g.

- https://github.com/mfem/mfem/blob/master/miniapps/adjoint/cvsRoberts_ASAi_dns.cpp#L201
- https://github.com/mfem/mfem/blob/master/miniapps/meshing/mesh-optimizer.cpp#L422-L423 (similar in `pmesh-optimizer`)
- https://github.com/mfem/mfem/blob/master/miniapps/meshing/polar-nc.cpp#L231 (similar in `klein-bottle` and `mobius-strip`)
- etc.

camierjs:
`std::abs` is fine but just `abs` is ambiguous.

tzanio:
First one is just `abs` :)

tzanio:
There are others, e.g. https://github.com/mfem/mfem/blob/master/linalg/solvers.cpp#L2275

jandrej:
This looks like a rare case where `<cmath>` or similar is not included which in C++ overloads `abs` with the correct type. E.g. in `solvers.cpp` you can write
```
double x = 1.0;
double abs_x = fabs(x);
abs_x = abs(x);
abs_x = std::abs(x);
```
which will all be the exact same code path most likely returning `__builtin_fabs(__x);`.

Using `fabs` here is correct.

jandrej (approved):

tzanio (approved):

tzanio:

PR #1835, opened 2020-10-23
Scaled Jacobian plot and locations of bad elements [scaled-jacobian-plot-dev]

wnissen:
@tzanio did all the work and gave me the magic to get the center and plot the scaled Jacobian [aka det(J)], this was very helpful for me in finding where I had poor mesh quality.

`mesh-explorer` now prints the id and center of the first 10 inverted elements:
```
--> x
subdivision factor ---> 0
det(J) < 0 = -1.59692e-09 in element 23684, centered at: 0.0746917 0.015 9.93346
det(J) < 0 = -6.23481e-09 in element 23688, centered at: 0.0771759 0.015 9.9364
[snip]
det(J) < 0 = -1.00524e-08 in element 23700, centered at: 0.0839758 0.015 9.94575
det(J) < 0 for 157 more elements not printed.

bad elements = 167  --  167 Cube(s)
min det(J)   = -4.94199e-08
max det(J)   = 2.16668e-07
global ratio = -4.38422
max el ratio = 559.434
min kappa    = 16.312
max kappa    = 96606.2
```
And plots the scaled (unitless) Jacobian values:
image
image

| [#1835](https://github.com/mfem/mfem/pull/1835) | @wnissen | @tzanio | @vladotomov + @kmittal2 + @tzanio + @v-dobrev | 10/24/20 | 10/29/20 | 11/03/20 | |


tzanio:
Math note:

- In 2D the scaled Jacobian for `J` with columns `v1` and `v2` is `sin ϕ`, where `ϕ` is the angle between the Jacobian columns, `ϕ = ∠(v1,v2)`. 

- In 3D, for `J = (v1 v2 v3)` the scaled Jacobian is `sin ϕ sin θ`, where `ϕ = ∠(v1,v2)` and `θ = ∠(v3, plane-of-v1-v2)`.

In both cases, if `det J >= 0` the scaled Jacobian is between 0 and 1, with 0 corresponding to a flat element, and 1 to orthogonal Jacobian columns.

v-dobrev (approved):

tzanio (approved):
I'm approving just to signify that I'm not planning any further changes to this.

kmittal2 (approved):
Looks good to me.

tzanio:
@vladotomov, do you want to take a quick look?

vladotomov (approved):

tzanio:

v-dobrev:
Merged in post v4.2 `next`.

tzanio:

tzanio:
Thank you Walt!

sshiraiwa:
Relating this, is there are way to "fix" a mesh when there is negative J element? I was guessing mesh-optimizer may be for it and trying various options. But, could not figure out how....Thank you for your advice.

tzanio:
There are various "untangling" approaches in Mesh Optimizer, but how well they work will depend on the problem. @vladotomov and @kmittal2 will know more.

sshiraiwa:
Thank you @tzanio. 

I am attaching one of mesh file I am trying to "fix". I am guessing mid=352 is the one to use. I tried various target_id, but couldn't find the one which works. I tend to see the maximum of metric value increases ( = det (J) becomes even more negative) after the
iteration. Also, sometimes I saw the warning saying that preconditioner is not positive-definite. 

[fs.msh.txt](https://github.com/mfem/mfem/files/5510714/fs.msh.txt)

vladotomov:
Thanks for the mesh @sshiraiwa, we'll run some tests and get back to you this week.

vladotomov:
Hi @sshiraiwa,

The best result we could get so far is [SS_bc_qo4.mesh.txt](https://github.com/mfem/mfem/files/5540257/SS_bc_qo4.mesh.txt), where two of the tets next to the holes are still inverted. 

This was generated by
`mpirun -np 6 pmesh-optimizer -m SS_bc.mesh -o 2 -rs 0 -mid 313 -tid 1 -ni 100 -ls 2 -li 50 -bnd -qt 1 -qo 4 -vl 1 -fd`
in the branch `tmop-inv-example`, which is still WIP. The `SS_bc.mesh` has the correct boundary attributes to allow tangential relaxation, except inside holes (which causes the issue).

We'll keep working on it next week and will let you know how it goes.

Vladimir

vladotomov:
Hi @sshiraiwa,

Here's a fully untangled version of your mesh [SS_bc_Q3_qo4.mesh.txt](https://github.com/mfem/mfem/files/5550694/SS_bc_Q3_qo4.mesh.txt).

We had to represent it as a 3rd order mesh, as the 2nd order didn't have enough DOFs to move in the restricted elements around the channels (10 nodes per element, 6 of which were fixed for `Q2`).

The mesh can be generated in the branch `tmop-inv-example` by 
`mpirun -np 6 pmesh-optimizer -m SS_bc.mesh -o 3 -rs 0 -mid 313 -tid 1 -ni 100 -ls 2 -li 50 -bnd -qt 1 -qo 4 -vl 1 -fd -no-vis`

I hope the mesh is useful,
Vladimir

tzanio:
Nice work  @vladotomov!

wnissen:
OK, that is pretty amazing. I haven't seen the `pmesh-optimizer` tool before. Are these parameters highly mesh-specific, or would they potentially work for many types of meshes? How much does it change the problem geometry? I see you have a bunch of examples in the source at http://mfem.github.io/doxygen/html/pmesh-optimizer_8cpp_source.html

vladotomov:
The parameters are not highly mesh specific, but some of them must be adjusted to the type of optimization we want to achieve. In this case we had a tangled 3D mesh and we wanted to get the best possible shape, hence we used a 3D untangling metric with ideal targets, ie `-mid 313 -tid 1`.

There is no change in the geometry -- all initial boundaries are preserved exactly. In this case the solver moved some boundary nodes tangentially (the ones on the straight boundaries), and the internal nodes to untangle around the channels. We have several options to preserve mesh features or adapt to solution features, but those were not used in this problem.

Screen Shot 2020-11-17 at 11 04 39 AM

sshiraiwa:
@vladotomov Fantastic. This is very very promising. Can you please explain the difference of tid=1, 2, and 3. It is labeled as IDEAL_SHAPE_UNIT_SIZE, IDEAL_SHAPE_EQUAL_SIZE, and IDEAL_SHAPE_GIVEN_SIZE. But not quite clear what these mean. 

Also, I have another question. -bnd option uses boundary attribute from 1-4 to enforce the tangential adjustment. Is it correct that the current constraints work only when the surface normal is orienting to a particular direction. How difficult do you think it is to extend it for a general curved surface,?

vladotomov:
IDEAL_SHAPE - this part means that the target shape is a regular tetrahedron.

- IDEAL_SHAPE_UNIT_SIZE - we should probably rename this to UNIT_SHAPE, because it is *not* supposed to be used for volume optimization. It's supposed to be combined with *shape*-only metrics.
- IDEAL_SHAPE_EQUAL_SIZE - this will try to produces tets with equal volume throughout the mesh.
- IDEAL_SHAPE_GIVEN_SIZE - this will try to produce tets that have the same volume as they originally had (this was named IDEAL_SHAPE_INITIAL_SIZE at some point).

The last two options are combined with metrics that see both *shape* and *size*.

For now we can do tangential relaxation only parallel to the axes (we force Dirichlet BC for the displacements). We are exploring methods for general surfaces, but in principle this will always produce some geometric error. We're approaching it as an algebraic method based on level sets, but we don't have anything to show yet.

PR #1842, opened 2020-10-26
Constraint solvers and examples [barker29/normal-bc]

barker29:
This PR introduces an abstract `ConstrainedSolver` object for solving problems of the form `A x = b` subject to the constraint `B x = r` where `A` is expected to be a usual square finite element operator and `B` is an operator with few rows that expresses a constraint, for example that the normal component of a vector at the boundary should be zero.

We currently have three implementations of `ConstrainedSolver`:
- `SchurConstrainedSolver` solves the saddle point problem and has the most general interface, so that it can be used for any abstract operators, in parallel, with no restrictions. It is also generally the least scalable and efficient. 
- `PenaltyConstrainedSolver` forms a penalized system. It needs `A` and `B` to be assembled matrices, but given that it also works in parallel for general constraints, and there is at least a chance that  preconditioner for `A` will also work for the penalized system.
- `EliminationSolver` requires that the constraints come in small invertible blocks that do not cross processor boundaries. Given that requirement, it is often the fastest option.

There are two examples in the PR:
- `ex-normal` computes a (global) L2 projection of a constant vector field with the constraint that the normal component along some (general, non-grid-aligned, curved) boundary is zero, testing the iteration counts and speed of the three solver variants. This example is just for reviewers to see and mess with, not intended to merge into master.
- `ex-sliding` is like `ex2p` but with normal displacements constrained to zero on two of the trapezoid's four surfaces, so that the object slides against but does not penetrate the surfaces it is resting on. This example is intended to be numbered and added as an example in `master`.

| [#1842](https://github.com/mfem/mfem/pull/1842) | @barker29 | @tzanio | @bslazarov + @jamiebramwell + @rcarson3 | 02/03/21 | 03/07/21 | 03/17/21 | |


barker29:
A sphere example:
sphereo2

linalg/constrained.hpp
```diff
@@ -0,0 +1,272 @@
...
+    subject to the constraint \f$ B x = r \f$ in an abstract setting.
+
+    Although this object may not use the below formulation, for
+    understanding some of its methods and notation you can think of
+    it as solving the saddle-point system
+
+     (  A   B^T  )  ( x )         (  f  )
+     (  B        )  ( lambda)  =  (  r  )
+
+     Not to be confused with ConstrainedOperator, which is totally
```
jamiebramwell, line 196: I have OPINIONS about that name! 😄

linalg/constrained.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+       @param prec Preconditioner for primal block.
+   */
+   void SetSchur(Solver& prec);
+
+   /** @brief Set the right-hand side r for the constraint B x = r
+
+       (r defaults to zero if you don't call this)
+   */
+   void SetDualRHS(const Vector& r);
```
jamiebramwell, line 216: Maybe `SetConstraintRHS`?

linalg/constrained.hpp
```diff
@@ -0,0 +1,272 @@
...
+       columns will be inverted (or approximately inverted in a
+       preconditioner, depending on options) to eliminate the constraint.
+
+       @todo this can be done with only secondary_dofs given in interface
+   */
+   void SetElimination(Array<int>& primary_dofs,
+                       Array<int>& secondary_dofs);
+
+   /** @brief Set up a penalty solver. */
+   void SetPenalty(double penalty);
```
jamiebramwell, line 231: I know jerome has brought this up, but it's possible that this penalty term could be different for every row (or even a sparse matrix itself). I think it would be fairly easy to expand on this interface for that option though.

linalg/constrained.hpp
```diff
@@ -0,0 +1,272 @@
...
+       need to set that with SetDualRHS(). Similarly, the output x is only
+       for the primal system, while if you want the Lagrange multiplier
+       solution you call GetDualSolution() after the solve. */
+   void Mult(const Vector& f, Vector& x) const;
+
+   /** @brief Return the Lagrange multiplier solution in lambda
+
+       Does not make sense unless you've already solved the constrained
+       system with Mult() */
+   void GetDualSolution(Vector& lambda) { lambda = dual_sol; }
```
jamiebramwell, line 246: Maybe `GetMultiplierSolution`?

linalg/constrained.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+     (  A   B^T  )  ( x )         (  f  )
+     (  B        )  ( lambda)  =  (  r  )
+
+     Not to be confused with ConstrainedOperator, which is totally
+     different. */
+class ConstrainedSolver : public Solver
+{
+public:
+   ConstrainedSolver(Operator& A, Operator& B);
```
jamiebramwell, line 201: Sam has some ideas on local constraint transformations instead of a global operator B. I think this API could still work for that, as long as B is just a generic operator. We should discuss this at some point.

tests/unit/linalg/test_constrainedsolver.cpp
```diff
@@ -0,0 +1,347 @@
...
+   }
+   else
+   {
+      Blocal.Add(0, 2*rank + 1, 1.0);
+      Blocal.Add(0, 2*rank + 2, 1.0);
+   }
+   Blocal.Finalize();
+   int row_starts_c[2] = { rank, rank + 1 };
+   int col_starts[2] = { 2*rank, 2 * (rank + 1) };
+   bmat = new HypreParMatrix(MPI_COMM_WORLD, 1, 4, 8, Blocal.GetI(),
```
jamiebramwell, line 229: I assume there's no way around a user explicitly constructing a HypreParMatrix in the parallel case. I can't think of a good way out of this, but I'm open to any ideas.

linalg/constrained.hpp
barker29: Done.

linalg/constrained.hpp
barker29: Done.

linalg/constrained.hpp
barker29: Yes we should!

tests/unit/linalg/test_constrainedsolver.cpp
barker29: The `SchurConstrainedSolver` takes an abstract `Operator` here, so if the user has something else in mind that's possible. But of course the more abstract the interface, the less we can do for preconditioning. It also may be possible to add a layer so the user provides (for example) a contact surface and we build the constraint `HypreParMatrix`.

barker29:
Thanks for looking at this @jamiebramwell ; I reorganized the interface a bit since you looked at it but I don't think it changes anything major.

jamiebramwell:
Yep! The overall interface looks great!

barker29:
Some very basic scaling results (iteration counts) for three different constraint solvers for a vector diffusion problem on `square-disc-p3.mesh` with the normal component of the solution constrained to zero on the curved boundary:

|refine | schur | elimination | pen=1.e+4 | pen=1e+6|
|- | - | - | - | -|
|1 | 43 | 11 | 12 | 63|
|2 | 77 | 13 | 11 | 64|
|3 | 100 | 19 | 9 | 53|
|4 | 176 | 25 | 9 | 34|

It is not clear the lower penalty is actually getting the right answer.

examples/ex-sliding.cpp
```diff
@@ -0,0 +1,523 @@
...
+    constraints that need to be imposed.
+
+    Each row of the returned matrix corresponds to a node that is
+    constrained. The rows are arranged in (contiguous) blocks corresponding
+    to the actual constraint; in 3D, a one-row constraint means the node
+    is free to move along a plane, a two-row constraint means it is free
+    to move along a line (eg the intersection of two normal-constrained
+    planes), and a three-row constraint is fully constrained (equivalent
+    to MFEM's usual essential boundary conditions). The lagrange_rowstarts
+    array is filled in to describe the structure of these constraints.
```
rcarson3, line 49: So, I'm having a bit of hard time understanding what is exactly is returned by lagrange_rowstarts. Are you returning the dofs that are being constrained or is it something else? Or maybe do you have a super tiny single 2D element example of what's going on?

examples/ex-sliding.cpp
```diff
@@ -0,0 +1,523 @@
...
+      }
+   }
+
+   // reorder so constraints eliminated together are grouped
+   // together in rows
+   // (this is perhaps not the best way to organize things)
+   std::map<int, int> reorder_rows;
+   int new_row = 0;
+   MFEM_VERIFY(lagrange_rowstarts.Size() == 0,
+               "rowstarts must be empty to start!");
```
rcarson3, line 119: If people are doing this process multiple times let's say for changing BCs over time you could maybe use the `Array::DeleteAll()` function to enforce this.

examples/ex-sliding.cpp
barker29: I've tried to improve the docs, let me know what you think. Thanks for taking a look!

.gitignore
```diff
@@ -50,6 +50,8 @@ examples/ex1[04-9]
 examples/ex1[0-9]p
 examples/ex2[0-9]
 examples/ex2[0-9]p
+examples/ex-normal
+examples/ex-sliding
```
tzanio, line 54: I think you can claim Example 28 and Example 29 at this point 😁

examples/ex-normal.cpp
```diff
@@ -0,0 +1,436 @@
+//                       MFEM Example normal-bc - Parallel Version
```
tzanio, line 1: Could there be a serial version too?

linalg/linalg.hpp
```diff
@@ -29,6 +29,7 @@
 #include "solvers.hpp"
 #include "handle.hpp"
 #include "invariants.hpp"
+#include "constrained.hpp"
```
tzanio, line 32: I'd prefer `constraints.hpp`

tests/unit/linalg/test_constrainedsolver.cpp
```diff
@@ -0,0 +1,813 @@
+// Copyright (c) 2010-2020, Lawrence Livermore National Security, LLC. Produced
```
tzanio, line 1: Here and in the other new files

```suggestion
// Copyright (c) 2010-2021, Lawrence Livermore National Security, LLC. Produced
```

linalg/constrained.hpp
```diff
@@ -0,0 +1,404 @@
...
+    different.
+
+    The height and width of this object as an IterativeSolver are for
+    the above saddle point system, but one can use its PrimalMult() method
+    to solve just \f$ Ax = f \f$ subject to a constraint with \f$ r \f$
+    defaulting to zero or set via SetConstraintRHS().
+
+    This abstract object unifies handling of the "dual" rhs \f$ r \f$
+    and the Lagrange multiplier solution \f$ \lambda \f$, so that derived
+    classses can reuse that code. */
```
tzanio, line 45: Nice documentation!

.gitignore
barker29: I made `ex-sliding` into `ex28`, but I am not sure `ex-normal` will end up as an example at all (it uses a mesh not in the main repo and IMO does not demonstrate a capability that's very interesting by itself). It's included right now so reviewers can mess with it.

examples/ex-normal.cpp
barker29: `ex-sliding` now has serial and parallel version `ex28` and `ex28p`. Note the serial solver in this case is pretty bad (no hypre).

linalg/linalg.hpp
barker29: Do you want the object changed to `ConstraintsSolver` or just the filename?

tzanio: No, I like `ConstrainedSolver` :)

tzanio: This reminds me that there may be some confusion with `ConstrainedOperator`... not sure what to do about thta

jamiebramwell: To throw my hat in the ring, I've never really liked the existing `ConstrainedOperator` name because most (well, at least me) people will assume it does this functionality.

tzanio: It has propagated through the code, but maybe it is not too late to rename it. What do you think will be a better name?

The description (in `linalg/operator.hpp`) is:

```c++
/** @brief Square Operator for imposing essential boundary conditions using only
    the action, Mult(), of a given unconstrained Operator.

    Square operator constrained by fixing certain entries in the solution to
    given "essential boundary condition" values. This class is used by the
    general, matrix-free system formulation of Operator::FormLinearSystem. */
class ConstrainedOperator : public Operator
```

tzanio:
Travis is failing

linalg/linalg.hpp
jamiebramwell: Hmm, I think the problem comes down to the fact that a solver is an operator, so semantically a `ConstrainedSolver` should be equivalent to a `ConstrainedOperator` (which it isn't). Maybe `FixedBoundaryOperator`? That's clunky, I'll think about this some more.

v-dobrev: Maybe we can extend `ConstrainedOperator` to support more general constraints, e.g. with different constructors. Then it represents the action of the forward constrained operator and the `ConstrainedSolver` class is a base class for solvers for `ConstrainedOperator`s.

linalg/constraints.hpp
```diff
@@ -0,0 +1,404 @@
...
+    subject to the constraint \f$ B x = r \f$.
+
+    Although implementations may not use the below formulation, for
+    understanding some of its methods and notation you can think of
+    it as solving the saddle-point system
+
+     (  A   B^T  )  ( x )          (  f  )
+     (  B        )  ( lambda )  =  (  r  )
+
+    Not to be confused with ConstrainedOperator, which is totally
```
jamiebramwell, line 35: Somewhere, we should include some documentation about how these two similarly named things are different.

linalg/constraints.hpp
```diff
@@ -0,0 +1,404 @@
...
+    different.
+
+    The height and width of this object as an IterativeSolver are for
```
jamiebramwell, line 38: As many times people don't care about the lagrange multipliers, I would have the default `Mult` and on the primal variables and a possible extra `fullMult` or something like that for the multipliers.

linalg/constraints.hpp
```diff
@@ -0,0 +1,404 @@
...
+
+
+/** @brief Perform elimination of a single constraint.
+
+    See EliminationProjection, EliminationCGSolver
+
+    This keeps track of primary / secondary tdofs and does small dense block
+    solves to eliminate constraints from a global system.
+
+    \f$ B_s^{-1} \f$ maps lagrange space into secondary displacements,
```
jamiebramwell, line 109: This function should reference abstract dofs instead of displacements.

linalg/constraints.hpp
```diff
@@ -0,0 +1,404 @@
...
+public:
+   /** @brief Constructor, with explicit splitting into primary/secondary dofs.
+
+       This constructor uses a single elimination block (per processor), which
+       provides the most general algorithm but is also not scalable
+
+       The secondary_dofs are eliminated from the system in this algorithm,
+       as they can be written in terms of the primary_dofs. */
+   EliminationSolver(HypreParMatrix& A, SparseMatrix& B,
+                     Array<int>& primary_dofs,
```
jamiebramwell, line 207: I assume these are local ordering given the local nature of this operator (and the fact they're `int`s vs `HYPRE_Int`s, but that should be explicitly stated.

linalg/constraints.hpp
```diff
@@ -0,0 +1,404 @@
...
+                     Array<int>& secondary_dofs);
+
+   /** @brief Constructor, elimination is by blocks.
+
+       The nonzeros in B are assumed to be in disjoint rows and columns; the
+       rows are identified with the lagrange_rowstarts array, the secondary
+       dofs are assumed to be the first nonzeros in the rows. */
+   EliminationSolver(HypreParMatrix& A, SparseMatrix& B,
```
jamiebramwell, line 215: This interface seems error prone given the implicit ordering assumption. What is gained by using this instead of the more explicit one above?

linalg/constraints.hpp
```diff
@@ -0,0 +1,404 @@
...
+
+/** @brief Solve constrained system with penalty method; see ConstrainedSolver.
+
+    Only approximates the solution, better approximation with higher penalty,
+    but with higher penalty the preconditioner is less effective. */
+class PenaltyConstrainedSolver : public ConstrainedSolver
+{
+public:
+   PenaltyConstrainedSolver(HypreParMatrix& A, SparseMatrix& B,
+                            double penalty_);
```
jamiebramwell, line 276: While I think it's goes beyond the scope of this initial PR, do you think it would be difficult to extend it to use different penalties for each constraint? In other words, the penalty parameter would be a vector instead of a double.

linalg/constraints.hpp
```diff
@@ -0,0 +1,404 @@
...
+   virtual IterativeSolver* BuildKrylov() const = 0;
+
+   double penalty;
+   Operator& constraintB;
+   HypreParMatrix * penalized_mat;
+   mutable Solver * prec;
+};
+
+
+/** Uses CG and a HypreBoomerAMG preconditioner for the penalized system. */
```
jamiebramwell, line 300: Could the iterative solvers and preconditioners be parameters instead of hard-coded?

examples/ex28p.cpp
```diff
@@ -0,0 +1,522 @@
...
+    constrained. The rows are arranged in (contiguous) blocks corresponding
+    to the actual constraint; in 3D, a one-row constraint means the node
+    is free to move along a plane, a two-row constraint means it is free
+    to move along a line (eg the intersection of two normal-constrained
+    planes), and a three-row constraint is fully constrained (equivalent
+    to MFEM's usual essential boundary conditions).
+
+    The lagrange_rowstarts array is filled in to describe the structure of
+    these constraints, so that constraint k is encoded in rows
+    lagrange_rowstarts[k] to lagrange_rowstarts[k + 1] - 1, inclusive,
```
jamiebramwell, line 53: I think using `constraint_rowstart` vs `lagrange_rowstart` would better describe what these operations are doing throughout the code.

examples/ex28p.cpp
```diff
@@ -0,0 +1,522 @@
...
+   Array<int> constraint_atts(2);
+   constraint_atts[0] = 1;  // attribute 1 bottom
+   constraint_atts[1] = 4;  // attribute 4 left side
+
+   ParLinearForm *b = new ParLinearForm(fespace);
+   b->AddBoundaryIntegrator(new VectorBoundaryLFIntegrator(f));
+   if (myid == 0)
+   {
+      cout << "r.h.s. ... " << flush;
+   }
```
rcarson3, line 394: Is this line really needed here?

examples/ex28p.cpp
```diff
@@ -0,0 +1,522 @@
...
+      push_force = 0.0;
+      push_force(1) = -5.0e-2; // index 1 attribute 2
+      f.Set(0, new PWConstCoefficient(push_force));
+   }
+
+   // 10. Set up constraint matrix to constrain normal displacement (but
+   //     allow tangential displacement) on specified boundaries.
+   Array<int> constraint_atts(2);
+   constraint_atts[0] = 1;  // attribute 1 bottom
+   constraint_atts[1] = 4;  // attribute 4 left side
```
rcarson3, line 387: Can this be moved down to where the `BuildNormalConstraints` function is actually called?

rcarson3 (commented):
I've only gotten a chance to go through part of the PR so far. I'll try and go through the actual `constraints.hpp` and `constraints.cpp` later in the week.

linalg/constraints.hpp
barker29: I first implemented it that way, but switched to this because it was used (in Smith) as the linear solver in a `NewtonSolver` object, which will always call with the `Mult()` method and has the saddle-point structure. I think that is probably the most expected use case?

linalg/constraints.hpp
barker29: That should be straightforward.

linalg/constraints.hpp
barker29: What I intend here is a real C++ interface where a user who wants different solvers and preconditioners implements another class (lines 300-331 are the entire implementation). For example `dimension` and `reorder` only make sense here for systems AMG - other solver/preconditioner combinations will also have their own particular options, and I don't want to handle all of those in a single class.

examples/ex28p.cpp
barker29: It's there in `ex2p` which I am mindlessly copying - I mean, which I am inspired by and closely following. I'm inclined to leave it for consistency but will happily remove it if reviewers want.

linalg/constraints.hpp
barker29: The other constructor eliminates the entire B matrix as one block constraint and is not scalable in matrix size. This one eliminates each (block) constraint individually, if they are small it is much more efficient. An alternative interface would take an `Array` of `Array`s or a similar data structure, so each constraint could be in arbitrary rows, but I'm not sure that's a better interface.

I've tried to improve the documentation to make this clearer, but it could probably be improved further. I'm also open to suggestions on the interfaces.

examples/ex28p.cpp
```diff
@@ -0,0 +1,522 @@
...
+    @param[in] constrained_att      Boundary attributes to constrain
+    @param[out] lagrange_rowstarts  The rowstarts for separately
+                                    eliminated constraints, possible
+                                    input to EliminationCGSolver
+
+    @return a constraint matrix
+
+    @todo use FiniteElementSpace instead of ParFiniteElementSpace, but
+    we need tdofs in parallel case. */
+SparseMatrix * BuildNormalConstraints(ParFiniteElementSpace& fespace,
```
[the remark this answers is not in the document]
barker29: This is done. I needed this push to do it right (ie, with the right understanding between serial/parallel numbering) so thanks for making me do it.

examples/ex28p.cpp
```diff
@@ -0,0 +1,522 @@
...
+      cout << "Size of linear system: " << A.GetGlobalNumRows() << endl;
+   }
+
+   // 13. Define and apply a parallel PCG solver for the constrained system
+   //     where the normal boundary constraints have been separately eliminated
+   //     from the system.
+   Array<int> lagrange_rowstarts;
+   SparseMatrix* local_constraints =
+      BuildNormalConstraints(*fespace, constraint_atts, lagrange_rowstarts);
+   EliminationCGSolver * solver = new EliminationCGSolver(A, *local_constraints,
```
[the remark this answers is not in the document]
barker29: I've added a `--penalty` option so you can use it here - again, this revealed a bug in the penalty solver in parallel so thanks for making me do it...

linalg/constraints.hpp
barker29: It it is in local truedof numbering, which is now documented.

examples/ex28p.cpp
```diff
@@ -0,0 +1,522 @@
...
+               int constraint = dof_constraint[truek];
+               int row = constraints[constraint][att];
+               for (int d = 0; d < dim; ++d)
+               {
+                  int inner_vdof = fespace.DofToVDof(k, d);
+                  int inner_truek = fespace.GetLocalTDofNumber(inner_vdof);
+                  // an arguably better algorithm does some kind of average
+                  // instead of just overwriting when two elements (with
+                  // potentially different normals) share a node.
+                  out->Set(row, inner_truek, nor[d]);
```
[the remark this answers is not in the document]
barker29: I implemented the averaging algorithm and moved this to the library. The averaging algorithm is a bit ugly and might be a memory hog for large problems - I'm not sure how much to think of this as a demonstration or as a real production capability.

barker29:
I *think* I have addressed all of the outstanding review comments at this point.

linalg/constraints.hpp
jamiebramwell: Hmmm. I'll think about this. When I wear my program hat, I think defaulting to only primal variables is more natural. When I wear my smith hat, I can see it the other way as you pointed out. Any thoughts @bslazarov @rcarson3 ?

examples/ex28p.cpp
jamiebramwell: Yeah, I think it's unnecessary noise and should be removed.

linalg/linalg.hpp
jamiebramwell: Any thoughts on if the file should be named `constrained_solver.hpp` to avoid confusion?

linalg/constraints.cpp
```diff
@@ -0,0 +1,893 @@
...
+   which MFEM sometimes calls a "vdof" - note that MFEM's naming conventions
+   regarding this are not entirely consistent. In parallel, this always
+   returns the "truedof" in parallel numbering. */
+int CanonicalNodeNumber(FiniteElementSpace& fespace,
+                        int node, bool parallel, int d=0)
+{
+#ifdef MFEM_USE_MPI
+   if (parallel)
+   {
+      try
```
jamiebramwell, line 700: Does MFEM allow try/catch in main code? A quick grep shows they only occur in zstr and adios files. If not, you should be able to get around this by casting to a pointer instead of a reference.

linalg/constraints.cpp
```diff
@@ -0,0 +1,893 @@
...
+   workb.MakeRef(const_cast<Vector&>(f_and_r), 0);
+   workx.MakeRef(x_and_lambda, 0);
+   Vector ref_constraint_rhs(f_and_r.GetData() + A.Height(), B.Height());
+   constraint_rhs = ref_constraint_rhs;
+   PrimalMult(workb, workx);
+   Vector ref_constraint_sol(x_and_lambda.GetData() + A.Height(), B.Height());
+   GetMultiplierSolution(ref_constraint_sol);
+}
+
+/* Helper routine to reduce code duplication - given a node (which MFEM
```
jamiebramwell, line 689: Thank you for the clear explanation!

linalg/constraints.cpp
```diff
@@ -0,0 +1,893 @@
...
+   sometimes calls a "dof"), this returns what normal people call a dof but
+   which MFEM sometimes calls a "vdof" - note that MFEM's naming conventions
+   regarding this are not entirely consistent. In parallel, this always
+   returns the "truedof" in parallel numbering. */
+int CanonicalNodeNumber(FiniteElementSpace& fespace,
```
jamiebramwell, line 694: This is so handy and not specific to constraints I wonder if it should go in `fespace.hpp`

jamiebramwell (approved):
Looks great! Thank you for putting this together. I'm approving it for now with the understanding that the consensus for using try/catch and moving a the canonical dof function to a common spot will be applied.

linalg/constraints.hpp
```diff
@@ -0,0 +1,462 @@
...
+public:
+   Eliminator(const SparseMatrix& B, const Array<int>& lagrange_dofs,
+              const Array<int>& primary_tdofs,
+              const Array<int>& secondary_tdofs);
+
+   const Array<int>& LagrangeDofs() const { return lagrange_tdofs; }
+   const Array<int>& PrimaryDofs() const { return primary_tdofs; }
+   const Array<int>& SecondaryDofs() const { return secondary_tdofs; }
+
+   /// Given primary displacements, return secondary displacements
```
rcarson3, line 125: You still use displacement within this class to describe the abstract dofs.

linalg/constraints.hpp
```diff
@@ -0,0 +1,462 @@
...
+   /** @brief Assemble this projector as a (processor-local) SparseMatrix.
+
+       Some day we may also want to try approximate variants. */
+   SparseMatrix * AssembleExact() const;
+
+   /** Given Lagrange multiplier right-hand-side \f$ g \f$, return
+       \f$ \tilde{g} \f$ */
+   void BuildGTilde(const Vector& g, Vector& gtilde) const;
+
+   /** After a solve, recover the Lagrange multiplier. */
```
rcarson3, line 182: It appears you still refer to a few variables here as displacement variables rather than the abstract dofs that Jamie had mentioned earlier in a comment.

linalg/constraints.cpp
```diff
@@ -0,0 +1,893 @@
...
+
+   Bp.SetSize(lagrange_tdofs.Size(), primary_tdofs.Size());
+   B.GetSubMatrix(lagrange_tdofs, primary_tdofs, Bp);
+
+   Bs.SetSize(lagrange_tdofs.Size(), secondary_tdofs.Size());
+   B.GetSubMatrix(lagrange_tdofs, secondary_tdofs, Bs);
+   BsT.Transpose(Bs);
+
+   ipiv.SetSize(Bs.Height());
+   Bsinverse.data = Bs.GetData();
```
rcarson3, line 40: In order to reduce potential issues for those using GPU code with these files, I would suggest for the time changing (where possible) all of these `GetData` in this file to either `HostRead()`, `HostReadWrite()`, or `HostWrite()` just so we know for sure the data is on the host and in-sync with everything. It should also make things easier on later on when trying to convert this over to run on the GPU.

linalg/constraints.cpp
```diff
@@ -0,0 +1,893 @@
...
+   projector = new EliminationProjection(hA, eliminators);
+}
+
+void EliminationSolver::PrimalMult(const Vector& rhs, Vector& sol) const
+{
+   if (!prec)
+   {
+      prec = BuildPreconditioner();
+   }
+   IterativeSolver * krylov = BuildKrylov();
```
rcarson3, line 315: It seems to me that it makes sense to only build the krylov solver once and then keep it alive on the class. I noticed you did this a couple of different places, and my suggestion would be change it in those other places as well.

Assuming these set operations aren't dynamic, you could probably also get away with only do them once as well.

linalg/constraints.cpp
```diff
@@ -0,0 +1,893 @@
...
+      }
+   }
+   else
+#endif
+   {
+      return fespace.DofToVDof(node, d);
+   }
+}
+
+SparseMatrix * BuildNormalConstraints(FiniteElementSpace& fespace,
```
rcarson3, line 720: Thanks for updating the comments within here. It's much easier to follow what's going within the code now :)

rcarson3 (commented):
Just a few more comments here and there but overall it's looking much better and I think almost there.

examples/ex28p.cpp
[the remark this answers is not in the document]
rcarson3: Thanks and I think what you have in the new version is fine for now. If it ends up being a bottle-neck, we can look at more efficient algorithms.

linalg/constraints.hpp
rcarson3: So from an application developer point of view, I personally would probably only care about the primal variables in order to advance my simulations. 

Although, I can't say I normally think of setting up my problems in this saddle-point kinda of way, so I can't say how much use I would have for the lagrange multipliers after the fact.

linalg/constraints.cpp
barker29: I changed it, it makes sense not to introduce this if it's not used.

rcarson3 (approved):
Overall, you've addressed my various concerns/comments I've had about this. I'm approving this, but I still think we still need to come to some conclusion in relation to @jamiebramwell 's earlier comment about in regards to whether the default Mult should just return the primal variable solution or if it should also include the lambda terms as well.

barker29:
Another option I've been reluctant to introduce: the `ConstrainedSolver` object could have a user-specified "mode" that determines whether the `Mult()` operation acts on just the primal variables/displacements or includes lagrange multipliers. I'm reluctant to do this because I think it may introduce a lot of complexity and confusion ("what mode is my operator in? what does that even mean?"), but perhaps it's the best option.

linalg/constraints.hpp
```diff
@@ -0,0 +1,473 @@
...
+#ifdef MFEM_USE_MPI
+/** @brief Basic saddle-point solver with assembled blocks (ie, the
+    operators are assembled HypreParMatrix objects.)
+
+    This uses a block-diagonal preconditioner that approximates
+    \f$ [ A^{-1} 0; 0 (B A^{-1} B^T)^{-1} ] \f$.
+
+    In the top-left block, we approximate \f$ A^{-1} \f$ with HypreBoomerAMG.
+    In the bottom-right, we approximate \f$ A^{-1} \f$ with the inverse of the
+    diagonal of \f$ A \f$, assemble \f$ B D^{-1} B^T \f$, and use
```
jandrej, line 401: D^{-1} = diag(A)^{-1}? If yes, could it be beneficial to do an l1 diagonal inverse?

linalg/constraints.cpp
```diff
@@ -0,0 +1,906 @@
...
+   sol += rtilde;
+}
+
+void PenaltyConstrainedSolver::Initialize(HypreParMatrix& A, HypreParMatrix& B)
+{
+   HypreParMatrix * hBT = B.Transpose();
+   HypreParMatrix * hBTB = ParMult(hBT, &B, true);
+   // this matrix doesn't get cleanly deleted?
+   // (hypre comm pkg)
+   (*hBTB) *= penalty;
```
jandrej, line 364: is this still an issue?

jandrej (approved):
I like the interface a lot. All choices seem reasonable. I think it is a good way to introduce the concept and we will see what limitations we run into.

rcarson3:
> Another option I've been reluctant to introduce: the `ConstrainedSolver` object could have a user-specified "mode" that determines whether the `Mult()` operation acts on just the primal variables/displacements or includes lagrange multipliers. I'm reluctant to do this because I think it may introduce a lot of complexity and confusion ("what mode is my operator in? what does that even mean?"), but perhaps it's the best option.

I would prefer if we could avoid the mode option as well, since as you put it that path seems like it could lead to confusion and potential errors down the road. It might be worth just having a short discussion / webex with some of the users/projects that will use this feature and decide on what makes sense for a default option and just make sure that it's well documented.

linalg/constraints.hpp
barker29: Maybe? This is a very easy change to make if you want me to, but I haven't tried it.

linalg/constraints.cpp
barker29: Yes, see #2035.

v-dobrev:
Another possibility is to say the `ConstrainedSolver` acts on the primal variables only and if you want an operator on both primal+multipliers then you can get that as another `Operator` with a method, e.g. `ConstrainedSolver::GetFullOperator` that returns a simple wrapper object around `ConstrainedSolver` that transfers calls to `Mult` to `FullMult`. That way both can exist independently and at the same time.

linalg/constraints.hpp
jandrej: It just came to my mind in the context of if its useful expose a custom preconditioning operator

liangcheng:
From a user perspective, I created a simple 3D model in MFEM mesh format, two cubes touch each other on a quarter of faces. Tet4 elements and tri3 boundaries. The attribute numbers are not starting from 0 and incremental, but it could be easily changed in this small model. I haven't gotten a chance to build and run the normal-bc branch, but will try it maybe tomorrow.

'
MFEM mesh v1.0

dimension
3

elements
12
11 4 2 1 7 3
11 4 1 4 7 5
11 4 4 7 2 0
11 4 2 7 1 0
11 4 1 7 4 0
11 4 4 2 7 6
12 4 10 9 15 11
12 4 9 12 15 13
12 4 12 15 10 8
12 4 10 15 9 8
12 4 9 15 12 8
12 4 12 10 15 14

boundary
8
-17 2 11 15 9
-17 2 15 13 9
17 2 2 0 4
17 2 4 6 2
22 2 3 7 1
22 2 7 5 1
23 2 10 8 12
23 2 12 14 10

vertices
16
3
1 1 1
1 1 0
1 0 1
1 0 0
0 1 1
0 1 0
0 0 1
0 0 0
1.5 1.5 2
1.5 1.5 1
1.5 0.5 2
1.5 0.5 1
0.5 1.5 2
0.5 1.5 1
0.5 0.5 2
0.5 0.5 1
'

liangcheng:
The ex28 solved on my computer (Win10+VS2017). Thanks for bringing MFEM and this great feature to our entire simulation community. I will try to implement my MPC routines early next month, with the ideas of ex28 and ConstrainedSolver. 

My following question is what is the best strategy to acquire the bonded nodes? The traditional FEM packages use the EQUATION or CE keywords with nodes and coefficients in the mesh file or input scripts, this method has been proved effective. Or do I need to write a contact detecting routine to find out the nodes and matrix B for the given pairs prior to solve? Since MFEM has AMR and Metis, as a newcomer, I am not quite sure the best way at this moment. Any suggestion? Thanks.

tzanio:
I'll merge this to start testing it, but lets discuss the interface and any outstanding issues at the PR meeting

tzanio:

tzanio:
LGTM

tzanio:
`ex28` artifacts are not properly listed in `.gitignore`:

```
?? examples/ex28_000000.mfem_root
?? examples/ex28_000000/
```

barker29:
@liangcheng For specifying constraints, at this point we are expecting the user to provide them, so I'm afraid you will have to write your own contact-detecting routine or use some other library.

liangcheng:
@barker29 I got it. Thank you.

v-dobrev:
I merged `master` here to see if the Gitlab CI diffs will go away.

v-dobrev:
LGTM

v-dobrev:
Re-merged in `next`.

v-dobrev:
Re-merged in `next`.

tzanio:
Travis is failing, can you take a look?

tzanio:

tzanio:
Can you please merge `master` in here and resolve the conflicts.

barker29:
This PR previously had an example that simply projected a vector field onto a space with zero normal components on the boundary, via a mass-matrix solve constrained by the methods introduced here. We removed that example before merging but you can browse it here: https://github.com/mfem/mfem/blob/2ff0021fcf696cb76853b95b146ace9f9368e7b4/examples/ex-normal.cpp

For reference you can find it at commit `2ff0021fcf696` in `examples/ex-normal.cpp`.

tzanio (approved):
This looks awesome, thanks @barker29!

Can you update `CHANGELOG` and remember tom document also in the `mfem/web` repo?

PR #1845, opened 2020-10-27
Remove FE spaces from Multigrid

chakshinglee:
I am writing a solver (#1538) that uses multigrid. However, the multigrid in the solver does NOT have a simple hierarchy of finite element spaces. Since I do want to make use of the current `Multigrid` implementation in MFEM, I am moving things around with the hope that the original setting is essentially preserved while it allows a multigrid to be constructed without `FiniteElementSpaceHierarchy`.

The main changes are:
1. Define `GeometricMultigrid`, which is derived from `Multigrid`. Move the members of `Multigrid` concerning finite elements to `GeometricMultigrid`.
2. Add a constructor of `Multgrid` that takes all the essential components (operators, smoothers, prolongations), as well as their ownerships as input.

Feel free to make suggestions to the class names.

| [#1845](https://github.com/mfem/mfem/pull/1845) | @chakshinglee | @tzanio | @barker29 + @drzisga + @delyank | 11/08/20 | 11/16/20 | 12/24/20 | |


delyank:
It's a simple and desirable PR. Even though it's simple, it modifies the interface and that will have impact on some codes that use the `Multigrid` class directly, which would need to make a simple switch to using `GeometricMultigrid` instead. I'll look at it a bit more before approving it. The only thing that I find unpleasant is the "ownership" situation and these associated arrays of boolean flags as arguments to the constructor. Let's think if we can improve on that somehow at least a bit.

barker29:
Do you want to add the`ready-for-review` label so this gets official deadlines for reviewing/merging, or are you looking for more informal feedback? (In my view it's pretty much ready for a formal review - I almost added the label myself.)

chakshinglee:
@barker29 Yes, it is ready for review, thanks for the reminder :)

barker29 (approved):
I think this looks good. I agree with @delyank that the `Array`s of bools for ownership is not ideal, but since we already have this structure for operators and smoothers, I actually think it makes sense to also have it for prolongations.

delyank:
@barker29, good point. Constancy is quite important. This is a good enough reason for me.

delyank (approved):

tzanio:
@drzisga, do you want to review this?

drzisga (approved):
Sorry it took so long. I really like the decoupling of `Multigrid` and `GeometricMultigrid`. The changes look good to me.

tzanio:

tzanio:
I have some concerns about this PR and will like to get more time to review it.

tzanio (approved):
@chakshinglee, @barker29, @drzisga and @delyank -- Please accept my apologies for the delay with my review 🙏 

I thought I had some concerns and I wanted to take a closer look, but I wasn't able to do it until now. Fortunately, I am actually OK with the PR as is (with minor editing) and will merge it now.

Thanks for adding this and for your patience!

PR #1879, opened 2020-11-16
Action of HypreSmoother::MultTranspose

chakshinglee:
The goal of this PR is to allow `HypreSmoother::MultTranspose` to be defined when the smoother is symmetric.

Specifically, it adds an `enum` type `TransposeAction` in `HypreSmoother`, a member `transpose_action`, and a public member function to set the value of `transpose_action` (default is `Undefined`). When `transpose_action` is set to be `SameAsMult` (e.g., when the user knows the smoother is symmetric), calling `MultTranspose` will be redirected to `Mult`. 

Later on one may add another option (something like `BasedOnType`) in `TransposeAction` if a specific action of transpose (that is different from `Mult`) is defined for certain types of smoothers.

I also thought about a simpler option (for achieving the goal of this PR), which is to simply use a boolean flag (something like `is_symmetric`), and when the flag is `true`, `Transpose` will just call `Mult`. In this case, again, the default value would be `false`. But I am a bit worried that this may potentially be misleading (for example, if I initialize a Jacobi smoother, by default the flag would still indicate that it is not symmetric).

| [#1879](https://github.com/mfem/mfem/pull/1879) | @chakshinglee | @tzanio | @barker29 + @tzanio + @v-dobrev | 11/16/20 | 01/07/21 | 01/14/21 | |

### DIFF
--- a/linalg/hypre.hpp
+++ b/linalg/hypre.hpp
@@ -591,6 +591,10 @@ void EliminateBC(HypreParMatrix &A, HypreParMatrix &Ae,
 /// Parallel smoothers in hypre
 class HypreSmoother : public Solver
 {
+public:
+   /// Options of what MultTranspose would do when it is called.
+   enum TransposeAction { Undefined, SameAsMult };
+
 protected:
    /// The linear system matrix
    HypreParMatrix *A;
@@ -638,6 +642,9 @@ protected:
    /// Combined coefficients for windowing and Chebyshev polynomials.
    double* fir_coeffs;
 
+   /// Indicate what MultTranspose would do when it is called.
+   TransposeAction transpose_action;
+
 public:
    /** Hypre smoother types:
        0    = Jacobi
@@ -684,6 +691,12 @@ public:
        entries in the associated matrix. */
    void SetPositiveDiagonal(bool pos = true) { pos_l1_norms = pos; }
 
+   /// Set the flag transpose_action that indicates the action of MultTranspose
+   /** By default, transpose_action is Undefined, in which case MultTranspose
+       should not be called. By setting transpose_action = SameAsMult, calling
+       MultTranspose will be redirected to Mult. */
+   void SetTransposeAction(TransposeAction act) { transpose_action = act; }
+
    /** Set/update the associated operator. Must be called after setting the
        HypreSmoother type and options. */
    virtual void SetOperator(const Operator &op);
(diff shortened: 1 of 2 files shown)


linalg/hypre.hpp
```diff
@@ -638,6 +642,9 @@ class HypreSmoother : public Solver
    /// Combined coefficients for windowing and Chebyshev polynomials.
    double* fir_coeffs;
 
+   /// Indicate what MultTranspose would do when it is called.
+   TransposeAction transpose_action;
+
 public:
    /** Hypre smoother types:
```
barker29, line 649: I suggest having the constructor(s) set the `transpose_action` flag if we know it's symmetric (Jacobi and l1-jacobi for sure, maybe some others.)

barker29 (approved):
This looks good. I think I like the `bool` option better than the currently implemented `enum` but either way is fine. The bool is simpler and if it's well documented I don't think it will cause confusion.

linalg/hypre.hpp
chakshinglee: Good suggestion. In that case, we should also set the flag in `HypreSmoother::SetType( ... )`.

tzanio:
Apologies for my late review.

I agree with @barker29 -- the `is_symmetric` Boolean flag seems more natural to me, and if you want you can set that to `true` for the diagonal types in `HypreSmoother::SetOperator`.

Will you consider refactoring the PR that way?

tzanio (changes requested):
Refactor using a Boolean flag?

chakshinglee:
Yes, I will make an update with Boolean flag instead.

chakshinglee:
I've made some update to the PR. Now `HypreSmoother` has new flag `A_is_symmetric`, which indicates if the linear system A (member of the class) is symmetric. It is `false` by default unless explicitly changed by `HypreSmoother::SetOperatorSymmetry(...)`.

If A is symmetric, then the smoother of any of the currently available types will also be symmetric, so 'MultTranspose' can be redirected to 'Mult'. Even if A is not symmetric, the smoothers of "Jacobi type" are still symmetric, so 'MultTranspose' will still be redirected to 'Mult' in those cases.

tzanio (approved):
Thanks for the contribution @chakshinglee. I like this interface.

barker29 (approved):
Everything looks good from my end (in case this was waiting on me, I sometimes lose track...)

tzanio:

Issue #1887, opened 2020-11-17
VTK prism orientation: example 1 doesn't work with beam-wedge.vtk


pazner:
Running `ex1` with the mesh `beam-wedge.vtk`, I get the following output:

```
% ./ex1 -m ../data/beam-wedge.vtk
Options used:
   --mesh ../data/beam-wedge.vtk
   --order 1
   --no-static-condensation
   --no-partial-assembly
   --device cpu
   --visualization
Device configuration: cpu
Memory configuration: host-std
Elements with wrong orientation: 8 / 8 (NOT FIXED)
Number of finite element unknowns: 19737
Size of linear system: 19737
   Iteration :   0  (B r, r) = -0.010132
PCG: The preconditioner is not positive definite. (Br, r) = -0.010132
```

It seems like this is an orientation issue. The VTK mesh reader does have a permutation for the prism vertex ordering, but perhaps it is not properly fixing the orientations?

mlstowell:
According to the VTK documentation the first three vertices form the base of the wedge and the normal to the base, using the right hand rule, should point away from the opposite face of the wedge.  The documentation agrees with the image:
![35021239-dbb37d7a-fae4-11e7-8534-167726495bd8](https://user-images.githubusercontent.com/6109571/99464176-58138800-28ec-11eb-8aa5-dc9613587796.png)

On the other hand the first element of our vtk file is:
```6 0 9 18 1 10 19```
With nodes 0, 9, and 18 given by:
```
0 0 0
...
0 1 0
...
0 0.5 1
...
```
Consequently, the normal for this face points in the positive x direction which is into the element i.e. towards the opposite face.  It seems to me that our `beam-wedge.vtk` does not contain valid VTK_WEDGE elements.

tzanio:
Yes, can you please fix it @mlstowell?

I don't remember doing it, but GitHub claims that I added this mesh and a few others in https://github.com/mfem/mfem/commit/bdc5415de918862c06085ac65dc765fdbef042a0.

Maybe I opened the `*.mesh` and used `Mesh::PrintVTK()` to save them. Shouldn't that work?

mlstowell:
I'd be happy to.  I'll check `Mesh::PrintVTK` as well.

tzanio:
Thank you Mark!

tzanio:
Discussion continued in #1888.

PR #1888, opened 2020-11-18
Fixing wedge elements in Mesh::PrintVTK and beam-wedge.vtk [vtk-wedge-fix]

mlstowell:
This PR addresses issue #1887 which pointed out that the `beam-wedge.vtk` mesh contains invalid wedge elements.

| [#1888](https://github.com/mfem/mfem/pull/1888) | @mlstowell | @tzanio | @tzanio + @pazner | 11/17/20 | 11/17/20 | 11/20/20 | |


pazner:
Awesome, thanks @mlstowell! I think `Mesh::PrintVTU` should probably be fixed too (around line 9255).

tzanio (approved):
Thanks for the quick fix @mlstowell !

tzanio:
I checked `data/fichera-mixed-p2.vtk` also and that appears to be OK.

mlstowell:
> Awesome, thanks @mlstowell! I think `Mesh::PrintVTU` should probably be fixed too (around line 9255).

Good catch.  I think we'll need to change this in `vtk.cpp`.  I'll need to take a closer look.

pazner:
> Good catch. I think we'll need to change this in `vtk.cpp`. I'll need to take a closer look.

The mappings in `vtk.cpp` are for high-order Lagrange VTK elements, and are (hopefully) already correct. It should be sufficient to do the same permutation as in `Mesh::PrintVTK`.

mlstowell:
> > Good catch. I think we'll need to change this in `vtk.cpp`. I'll need to take a closer look.
> 
> The mappings in `vtk.cpp` are for high-order Lagrange VTK elements, and are (hopefully) already correct. It should be sufficient to do the same permutation as in `Mesh::PrintVTK`.

I see.  Do we have a way to check the output from `Mesh::PrintVTU`?

pazner:
> I see. Do we have a way to check the output from `Mesh::PrintVTU`?

We can load the mesh in ParaView and see if it looks correct (I checked and it does). The only thing is I think the "wrong" ordering still looked fine, so it's not really a comprehensive test. I did look at the ASCII output and verified that the ordering was what we expect.

Now that I am looking into it, the way we save VTU meshes, the vertices are always duplicated between elements, which means that they are not suitable for reading. So maybe this is not so consequential.

I pushed a minor change simplifying the code slightly and introducing `vtk_prism_perm` to avoid manually reproducing this permutation in many places in the code. This same permutation appears in the VTK mesh reader, so I am planning on using this mapping in my work on the high-order VTK reader, and so this seems like a good place to introduce it.

If you prefer your earlier version, feel free to revert my commit.

mlstowell:
> If you prefer your earlier version, feel free to revert my commit.

No, this is a good idea.  Thanks!

pazner (approved):

tzanio (approved):

tzanio:

PR #1894, opened 2020-11-19
XML VTK Support

pazner:
(Note target branch is not master).

Adds support for reading XML VTK meshes (including high-order meshes), using the [TinyXML2](https://github.com/leethomason/tinyxml2) library for parsing the XML files.

The legacy and XML readers are structured so that they load the same data arrays, which are then passed to `Mesh:: CreateVTKMesh`, which handles the logic of mesh creation, node index reordering, etc.


pazner:
@v-dobrev, @tzanio: what is the consensus on including the TinyXML library? Should I merge this PR to the other VTK mesh branch?

tzanio:
Yes, I think we should include it and merge this in `feature/vtk_high_order`.

Can you just mention it at the end of the `NOTICE` file?

tzanio:
can we also delete the branch?

PR #1904, opened 2020-11-24
AmgX configuration method

artv3:
This PR adds a configuration method when using non-default preconditioner/solver modes.

If AmgX is configured as a preconditioner the check for convergence is skipped. 

Used to make adjustments in the ceed benchmarks: 
https://github.com/CEED/benchmarks/pull/14

| [#1904](https://github.com/mfem/mfem/pull/1904) | @artv3 | @tzanio | @dylan-copeland + @acfisher + @barker29 | 11/30/20 | 12/22/20 | 12/28/20 | |


dylan-copeland (approved):
I ran the unit tests and `amgx/ex1(p)` sample runs on `lassen`, and everything looks fine. I think the `ConfigureAs()` calls could be eliminated just by setting the mode in the constructor (at least for the examples), but it is still good to have the new method.

linalg/amgxsolver.hpp
```diff
@@ -162,6 +162,11 @@ class AmgXSolver : public Solver
    */
    void DefaultParameters(const AMGX_MODE amgxMode_, const bool verbose);
 
+   // Configure as AmgXSolver::PRECONDITIONER or AmgXSolver::SOLVER.
+   // When configuring as a preconditioner, the check for convergence
+   // or divergence will be skipped after applying Mult.
+   void ConfigureAs(const AMGX_MODE amgxMode_);
```
barker29, line 168: In my view this name is wrong, because calling this does not in fact reconfigure the object. The name makes me expect that if I construct with `AmgXSolver(PRECONDITIONER)` and then call `.ConfigureAs(SOLVER)` it will change things, but that's not the case. I suggest adding a `check_diverged` flag or something similar and change this method to `SetConvergenceCheck()`. Then `DefaultParameters()` can still set the new flag based on the mode. Does that make sense?

barker29 (changes requested):
I hate to be picky about this small (and needed) PR, but I do think my suggested change makes it better. Will be happy to approve and see this merged after that's done.

linalg/amgxsolver.hpp
artv3: Yes! That is more clear, I will update it accordingly.

linalg/amgxsolver.cpp
```diff
@@ -30,10 +30,15 @@ int AmgXSolver::count = 0;
 
 AMGX_resources_handle AmgXSolver::rsrc = nullptr;
 
+AmgXSolver::AmgXSolver()
+   : ConvergenceCheck(false) {};
+
 AmgXSolver::AmgXSolver(const AMGX_MODE amgxMode_, const bool verbose)
```
barker29, line 36: I think you need to initialize `ConvergenceCheck` to `false` in every constructor (or have an `else` in the `if amgxMode` line).

linalg/amgxsolver.hpp
```diff
@@ -162,10 +165,8 @@ class AmgXSolver : public Solver
    */
    void DefaultParameters(const AMGX_MODE amgxMode_, const bool verbose);
 
-   // Configure as AmgXSolver::PRECONDITIONER or AmgXSolver::SOLVER.
-   // When configuring as a preconditioner, the check for convergence
-   // or divergence will be skipped after applying Mult.
-   void ConfigureAs(const AMGX_MODE amgxMode_);
+   // Add a check for convergence after applying Mult.
```
barker29, line 168:
```suggestion
   /// Add a check for convergence after applying Mult.
```

barker29 (approved):
Very minor comments, but this looks good to me.

linalg/amgxsolver.cpp
artv3: Ah that is right, fixed!

tzanio:
@acfisher, can you take a quick look?

acfisher (approved):

tzanio:

examples/amgx/ex1.cpp
```diff
@@ -205,10 +205,14 @@ int main(int argc, char *argv[])
 
       if (amgx_solver)
       {
+         amgx.SetConvergenceCheck(true);
          amgx.Mult(B,X);
       }
       else
       {
+         //Omit convergence check at the AmgX level
+         //when using as a preconditioner
```
tzanio: Styling comment: in the mfem source we leave a space after `//` and align on 80 characters 😁

tzanio (approved):
Thanks @artv3 !

Issue #385, opened 2017-12-12
DiscreteLinearOperator issue


stefanozampini:
This issue has been created to keep track of the following problem: this code gives the wrong interpolation operator from L2 to H1

      FiniteElementCollection *H1fec = new H1_FECollection(order, dim);
      ParFiniteElementSpace *fespace = new ParFiniteElementSpace(pmesh, H1fec);
      FiniteElementCollection *l2fec = new L2_FECollection(order-1, dim);
      ParFiniteElementSpace *l2fespace = new ParFiniteElementSpace(pmesh, l2fec);
      ParDiscreteLinearOperator *id;
      id = new ParDiscreteLinearOperator(l2fespace, fespace);
      id->AddDomainInterpolator(new IdentityInterpolator);
      id->Assemble();
      id->Finalize();

 Here is Veselin's reason for the issue

> This is actually a case that is not supported by MFEM. The reason is that, in the “Assemble” method, we need to treat this case differently. Simply overwriting the element matrices in the global matrix (this is what we do right now) does not do the right thing. Instead it leads to multiple input values from neighbor elements adding to shared dofs in the range space. Essentially, the current approach works only if the range space has the same or less overlap than the domain space.

Issue #88, opened 2016-10-04
Boundary element method


jcharris:
Has MFEM ever been used for a boundary element method code?  Would it be difficult for a person to make this modification, or is this not worth the effort?  It appears that it would be a question of rewriting a Bilinear form for the assembly and solve phases, at a minimum?

Have been developing a boundary element code for some years and working to update the higher-order elements, and it would be advantageous to have some of the features that are already developed in MFEM, I wasn't sure if it was easier to write from scratch the needed parts, or if it would be possible to make use of a code like this.

mlstowell:
As far as I know MFEM has not been used for a boundary element code.

I have seen finite element libraries used for boundary element codes in the past and it is possible. The mesh and basis function classes could probably be used with very minor modifications.  A boundary element code would require much more advanced integration rules which could properly handle singular Green's function kernels. Typically, efficiently computing and possibly caching the jacobians is important in a boundary element code so this may need to be added.  Other issues I've seen deal with non-trivial topologies of surface meshes such as three elements meeting at a single edge in a T-shape.  Also, boundary element codes are typically written with complex valued fields whereas MFEM is real valued.

In short, it would require a great deal of specialized code but it should be possible for a subset of boundary element applications at least.

PR #1931, opened 2020-12-08
Broadening support for linear interpolators [test-lin-interp-dev]

mlstowell:
Broaden the testing of the linear interpolators to include the `IdentityInterpolator` and the `Grad`, `Curl`, and `Div` interpolators.  Also expanded these unit tests to validate the relevant operators on various element types.

The impetus for this was the discovery of limitations involving `L2` elements with `INTEGRAL` finite element type and also mapping to vector finite element types from either `ND`, `RT`, or vector versions of `H1` or `L2`.

Support for the various combinations of finite elements when using the `IdentityInterpolator` is still not complete.  Specifically, mapping to `H1^d` or `L2^d` is not currently possible and adding this support seems to be nontrivial. 

| [#1931](https://github.com/mfem/mfem/pull/1931) | @mlstowell | @tzanio | @dylan-copeland + @psocratis | 12/15/20 | 01/04/21 | 01/12/21 | |

### DIFF
--- a/fem/bilininteg.hpp
+++ b/fem/bilininteg.hpp
@@ -3071,6 +3071,22 @@ protected:
    VectorCoefficient *VQ;
 };
 
+/** Interpolator of the 2D cross product between a vector coefficient and an
+    H(curl)-conforming field onto an L2-conforming field. */
+class ScalarCrossProductInterpolator : public DiscreteInterpolator
+{
+public:
+   ScalarCrossProductInterpolator(VectorCoefficient & vc)
+      : VQ(&vc) { }
+
+   virtual void AssembleElementMatrix2(const FiniteElement &nd_fe,
+                                       const FiniteElement &l2_fe,
+                                       ElementTransformation &Trans,
+                                       DenseMatrix &elmat);
+protected:
+   VectorCoefficient *VQ;
+};
+
 /** Interpolator of the cross product between a vector coefficient and an
     H(curl)-conforming field onto an H(div)-conforming field. The range space
     can also be vector L2. */
--- a/fem/fe.cpp
+++ b/fem/fe.cpp
@@ -722,17 +722,34 @@ void NodalFiniteElement::Project(
 {
    if (fe.GetRangeType() == SCALAR)
    {
-      MFEM_ASSERT(map_type == fe.GetMapType(), "");
-
       Vector shape(fe.GetDof());
 
       I.SetSize(dof, fe.GetDof());
-      for (int k = 0; k < dof; k++)
+      if (map_type == fe.GetMapType())
       {
-         fe.CalcShape(Nodes.IntPoint(k), shape);
-         for (int j = 0; j < shape.Size(); j++)
+         for (int k = 0; k < dof; k++)
          {
-            I(k,j) = (fabs(shape(j)) < 1e-12) ? 0.0 : shape(j);
+            fe.CalcShape(Nodes.IntPoint(k), shape);
+            for (int j = 0; j < shape.Size(); j++)
+            {
+               I(k,j) = (fabs(shape(j)) < 1e-12) ? 0.0 : shape(j);
+            }
+         }
+      }
+      else
+      {
+         for (int k = 0; k < dof; k++)
+         {
+            Trans.SetIntPoint(&Nodes.IntPoint(k));
+            fe.CalcPhysShape(Trans, shape);
+            if (map_type == INTEGRAL)
+            {
+               shape *= Trans.Weight();
+            }
+            for (int j = 0; j < shape.Size(); j++)
+            {
+               I(k,j) = (fabs(shape(j)) < 1e-12) ? 0.0 : shape(j);
+            }
          }
       }
    }
--- a/fem/fe.cpp
+++ b/fem/fe.cpp
@@ -1048,6 +1048,8 @@ void VectorFiniteElement::Project_RT(
 
          fe.CalcShape(ip, shape);
          Trans.SetIntPoint(&ip);
+         // Transform RT face normals from reference to physical space
+         // vk = adj(J)^T nk
          Trans.AdjugateJacobian().MultTranspose(nk + d2n[k]*dim, vk);
          if (fe.GetMapType() == INTEGRAL)
          {
@@ -1065,6 +1067,8 @@ void VectorFiniteElement::Project_RT(
             {
                s = 0.0;
             }
+            // Project scalar basis function multiplied by each coordinate
+            // direction onto the transformed face normals
             for (int d = 0; d < sdim; d++)
             {
                I(k,j+d*shape.Size()) = s*vk[d];
@@ -1086,8 +1090,12 @@ void VectorFiniteElement::Project_RT(
          const IntegrationPoint &ip = Nodes.IntPoint(k);
 
          Trans.SetIntPoint(&ip);
+         // Transform RT face normals from reference to physical space
+         // vk = adj(J)^T nk
          Trans.AdjugateJacobian().MultTranspose(nk + d2n[k]*dim, vk);
+         // Compute fe basis functions in physical space
          fe.CalcVShape(Trans, vshape);
+         // Project fe basis functions onto transformed face normals
          vshape.Mult(vk, vshapenk);
          if (!square_J) { vshapenk /= Trans.Weight(); }
          for (int j=0; j<vshapenk.Size(); j++)
@@ -1255,6 +1263,8 @@ void VectorFiniteElement::Project_ND(
 
          fe.CalcShape(ip, shape);
          Trans.SetIntPoint(&ip);
+         // Transform ND edge tengents from reference to physical space
+         // vk = J tk
          Trans.Jacobian().Mult(tk + d2t[k]*dim, vk);
          if (fe.GetMapType() == INTEGRAL)
          {
@@ -1272,6 +1282,8 @@ void VectorFiniteElement::Project_ND(
             {
                s = 0.0;
             }
+            // Project scalar basis function multiplied by each coordinate
+            // direction onto the transformed edge tangents
             for (int d = 0; d < sdim; d++)
             {
                I(k, j + d*shape.Size()) = s*vk[d];
@@ -1292,8 +1304,12 @@ void VectorFiniteElement::Project_ND(
          const IntegrationPoint &ip = Nodes.IntPoint(k);
 
          Trans.SetIntPoint(&ip);
+         // Transform ND edge tangents from reference to physical space
+         // vk = J tk
          Trans.Jacobian().Mult(tk + d2t[k]*dim, vk);
+         // Compute fe basis functions in physical space
          fe.CalcVShape(Trans, vshape);
+         // Project fe basis functions onto transformed edge tangents
          vshape.Mult(vk, vshapetk);
          for (int j=0; j<vshapetk.Size(); j++)
          {
--- a/fem/fe.cpp
+++ b/fem/fe.cpp
@@ -1074,7 +1074,27 @@ void VectorFiniteElement::Project_RT(
    }
    else
    {
-      mfem_error("VectorFiniteElement::Project_RT (fe version)");
+      int sdim = Trans.GetSpaceDim();
+      double vk[Geometry::MaxDim];
+      DenseMatrix vshape(fe.GetDof(), sdim);
+      Vector vshapenk(fe.GetDof());
+      const bool square_J = (dim == sdim);
+
+      I.SetSize(dof, fe.GetDof());
+      for (int k = 0; k < dof; k++)
+      {
+         const IntegrationPoint &ip = Nodes.IntPoint(k);
+
+         Trans.SetIntPoint(&ip);
+         Trans.AdjugateJacobian().MultTranspose(nk + d2n[k]*dim, vk);
+         fe.CalcVShape(Trans, vshape);
+         vshape.Mult(vk, vshapenk);
+         if (!square_J) { vshapenk /= Trans.Weight(); }
+         for (int j=0; j<vshapenk.Size(); j++)
+         {
+            I(k,j) = vshapenk(j);
+         }
+      }
    }
 }
 
@@ -1261,7 +1281,25 @@ void VectorFiniteElement::Project_ND(
    }
    else
    {
-      mfem_error("VectorFiniteElement::Project_ND (fe version)");
+      int sdim = Trans.GetSpaceDim();
+      double vk[Geometry::MaxDim];
+      DenseMatrix vshape(fe.GetDof(), sdim);
+      Vector vshapetk(fe.GetDof());
+
+      I.SetSize(dof, fe.GetDof());
+      for (int k = 0; k < dof; k++)
+      {
+         const IntegrationPoint &ip = Nodes.IntPoint(k);
+
+         Trans.SetIntPoint(&ip);
+         Trans.Jacobian().Mult(tk + d2t[k]*dim, vk);
+         fe.CalcVShape(Trans, vshape);
+         vshape.Mult(vk, vshapetk);
+         for (int j=0; j<vshapetk.Size(); j++)
+         {
+            I(k, j) = vshapetk(j);
+         }
+      }
    }
 }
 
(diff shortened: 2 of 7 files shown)


fem/fe.cpp
```diff
@@ -1057,7 +1074,27 @@ void VectorFiniteElement::Project_RT(
    }
    else
    {
-      mfem_error("VectorFiniteElement::Project_RT (fe version)");
```
dylan-copeland, line 1060: It is not easy to follow the Project* functions, with their different cases (based on map type, square_J, etc.). Before this PR, there was a lack of documentation, so this is not a new issue. It would help me understand things for this review, and others in general, if there were some documentation on the transformations being used and the definition of `I` in the different cases. Some comments about the meaning of `d2n` and `nk` would also be helpful (otherwise, one has to look up the constructor for a particular `FiniteElement` to see what they are).

tests/unit/fem/test_lin_interp.cpp
```diff
@@ -59,164 +96,1226 @@ double FdotG3(const Vector & x)
...
+              }
+         */
+      }
+   }
+}
+
+TEST_CASE("3D Identity Linear Interpolators",
+          "[IdentityInterpolator]")
+{
+   int order_h1 = 1, order_nd = 2, order_rt = 1, order_l2 = 1, n = 3, dim = 3;
```
dylan-copeland, line 507: It looks like there is a lot of code duplication between this case and the 2D one. The main differences seem to be the declaration of `mesh` and the loop over elements. It is common in unit tests to have a loop over dimensions and a switch for the `mesh` declaration. Specifying the element loop in 2D and 3D could be done by setting the lower and upper bounds for the loop as variables. It would also be possible to put this in a function called by the `TEST_CASE` for each dimension and element type (one test case for 2D and 3D). Optionally, if you want to be really fancy, you could use `GENERATE`, cf. test_pa_kernels.cpp.

tests/unit/fem/test_lin_interp.cpp
```diff
@@ -59,164 +96,1226 @@ double FdotG3(const Vector & x)
...
+      Mesh mesh(n, (Element::Type)type, 1, 2.0);
+
+      SECTION("Operators on H1 for element type " + std::to_string(type))
+      {
+         H1_FECollection    fec_h1(order_h1, dim);
+         FiniteElementSpace fespace_h1(&mesh, &fec_h1);
+
+         GridFunction f0(&fespace_h1);
+         f0.ProjectCoefficient(fCoef);
+
```
psocratis, line 119: Just for consistency with the rest of the tests, testing the H1(order)->H1(order+1) embedding seems appropriate to be added here

fem/fe.cpp
psocratis: I agree with @dylan-copeland about the documentation. As it is mentioned, I too don't think this is something that needs to be addressed in this PR.

tests/unit/fem/test_lin_interp.cpp
psocratis: I see @dylan-copeland 's point, the test code could be a bit shorter, but since these are just unit tests, though longer, they are easily readable. I am fine either way.

tests/unit/fem/test_lin_interp.cpp
```diff
@@ -59,164 +96,1226 @@ double FdotG3(const Vector & x)
...
+            GridFunction f0p(&fespace_h1p);
+
+            DiscreteLinearOperator Op(&fespace_h1,&fespace_h1p);
+            Op.AddDomainInterpolator(new IdentityInterpolator());
+            Op.Assemble();
+
+            Op.Mult(f0,f0p);
+
+            REQUIRE( f0p.ComputeL2Error(fCoef) < tol );
```
psocratis, line 191: Here and in some other cases, it would be more appropriate to check the H1/Hcurl/Hdiv error instead of the L2.

psocratis (approved):
This looks very good overall. All unit tests seem fine too. One thing I'd like to point out is that the unit tests only cover one of the two use cases of the `IdentityInterpolator` (i.e, the embedding but not the projection. I guess testing the other way would be more elaborate (checking rates etc.) so maybe this could be done in a separate test (or we just skip it).

tests/unit/fem/test_lin_interp.cpp
mlstowell: I agree with you, @dylan-copeland, I had considered performing a loop over the dimension but the function coefficients stopped me.  In most cases different coefficients are needed in different dimensions.  This could be solved in some cases by making the functions aware of the dimension, replacing some coefficients by pointers to coefficients (and pointing to different things based on dimension), or using conditionals when using coefficients in yet other cases.  I chose not to proceed with these approaches so that I could keep the flow of these tests easier to follow.

Clearly there are competing priorities here and perhaps I've let these get too long.  I'd be happy to try for a compromise if you like.

tests/unit/fem/test_lin_interp.cpp
mlstowell: OK, this sounds good.  Should we uses these measures whenever the target spaces allow them or should we be more selective?

psocratis: I am thinking using the norm induced by the range space. This will also require a small change in 2D Hcurl problems where the CurlCoeff will have to be changed from Scalar to Vector of dim = 1.

tests/unit/fem/test_lin_interp.cpp
dylan-copeland: @mlstowell I did not notice the difference in function coefficients. As you mentioned, this could be simplified by making the functions aware of the dimension. This is done, for example, in test_pa_coeff.cpp. It seems like a simple change to greatly reduce the amount of code. On one hand, it may be a little easier to follow with separate 2D and 3D versions, but on the other hand it is easier to maintain and update the tests in the future if code is not duplicated, so I am in favor of reducing to one test for 2D/3D with loops and coefficient functions that switch based on the dimension.

mlstowell: This is a good point.  Coefficients are more of a hassle when derivatives and products with material coefficients are involved.  For these tests the coefficients should be much easier to handle.  I'll take a look at collapsing these tests into loops.

fem/fe.cpp
mlstowell: @dylan-copeland and @psocratis , I added short comments to both the header and source.  Hopefully they'll make the project methods easier to follow.  Let me know if we need more.

tests/unit/fem/test_lin_interp.cpp
mlstowell: Collapsing into a loop cut nearly 400 lines of code and I don't think it's any more difficult to follow.

tests/unit/fem/test_lin_interp.cpp
mlstowell: This has been fixed as a consequence of collapsing different dimensional cases into a larger loop.

tests/unit/fem/test_lin_interp.cpp
mlstowell: I think I found them all.  This was a good idea.  Thanks, @psocratis !

fem/fe.cpp
```diff
@@ -722,17 +722,34 @@ void NodalFiniteElement::Project(
 {
    if (fe.GetRangeType() == SCALAR)
    {
-      MFEM_ASSERT(map_type == fe.GetMapType(), "");
```
dylan-copeland, line 725: In the non-scalar (vector) case, what happens if `map_type != fe.GetMapType()`? Is that case supported, or does there need to be an assert?

mlstowell: In the vector case `fe.GetMapType()` will either be H_CURL or H_DIV whereas `map_type` will be either VALUE or INTEGRAL.  So, these are guaranteed to be different.  We handle both H_CURL and H_DIV by calling `fe.CalcVShape` with the `ElementTransformation` argument which returns shape functions transformed to the physical space where the map type distinction doesn't matter.  The VALUE or INTEGRAL type for the nodal finite element basis functions is handled by scaling with the Jacobian determinant on line 767.  In short, all possibilities are handled but I grant you that this is far from obvious.

fem/fe.cpp
```diff
@@ -1057,7 +1078,31 @@ void VectorFiniteElement::Project_RT(
...
+      I.SetSize(dof, fe.GetDof());
+      for (int k = 0; k < dof; k++)
+      {
+         const IntegrationPoint &ip = Nodes.IntPoint(k);
+
+         Trans.SetIntPoint(&ip);
+         // Transform RT face normals from reference to physical space
+         // vk = adj(J)^T nk
+         Trans.AdjugateJacobian().MultTranspose(nk + d2n[k]*dim, vk);
+         // Compute fe basis functions in physical space
```
dylan-copeland, line 1096: Reference space?

mlstowell: Meaning the coordinate system used in the reference elements e.g. the unit cube with one vertex at the origin and the opposite vertex at (1,1,1) is the reference element for all Hexahedra in MFEM.

fem/fe.cpp
```diff
@@ -1057,7 +1078,31 @@ void VectorFiniteElement::Project_RT(
...
+         fe.CalcVShape(Trans, vshape);
+         // Project fe basis functions onto transformed face normals
+         vshape.Mult(vk, vshapenk);
+         if (!square_J) { vshapenk /= Trans.Weight(); }
```
dylan-copeland, line 1100: Has this been tested in the non-square case, e.g. 2D mesh in 3D space? It would be helpful to document this weight (what is the transformation and why is the weight needed only in the square case; and note that the `Weight()` is `det` in the square case, something else in the non-square case).

fem/fe.hpp
```diff
@@ -860,6 +901,19 @@ class VectorFiniteElement : public FiniteElement
...
+       @param fe    Vector-valued finite element basis
+       @param Trans Transformation from reference to physical coordinates
+       @param I     Expansion coefficients for the approximation of each basis
+                    function
+
+       Note: If the FiniteElement, fe, is scalar-valued the projection will
+             assume that a FiniteElementSpace is being used to define a vector
+             field using the scalar basis functions for each component of the
+             vector field.
+   */
```
dylan-copeland, line 916: Thanks for adding such informative documentation!

tests/unit/fem/test_lin_interp.cpp
```diff
@@ -59,164 +106,851 @@ double FdotG3(const Vector & x)
...
+
+   FunctionCoefficient       fCoef(f3);
+   VectorFunctionCoefficient FCoef(dim, F3);
+
+   VectorFunctionCoefficient GradfCoef(dim, Grad_f3);
+   VectorFunctionCoefficient CurlFCoef(dim, CurlF3);
+   FunctionCoefficient       DivFCoef(DivF3);
+
+   for (int type = (int)Element::TETRAHEDRON;
+        type <= (int)Element::HEXAHEDRON; type++)
```
dylan-copeland, line 681: Could the 1D-3D cases be collapsed into this loop like the other test?

fem/fe.cpp
dylan-copeland: Right, but I mean should this comment be `Compute fe basis functions in reference space`?

fem/fe.cpp
mlstowell: This is a good question.  I'm not aware of any places where we test the support of curves in 2D or 3D and surfaces in 3D.  We should certainly add unit tests for these cases but that's beyond the scope of this PR.

I'm not sure of the best place to document this (nor am I certain I fully understand it).  We should probably document `DenseMatrix::CalcAdjugate`, `DenseMatrix::Weight`, and `ElementTransformation::AdjugateJacobian` but people would still have questions.

I'm tempted to kick this can down the road if you don't mind.

dylan-copeland: Ok, we should document this in the future, and I agree the documentation needs to be in several places beyond this function. I have used 2D spaces in 3D in my DD code (on interfaces), so I figured this out at one point (but don't recall all the details now). Maybe I will review it soon and try to document everything.

acfisher (approved):

tests/unit/fem/test_lin_interp.cpp
mlstowell: Yes, it could.

mlstowell:
As long as we're on a roll I added 1D and 2D product interpolator tests (including a new scalar cross product in 2D).

dylan-copeland (approved):

tzanio:

fem/bilininteg.hpp
```diff
@@ -3071,6 +3071,22 @@ class VectorScalarProductInterpolator : public DiscreteInterpolator
    VectorCoefficient *VQ;
 };
 
+/** Interpolator of the 2D cross product between a vector coefficient and an
+    H(curl)-conforming field onto an L2-conforming field. */
+class ScalarCrossProductInterpolator : public DiscreteInterpolator
```
tzanio, line 3076: Add to the documentation in https://mfem.org/lininterp/?

fem/fe.hpp
```diff
@@ -827,6 +841,19 @@ class VectorFiniteElement : public FiniteElement
...
```
tzanio, line 856: Thanks for adding this @mlstowell !

fem/gridfunc.cpp
```diff
@@ -462,15 +462,26 @@ const
    fes->GetElementDofs(i, dofs);
    fes->DofsToVDofs(vdim-1, dofs);
    const FiniteElement *FElem = fes->GetFE(i);
-   MFEM_ASSERT(FElem->GetMapType() == FiniteElement::VALUE,
-               "invalid FE map type");
    int dof = FElem->GetDof();
    Vector DofVal(dof), loc_data(dof);
    GetSubVector(dofs, loc_data);
-   for (int k = 0; k < n; k++)
+   if (FElem->GetMapType() == FiniteElement::VALUE)
```
tzanio, line 468: Does this address https://github.com/mfem/mfem/issues/1037?

tzanio:
> Specifically, mapping to H1^d or L2^d is not currently possible and adding this support seems to be nontrivial.

Is this https://github.com/mfem/mfem/issues/385?

tzanio:
Should we document in `CHANGELOG`?

fem/gridfunc.cpp
mlstowell: Partially.  This PR extended `GridFunction::GetValue`.  I added a commit today which addresses the limitation in `CurlInterpolator`.

Issue #1037 also mentions a problem in the `VectorFECurlIntegrator` (and possibly others).  I'll try to address these in a separate PR.

fem/bilininteg.hpp
mlstowell: Will do, thanks!

mlstowell:
> > Specifically, mapping to H1^d or L2^d is not currently possible and adding this support seems to be nontrivial.
> 
> Is this #385?

No, I believe it is separate.  Issue #385 seeks to compute H1 values by interpolating an L2 field.  Strictly speaking, this is not possible for two reasons.  First, L2 fields are discontinuous whereas H1 are continuous so they cannot unambiguously represent an L2 field.  Secondly, if the domain has a boudnary the boundary values would need to be extrapolated.  In either case a simple interpolation will not be possible.

The issue with mapping to H1^d or L2^d mentioned here has to do with the manner in which we implement our interpolators.  Mapping from H1^d, H(Curl), or H(Div) to L2^d should be possible (mathematically).  However, our interpolators use basis functions of the target space to project the basis functions of the source space. The L2^d basis functions are scalar-valued so they cannot correctly interpret vector-valued fields.  We would need to project the components of the vector field onto the L2 basis functions separately and build the appropriate local matrix from the results.  I'm sure this is possible but this is what I referred to as non-trivial.

fem/bilininteg.hpp
mlstowell: Just created PR [#88](https://github.com/mfem/web/pull/88) to add documentation for this (and the enhanced curl interpolator).

tzanio:
Thanks Mark!

fem/bilininteg.hpp
tzanio: Merged, thanks!

PR #1940, opened 2020-12-12
Add wrapper for FormRectangular functions [wrapper-bform-dev]

alecampos2009:
The ```ParBilinearForm``` class has functions that form the linear system and system matrix using as input either an operator handle or an operator type. The ```ParMixedBilinearForm``` class only has functions to form the linear system and system matrix using an operator handle, but not an operator type.

In this PR I've added the wrappers to extend the functionality in ```ParMixedBilinearForm```, so as to be analogous to that in ```ParBilinearForm```.

| [#1940](https://github.com/mfem/mfem/pull/1940) | @alecampos2009 | @tzanio | @cjvogl + @rcarson3 | 01/06/21 | 01/14/21 | 01/24/21 | |


alecampos2009:
@tzanio @cjvogl @homijan  To test these new wrappers, I have temporarily added a couple of lines to ex5p.cpp. This test, however, shows that the new wrappers give a seg fault. I'm not quite sure what the source of the seg fault is, or how to track it down.

examples/ex5p.cpp
```diff
@@ -260,6 +260,9 @@ int main(int argc, char *argv[])
    }
    else
    {
+      HypreParMatrix matB;
+      bVarf->FormRectangularSystemMatrix(empty_tdof_list, empty_tdof_list, matB);
```
cjvogl, line 264: @alecampos2009, if I had to guess, I would suspect as issue from ‘bVarf’ being assembled twice: once on this line as part of the call to ‘FormRectangularSystemMatrix’ and again on line 267 that call ‘ParallelAssemble’.

It might be worth commenting out line 267 and replacing ‘matB’ with ‘B’ on this line

cjvogl, line 264: The segfault occurs on line 267 where `bVarf` is assembled after it is assembled on this line.  The segfault can be bypassed with the following steps:
- replace line 263 with `B = new HypreParMatrix();`
- replace line 264 with `bVarf->FormRectangularSystemMatrix(empty_tdof_list, empty_tdof_list, *B);`
- remove line 267
- change line 264 to `

alecampos2009:
Tested the wrapper using ex5p.cpp, following @cjvogl suggestion. No segfaults are now encountered and results seem equal to those from the original ex5p.cpp. The file ex5p.cpp has now been restored to its original state.

cjvogl (approved):
While generally I would request some code coverage (e.g. inclusion in an example or unit test code), I think it's ok to forgo that considering this is only a namespace inclusion change (and it compiles without issue).

rcarson3 (approved):
Sorry about the delayed review. I don't see anything wrong with these minor changes.

tzanio:

tzanio:
The Travis [failure](https://github.com/mfem/mfem/pull/1940/checks?check_run_id=1716074184) for b090867 was interesting -- it was not due to the code in the PR itself, but rather due to the merging with `master`, which Travis performs automatically.

tzanio:

tzanio (approved):
Thanks for the contribution @alecampos2009!

PR #1951, opened 2020-12-15
Minor bugfixes and improvements

najlkin:
Collection of various bugfixes and minor improvements that would be too small for separate PRs. These are:

- added `AddMultAtB()` and `AddMult_a_AtB()` for dense matrices
- added `ParBilinearForm::ParInnerProduct()` and `ParBilinearForm::TrueInnerProduct()` for calculation of the global inner product (analogously to `BilinearForm::InnerProduct()`)
- fixed `FGMRESSolver` final number of iterations and norm when it has not converged (analogously to other solvers) and its false reporting of non-convergence occasionally
- added `StopWatch` copy constructor making a deep copy
- ~~added parameters setters for ParaSails and documented `SetSymmetry()`~~ documented ParaSails according to [hypre documentation](https://hypre.readthedocs.io/en/latest/solvers-parasails.html)
- improved `BlockOperator` to not perform multiplications when the multiplicative coefficient is zero
- added const qualifier to `Array::IsSorted()`
- fixed `ParGridFunction::ProjectBdrCoefficient()` and `ParGridFunction::ProjectBdrCoefficientTangent()` to set consistent results through all MPI processes (this might be crucial in some cases). Update: this has been fixed in #4151 already, but an alternative approach is offered here.
- fixed `HypreSmoother` for GS type, where an auxiliary array is required (otherwise gives segfaults)
- ~~added const qualifier to `ParGridFunction::SaveAsOne()`~~

| [#1951](https://github.com/mfem/mfem/pull/1951) | @najlkin | @v-dobrev | @kmittal2 + @jakubcerveny + @psocratis | 12/21/20 | 4/24/24 | 4/28/24 | |


najlkin:
I have fixed the Joule solver to work as before, but it is wrong. Inner product is not a linear operation, so the sum of the local contributions is not equal to the global inner product (except DG spaces). I added a note there, saying it is approximate, so if you were aware of this, it is then ok. If not, the code must be further changed.
A similar problem was in example 10p, but I replaced the sum of the local ones by the global one and it seems to be working fine.

v-dobrev:
There seems to be some misunderstanding regarding the use of `BilinearForm::InnerProduct()`, e.g. there is no bug in ex10p as it is in `master`. The inputs to `BilinearForm::InnerProduct()` are assumed to be L-vectors, i.e. `GridFunction`/`ParGridFunction`-size (local size) vectors, so it naturally returns just the local inner product (which corresponds to the integral over the local mesh.

Anyway, we can discuss this in more detail during the review process.

najlkin:
Yes, @v-dobrev , you are right that inner product is linear in the matrices, but not in the vectors, so you have to assume that you do not have very general L-vectors, but prolongations of some T-vectors. However, this is the case in example 10p and Joule as well. I admit, I confused myself a bit ... that happens when you cumulate the patches over the year to give them as Christmas gifts :smile: :gift: . Anyway, I will remove that comment from Joule, but about example 10p I am not sure. It is cleaner this way than a hardcoded MPI reduce, but there is a considerable overhead. Maybe we may do it that way that the local inner product + reduce would be implementation of `ParBilinearForm::InnerProduct(const ParGridFunction &, const ParGridFunction &)`, but my idea was to not assume anything about the grid functions and project/restrict them instead, so I do not know.

linalg/densemat.hpp
```diff
@@ -425,13 +425,20 @@ bool LinearSolve(DenseMatrix& A, double* X, double TOL = 1.e-9);
 /// Matrix matrix multiplication.  A = B * C.
 void Mult(const DenseMatrix &b, const DenseMatrix &c, DenseMatrix &a);
 
+/// Matrix matrix multiplication.  A = B^t * C.
+void MultTranspose(const DenseMatrix &b, const DenseMatrix &c, DenseMatrix &a);
```
psocratis, line 429: This functionality already exists. See 

https://github.com/mfem/mfem/blob/caf5d7b471571b2094ccdfd4b2ff772d76fc778c/linalg/densemat.hpp#L482-L483

fem/pbilinearform.cpp
```diff
@@ -295,6 +295,39 @@ const
...
+double ParBilinearForm::InnerProduct(const ParGridFunction &x,
+                                     const ParGridFunction &y)
+{
+   MFEM_ASSERT(x.ParFESpace() == pfes, "the parallel spaces must agree");
+   MFEM_ASSERT(y.ParFESpace() == pfes, "the parallel spaces must agree");
+
+   HypreParVector *x_p = x.ParallelProject();
+   HypreParVector *y_p = y.ParallelProject();
+
+   double res = InnerProduct(*x_p, *y_p);
```
psocratis, line 307: Something is probably missing here. The intent is to do `x^t A y` right? I don't see the action of the `ParBilinearForm`.

fem/pbilinearform.hpp
```diff
@@ -150,6 +150,16 @@ class ParBilinearForm : public BilinearForm
...
 
+   using BilinearForm::InnerProduct;
+   /// Compute inner product x^t A y (grid function version)
+   /** @note It is assumed that the parallel system matrix is assembled,
+       see FormSystemMatrix(). */
+   double InnerProduct(const ParGridFunction &x, const ParGridFunction &y);
+   /// Compute inner product x^t A y (vector version)
+   /** @note It is assumed that the parallel system matrix is assembled,
+       see FormSystemMatrix(). */
+   double InnerProduct(HypreParVector &x, HypreParVector &y);
```
psocratis, line 161: I have a few comments/objections with regards to these two `InnerProduct` functions. 
1. A `ParGridFunction` operates on `ldofs` therefore it's missleading to actually perform the dot product on true dofs. Maybe a better name would be `TrueInnerProduct`. This however can be done by the existing method `void ParBilinearForm::TrueAddMult`.
2. More importantly, the naming "InnerProduct" implies a positive definite and symmetric form which is not always true.

linalg/densemat.hpp
najlkin: Hmm, I had not noticed, thanks. The interface is somewhat inconsistent in the naming, so I removed the implementation, but I would prefer to keep the function.

linalg/hypre.hpp
```diff
@@ -967,7 +967,15 @@ class HypreParaSails : public HypreSolver
...
 
+   /// Set symmetry parameter
+   /** The recognized options are:
+    *  0 = nonsymmetric and/or indefinite problem, and nonsymmetric preconditioner
+    *  1 = SPD problem, and SPD (factored) preconditioner
+    *  2 = nonsymmetric, definite problem, and SPD (factored) preconditioner
+    */
    void SetSymmetry(int sym);
+   /// Set the logging parameter
+   void SetLogging(int logging);
```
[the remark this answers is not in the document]
najlkin: I agree, but there is not much detail about this in the hypre documentation. Still, I added the short description from the [reference manual](https://computing.llnl.gov/sites/default/files/public/hypre-2.11.2_ref_manual.pdf). In addition, I added the rest of the parameters, but one should be careful with them.

fem/pbilinearform.cpp
najlkin: It calls the new `ParBilinearForm::InnerProduct()` with `HypreParVector` parameters, I believe :wink:

psocratis: > It calls the new `ParBilinearForm::InnerProduct()` with `HypreParVector` parameters, I believe 😉

Yes ... I missed that. Thanks !

fem/pbilinearform.hpp
najlkin: 1. I would not say that `ParGridFunction` solely operates on ldofs. For that, you could use just plain `GridFunction`, but `ParGridFunction` represents a global thing. However, it is assumed most of the time that the grid function is "distributed" through `ParGridFunction::Distribute()`, i.e., it is a prolonged T-vector, so you may easily use the L-vector. Anyway, I think it is probably the first thing one would expect there (but somewhat dangerous), so I would agree with your suggestion. The version based on the local inner product would here instead.
2. Well, you are mathematically right, but it is just a name used throughout MFEM. It is  on user to apply it on appropriate matrices. If you really want, I may add a note there...

fem/pbilinearform.hpp
```diff
@@ -150,6 +150,16 @@ class ParBilinearForm : public BilinearForm
        vectors on the true dofs. */
    void TrueAddMult(const Vector &x, Vector &y, const double a = 1.0) const;
 
+   using BilinearForm::InnerProduct;
```
jakubcerveny, line 153: Do we need this here? `BilinearForm::InnerProduct(const Vector &, const Vector &)` should be accessible without problems, or am I missing something?

fem/pbilinearform.cpp
```diff
@@ -295,6 +295,39 @@ const
    pfes->Dof_TrueDof_Matrix()->MultTranspose(a, Y, 1.0, y);
 }
 
+double ParBilinearForm::InnerProduct(const ParGridFunction &x,
+                                     const ParGridFunction &y)
+{
+   MFEM_ASSERT(x.ParFESpace() == pfes, "the parallel spaces must agree");
```
jakubcerveny, line 312: Maybe use the word "match" instead of "agree"?

examples/ex10p.cpp
```diff
@@ -627,10 +627,7 @@ double HyperelasticOperator::ElasticEnergy(const ParGridFunction &x) const
 
 double HyperelasticOperator::KineticEnergy(const ParGridFunction &v) const
 {
-   double loc_energy = 0.5*M.InnerProduct(v, v);
-   double energy;
-   MPI_Allreduce(&loc_energy, &energy, 1, MPI_DOUBLE, MPI_SUM,
-                 fespace.GetComm());
+   double energy = 0.5*M.InnerProduct(v, v);
```
jakubcerveny, line 630: I am also concerned that there is a hidden `ParallelProject` in the new version, so the new version is not really the same here (?)

examples/ex10p.cpp
```diff
@@ -627,10 +627,7 @@ double HyperelasticOperator::ElasticEnergy(const ParGridFunction &x) const
...
    return energy;
```
jakubcerveny, line 631: An alternative could perhaps be something like `return mfem::ReduceSum(loc_energy, fespace.GetComm())`?

linalg/hypre.hpp
```diff
@@ -967,7 +967,48 @@ class HypreParaSails : public HypreSolver
...
+    * preconditioners are also more expensive per iteration. The default
+    * values are @a thresh = 0.1 and @a nlevels = 1.
+    */
+   void SetParameters(double thresh, int nlevels);
+   /// Set the filter parameter
+   /** The filter parameter is used to drop small nonzeros in the preconditioner,
+    * to reduce the cost of applying the preconditioner. Values from 0.055
+    * to 0.1 are recommended. The default value is 0.1.
+    */
+   void SetFilterParameter(double filter);
```
jakubcerveny, line 984: Naming: maybe `SetFilter` would be sufficient?

linalg/hypre.hpp
```diff
@@ -967,7 +967,48 @@ class HypreParaSails : public HypreSolver
...
+    * is very sparse and fast to construct. The default value is 0.
+    */
+   void SetLoadbalancing(double loadbal);
+   /// Set the pattern reuse parameter
+   /** A nonzero value indicates that the pattern of the preconditioner
+    * should be reused for subsequent constructions of the proconditioner.
+    * A zero value inicates that the peconditioner should be constructed
+    * from scratch. The default value is 0.
+    */
+   void SetReuseParameter(int reuse);
```
jakubcerveny, line 1006: Maybe just `SetReuse`?

linalg/hypre.hpp
```diff
@@ -967,7 +967,48 @@ class HypreParaSails : public HypreSolver
...
+    */
+   void SetFilterParameter(double filter);
+   /// Set symmetry parameter
+   /** The recognized options are:
+    *  0 = nonsymmetric and/or indefinite problem, and nonsymmetric preconditioner
+    *  1 = SPD problem, and SPD (factored) preconditioner
+    *  2 = nonsymmetric, definite problem, and SPD (factored) preconditioner
+    */
    void SetSymmetry(int sym);
+   /// Set the load balance parameter
```
jakubcerveny, line 992: The doc-strings are excellent. Perhaps just add an empty line before -- somehow I keep visually associating the documentation with the preceding method...

fem/pbilinearform.hpp
najlkin: 1. Ok, I just implemented your suggestion, so there is `InnerProduct` and `TrueInnerProduct`, where the user is warned about the necessary consistency of the L-vectors in the former.
2. Regarding the second comment, that is bit tricky. A note in the code about the mathematical requirements on the forms would help beginners definitely, but it would confuse experts, because these properties are not guaranteed on the numerical level. Therefore, I went a way around and followed the description from `BilinearForm`, where it is simply not stated in the description that it is an inner product and just the applied formula appears there. (For consistency, I swapped the arguments.)

examples/ex10p.cpp
najlkin: Solved by distinguishing between `InnerProduct()` and `TrueInnerProduct()`. We just collided in the development/review process :smile: .

examples/ex10p.cpp
najlkin: It is just `InnerProduct()` now, see above :wink:

linalg/hypre.hpp
najlkin: I agree, done.

linalg/hypre.hpp
najlkin: Ok, I preferred the word parameter to indicate that no objects are passed, just parameters, but for consistency with hypre it is maybe better without it.

linalg/hypre.hpp
najlkin: The same as above, but if you wish, I changed that :wink:

fem/pbilinearform.hpp
najlkin: True, it is not needed at the moment.

fem/pbilinearform.cpp
najlkin: Ok, that is maybe better.

fem/pbilinearform.cpp
```diff
@@ -295,6 +295,50 @@ const
...
+}
+
+double ParBilinearForm::TrueInnerProduct(HypreParVector &x,
+                                         HypreParVector &y) const
+{
+   MFEM_VERIFY(p_mat.Ptr() != NULL, "parallel matrix must be assembled");
+
+   HypreParVector *Ax = new HypreParVector(pfes);
+   HypreParMatrix *A = p_mat.As<HypreParMatrix>();
+   A->Mult(x, *Ax);
```
psocratis, line 333: This doesn't really need a  casting to a `HypreParMatrix`. It can simply be `p_mat->Mult(x,*Ax)`

linalg/densemat.hpp
psocratis: I would prefer to have just the existing `MultAtb` and rename AddMultTranspose to `AddMultAtb`, I guess it's a matter of taste, though I would go with the most popular vote.

fem/pbilinearform.hpp
```diff
@@ -150,6 +150,27 @@ class ParBilinearForm : public BilinearForm
...
    void TrueAddMult(const Vector &x, Vector &y, const double a = 1.0) const;
 
+   /// Compute \f$ y^T M x \f$
+   /** @warning The calculation is performed on local dofs, assuming that
+       the local vectors are consistent with the prolongations of the true
+       vectors (see ParGridFunction::Distribute()). If this is not the case,
+       use TrueInnerProduct(const ParGridFunction &, const ParGridFunction &)
+       instead. */
+   double InnerProduct(const ParGridFunction &x,
+                       const ParGridFunction &y) const;
```
psocratis, line 160: Here is an issue I came across while testing this, though this is an existing issue not introduced by this PR. 
`ParBilinearForm::InnerProduct` calls `BilinearForm::InnerProduct` and this in turn calls `SparseMatrix::InnerProduct` for the internal `SparseMatrix * mat`. In case 
 `ParBilinearForm::FormSystemMatrix` was called before, then it segfaults because  `mat` had been deleted. For the same reason,  all `[Par]BilinearForm::[True]InnerProduct` calls cannot work with `-pa`.  I think this shouldn't affect this PR, but it has to be fixed either by adding a check, or we can simply implement it in a separate PR. I'll wait to see what others think.

fem/pgridfunc.cpp
```diff
@@ -594,9 +594,10 @@ void ParGridFunction::ProjectBdrCoefficient(
    // Count the values globally.
    GroupCommunicator &gcomm = pfes->GroupComm();
    gcomm.Reduce<int>(values_counter, GroupCommunicator::Sum);
+   gcomm.Bcast<int>(values_counter);
    // Accumulate the values globally.
    gcomm.Reduce<double>(values, GroupCommunicator::Sum);
-   // Only the values in the master are guaranteed to be correct!
+   gcomm.Bcast<double>(values);
```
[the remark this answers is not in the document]
najlkin: :wink: It had been there for years! I rather do not want to know what consequences it might had.

linalg/densemat.hpp
najlkin: There is a similar thing for sparse matrices `TransposeMult()`, so that is another option :smile: We should really vote about this.

fem/pbilinearform.cpp
najlkin: It does not, but if you look into `HypreParMatrix::Mult(const Vector&, Vector &)`, it constructs `HypreParVector`s by wrapping the T-vectors, so passing `HypreParVector`s directly saves unnecessary overhead.

fem/pbilinearform.hpp
najlkin: I added an assert and a note that the local matrix is assumed to be assembled, because the situation is different compared to `BilinearForm`, where the matrix is not deleted normally and it is a (nearly) clear mistake of the user to call operations on non-assembled form (of course an assert would be better than a segfault, but that is another matter...). In contrast, `ParBilinearForm` "replaces" the local matrix by the parallel during normal operation, when that `FormSystemMatrix()` is called, so it is reasonable to put an assert here.
Regarding operator-based forms (partial/element assembly/matrix free), that is another thing and there are many methods like that in `BilinearForm`, operating only on `SparseMatrix`. This should be solved on the side of `BilinearForm` and gets beyond the scope of this PR.

fem/pbilinearform.cpp
psocratis: I was refering to the line  `HypreParMatrix *A = p_mat.As<HypreParMatrix>();`

najlkin: I too :wink: , but `p_mat` is not a matrix (confusingly), but `OperatorHandle`, so the multiplication would call the general routine `Mult(const Vector&, Vector&)` instead of the specific `Mult(const HypreParVector&, HypreParVector &)`.

psocratis (approved):
This looks good! 
All tests also look good!
Thanks @najlkin

fem/pgridfunc.cpp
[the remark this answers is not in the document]
psocratis: I am sure it was not affecting any actual computations performed on true dofs. But I do see the difference when you try to visualize a `ParGridFunction` right after you project a `FunctionCoefficient` (e.g for an essential BC)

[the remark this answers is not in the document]
najlkin: But it could affect computation on L-dofs like that `InnerProduct` for example.

jakubcerveny (approved):

tzanio:
@v-dobrev, any objections to merge in `next`?

v-dobrev:
Looking briefly at this, I see some discussions that are not resolved yet, so I'll need to look closer.

@psocratis, @jakubcerveny correct me if I'm wrong.

psocratis:
> Looking briefly at this, I see some discussions that are not resolved yet, so I'll need to look closer.
> 
> @psocratis, @jakubcerveny correct me if I'm wrong.

I expressed some of my opinions, mostly on minor things, and I was waiting to hear from others too. Either way, I am good with the current state of this.

najlkin:
Yes, there was just discussion about name of a method, where may vote about it, but i would be fine with any variant, and the second "unresolved conversation" is a proposition for a new PR eventually. However, there is nothing truly unresolved within this PR.

tzanio:
> @tzanio , this one still has not been merged 😔 . I just merged in master, so should be mergeable again 😉 . There is just a slight change of the fix of FGMRES, so it is more consistent with the current state of the code.

ping: @v-dobrev

najlkin:
@v-dobrev , it is waiting for you to put it into `next` (if you are fine with the little change of the FGMRES fix). I am still hopeful it will get after all those years eventually merged... 🙏 😄

najlkin:
I merged in `master` again @v-dobrev , so should be smoothly mergeable again... 😉 Moreover, I just noticed it might help with the long-standing #2226 .

najlkin:
Merged `master`, so it can go to `mfem-4.7` 😉

najlkin:
ping @psocratis

psocratis (approved):

fem/pbilinearform.cpp
```diff
@@ -368,6 +368,53 @@ const
...
+   return res;
+}
+
+real_t ParBilinearForm::TrueInnerProduct(HypreParVector &x,
+                                         HypreParVector &y) const
+{
+   MFEM_VERIFY(p_mat.Ptr() != NULL, "parallel matrix must be assembled");
+
+   HypreParVector *Ax = new HypreParVector(pfes);
+   HypreParMatrix *A = p_mat.As<HypreParMatrix>();
```
v-dobrev, line 408: In some cases `p_mat` may not be a `HypreParMatrix`, see the method `SetOperatorType`. So it is better to use the general `Operator::Mult()` interface, or use a switch on `p_mat.Type()` to handle the case of `HypreParMatrix` and maybe `PetscParMatrix` more efficiently with a fallback to the general interface.

fem/pbilinearform.hpp
```diff
@@ -173,6 +173,30 @@ class ParBilinearForm : public BilinearForm
...
+       @note It is assumed that the local matrix is assembled and it has
+       not been replaced by the parallel matrix through FormSystemMatrix().
+       @see TrueInnerProduct(const ParGridFunction&, const ParGridFunction&) */
+   real_t InnerProduct(const ParGridFunction &x,
+                       const ParGridFunction &y) const;
+
+   /// Compute $ y^T M x $ on true dofs (grid function version)
+   /** @note It is assumed that the parallel system matrix is assembled,
+       see FormSystemMatrix().
+       @see InnerProduct(const ParGridFunction&, const ParGridFunction&) */
```
v-dobrev, line 191: It is good to clarify in the doxygen that the input `ParGridFunction`s are restricted to true-vectors and then used to compute the inner product.

fem/pbilinearform.hpp
```diff
@@ -173,6 +173,30 @@ class ParBilinearForm : public BilinearForm
...
+   real_t TrueInnerProduct(const ParGridFunction &x,
+                           const ParGridFunction &y) const;
+
+   /// Compute $ y^T M x $ on true dofs (vector version)
+   /** @note It is assumed that the parallel system matrix is assembled,
+       see FormSystemMatrix(). */
+   real_t TrueInnerProduct(HypreParVector &x, HypreParVector &y) const;
```
v-dobrev, line 198: We may need a version of this with `const Vector &` arguments to support the case when `p_mat` is not a `HypreParMatrix`.

linalg/densemat.hpp
v-dobrev: I agree with @psocratis, I think the names `MultAtB` and `AddMultAtB` are more precise and so a better choice here. Also, renaming the arguments to `A`, `B`, and `C` (in this order) will make things consistent with the function names.

linalg/hypre.cpp
```diff
@@ -3539,6 +3539,10 @@ void HypreSmoother::SetOperator(const Operator &op)
       ones = 1.0;
       A->Mult(ones, diag);
    }
+   else if (type == 6)
+   {
+      Z = new HypreParVector(*A);
+   }
```
v-dobrev, line 3545: For what version of hypre was this necessary? Looking at v2.29.0, it seems like `Z` is not used for type 6 on CPU, however, it is used on GPU. Was on GPU where you saw segfaults?

linalg/solvers.cpp
```diff
@@ -1317,7 +1317,9 @@ void FGMRESSolver::Mult(const Vector &b, Vector &x) const
       if (v[i]) { delete v[i]; }
       if (z[i]) { delete z[i]; }
    }
-   converged = false;
+
+   final_norm = beta;
+   final_iter = j;
```
v-dobrev, line 1322: Small tweak: if `converged` is `false`, then `j` will be `max_iter+1`, however, we actually want `final_iter` to be `max_iter`.
```suggestion
   final_iter = converged ? j : max_iter;
```

examples/ex10p.cpp
```diff
@@ -646,10 +646,7 @@ real_t HyperelasticOperator::ElasticEnergy(const ParGridFunction &x) const
 
 real_t HyperelasticOperator::KineticEnergy(const ParGridFunction &v) const
 {
-   real_t loc_energy = 0.5*M.InnerProduct(v, v);
-   real_t energy;
-   MPI_Allreduce(&loc_energy, &energy, 1, MPITypeMap<real_t>::mpi_type,
-                 MPI_SUM, fespace.GetComm());
+   real_t energy = 0.5*M.InnerProduct(v, v);
```
v-dobrev, line 649: This kind of change will break user codes since the meaning of `InnerProduct` becomes overloaded and it will call the version that does the global reduction while before there was no global reduction.

A simple solution is to use a different name for the new method, e.g. `ParInnerProduct`.

This same observation affects the change in `examples/sundials/ex10p.cpp` and in `miniapps/electromagnetics/joule_solver.cpp`.

najlkin: It is not exactly overloaded, because it takes `ParGridFuction`s as arguments, but it is true it may break something. It is the kind of thing that makes sense from the design point of view, but it was better to do in the early days and it is hard to go back now 😕 . Ok, will change it... 😉

fem/pbilinearform.hpp
najlkin: That true dofs in the short description say that, but it can be more verbose 😉 .

linalg/densemat.hpp
najlkin: Ok, I wanted it to follow the terminology of the methods of `DenseMatrix`, but this also makes sense 😉 . Besides that we are missing `_a_` version here, so I will add also that one 👍 .

fem/pbilinearform.hpp
najlkin: Done 👇

fem/pbilinearform.cpp
najlkin: Done 😉

najlkin:
I also added a unit test for the dense matrix multiplication `A^T * B` 😉

linalg/solvers.cpp
najlkin: Done + corrected the printout 😉 .

linalg/hypre.cpp
najlkin: Yes, they used `Z` on CPU in Hypre 2.20, but anymore not from 2.21. Still, it is used for GPUs.

najlkin: The rabbit hole goes much further, checking the code, `Z` was used at multiple other places in v2.20, which we had not covered. From v2.21, we are missing it for GPUs in many cases, which partially overlap, but not fully to not make things easy 😩 . So should I care about v2.20? And should I fix that GPUs?

najlkin: Ok, how about now? But the list might grow as they implement more cases on GPUs, so it might be better to allocate it always for GPUs for future compatibility 🤔

linalg/hypre.cpp
```diff
@@ -3573,6 +3573,33 @@ void HypreSmoother::SetOperator(const Operator &op)
...
+   {
+      case 3:
+      case 6:
+      case 8:
+      case 10:
+      case 13:
+      case 14:
+         Z = new HypreParVector(*A);
+   }
+#elif defined(HYPRE_USING_GPU) || defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
```
v-dobrev, line 3587: This should be equivalent:
```suggestion
#elif defined(HYPRE_USING_GPU)
```

linalg/hypre.cpp
najlkin: Ok, inverted the logic to allocate Z always on GPUs except the specified cases 😉

linalg/hypre.cpp
najlkin: Done 👍

najlkin:
Returned to the single bcast of #4151 when running optimized, so the debug test runs the second one 😉 .

v-dobrev (approved):
Looks good, thanks @najlkin!

v-dobrev:

PR #1955, opened 2020-12-17
Document GridFunction::ProjectCoefficient and FiniteElement::Project.

dylan-copeland:
Improving documentation.

| [#1955](https://github.com/mfem/mfem/pull/1955) | @dylan-copeland | @tzanio | @mlstowell + @rw-anderson + @tzanio | 12/22/20 | 12/28/20 | 01/02/21 | |

### DIFF
--- a/fem/fe.hpp
+++ b/fem/fe.hpp
@@ -504,18 +504,18 @@ public:
    /** @brief Given a coefficient and a transformation, compute its projection
        (approximation) in the local finite dimensional space in terms
        of the degrees of freedom. */
-   /** Depending on the derived class that implements this, the degrees of
-       freedom may be computed by interpolation, i.e. evaluation at nodes, which
-       generally differs from L2-projection. */
+   /** The approximation used to project is usually local interpolation of
+       degrees of freedom. The derived class could use other methods not
+       implemented yet, e.g. local L2 projection. */
    virtual void Project(Coefficient &coeff,
                         ElementTransformation &Trans, Vector &dofs) const;
 
    /** @brief Given a vector coefficient and a transformation, compute its
        projection (approximation) in the local finite dimensional space
        in terms of the degrees of freedom. (VectorFiniteElements) */
-   /** Depending on the derived class that implements this, the degrees of
-       freedom may be computed by interpolation, i.e. evaluation at nodes, which
-       generally differs from L2-projection. */
+   /** The approximation used to project is usually local interpolation of
+       degrees of freedom. The derived class could use other methods not
+       implemented yet, e.g. local L2 projection. */
    virtual void Project(VectorCoefficient &vc,
                         ElementTransformation &Trans, Vector &dofs) const;
 
--- a/fem/fe.hpp
+++ b/fem/fe.hpp
@@ -504,12 +504,18 @@ public:
    /** @brief Given a coefficient and a transformation, compute its projection
        (approximation) in the local finite dimensional space in terms
        of the degrees of freedom. */
+   /** Depending on the derived class that implements this, the degrees of
+       freedom may be computed by interpolation, i.e. evaluation at nodes, which
+       generally differs from L2-projection. */
    virtual void Project(Coefficient &coeff,
                         ElementTransformation &Trans, Vector &dofs) const;
 
    /** @brief Given a vector coefficient and a transformation, compute its
        projection (approximation) in the local finite dimensional space
        in terms of the degrees of freedom. (VectorFiniteElements) */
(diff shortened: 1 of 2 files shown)


fem/gridfunc.hpp
```diff
@@ -345,14 +345,32 @@ class GridFunction : public Vector
...
    void ProjectGridFunction(const GridFunction &src);
 
+   /** @brief Project @a coeff Coefficient element-wise to @a this GridFunction.
+       The projection computation used on each element depends on the choice of
+       FiniteElementSpace #fes. Note that this is not generally an orthogonal
+       projection but may be an element-wise interpolation at element
+       integration points. */
    virtual void ProjectCoefficient(Coefficient &coeff);
 
+   /** @brief Project @a coeff Coefficient element-wise to @a this GridFunction,
```
mlstowell, line 355: It was not clear to me what this method does.  It might be useful to replace `element-wise` with `dof-wise` or something similar.  Perhaps it should also be stated that only degrees of freedom listed in the `dofs` array will be altered all others will remain unchanged.

fem/gridfunc.hpp
```diff
@@ -345,14 +345,32 @@ class GridFunction : public Vector
        projection matrix. */
    void ProjectGridFunction(const GridFunction &src);
 
+   /** @brief Project @a coeff Coefficient element-wise to @a this GridFunction.
+       The projection computation used on each element depends on the choice of
+       FiniteElementSpace #fes. Note that this is not generally an orthogonal
+       projection but may be an element-wise interpolation at element
+       integration points. */
```
mlstowell, line 352: I wonder if the term "integration points" will be misleading.  It seems that you're trying to make the point that this is not computing integrals of the coefficient.  Perhaps something more vague like "at select points within each element".

fem/gridfunc.hpp
```diff
@@ -345,14 +345,32 @@ class GridFunction : public Vector
...
+   /** @brief Project @a coeff Coefficient element-wise to @a this GridFunction,
+       using one element for each degree of freedom and nodal interpolation on
+       that element. */
    void ProjectCoefficient(Coefficient &coeff, Array<int> &dofs, int vd = 0);
 
+   /** @brief Project @a vcoeff VectorCoefficient element-wise to @a this
+       GridFunction. The projection computation used on each element depends on
+       the choice of FiniteElementSpace #fes. Note that this is not generally an
+       orthogonal projection but may be an element-wise interpolation at element
+       integration points. */
```
mlstowell, line 364: The comment for the scalar coefficient version applies here as well.

fem/gridfunc.hpp
```diff
@@ -345,14 +345,32 @@ class GridFunction : public Vector
...
    void ProjectCoefficient(VectorCoefficient &vcoeff);
 
+   /** @brief Project @a vcoeff VectorCoefficient element-wise to @a this
```
mlstowell, line 367: The comment for the scalar coefficient version also applies here.

mlstowell (approved):
I've made a couple small comments but I'll go along with whatever @dylan-copeland , @rw-anderson, and @tzanio decide.

fem/fe.hpp
```diff
@@ -504,12 +504,18 @@ class FiniteElement
    /** @brief Given a coefficient and a transformation, compute its projection
        (approximation) in the local finite dimensional space in terms
        of the degrees of freedom. */
+   /** Depending on the derived class that implements this, the degrees of
+       freedom may be computed by interpolation, i.e. evaluation at nodes, which
+       generally differs from L2-projection. */
    virtual void Project(Coefficient &coeff,
```
tzanio, line 510: This sounds a bit confusing to me, suggested rewording:

> Depending on the derived class, the approximation can be either local interpolation in the 
degrees of freedom (most common) or local L2 projection (see ExampleClass).

dylan-copeland: Thanks, this is an improvement, but I cannot find any example class that does local L2 projection. After searching fe.cpp, I think every case uses DOF interpolation. There are other `Project` functions that seem to compute an L2 projection matrix, but not this function. I have updated the comments, so please take another look.

tzanio: That sounds good, thanks

tzanio:
I updated the wording a bit in 50197feee0d2b13afbaee4776a1acf2f28539c66. Please revert if you don't like it :)

tzanio (approved):
Thanks for improving the documentation @dylan-copeland !

tzanio:

PR #1983, opened 2021-01-07
Add convective and skew-symmetric integrator for Navier-Stoke equations [NSE-nonlinear-update]

nnpp-schneier:
Adds the convective form and skew symmetric form of the nonlinear term arising in the Navier-Stokes equations. The current implementation within MFEM, which does not require partial assembly can only be used for obtaining the Gateaux derivative. However, other versions of the nonlinear form are needed for approaches such as a Picard iterations or IMEX time stepping schemes.

| [#1983](https://github.com/mfem/mfem/pull/1983) | @nnpp-schneier | @tzanio | @jandrej + @rcarson3 + @YohannDudouit | 01/09/21 | 02/01/21 | 02/07/21 | |

### DIFF
--- a/fem/nonlininteg.cpp
+++ b/fem/nonlininteg.cpp
@@ -914,12 +914,7 @@ void SkewSymmetricVectorConvectionNLFIntegrator::AssembleElementGrad(
    double w;
    Vector vec1(dim), vec2(dim), vec3(nd), vec4(dim), vec5(nd);
 
-   const IntegrationRule *ir = IntRule;
-   if (ir == nullptr)
-   {
-      int order = 2 * el.GetOrder() + trans.OrderGrad(&el);
-      ir = &IntRules.Get(el.GetGeomType(), order);
-   }
+   const IntegrationRule *ir = IntRule ? IntRule : &GetRule(el, trans);
 
    elmat = 0.0;
    elmat_comp_T = 0.0;
--- a/fem/nonlininteg.cpp
+++ b/fem/nonlininteg.cpp
@@ -876,8 +876,8 @@ void SkewSymmetricVectorConvectionNLFIntegrator::AssembleElementGrad(
    const Vector &elfun,
    DenseMatrix &elmat)
 {
-   int nd = el.GetDof();
-   int dim = el.GetDim();
+   const int nd = el.GetDof();
+   const int dim = el.GetDim();
 
    shape.SetSize(nd);
    dshape.SetSize(nd, dim);
--- a/fem/nonlininteg.cpp
+++ b/fem/nonlininteg.cpp
@@ -728,10 +728,8 @@ void VectorConvectionNLFIntegrator::AssembleElementVector(
       el.CalcShape(ip, shape);
       el.CalcPhysDShape(T, dshape);
       double w = ip.weight * T.Weight();
-      if (Q)
-      {
-         w *= Q->Eval(T, ip);
-      }
+      if (Q) { w *= Q->Eval(T, ip); }
+
       MultAtB(EF, dshape, gradEF);
       EF.MultTranspose(shape, vec1);
       gradEF.Mult(vec1, vec2);
@@ -746,8 +744,8 @@ void VectorConvectionNLFIntegrator::AssembleElementGrad(
    const Vector &elfun,
    DenseMatrix &elmat)
 {
-   int nd = el.GetDof();
-   int dim = el.GetDim();
+   const int nd = el.GetDof();
+   const int dim = el.GetDim();
 
    shape.SetSize(nd);
    dshape.SetSize(nd, dim);
@@ -823,8 +821,8 @@ void ConvectiveVectorConvectionNLFIntegrator::AssembleElementGrad(
    const Vector &elfun,
    DenseMatrix &elmat)
 {
-   int nd = el.GetDof();
-   int dim = el.GetDim();
+   const int nd = el.GetDof();
+   const int dim = el.GetDim();
 
    shape.SetSize(nd);
    dshape.SetSize(nd, dim);
--- a/fem/nonlininteg.cpp
+++ b/fem/nonlininteg.cpp
@@ -728,7 +728,10 @@ void VectorConvectionNLFIntegrator::AssembleElementVector(
       el.CalcShape(ip, shape);
       el.CalcPhysDShape(T, dshape);
       double w = ip.weight * T.Weight();
-      if (Q) { w *= Q->Eval(T, ip); }
+      if (Q)
+      {
+         w *= Q->Eval(T, ip);
+      }
       MultAtB(EF, dshape, gradEF);
       EF.MultTranspose(shape, vec1);
       gradEF.Mult(vec1, vec2);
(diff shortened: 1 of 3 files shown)


fem/nonlininteg.hpp
```diff
@@ -341,6 +354,56 @@ class VectorConvectionNLFIntegrator : public NonlinearFormIntegrator
    virtual void AddMultPA(const Vector &x, Vector &y) const;
 };
 
+/** @brief This class is used to assemble the skew-symmetric form of the
+nonlinear  term arising in the Navier Stokes equations
+\f$(u \cdot \nabla v, w )\f$*/
+class ConvectiveVectorConvectionNLFIntegrator : public NonlinearFormIntegrator
```
YohannDudouit, line 360: If my understanding is correct, this class is equivalent to `VectorConvectionNLFIntegrator` but the gradient is computed differently, would it makes sense to inherit from `VectorConvectionNLFIntegrator`?
```suggestion
class ConvectiveVectorConvectionNLFIntegrator : public VectorConvectionNLFIntegrator
```
If this is the case you can also remove `GetRule` and the private attributes.

fem/nonlininteg.cpp
```diff
@@ -813,4 +816,149 @@ void VectorConvectionNLFIntegrator::AssembleElementGrad(
...
+      {
+         w *= Q->Eval(trans, ip);
+      }
+
+      EF.MultTranspose(shape, vec1); // u^n
+
+      trans.AdjugateJacobian().Mult(vec1, vec2);
+
+      vec2 *= w;
+      dshape.Mult(vec2, vec3); // (u^n \cdot grad u^{n+1})
```
YohannDudouit, line 876: Could you develop a bit how you derive this sequence of operations, I don't follow how `dshape * w * det(J) * J^-1 * shape * u = u \cdot grad u`.

fem/nonlininteg.cpp
```diff
@@ -813,4 +816,149 @@ void VectorConvectionNLFIntegrator::AssembleElementGrad(
...
+   elmat = 0.0;
+   for (int i = 0; i < ir->GetNPoints(); i++)
+   {
+      const IntegrationPoint &ip = ir->IntPoint(i);
+      trans.SetIntPoint(&ip);
+
+      el.CalcShape(ip, shape);
+      el.CalcDShape(ip, dshape);
+
+      w = ip.weight;
```
YohannDudouit, line 864: Why is it not 
```suggestion
      w = ip.weight * T.Weight();
```
as in `VectorConvectionNLFIntegrator`?

fem/nonlininteg.cpp
```diff
@@ -813,4 +816,149 @@ void VectorConvectionNLFIntegrator::AssembleElementGrad(
...
+
+   double w;
+   Vector vec1(dim), vec2(dim), vec3(nd), vec4(dim), vec5(nd);
+
+   const IntegrationRule *ir = IntRule;
+   if (ir == nullptr)
+   {
+      int order = 2 * el.GetOrder() + trans.OrderGrad(&el);
+      ir = &IntRules.Get(el.GetGeomType(), order);
+   }
```
YohannDudouit, line 924:
```suggestion
   const IntegrationRule *ir = IntRule ? IntRule : &GetRule(el, trans);
```

fem/nonlininteg.cpp
nnpp-schneier: I believe you are right, I've gone ahead and removed those comments.

fem/nonlininteg.cpp
nnpp-schneier: I believe you are looking AssembleElementVector in VectorConvectionNLFIntegrator, they do not use that for AssembleElementGrad which is what is being implemented here.

fem/nonlininteg.hpp
nnpp-schneier: I agree this makes more sense.

fem/nonlininteg.cpp
YohannDudouit: My bad, thank you.

fem/nonlininteg.hpp
```diff
@@ -341,6 +354,49 @@ class VectorConvectionNLFIntegrator : public NonlinearFormIntegrator
    virtual void AddMultPA(const Vector &x, Vector &y) const;
 };
 
+/** @brief This class is used to assemble the skew-symmetric form of the
```
rcarson3, line 357: The comment here appears to be copy and paste from the `SkewSymmetricVectorConvectionNLFIntegrator` class. I imagine skew-symmetric form here should be convective from.

fem/nonlininteg.cpp
```diff
@@ -813,4 +816,121 @@ void VectorConvectionNLFIntegrator::AssembleElementGrad(
...
+   dshapex.SetSize(nd, dim);
+   elmat.SetSize(nd * dim);
+   elmat_comp.SetSize(nd);
+   gradEF.SetSize(dim);
+
+   DenseMatrix elmat_comp_T(nd);
+
+   EF.UseExternalData(elfun.GetData(), nd, dim);
+
+   double w;
```
rcarson3, line 895: I would have this declared within the loop. You could even make use of a ternary operation internally to make it a `const double` as well by doing something like:

```c++
const double w = Q ? Q->Eval(trans, ip) * ip.weight : ip.weight;
```

Although, I would say that sacrifices readability of the code a bit.

fem/nonlininteg.cpp
```diff
@@ -813,4 +816,121 @@ void VectorConvectionNLFIntegrator::AssembleElementGrad(
...
+
+
+void SkewSymmetricVectorConvectionNLFIntegrator::AssembleElementGrad(
+   const FiniteElement &el,
+   ElementTransformation &trans,
+   const Vector &elfun,
+   DenseMatrix &elmat)
+{
+   int nd = el.GetDof();
+   int dim = el.GetDim();
```
rcarson3, line 882: These both can be made `const int`.

fem/nonlininteg.cpp
```diff
@@ -813,4 +816,121 @@ void VectorConvectionNLFIntegrator::AssembleElementGrad(
...
+   shape.SetSize(nd);
+   dshape.SetSize(nd, dim);
+   dshapex.SetSize(nd, dim);
+   elmat.SetSize(nd * dim);
+   elmat_comp.SetSize(nd);
+   gradEF.SetSize(dim);
+
+   EF.UseExternalData(elfun.GetData(), nd, dim);
+
+   double w;
```
rcarson3, line 838: I would have this declared within the loop. You could even make use of a ternary operation internally to make it a const double as well by doing something like:

```c++
const double w = Q ? Q->Eval(trans, ip) * ip.weight : ip.weight;
```

Although, I would say that sacrifices readability of the code a bit.

fem/nonlininteg.cpp
```diff
@@ -813,4 +816,121 @@ void VectorConvectionNLFIntegrator::AssembleElementGrad(
...
 
+
+void ConvectiveVectorConvectionNLFIntegrator::AssembleElementGrad(
+   const FiniteElement &el,
+   ElementTransformation &trans,
+   const Vector &elfun,
+   DenseMatrix &elmat)
+{
+   int nd = el.GetDof();
+   int dim = el.GetDim();
```
rcarson3, line 827: These both can be made `const int`

fem/nonlininteg.hpp
```diff
@@ -34,10 +34,16 @@ class NonlinearFormIntegrator
...
+   {
+      IntRule = ir;
+   }
 
    /// Prescribe a fixed IntegrationRule to use.
-   void SetIntegrationRule(const IntegrationRule &irule) { IntRule = &irule; }
+   void SetIntegrationRule(const IntegrationRule &irule)
+   {
+      IntRule = &irule;
+   }
```
rcarson3, line 46: Why were these modified?

fem/nonlininteg.hpp
```diff
@@ -145,7 +151,10 @@ class HyperelasticModel
    /// evaluate Coefficient%s.
    /** @note It is assumed that _Ttr.SetIntPoint() is already called for the
        point of interest. */
-   void SetTransformation(ElementTransformation &_Ttr) { Ttr = &_Ttr; }
+   void SetTransformation(ElementTransformation &_Ttr)
+   {
+      Ttr = &_Ttr;
+   }
```
rcarson3, line 157: Same as above why did this change?

fem/nonlininteg.hpp
```diff
@@ -214,7 +223,10 @@ class NeoHookeanModel : public HyperelasticModel
 
 public:
    NeoHookeanModel(double _mu, double _K, double _g = 1.0)
-      : mu(_mu), K(_K), g(_g), have_coeffs(false) { c_mu = c_K = c_g = NULL; }
+      : mu(_mu), K(_K), g(_g), have_coeffs(false)
+   {
+      c_mu = c_K = c_g = NULL;
+   }
```
rcarson3, line 229: Again about why was this modified?

fem/nonlininteg.cpp
```diff
@@ -728,7 +728,10 @@ void VectorConvectionNLFIntegrator::AssembleElementVector(
       el.CalcShape(ip, shape);
       el.CalcPhysDShape(T, dshape);
       double w = ip.weight * T.Weight();
-      if (Q) { w *= Q->Eval(T, ip); }
+      if (Q)
+      {
+         w *= Q->Eval(T, ip);
+      }
```
rcarson3, line 734: Why was this modified?

fem/nonlininteg.cpp
nnpp-schneier: The implementation is based on VectorConvectionNLFIntegrator::AssembleElementGrad, which I did not write. I'm open to making this change, however I think it should be consistent with VectorConvectionNLFIntegrator::AssembleElementGrad, which I think is beyond the scope of this PR (even if it's a relatively minor change).

fem/nonlininteg.cpp
nnpp-schneier: The implementation is based on VectorConvectionNLFIntegrator::AssembleElementGrad, which I did not write. I'm open to making this change, however I think it should be consistent with VectorConvectionNLFIntegrator::AssembleElementGrad, which I think is beyond the scope of this PR (even if it's a relatively minor change).

fem/nonlininteg.cpp
nnpp-schneier: fixed

fem/nonlininteg.cpp
nnpp-schneier: fixed

fem/nonlininteg.hpp
nnpp-schneier: fixed

fem/nonlininteg.hpp
nnpp-schneier: fixed

fem/nonlininteg.hpp
nnpp-schneier: Fixed this to correctly state it is assembling the convective form.

fem/nonlininteg.cpp
rcarson3: @nnpp-schneier so you could definitely make the changes within the classes you've created in this PR. I don't think moving a variable from outside a loop into would make it any harder make to compare the child class against the parent class. 

The ternary operation we should probably get @YohannDudouit and @jandrej thoughts on as the other reviewers. I mean I like them just because they can allow us to make a variable const, but I know not everyone is always as big of a fan of them.

fem/nonlininteg.hpp
```diff
@@ -341,6 +342,49 @@ class VectorConvectionNLFIntegrator : public NonlinearFormIntegrator
...
+
+   virtual void AssembleElementGrad(const FiniteElement &el,
+                                    ElementTransformation &trans,
+                                    const Vector &elfun,
+                                    DenseMatrix &elmat);
+
+};
+/** @brief This class is used to assemble the skew-symmetric form of the nonlinear
+term arising in the Navier Stokes equations \f$.5*(u \cdot \nabla v, w ) -
+.5*(u \cdot \nabla w, v )\f$ */
```
jandrej, line 368: Besides the comment formatting, the PR looks ok.

fem/nonlininteg.cpp
nnpp-schneier: Since no one seemed to have an issue with it I went ahead and just made this suggested change.

fem/nonlininteg.cpp
nnpp-schneier: Since no one seemed to have an issue with it I went ahead and just made this suggested change.

fem/nonlininteg.hpp
nnpp-schneier: I believe I fixed the formatting and reran astyle. I think it looks good now.

rcarson3 (approved):

YohannDudouit (approved):

jandrej (approved):

tzanio:

tzanio (approved):
@nnpp-schneier, thank you for this contribution!

@jandrej, @rcarson3 and @YohannDudouit -- thank you for reviewing!

PR #1989, opened 2021-01-09
Custom basis for 3D HCurl and HDiv in 1D or 2D [lo-dim-em-dev]

mlstowell:
Adding ND and RT basis functions for 3D vector fields on 1D or 2D elements.  These are intended for modeling electromagnetic fields on 1D or 2D domains in the presence of anisotropic material properties.   In such cases the in-domain and out-of-domain polarizations can mix.  These basis functions avoid a great deal of code complexity needed to handle this type of mixing.

Anisotropic materials can arise in wave propagation through crystals or magnetized plasmas.

This PR adds three examples analogous to examples ex3p, ex6p, and ex13p but employing the new basis functions. Serial versions of two of these could also be added if desired (the third is an eigenvalue example which would remain parallel).

To facilitate the use of these basis functions I have added two new methods to the `FiniteElement` base class which specify the dimensions of the vector field and its curl.  Previously we made the assumption that `vdim == dim` or `vdim == sdim` but this is not the case for these bases.  Also the curl could have dimension 1 or 3 in 2D.  We could make assumptions about the dimension of the curl but I felt it was cleaner to declare it in a more obvious manner.

Although this PR adds many lines of code the vast majority of these lines are in unit tests so the library itself is not significantly larger.

Background:
I'm sure we're all aware that modern electromagnetic modeling is often accomplished with Nedelec and Raviart-Thomas finite elements (or an equivalent Yee lattice when finite differences will suffice). Unlike position or velocity fields which map `R^d` to `R^d` where `d`, clearly, is the dimension of the space, electromagnetic fields are inherently three dimensional. In other words the electric and magnetic vector fields always map `R^d` to `R^3`.

There are, of course, many special cases where symmetries of the problem ensure that one or more components of the electric or magnetic field will be zero.  For example, if the problem is invariant with respect to translations in the z direction, and the materials are isotropic, the electric field component in the z direction might be guaranteed to be zero and the standard Nedelec basis functions defined on triangles and/or quadrilaterals would suffice to describe such a field. Alternatively, the x and y components of the electric field might be guaranteed to be zero and the z component of the field could be accurately described by an H1 scalar finite element on triangles and/or quadrilaterals. In some cases none of the field components are guaranteed to be zero but the symmetries of the problem ensure that the x and y components are decoupled from the z component. In these cases the full electric field is often computed by two solving separate PDEs; one for the x and y components, and one for the z component, and combining the results during post-processing if at all.

Such special cases only exist if the material coefficients possess symmetries which are compatible with the translational or rotational symmetries of the problem domain, boundary conditions, and sources. In crystallography and plasma physics the material coefficients are generally anisotropic and not aligned with the other symmetries. This prevents the decoupling of the field components.

Motivation:
In order to model electromagnetic fields in reduced dimensions with anisotropic materials in MFEM we need to build `BlockOperators` and `BlockVectors` to describe our coupled system of PDEs. This can be done with the existing operators but it is far from convenient and there are many places where user error can lead to incorrect systems of PDEs, boundary conditions, or source functions.

Using `BlockVectors` in this manner is a bit unnatural. Building block systems to simultaneously solve for say density and velocity in the Darcy problem makes perfect sense because these are separate fields. For the electromagnetic fields this splitting into `BlockVectors` is inconvenient and unnatural. It would seem to be more logical to keep E_xy and E_z tightly coupled in a single `GridFunction` so that its curl, energy, etc. could be easily computed in a single operation.

The alternative implemented in this PR allows a user to setup the same linear system that would be used in a 3D simulation using the same integrators. This is accomplished by creating a set of basis functions which, in a sense, exist in between the standard 3D and 2D basis functions. We can think of the standard 2D Nedelec and Raviart-Thomas basis functions as being obtained from their 3D counterparts by first restricting the domain of the 3D basis functions to 2D and then projecting the vector-valued basis functions onto the 2D plane. This pair of operations eliminates many of the 3D basis functions by reducing effectively them to zero. The basis functions introduced in this PR perform the Restriction to 2D, hence the label `R2D`, but not the projection. Consequently they retain basis functions which are directed in the z direction.

Another way to look at the new Nedelec basis functions is that the `ND_R2D` basis functions are the direct sum of the standard 2D Nedelec basis functions and the 2D H1 basis functions multiplied by the unit vector in the z direction. Similarly the `RT_R2D` basis functions are a direct sum of 2D RT basis functions and 2D L2 basis functions (with map type INTEGRAL) multiplied by the z directed unit vector. The new `R1D` basis functions can be similarly constructed using two copies of the H1 or L2 basis functions for the y and z components. From this point of view they indeed look strange but I feel that the "restriction" view point described above is quite natural.

I think it's rather impressive that so much of the existing MFEM machinery works seamlessly with these new basis functions. The changes needed in some of the integrators were really quite small and, I think, make those integrators more robust.

| [#1989](https://github.com/mfem/mfem/pull/1989) | @mlstowell | @v-dobrev | @brendankeith + @dylan-copeland | 04/14/21 | 03/08/22 | 03/11/22 | |


v-dobrev:

v-dobrev:
@mlstowell, can you provide some background on these new finite elements, maybe a paper reference? I tried to look at the implementation but it all seems quite strange and non-standard.

fem/bilininteg.cpp
```diff
@@ -526,16 +527,58 @@ void MixedVectorIntegrator::AssembleElementMatrix2(
...
          for (int j=0; j<test_nd; j++)
          {
-            test_shape_tmp(j,0) = test_shape(j,1) * V(2) -
-                                  test_shape(j,2) * V(1);
-            test_shape_tmp(j,1) = test_shape(j,2) * V(0) -
-                                  test_shape(j,0) * V(2);
-            test_shape_tmp(j,2) = test_shape(j,0) * V(1) -
-                                  test_shape(j,1) * V(0);
+            // Compute shape_tmp = test_shape x V
+            // V will always be of length 3
```
dylan-copeland, line 534: Should there be `MFEM_VERIFY(VQ->GetVDim() == 3, "")`?

fem/fe.hpp
```diff
@@ -3282,6 +3293,427 @@ class ND_SegmentElement : public VectorFiniteElement
...
+
+   virtual void Project(const FiniteElement &fe,
+                        ElementTransformation &Trans,
+                        DenseMatrix &I) const;
+
+   virtual void ProjectCurl(const FiniteElement &fe,
+                            ElementTransformation &Trans,
+                            DenseMatrix &curl) const;
+   /*
+    virtual void GetLocalInterpolation(ElementTransformation &Trans,
```
dylan-copeland, line 3441: Should the commented code here be kept?

fem/fe.cpp
```diff
@@ -12627,6 +12637,2047 @@ void ND_SegmentElement::CalcVShape(const IntegrationPoint &ip,
...
+
+void ND_R1D_PointElement::CalcVShape(const IntegrationPoint &ip,
+                                     DenseMatrix &shape) const
+{
+   shape(0,0) = 1.0;
+   shape(0,1) = 0.0;
+
+   shape(1,0) = 0.0;
+   shape(1,1) = 1.0;
+}
```
dylan-copeland, line 12657: According to the documentation, this is an arbitrary order element that is constant in the second and third dimension. Why is just a constant 2D basis of dimension 2?

fem/fe.hpp
```diff
@@ -3282,6 +3293,427 @@ class ND_SegmentElement : public VectorFiniteElement
...
+
+   virtual void CalcVShape(ElementTransformation &Trans,
+                           DenseMatrix &shape) const;
+};
+
+/// Arbitrary order Nedelec 3D elements in 1D on a segment
+/** ND_R1D_SegmentElement provides a representation of a 3D Nedelec
+    basis where the vector field is assumed constant in the second and
+    third dimensions.
+*/
```
dylan-copeland, line 3322: The basis functions vary in all 3 components. Would it be clearer to say that the vector field is three-dimensional and depends only on the first dimension of position?

fem/fe.cpp
```diff
@@ -12627,6 +12637,2047 @@ void ND_SegmentElement::CalcVShape(const IntegrationPoint &ip,
...
+      }
+   for (int j = 0; j < pp1; j++)
+      for (int i = 0; i < pp1; i++)
+      {
+         int idx = dof_map[o++];
+         divshape(idx) = 0.;
+      }
+}
+
+
```
dylan-copeland, line 14680: This file is getting very long. Should the new code be put in a different `cpp` file?

fem/fe_coll.cpp
```diff
@@ -2751,6 +2751,451 @@ ND_Trace_FECollection::ND_Trace_FECollection(const int p, const int dim,
...
+}
+
+ND_R1D_FECollection::~ND_R1D_FECollection()
+{
+   for (int g = 0; g < Geometry::NumGeom; g++)
+   {
+      delete ND_Elements[g];
+   }
+}
+
```
dylan-copeland, line 2821: Extra line

fem/gridfunc.cpp
```diff
@@ -328,7 +328,28 @@ int GridFunction::VectorDim() const
...
+      { Geometry::SEGMENT, Geometry::TRIANGLE, Geometry::TETRAHEDRON };
+      fe = fec->FiniteElementForGeometry(geoms[fes->GetMesh()->Dimension()-1]);
+   }
+   else
+   {
+      fe = fes->GetFE(0);
+   }
+   if (!fe || fe->GetRangeType() == FiniteElement::SCALAR)
+   {
+      return 2 * fes->GetMesh()->SpaceDimension() - 3;
```
dylan-copeland, line 350: Is this correct in 1D, returning -1?

tests/unit/fem/test_r1d_bilininteg.cpp
```diff
@@ -0,0 +1,4753 @@
...
+            L2_FECollection fec_l2(order - 1, dim);
+            FiniteElementSpace fespace_l2(&mesh, &fec_l2);
+
+            GridFunction df_l2(&fespace_l2); df_l2.ProjectCoefficient(divF_coef);
+
+            FunctionCoefficient divF3_coef(DivF3);
+            REQUIRE( df_l2.ComputeL2Error(divF3_coef) < tol );
+         }
+         /*
+              SECTION("Divergence of RT")
```
dylan-copeland, line 1994: Why is this commented?

tests/unit/fem/test_r2d_bilininteg.cpp
```diff
@@ -0,0 +1,4671 @@
...
+            L2_FECollection fec_l2(order - 1, dim);
+            FiniteElementSpace fespace_l2(&mesh, &fec_l2);
+
+            GridFunction df_l2(&fespace_l2); df_l2.ProjectCoefficient(divF_coef);
+
+            FunctionCoefficient divF3_coef(DivF3);
+            REQUIRE( df_l2.ComputeL2Error(divF3_coef) < tol );
+         }
+         /*
+              SECTION("Divergence of RT")
```
dylan-copeland, line 1999: Another commented section?

tests/unit/fem/test_r2d_bilininteg.cpp
```diff
@@ -0,0 +1,4671 @@
...
+   VectorFunctionCoefficient  F3_coef(vdim, F3);
+   FunctionCoefficient        q3_coef(q3);
+   VectorFunctionCoefficient  D3_coef(vdim, V3);
+   MatrixFunctionCoefficient  M3_coef(vdim, M3);
+   MatrixFunctionCoefficient MT3_coef(vdim, MT3);
+   VectorFunctionCoefficient qF3_coef(vdim, qF3);
+   VectorFunctionCoefficient DF3_coef(vdim, DF3);
+   VectorFunctionCoefficient MF3_coef(vdim, MF3);
+
+   for (int type = (int)Element::TRIANGLE;
```
dylan-copeland, line 518: It looks like the files `test_r1d_bilininteg.cpp` and `test_r2d_bilininteg.cpp` are similar enough that they could be combined into one file, with a loop over all types of elements. Depending on the dimension, a pointer to a `FiniteElementCollection` could be set to a new 1D or 2D collection. Maybe there are complications I don't see, but if this could be done, it would save 4600+ lines of code.

examples/ex29p_proposed.cpp
```diff
@@ -0,0 +1,540 @@
...
+//               solution E and compute the corresponding r.h.s. f.
+//               We discretize with Nedelec finite elements in 1D, 2D, or 3D.
+//
+//               The example demonstrates the use of restricted H(curl) finite
+//               element spaces with the curl-curl and the (vector finite
+//               element) mass bilinear form, as well as the computation of
+//               discretization error when the exact solution is known. These
+//               restricted spaces allow the solution of 1D or 2D
+//               electromagnetic problems which involve 3D field vectors.  Such
+//               problems arise in plasma physics and crystalography.
```
dylan-copeland, line 28: crystallography

fem/fe.hpp
```diff
@@ -301,13 +303,21 @@ class FiniteElement
...
+                 int F = FunctionSpace::Pk, int VD = 0, int CD = 0);
 
    /// Returns the reference space dimension for the finite element
    int GetDim() const { return dim; }
 
+   /// Returns the vector dimension for vector-valued finite elements
+   int GetVDim() const { return vdim; }
+
+   /// Returns the dimension of the curl for vector-valued finite elements
+   int GetCDim() const { return cdim; }
```
dylan-copeland, line 319: `GetCurlDim` would be more readable, but I don't have a strong preference.

examples/ex29p_proposed.cpp
```diff
@@ -0,0 +1,540 @@
...
+   VectorFunctionCoefficient CurlE(3, CurlE_exact);
+   sol.ProjectCoefficient(E);
+
+   // 10. Set up the parallel bilinear form corresponding to the EM diffusion
+   //     operator curl muinv curl + sigma I, by adding the curl-curl and the
+   //     mass domain integrators.
+   DenseMatrix sigmaMat(3);
+   sigmaMat(0,0) = 2.0; sigmaMat(1,1) = 2.0; sigmaMat(2,2) = 2.0;
+   sigmaMat(0,2) = 0.0; sigmaMat(2,0) = 0.0;
+   sigmaMat(0,1) = M_SQRT1_2; sigmaMat(1,0) = M_SQRT1_2;
```
dylan-copeland, line 169: This is the first time I ever saw constants like `M_SQRT1_2`, but Google helped me learn what they are. Are other people usually familiar with these, or should there be a comment about the definition?

mlstowell:
> @mlstowell, can you provide some background on these new finite elements, maybe a paper reference? I tried to look at the implementation but it all seems quite strange and non-standard.

Sorry for the delayed response. I missed this question until @dylan-copeland pointed it out this morning.

I am not aware of any papers that describe these finite elements and a quick literature search didn't turn up any. These finite elements arose out of a few short discussions that I had with @homijan in the distant past i.e. before the pandemic. I would be very interested to hear his feedback on this PR as well.

I'll add more background and motivation to the comments at the start of this PR so that they'll be easy to find.

fem/bilininteg.cpp
mlstowell: Good point

fem/fe.cpp
```diff
@@ -22,10 +22,12 @@ namespace mfem
 
 using namespace std;
 
-FiniteElement::FiniteElement(int D, Geometry::Type G, int Do, int O, int F)
+FiniteElement::FiniteElement(int D, Geometry::Type G,
+                             int Do, int O, int F, int VD, int CD)
    : Nodes(Do)
 {
    dim = D ; geom_type = G ; dof = Do ; order = O ; func_space = F;
+   vdim = VD ; cdim = CD;
```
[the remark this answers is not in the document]
mlstowell: I agree but I'm following the style set in the previous line. I can change both lines if people agree but in this case I think the spaces do make it more readable.

fem/fe.cpp
mlstowell: It looks weird, doesn't it? This finite element is analogous to the `PointFiniteElement` which appears alongside the `H1_FiniteElements`. It's "arbitrary order" in the sense that it is compatible with 1D Nedelec finite elements of any order. In the constructor I pass an integer `p` even though this is not really relevant which is the "order" used in the `ND_R1D_FECollection`. Following the discussion in PR #2163 I'm really not sure what to do with these integers.

fem/fe.cpp
mlstowell: I would support splitting `fe.cpp` and `fe.hpp` into multiple files. Perhaps one for H1, one for ND, one for RT, one for L2, and keep common base classes in `fe.hpp` and `fe.cpp`. Other basis functions could either stay in the current files or be placed in a miscellaneous file.

fem/fe_coll.cpp
mlstowell: That's a matter of opinion, I suppose. This file often uses a double blank line between classes. See lines 1531, 1918, 1938, etc..

fem/fe.cpp
dylan-copeland: That sounds good to me. We already have similar splittings of the `bilininteg_*` files.

fem/fe.cpp
dylan-copeland: This is the first time I've looked at finite elements at a point, and it doesn't make sense to me yet. I see that `H1_FECollection` has `PointFiniteElement`, and it seems to be just 0-order (constant). How is it actually used, and why? Does it just get used for evaluating a point value of a `GridFunction` or something, so it only needs to be a constant that gets multiplied by a value given by a coefficient?

fem/gridfunc.cpp
mlstowell: I'm not sure it needs to be correct in 1D but I could be mistaken. This `if` statements handles the case where we have multiple copies of a scalar field and our current assumptions are that such a field is either 2 or 3 dimensional. I believe a 1D scalar field, with vdim=1, would throw an error before calling this method.

Of course, it wouldn't hurt to simplify this logic and maybe even document this method so that we know what to expect from it...

tests/unit/fem/test_r1d_bilininteg.cpp
mlstowell: Good question. I don't see a reason so keep it.

tests/unit/fem/test_r2d_bilininteg.cpp
mlstowell: Hmm... I'll take it out

examples/ex29p_proposed.cpp
mlstowell: Good catch

fem/fe.hpp
mlstowell: I agree and I also don't have a strong preference. Let's see if anyone else has a preference. I'll be happy to change it if that's the consensus.

examples/ex29p_proposed.cpp
mlstowell: This may be one of the more obscure ones. I can add a comment at the end of the line if that would help to clarify things.

tests/unit/fem/test_r2d_bilininteg.cpp
mlstowell: I suspected you might comment on this :smirk:

It would probably be possible. One issue would be the tests which mix the new classes with derivatives of the preexisting classes such as gradient of an H1. These tend to involve multiplication with rectangular matrices like 3x1 or 3x2 mapping matrices. The R1D versions tend to have more pieces than the R2D versions which may not be trivial to merge. Another issue is the analytic functions at the top of the page. These are generally named the same in the two files but they have different definitions.  Renaming and merging these would not be fun although it could certainly be done.

The last issue that comes to mind might be the most important for me. In the current implementation the `TEST_CASE` statements can be used to run subsets of the unit tests and this is _so_ convenient when tracking down problems. If we merged these unit tests into a single file we'd need to merge the names appearing in these `TEST_CASE` statements and we would be able to just run R1D or R2D cases separately. This issues has already bit me after we merged other tests and I've had to temporarily edit files so that they skip certain tests so that I can focus on more urgent things. Clearly there are workarounds for this issue and I could manage if the size of the unit test files is something we need to reduce.

In short, I am willing to merge these but I'd rather not.

fem/fe.cpp
mlstowell: The only uses I can think of are as boundary elements. For example, when preparing a Neumann BC for a PDE on a 1D domain. This involves an integral over the boundary which in this case is just evaluating the BC function and the basis function at a single point. Rather than treating 0D boundaries as special cases we have 0D basis functions which can be plugged into the usual machinery and produce the proper results.

examples/ex29p_proposed.cpp
dylan-copeland: Yes, it can be a short comment like `// 1/sqrt(2) in cmath`

tests/unit/fem/test_r2d_bilininteg.cpp
dylan-copeland: It sounds reasonable to keep them separate then. I think we should merge unless there are good reasons not to, and you provided some good reasons.

fem/fe.cpp
dylan-copeland: Then maybe the source of confusion is the comment in the hpp file that the elements are "arbitrary order." The documentation for `PointFiniteElement` does not say anything about the order. Would it be more accurate to say it is 0-order and works for 1D boundary evaluations with arbitrary order 1D elements? Or is it better just not to mention order, since it can be confusing without a complete explanation of how these elements are used?

tests/unit/fem/test_r2d_bilininteg.cpp
mlstowell: That's a relief 😌. Thanks, @dylan-copeland !

fem/fe.cpp
mlstowell: Agreed. I rewrote the comment to remove "order" and hopefully summarized its purpose a little better.

brendankeith:
> > @mlstowell, can you provide some background on these new finite elements, maybe a paper reference? I tried to look at the implementation but it all seems quite strange and non-standard.
> 
> Sorry for the delayed response. I missed this question until @dylan-copeland pointed it out this morning.
> 
> I am not aware of any papers that describe these finite elements and a quick literature search didn't turn up any. These finite elements arose out of a few short discussions that I had with @homijan in the distant past i.e. before the pandemic. I would be very interested to hear his feedback on this PR as well.
> 
> I'll add more background and motivation to the comments at the start of this PR so that they'll be easy to find.

The new background and motivation is quite nice! :)

examples/ex30p_proposed.cpp
```diff
@@ -0,0 +1,532 @@
+//                       MFEM Example 30 - Parallel Version
+//
+// Compile with: make ex30p
+//
+// Sample runs:  mpirun -np 4 ex30p -m ../data/inline-segment.mesh
```
brendankeith, line 5: `❯ ./ex30p_proposed` runs fine. However, when I execute the first suggested sample run, `❯ ./ex30p_proposed -m ../data/inline-segment.mesh`, I receive a segmentation fault. The segmentation also appears when running `❯ mpirun -np 4 ex30p_proposed -m ../data/inline-segment.mesh`. I have not checked the remaining sample runs in this file.

mlstowell: Actually I had to tweak HYPRE's implementation of AMS so that it would support 1D problems. I had forgotten about this. I have a pull request ([#269](https://github.com/hypre-space/hypre/pull/269)) with HYPRE which I should really followup on.

examples/ex32p_proposed.cpp
```diff
@@ -0,0 +1,455 @@
+//                       MFEM Example 32 - Parallel Version
+//
+// Compile with: make ex32p
+//
+// Sample runs:  mpirun -np 4 ex32p -m ../data/square-disc.mesh -o 1
```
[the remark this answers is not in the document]
brendankeith: More than one of sample runs have visualization problems, similar to the case reported above, when they are run with MPI.

It appears that all of the sample runs work when run in serial. However, some of the runs deliver very suspicious refinement patterns. It is possible that the error estimator being used here is not appropriate "right out of the box" for practical applications. Are you sure that this example should be merged into the main code? Another option is stating somewhere that there is "experimental support" for AMR/a posteriori error estimation.

examples/ex30p_proposed.cpp
dylan-copeland: Nice catch, I didn't try this sample run. @mlstowell should this sample run be removed until the hypre PR is merged, since otherwise the MFEM testing will fail?

examples/ex31p_proposed.cpp
```diff
@@ -0,0 +1,531 @@
+
+//                       MFEM Example 31 - Parallel Version
+//
+// Compile with: make ex31p
+//
+// Sample runs:  mpirun -np 4 ex31p -m ../data/inline-segment.mesh
```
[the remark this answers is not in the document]
brendankeith: At this point, it's clear why this failed. I'm setting this as "resolved."

examples/ex32p_proposed.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+   kappa = freq * M_PI;
+
+   // 3. Read the (serial) mesh from the given mesh file on all processors.  We
+   //    can handle triangular, quadrilateral, tetrahedral, hexahedral, surface
+   //    and volume meshes with the same code.
+   Mesh *mesh = new Mesh(mesh_file, 1, 1);
+   dim = mesh->Dimension();
+
+   MFEM_VERIFY(dim == 2, "This exmaple requires a 2D mesh.");
```
brendankeith, line 78: This **example** requires...

examples/ex32p_proposed.cpp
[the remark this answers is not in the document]
mlstowell: Yes, I noticed this a few days ago and I've been trying to track down the problem (without much luck). I may pull `ex32p_proposed ` out of this PR. Sorry for the hassle.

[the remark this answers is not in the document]
psocratis: > Yes, I noticed this a few days ago and I've been trying to track down the problem (without much luck). I may pull `ex32p_proposed ` out of this PR. Sorry for the hassle.

I was playing with this too, just now, and what happens in parallel runs is that the solver produces NANs, in a non-deterministic way. I noticed that the source of this is the call to 
```cpp
sol.ProjectBdrCoefficientTangent(E, ess_bdr);
````
If replaced with `sol.ProjectCoefficient(E);` the non-deterministic behavior goes away. There is still a Bad termination at the 17th AMR iteration but it's deterministic and up to that point everything looks good.

**Comment by mlstowell** [comment]
I've removed the 1D sample runs to await possible support in AMS.

I've also moved the AMR example to a separate PR since it's not working in parallel for some unknown reason.

File: examples/ex30p_proposed.cpp (hunk 15 above)
Comment (mlstowell) [reply examples/ex30p_proposed.cpp]: I have removed the 1D sample runs.

File: fem/fe.hpp (hunk 12 above)
Comment (brendankeith) [reply fem/fe.hpp:319]: I prefer `GetCurlDim`

**Comment by mlstowell** [comment]
Added a serial version of ex30 (and this one works on 1D meshes since it doesn't rely on HYPRE)

File: examples/ex32p_proposed.cpp (hunk 16 above)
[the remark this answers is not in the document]
Comment (mlstowell) [reply examples/ex32p_proposed.cpp]: This sounds like a useful clue. Thanks, @psocratis!

Ever since these `ProjectBdr*` methods started using `AccumulateAndCount*` they keep causing me trouble. We should see if we have any unit tests that cover these. Not in this PR but soon...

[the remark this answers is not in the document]
Comment (mlstowell) [reply examples/ex32p_proposed.cpp]: I created an issue (#2226) so that we don't forget to resolve this.

File: examples/ex30_proposed.cpp [hunk 19]
```diff
@@ -0,0 +1,466 @@
+//                                MFEM Example 30
+//
+// Compile with: make ex30
+//
+// Sample runs:  ex30 -m ../data/inline-segment.mesh -o 2
+//               ex30 -m ../data/hexagon.mesh -o 2
+//               ex30 -m ../data/star.mesh -o 2
+//               ex30 -m ../data/fichera.mesh -o 3 -r 1
```
brendankeith, line 8: Although the `../data/amr-hex.mesh -r 1` and `ex30 -m ../data/amr-hex.mesh -r 1` execute without errors, I am not sure whether they should be here. I thought that this basis was meant for 1D and 2D domains. On that note, a sample run where the domain is an embedded 2D surface may be interesting. Thoughts?

examples/ex31p_proposed.cpp
```diff
@@ -0,0 +1,530 @@
+
+//                       MFEM Example 31 - Parallel Version
+//
+// Compile with: make ex31p
+//
+// Sample runs:  mpirun -np 4 ex31p -m ../data/hexagon.mesh -o 2
+//               mpirun -np 4 ex31p -m ../data/star.mesh
+//               mpirun -np 4 ex31p -m ../data/square-disc.mesh -o 2 -n 4 -rs 1
```
brendankeith, line 8: Suspicious results when running `mpirun -np 4 ex31p_proposed -m ../data/square-disc.mesh -o 2 -n 4 -rs 1`
Screen Shot 2021-05-06 at 12 01 10 PM

brendankeith: Is the genus of the domain being properly handled here? The cohomology is non-trivial here....

examples/ex31p_proposed.cpp
```diff
@@ -0,0 +1,530 @@
+
+//                       MFEM Example 31 - Parallel Version
+//
+// Compile with: make ex31p
+//
+// Sample runs:  mpirun -np 4 ex31p -m ../data/hexagon.mesh -o 2
```
brendankeith, line 6: The eigenmodes are very pretty :star-struck:, however, they do not appear to preserve the symmetries of the domain. Could this be a visualization issue or a convergence issue?  Or is it possibly something else?
Screen Shot 2021-05-06 at 12 12 21 PM
Screen Shot 2021-05-06 at 12 11 34 PM

I'm not sure whether there should be repeated eigenvalues with this type of problem. @mlstowell do you know?

examples/ex31p_proposed.cpp
brendankeith: > Not that different results are produced from `mpirun -np 1 ex31p_proposed -m ../data/square-disc.mesh -o 2 -n 4 -rs 1`

This also could be related to https://github.com/mfem/mfem/issues/2226. Thanks @psocratis 
Screen Shot 2021-05-06 at 12 56 56 PM

examples/ex31p_proposed.cpp
mlstowell: This problem uses an anisotropic material tensor in the mass matrix. The strongest eigenvalue of this tensor is directed in the (1,1,1) direction, the next in the (0,0,1) direction, and the smallest in the (1,-1,1) direction. This is why the modes tend to be aligned with +/-45 degrees.

examples/ex31p_proposed.cpp
mlstowell: GLVis_s01
This result looks much better on the NURBs mesh. In any case this is certainly a good question. HypreAME should be using the discrete gradient operator to weed out gradients of scalar fields. This should leave the rotational fields we're interested in and a single irrotational field due to the hole in the topology. This irrotational field doesn't wrap around the hole as I would expect but I cannot say if it's correct or not.

As for the other three panels; they look ugly. I was considering rescaling the panels so that the XY panel and the Z panel use the same color bar. I think this would help to alleviate some of the confusion caused by this set of images.

Do you think this rescaling is a good idea in general? Should we avoid this topology in particular? Should we switch to the NURBs mesh (I think we should)?

brendankeith: It's really interesting that the results are so much poorer with the high order mesh than with the NURBS mesh. I would suggest that the default behaviour be not to plot the zero eigenmodes. That would bypass the need for any rescaling.

brendankeith: It may be good to keep this there, alongside the NURBS mesh example. There is some implicit information in there. If someone runs both, they may decide right away that it's worth using a NURBS description of their domain geometry.

examples/ex30_proposed.cpp
brendankeith: @mlstowell Any thoughts on this?

examples/ex31p_proposed.cpp
brendankeith: > GLVis_s01
> This result looks much better on the NURBs mesh. In any case this is certainly a good question. HypreAME should be using the discrete gradient operator to weed out gradients of scalar fields. This should leave the rotational fields we're interested in and a single irrotational field due to the hole in the topology. This irrotational field doesn't wrap around the hole as I would expect but I cannot say if it's correct or not.
> 
> As for the other three panels; they look ugly. I was considering rescaling the panels so that the XY panel and the Z panel use the same color bar. I think this would help to alleviate some of the confusion caused by this set of images.
> 
> Do you think this rescaling is a good idea in general? Should we avoid this topology in particular? Should we switch to the NURBs mesh (I think we should)?

@mlstowell Can you share the command you used to produce this figure? I am not seeing such good results with `../data/square-disc-nurbs.mesh` anymore :/

brendankeith: For example, `❯ mpirun -np 4 ex31p_proposed -m ../data/square-disc-nurbs.mesh -o 2 -n 4 -rs 1` produces
Screen Shot 2021-05-18 at 10 53 20 AM

mlstowell: I don't recall exactly but `./ex31p_proposed -m ../data/square-disc-nurbs.mesh -rs 4 -o 2` looks quite similar. This one's not bad either `./ex31p_proposed -m ../data/square-disc-nurbs.mesh -rs 3 -o 3`.

brendankeith: OK. Perhaps the default should be `./ex31p_proposed -m ../data/square-disc-nurbs.mesh -rs 3 -o 3`
`mpirun -np 4 ex31p_proposed -m ../data/square-disc-nurbs.mesh` doesn't look very nice

brendankeith (approved):

mlstowell:
@v-dobrev, et al, Dylan was concerned that the `fe.cpp` and `fe.hpp` files are becoming too long and I tend to agree. I'd like to propose the following split:

- `fe.[ch]pp` to contain the `FiniteElement` base class and any other common definitions or code
- `fe_h1.[ch]pp` to contain the H1 basis functions
- `fe_l2.[ch]pp` to contain the L2 basis functions
- `fe_nd.[ch]pp` to contain the ND basis functions
- `fe_rt.[ch]pp` to contain the RT basis functions 
- `fe_misc.[ch]pp` to contain other classes that don't fit naturally in the above files

This is probably the start of a longer discussion but the real question here is should this splitting (or something similar) be done in this PR?

tzanio:
I think it is time to split the `fe` files, there is a lot there and it may be difficult to find to separate the important from less frequently used stuff.

v-dobrev:
> This is probably the start of a longer discussion but the real question here is should this splitting (or something similar) be done in this PR?

I'd prefer to do this in a separate PR.

dylan-copeland (approved):

homijan:
@mlstowell I think that adding the `RT_R1D`/`RT_R2D` examples is unnecessary it it holds back this PR. Adding it in a future PR makes more sense.

homijan (approved):

mlstowell:
@v-dobrev, I think this now incorporates all the changes we discussed the other day. Thanks again for the useful feedback!

fem/fe/fe_base.hpp
```diff
@@ -360,7 +368,8 @@ class FiniteElement
    /** @brief Evaluate the values of all shape functions of a scalar finite
        element in physical space at the point described by @a Trans. */
    /** The size (#dof) of the result Vector @a shape must be set in advance. */
-   void CalcPhysShape(ElementTransformation &Trans, Vector &shape) const;
+   virtual void CalcPhysShape(ElementTransformation &Trans,
+                              Vector &shape) const;
```
v-dobrev, line 372: Why make this method virtual? To me it looks like it does not need to be.

v-dobrev:
@mlstowell,

can you also:
- [ ] Update CHANGELOG
- [ ] List the new example in doc/CodeDocumentation.dox

Also, I'll open an issue to document the new examples in the MFEM webpage repo, https://github.com/mfem/web.

fem/fe/fe_base.hpp
mlstowell: I don't think it makes a difference for this PR. I can change it back. I probably just didn't see a reason why it shouldn't be virtual.

fem/bilininteg.cpp
```diff
@@ -600,17 +664,17 @@ void MixedScalarVectorIntegrator::AssembleElementMatrix2(
    int trial_nd = trial_fe.GetDof(), test_nd = test_fe.GetDof(), i;
    int sca_nd = sca_fe->GetDof();
    int vec_nd = vec_fe->GetDof();
-   int spaceDim = Trans.GetSpaceDim();
+   int vdim = GetVDim(*vec_fe);
```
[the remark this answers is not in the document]
mlstowell: This is a tricky one. In this case `GetVDim` is a member function of the integrator which is meant to be overridden by derived classes. So, we don't want to supersede choices made in the derived classes. Perhaps the best resolution is to store the space dimension as member data so that we can compute the `max` inside `GetVDim` by default but still allow derived classes to make their own determination.

v-dobrev:

adam-sim-dev:
Dear Mark, @mlstowell , I have a question about the ND_R2D element. Does it support the 3D source? For example, a horizontal electric dipole along the x-axis in the xy plane, or a vertical magnetic dipole on the xy plane. I do see in ex31, the RHS is 3D.

adam-sim-dev:
Another question, are the three dimensions of the shape functions in ND_R2D_TriangleElement::CalcVShape and ND_R2D_TriangleElement::CalcCurlShape corresponding to x,y, and z, where x, y are the two dimensions of the 2D model?

PR #2012, opened 2021-01-23
Fix Intel compiler warnings

### LINKED ISSUE #2003: Warning: overloaded virtual function
fem/bilininteg.cpp(3679): warning: overloaded virtual function "mfem::VectorCoefficient::Eval" is only partially overridden in class "VCrossVShapeCoefficient"

tzanio:
Fixes #2003

| [#2012](https://github.com/mfem/mfem/pull/2012) | @tzanio | @tzanio | @camierjs + @mlstowell | 01/23/21 | 02/02/21 | 02/04/21 | |

### DIFF
--- a/fem/bilininteg.cpp
+++ b/fem/bilininteg.cpp
@@ -3687,6 +3687,8 @@ ScalarCrossProductInterpolator::AssembleElementMatrix2(
          : VectorCoefficient(fe_.GetDof()), VQ(vq), fe(fe_),
            vshape(vdim, vq.GetVDim()), vc(vq.GetVDim()) { }
 
+      using VectorCoefficient::Eval;
+
       virtual void Eval(Vector &V, ElementTransformation &T,
                         const IntegrationPoint &ip)
       {
--- a/fem/bilininteg.cpp
+++ b/fem/bilininteg.cpp
@@ -3688,7 +3688,6 @@ ScalarCrossProductInterpolator::AssembleElementMatrix2(
            vshape(vdim, vq.GetVDim()), vc(vq.GetVDim()) { }
 
       using VectorCoefficient::Eval;
-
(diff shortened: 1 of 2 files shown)


tzanio:
It looks like the Intel compiler needs `#pragma unroll (n)` instead of `#pragma unroll n`.

A quick google search seem to indicate that clang supports both and that gcc needs `#pragma GCC unroll (n)`.

Any suggestions how to handle this in a portable way?

mlstowell:
There are six of these `structs` in `bilininteg.cpp`; three `VectorCoefficients` and three `MatrixCoefficients`.  Only the `VectorCoefficient` versions require these `using` declarations.  Do we understand why this is the case?

tzanio:
> There are six of these `structs` in `bilininteg.cpp`; three `VectorCoefficients` and three `MatrixCoefficients`.  Only the `VectorCoefficient` versions require these `using` declarations.  Do we understand why this is the case?

Yes 😁 

The issue is that `VectorCoefficients` has two virtual functions named `Eval` and the 3 derived classes in `bilininteg.cpp` override only one of them. This is a [problem](https://stackoverflow.com/questions/21462908/warning-overloaded-virtual-function-baseprocess-is-only-partially-overridde) with some C++ compilers. 

The `using VectorCoefficient::Eval;` is actually about the second virtual `Eval` not the one that is being overwritten.

`MatrixCoefficients` has only one `Eval`, so there is no such issue there.

Does this make sense?

mlstowell:
That does make sense.  Thanks, @tzanio!

Although perhaps we should (one day) standardize the `Eval` methods of the three coefficient types so they match...

mlstowell (approved):

camierjs (approved):

PR #2031, opened 2021-02-01
Allow first SparseMatrix to not use CUDA

### LINKED ISSUE #2026: CUDA Error
I'm trying to use CUDA to solve my problem and get the following errors on my terminal
```
 ** On entry to cusparseSpMV_bufferSize() parameter number 1 (handle) had an illegal value: NULL pointer

 ** On entry to cusparseSpMV() parameter number 1 (handle) had an illegal value: NULL pointer

```

I've traced it back to something happening in `ParBilinearForm::FormLinearSystem()` but I'm not sure where else to look.

wcdawn:
Currently, if the first `SparseMatrix` is generated without using CUDA, no subsequent `SparseMatrix` can use CUDA because the CUDA handle will not be initialized. This is fixed here.

Resolves #2026 

| [#2031](https://github.com/mfem/mfem/pull/2031) | @wcdawn | @tzanio | @artv3 + @YohannDudouit | 02/19/21 | 02/19/21 | 02/26/21 | |

### DIFF
--- a/linalg/sparsemat.cpp
+++ b/linalg/sparsemat.cpp
@@ -30,7 +30,7 @@ using namespace std;
 
 #ifdef MFEM_USE_CUDA
 int SparseMatrix::SparseMatrixCount = 0;
-cusparseHandle_t SparseMatrix::handle;
+cusparseHandle_t SparseMatrix::handle = nullptr;
 size_t SparseMatrix::bufferSize = 0;
 void * SparseMatrix::dBuffer = nullptr;
 #endif
@@ -39,10 +39,15 @@ void SparseMatrix::InitCuSparse()
 {
    // Initialize cuSPARSE library
 #ifdef MFEM_USE_CUDA
-   SparseMatrixCount++;
-   if (SparseMatrixCount == 1 && Device::Allows(Backend::CUDA_MASK))
+   if (Device::Allows(Backend::CUDA_MASK))
    {
-      cusparseCreate(&handle);
+      if (!handle) { cusparseCreate(&handle); }
+      useCuSparse=true;
+      SparseMatrixCount++;
+   }
+   else
+   {
+      useCuSparse=false;
    }
 #endif
 }
--- a/linalg/sparsemat.hpp
+++ b/linalg/sparsemat.hpp
@@ -619,12 +619,24 @@ public:
    {
       Destroy();
 #ifdef MFEM_USE_CUDA
-      if (handle && SparseMatrixCount==1 && Device::Allows(Backend::CUDA_MASK))
+      if (useCuSparse)
       {
-         cusparseDestroy(handle);
-         CuMemFree(dBuffer);
+         if (SparseMatrixCount==1)
+         {
+            if (handle)
+            {
+               cusparseDestroy(handle);
+               handle = nullptr;
+            }
+            if (dBuffer)
+            {
+               CuMemFree(dBuffer);
+               dBuffer = nullptr;
+               bufferSize = 0;
+            }
+         }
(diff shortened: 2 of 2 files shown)


linalg/sparsemat.cpp
```diff
@@ -40,7 +40,7 @@ void SparseMatrix::InitCuSparse()
    // Initialize cuSPARSE library
 #ifdef MFEM_USE_CUDA
    SparseMatrixCount++;
-   if (SparseMatrixCount == 1 && Device::Allows(Backend::CUDA_MASK))
+   if (!handle && Device::Allows(Backend::CUDA_MASK))
```
artv3, line 43: To be extra safe can we set ``cusparseHandle_t SparseMatrix::handle = nullptr``

wcdawn: In the class initalizer? I like this idea.

artv3: Yup, here 
https://github.com/mfem/mfem/blob/b9bc24bb27be115f3ad3dd3dbfa892d165ccc8a3/linalg/sparsemat.cpp#L33

tomstitt: It doesn't look like the `= nullptr` update was pushed, could you double-check that @wcdawn  ?

wcdawn: Should be fixed now. Thanks!

tomstitt (approved):

YohannDudouit (approved):
Thank you for this contribution @wcdawn !

artv3 (approved):

tzanio:

tzanio:
There are number of [autotest failures](https://mybitbucket.llnl.gov/projects/MFEM/repos/autotest/browse/tux429/2021-02-20-github-next/sample-runs-devices-tux429.diff) that could be due to this PR. 

For example, `ex6 -pa -d cuda -no-vis` reports

```
+ ** On entry to cusparseSpMV_bufferSize() parameter number 1 (handle) had an illegal value: bad initialization or already destroyed
```

artv3:
@wcdawn , for the sparse matrix that is not using CuSparse I think we have to set ``useCuSparse = false`` and if its not using CuSparse then skip the clean up parts of using CuSparse.

wcdawn:
Turned out to be a challenge to track down. This was not associated with the CUDA `handle`, but the buffer `dBuffer`. I have passed `ex6` but I can't see what other tests were failing. I also switched `CUSPARSE_CSRMV_ALG1` to `CUSPARSE_SPMV_CSR_ALG1` to suppress a deprecation warning.

tzanio:

YohannDudouit:
`linalg/sparsemat.cpp(682): error: identifier "CUSPARSE_SPMV_CSR_ALG1" is undefined`

YohannDudouit:
@wcdawn Shouldn't you check the Cuda version to choose between `CUSPARSE_CSRMV_ALG1` and `CUSPARSE_SPMV_CSR_ALG1`?

wcdawn:
@YohannDudouit thank you for the catch! It looks like the deprecation/change was made in CUDA v11.0. There is currently no precompiler variable in MFEM with the CUDA version. It looks like it could be added by checking `CUDART_VERSION`.

This PR isn't to update MFEM for CUDA v11.0 features so that can be done later. For now, I've reverted to the old format.

tzanio:
@wcdawn, can you avoid force-pushing? This complicates the review and our internal testing.

tzanio:

wcdawn:
@tzanio Will do! Thanks!

PR #2052, opened 2021-02-13
Some fixes related to field-diff.

vladotomov:
| [#2052](https://github.com/mfem/mfem/pull/2052) | @vladotomov | @tzanio | @tzanio + @kmittal2 | 02/13/21 | 02/19/21 | 02/22/21 | |

### DIFF
--- a/fem/gslib.cpp
+++ b/fem/gslib.cpp
@@ -296,7 +296,10 @@ void FindPointsGSLIB::GetQuadHexNodalCoordinates()
    const TensorBasisElement *tbe =
       dynamic_cast<const TensorBasisElement *>(fes->GetFE(0));
    MFEM_VERIFY(tbe != NULL, "TensorBasis FiniteElement expected.");
-   const Array<int> &dof_map = tbe->GetDofMap();
+   Array<int> dof_map(dof_cnt);
+   const Array<int> &dm = tbe->GetDofMap();
+   if (dm.Size() > 0) { dof_map = dm; }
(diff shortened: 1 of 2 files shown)


tangqi:
Thanks again for the quick fix. It looks good and I only have one question.

How could I correlate the diff grid function with the max diff it reports? Attached is the result using the data I provided. The max diff is greater than the diff plot. Is the discrepancy due to the order of computing the diff GridFunction (project first and then diff) or just a result of visualization?

```
./field-diff -m1 amr.mesh -s1 j.amr -m2 full.mesh -s2 j.full -p 800
...
Searched 640000 points.
Max diff: 0.008342
Avg diff: 7.61094e-05
Vol diff: 1.15513e-05
```

Screen Shot 2021-02-14 at 10 39 34 AM

vladotomov:
> How could I correlate the diff grid function with the max diff it reports?

You shouldn't try to correlate them, because they are sampling at different points. The visualization should only be used to get some idea where the differences are.

The `diff` numbers that are printed in the terminal depend on the number of equidistant points you sample (by setting `-p`). They are the errors at those exact points.

To visualize a `GridFunction`, we always take the nodes of `mesh_1` (the vertices for a linear mesh), compute the difference at those nodes, and construct a GridFunction using these numbers.

tangqi (approved):

tzanio (approved):

kmittal2 (approved):

tzanio:

PR #2058, opened 2021-02-18
Bugfix for periodic NURBS meshes in parallel [periodic-nurbs-dev]

IdoAkkerman:
This PR fixes a bug that precluded running periodic BCs in parallel.
It also fixes some issues when specifying the periodicity in the mesh file instead of a seperate file.

Please not, one important issue here is that the fespace for the nodal values  remains non-periodic. While the fespaces for the solutions are, that is the reason for the extra "ConnectBoundaries()" calls. This could potentially also been done by including a flag to the constructor params.

### DIFF
--- a/fem/gridfunc.hpp
+++ b/fem/gridfunc.hpp
@@ -99,7 +99,7 @@ public:
    /** The content of @a input should be in the format created by the method
        Save(). The reconstructed FiniteElementSpace and FiniteElementCollection
        are owned by the GridFunction. */
-   GridFunction(Mesh *m, std::istream &input);
+   GridFunction(Mesh *m, std::istream &input, bool connect = false);
 
    GridFunction(Mesh *m, GridFunction *gf_array[], int num_pieces);
 
--- a/data/periodic-square-nurbs.mesh
+++ b/data/periodic-square-nurbs.mesh
@@ -1,0 +1,62 @@
+MFEM NURBS mesh v1.0
+
+#
+# MFEM Geometry Types (see mesh/geom.hpp):
+#
+# SEGMENT     = 1
+# SQUARE      = 3
+# CUBE        = 5
+#
+
+dimension
+2
+
+elements
+1
+1 3 0 1 2 3
+
+boundary
+4
+1 1 0 1
+3 1 2 3
+4 1 3 0
+2 1 1 2
+
+
+
+edges
+4
+0 0 1
+0 3 2
+1 0 3
+1 1 2
+
+vertices
+4
+
+knotvectors
+2
+1 2 0 0 1 1
+1 2 0 0 1 1
+
+periodic
+1 1
+1 3
+
+weights
+1
+1
+1
+1
+
+MFEM FiniteElementSpace v1.0
+FiniteElementCollection: NURBS
+VDim: 2
+Ordering: 1
+End: MFEM FiniteElementSpace v1.0
+
+0 0
+1 0
+1 1
+0 1
+
--- a/data/periodic-square-nurbs.mesh
+++ b/data/periodic-square-nurbs.mesh
@@ -22,8 +22,6 @@ boundary
 4 1 3 0
 2 1 1 2
 
-
-
 edges
 4
 0 0 1
--- a/fem/datacollection.cpp
+++ b/fem/datacollection.cpp
@@ -620,7 +620,7 @@ void VisItDataCollection::LoadFields()
       {
          if ((it->second).association == "nodes")
          {
-            field_map.Register(it->first, new GridFunction(mesh, file), own_data);
+            field_map.Register(it->first, new GridFunction(mesh, file, true), own_data);
          }
          else if ((it->second).association == "elements")
          {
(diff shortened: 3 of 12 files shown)


tzanio:
Hi @IdoAkkerman,

This slipped through the cracks because it was not listed as `ready-for-review` or `wip`.

If it is ready to go, can you merge `master` in it and mark it with  `ready-for-review`?

Thanks!
Tzanio

tzanio:
ping: @IdoAkkerman

tzanio:

IdoAkkerman:
I have been absent for a while but am picking things up again.

tzanio:

marc-85:
Hi all,

Thanks @IdoAkkerman. I have a question with regards to the constructor 
```cpp
ParNURBSExtension::ParNURBSExtension(MPI_Comm comm, NURBSExtension *parent,
                                     int *part, const Array<bool> &active_bel)
```
In it, it is assumed that the parent `NURBSextension` owns the mesh topology and if it doesn't it throws an error. Is this strictly necessary? If this wasn't necessary the constructor could add an additional argument to copy the mesh topology or a pointer to a mesh topology?

Other than that it looks good to me.

IdoAkkerman:
@marc-85 
Good question. How the different constructors handle the own_topo can probably improved and made more uniform.
In the mentioned routine we can perhaps make the following simplification:
Replace:
```
   // steal ownership of patchTopo from the 'parent' NURBS extension
   if (!parent->own_topo)
   {
      mfem_error("ParNURBSExtension::ParNURBSExtension :\n"
                 "  parent does not own the patch topology!");
   }
   own_topo = 1;
   parent->own_topo = 0;
```
with
```
   own_topo =  parent->own_topo ;
   parent->own_topo = 0;
```
However, I am reluctant to make these changes here. As this is not changed in this PR and due to  refinement and order elevation routines I can fully oversee the consequences at this stage. This might be better attempted in a small dedicated PR.

IdoAkkerman:
The NURBSext->ConnectBoundaries(); call seems problematic. It is not called when reading in sol.gf in glvis. I will think of a mechanism of how to call/not-call this function.
Summary:

- For the Nodes Gridfunction do NOT call, coordinates on each side of periodic BC are different
- For the solution GridFunctions it should be called, solution on each side should be the same

JacobLotz:
If I understand correctly, after this the periodic boundaries are read correctly, but the actual connecting of the boundaries should be done manually after creating a mesh/NURBSExtension. This was partly done automatically before this and might give problems with backwards compatibility?

Changes to the miniapps look logical to me.

I have compiled Visit using this branch. Comparing the results of miniapp nurbs_ex1p gives the expected results for the periodic mesh "periodic-square-nurbs.mesh" compared to the non periodic mesh "square-nurbs.mesh". 

```
mpirun -n 4 ./nurbs_ex1p -no-vis -m ../../data/periodic-square-nurbs.mesh -o 2 -r 1
mpirun -n 4 ./nurbs_ex1p -no-vis -m ../../data/square-nurbs.mesh -o 2 -r 1
```

However performing calculations for the same meshes for miniapp nurbs_ex1p does not give the expected results for the periodic mesh. 

```
./nurbs_ex1 -no-vis -m ../../data/periodic-square-nurbs.mesh -o 2 -r 1
./nurbs_ex1 -no-vis -m ../../data/square-nurbs.mesh -o 2 -r 1
```
Periodic case
Screenshot from 2021-12-22 14-24-36

Not periodic case
Screenshot from 2021-12-22 14-25-48

Unfortunately performing a same test in the master branch throws an error. So im not able to check if this is a branch specific thing.

```
./nurbs_ex1 -no-vis -m ../../data/square-nurbs.mesh -o 2 -r 1 --master 1 --slave 3

```

tzanio:
@IdoAkkerman and @JacobLotz any chance you can pick this up again?

I'd like to merge it in, now that mfem-4.4 is out 😄 

Thanks!
Tzanio

tzanio:
I tried to fix merging and the new check for shadow variables, but one of the tests is failing... can you take a look?

IdoAkkerman:
@tzanio I think I fixed it.

fem/gridfunc.hpp
```diff
@@ -103,7 +103,7 @@ class GridFunction : public Vector
    /** The content of @a input should be in the format created by the method
        Save(). The reconstructed FiniteElementSpace and FiniteElementCollection
        are owned by the GridFunction. */
-   GridFunction(Mesh *m, std::istream &input);
+   GridFunction(Mesh *m, std::istream &input, bool connect = false);
```
mlstowell, line 106: I must say this type of change to a very commonly used class makes me very nervous. What exactly is the purpose of this new argument? At the very least we need to document this argument here and in the other constructors with new arguments. Beyond that I wonder what impact this might have in a non-NURBS context. Even if that impact is purely psychological i.e. will users expect this argument to activate some new feature in non-NURBS settings?

IdoAkkerman: I had a look at `FormLinearSystem` and its probably a bit more involved than I anticipated...

IdoAkkerman:
I will check if this PR can be replaced by making correct use of
`   static Mesh MakePeriodic(const Mesh &orig_mesh, const std::vector<int> &v2v);`

PR #2074, opened 2021-03-01
Bugfix - CoarseToFineMap currently fails for derefinement transform

kmittal2:
```
void GetCoarseToFineMap(const Mesh &fine_mesh,
                           Table &coarse_to_fine,
                           Array<int> &coarse_to_ref_type,
                           Table &ref_type_to_matrix,
                           Array<Geometry::Type> &ref_type_to_geom) const;
```
It seems like `GetCoarseToFineMap` was originally intended for use with `GetRefinementTransforms`. It currently segfaults with `GetDerefinementTransforms` because it assumes that we are working with the `fine_mesh`. Nonetheless, since it still returns the right output for `coarse_to_fine` with `GetDerefinementTransforms`, I have added a fix to circumvent the segfault because we need the `coarse_to_fine` map for hr-adaptivity and the DRL project.

Additionally, for `RefinementTransform`, currently the embedding size is set based on `leaf_elements` (which includes `ghost_elements`), but the embedding information is filled using the `coarse_elements` (which does not include `ghost_elements`). This results in garbage for the embeddings corresponding to the ghost elements.

| [#2074](https://github.com/mfem/mfem/pull/2074) | @kmittal2 | @tzanio | @jakubcerveny + @rw-anderson | 03/02/21 | 07/04/21 | 07/06/21 | |


tzanio:

mesh/ncmesh.cpp
```diff
@@ -4261,7 +4261,7 @@ const CoarseFineTransformations& NCMesh::GetRefinementTransforms()
    if (!transforms.embeddings.Size())
    {
       transforms.Clear();
-      transforms.embeddings.SetSize(leaf_elements.Size());
+      transforms.embeddings.SetSize(leaf_elements.Size() - GetNumGhostElements());
```
jakubcerveny, line 4264: I think this can now (after `par-ncmesh-saving-dev`) be done more simply as `SetSize(NElements)`.

mesh/ncmesh.hpp
```diff
@@ -331,6 +331,7 @@ class NCMesh
 
    /// Return the number of root elements.
    int GetNumRootElements() { return root_state.Size(); }
+   virtual int GetNumGhostElements() const { return 0; }
```
jakubcerveny, line 334: Please don't add this. In fact please remove `ParNCMesh::GetNumGhostElements` and `ParNCMesh::GetNumGhostVertices`  for me (I forgot to get rid of those in `par-ncmesh-saving-dev`).

mesh/ncmesh.cpp
```diff
@@ -4435,6 +4435,11 @@ void CoarseFineTransformations::GetCoarseToFineMap(
       coarse_to_fine.GetJ()[i] = cf_j[i].two;
    }
 
+   if (fine_mesh.GetLastOperation() == Mesh::Operation::DEREFINE)
```
jakubcerveny, line 4438: This seems like a hacky solution. Unfortunately I'm not the author of `CoarseFineTransformations::GetCoarseToFineMap` so I'm not sure what better solution to suggest (the method is not documented and I'm having a hard time understanding it). Maybe split the method into several methods and avoid calling those that don't make sense for derefinement? Maybe @v-dobrev or @pazner can advise here?

jakubcerveny (changes requested):

tzanio:
@v-dobrev, @pazner -- can you help with `GetCoarseToFineMap`?

kmittal2:
@v-dobrev I made the changes that we discussed.

tzanio:
@jakubcerveny, can you maybe take another look?

mesh/ncmesh.hpp
```diff
@@ -68,7 +68,20 @@ struct CoarseFineTransformations
                            Table &coarse_to_fine,
                            Array<int> &coarse_to_ref_type,
                            Table &ref_type_to_matrix,
-                           Array<Geometry::Type> &ref_type_to_geom) const;
+                           Array<Geometry::Type> &ref_type_to_geom,
+                           bool get_coarse_to_fine_only = false) const;
+
+   void GetCoarseToFineMap(const Mesh &fine_mesh,
+                           Table &coarse_to_fine) const
+   {
```
jakubcerveny, line 76: Would it be possible to move the implementation to the `cpp` file to keep the header concise?

jakubcerveny (approved):
Looks OK, only a minor request below.

mesh/ncmesh.hpp
kmittal2: Done. Thanks @jakubcerveny

tzanio:
thanks for the review @jakubcerveny 

@v-dobrev can you confirm you are OK with these changes by approving?

v-dobrev (approved):

tzanio:

PR #2077, opened 2021-03-03
Basis functions for LOR preconditioners on GLL meshes

dylan-copeland:
Adding new option to use transformed 1D basis functions for tensor elements.

### DIFF
--- a/fem/fe.cpp
+++ b/fem/fe.cpp
@@ -7275,6 +7275,26 @@ void Poly_1D::Basis::Eval(const double y, Vector &u, Vector &d,
    }
 }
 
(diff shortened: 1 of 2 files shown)


dylan-copeland:
Note that the new basis has to be enabled by setting `bool useTransformedShapes=true;` in fe.hpp. We can improve the interface after getting things working, so that the option can be specified from a higher level (from the `FiniteElementCollection`). With the new basis for the Nedelec space, ex3p.cpp does not work for order greater than 1, even with just the mass matrix. Are we missing something @crdohrm?

pazner:
Superseded by #2089.

PR #2089, opened 2021-03-08
Improved LOR solvers, including for H(curl) and H(div) [lor-hcurl-hdiv]

pazner:
# New LOR Functionality

* New "integrated" basis type for vector finite elements
* Spectrally equivalent LOR spaces for H(curl) and H(div)
* User-friendly LOR discretization and solver classes
* New LOR solver miniapp for all spaces: H1, H(curl), H(div), L2
* "Partially conforming" LOR discretizations on AMR meshes:
    * Spectrally equivalent LOR solvers for high-order problems on AMR meshes

Minor changes:
* Makes everything in `QuadratureFunctions1D` static
* Adds some convenience functions to `Geometry` for tensor-product geometries
* Add const accessor to `OperatorHandle`
* Fix some bugs in `BasisType`
* Implements `GetTransferMatrix` based on L2 projection when using the new basis with AMR meshes

## Some information about the LOR classes

(This is also available as documentation for users in `miniapps/solvers/lor_docs.md`).

This PR introduces some new LOR classes intended to make working with low-order refined discretizations and solvers more convenient.
The interface is designed to be convenient, flexible, and powerful: solvers with "default" options can be created in only one line of code, but much more customization is available if desired.

### Creating solvers with default options

The most basic LOR solver can be created as a one liner. For example, the following creates a direct solver based on a serial LOR discretization:

```c++
// Given BilinearForm `a` and list of essential DOFs, create direct LOR solver
LORSolver<UMFPackSolver> prec(a, ess_tdof_list);
```
This will automatically create a LOR discretization corresponding to `a`, assemble the system, and create a solver whose type is given by the template parameter.
For example, in parallel, one could do the following to create an AMG preconditioner instead of direct solver:
```c++
LORSolver<HypreBoomerAMG> prec(a, ess_tdof_list);
```

#### What does `LORSolver` do?

In the above examples, `LORSolver<T>` does a couple of things:

* Creates a LOR mesh and finite element space
* Assembles a LOR version of the bilinear form `a`
* Creates a preconditioner of type `T` using the LOR-assembled version
* Creates a DOF permutation (if required) to transition between the LOR and high-order DOF numbering

### Creating discretizations

If one is interested in using the underlying LOR discretization, it can be created as follows:
```c++
LOR lor(a, ess_tdof_list);
SparseMatrix &A_lor = lor.GetAssembledMatrix();
```
The low-order refined finite element space can be accessed by `lor.GetFESpace()`.
In parallel, there is an analogous `ParLOR` class:
```c++
ParLOR lor(a, ess_tdof_list);
HypreParMatrix &A_lor = lor.GetAssembledMatrix();
```

### Passing options to solvers

Some solvers require special options passed to the constructor.
For example, this can be useful when creating AMS preconditioners, which require access to the finite element space:
```c++
ParLOR lor(a, ess_tdof_list);
LORSolver<HypreAMS> ams(lor, &lor.GetParFESpace());
```
In this case, parameters passed to the `LORSolver` constructor will be forwarded to the `HypreAMS` constructor.
Also, note that the underlying solver object can be accessed with the `*` and `->` operators.

### Assembling custom bilinear forms

In all of the above examples, the LOR discretization is based off of the high-order bilinear form: the same form is assembled on a LOR finite element space.
In some circumstances, it is desirable to assemble a custom bilinear form on the LOR space and *not* to reuse the high-order form.
For example, one may want to assemble several LOR bilinear forms, and avoid the overhead of creating a mesh and finite element space for each of them.
These capabilities can be accomplished with the `LOR` and `LORSolver` classes:
```c++
LOR lor(fes); // Create LOR version of fes, don't assemble any forms
BilinearForm a_lor(&lor.GetFESpace());
a_lor.AddDomainIntegrator(new MassIntegrator);
a_lor.Assemble();
LORSolver<DSmoother> D(a_lor.SpMat(), lor);
```
In this case, the object `lor` does not assemble any forms, and does not represent any particular operator.
Instead, it simply represents a LOR mesh and finite element space, which can be used to create custom bilinear forms.
Once assembled, these operators can be passed to `LORSolver`, which will take care of any low-order-to-high-order DOF permutations, if needed.

| [#2089](https://github.com/mfem/mfem/pull/2089) | @pazner | @tzanio | @barker29 + @dylan-copeland + @tzanio | 04/27/21 | 05/16/21 | 05/19/21 | |


barker29:
Just so you know, I will be holding up this review until #1781 is merged into master, just so I can say, for a few days, that I wrote the best high-order H(curl) solver in MFEM.

More seriously, thanks for doing this and I'm looking forward to seeing how it performs.

tzanio:
`make style` :)

crdohrm:
Wow! Looks like a lot got done since our last discussion and my getting back from vacation today. Based on a quick look at the commits, it seems like you have both the degree of freedom and sign (+1 or -1) maps between the HO and LOR discretizations. Do I have that correct? Looking forward to hearing how this works out.

pazner:
> Wow! Looks like a lot got done since our last discussion and my getting back from vacation today. Based on a quick look at the commits, it seems like you have both the degree of freedom and sign (+1 or -1) maps between the HO and LOR discretizations. Do I have that correct? Looking forward to hearing how this works out.

Yes, I think we've got the DOF numbering and orientations figured out now. Thanks for your help with that! We'll keep you in the loop as we progress with this

barker29 (commented):
This is already very impressive. I'm seeing quite nice convergence and fast implementation at high order.

I am probably jumping the gun providing a "review" of a WIP branch, but I was excited to try this out and wanted to report what I noticed.

Can you suggest a command line for `plor` where `--hybridization` works? I can't get it to converge.

fem/fe.hpp
```diff
@@ -38,7 +38,8 @@ class BasisType
       OpenHalfUniform = 5,  ///< Nodes: x_i = (i+1/2)/n,   i=0,...,n-1
       Serendipity     = 6,  ///< Serendipity basis (squares / cubes)
       ClosedGL        = 7,  ///< Closed GaussLegendre
-      NumBasisTypes   = 8   /**< Keep track of maximum types to prevent
+      Integrated      = 8,  ///< Integrated indicator functions
```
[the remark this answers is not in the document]
pazner: Yeah, I wasn't so happy with the name myself. Think of it as a placeholder for now. I'm not sure `Gerritsma` is any better, since it's not standard terminology, and you'd have to read the paper anyway to know what it means.

We could try to think of another term that suggests that instead of a *nodal* basis, it is a basis that satisfies a sort of "integral indicator function" condition. The other thing to note is that it is also dependent on the partitioning of the reference interval, which in this case is Gauss-Lobatto.

If we wanted to go for maximum verbosity, we could have `GaussLobattoIntegralIndicator`, which conveys the meaning but it awkward to read and write.

pazner:
> This is already very impressive. I'm seeing quite nice convergence and fast implementation at high order.
> 
> I am probably jumping the gun providing a "review" of a WIP branch, but I was excited to try this out and wanted to report what I noticed.
> 
> Can you suggest a command line for `plor` where `--hybridization` works? I can't get it to converge.

Thanks taking a look at this! Unfortunately hybridization is not working, and it's not clear (from a linear algebra perspective, not from an implementation perspective) how to get it to work well in this case. If you have any great ideas I'd be happy to hear them :)

crdohrm:
I'm happy to see that work is progressing right long. I was wondering about peoples' thoughts on how this is working out so far. If everything is working properly, I would expect a sparse direct solver for the LOR system to work well as a preconditioner for the HO system based on the spectral equivalence of the two systems. Have you observed this for H(curl) and H(div) problems? Also, have you looked into using AMS or ADS preconditioners for the LOR system as a preconditioner for the HO system? I'm hoping that this all works out and will open the door for the investigation of other preconditioning strategies as well (I like both multigrid and domain decomposition). Thanks for all the nice work.

pazner:
> I'm happy to see that work is progressing right long. I was wondering about peoples' thoughts on how this is working out so far. If everything is working properly, I would expect a sparse direct solver for the LOR system to work well as a preconditioner for the HO system based on the spectral equivalence of the two systems. Have you observed this for H(curl) and H(div) problems? Also, have you looked into using AMS or ADS preconditioners for the LOR system as a preconditioner for the HO system? I'm hoping that this all works out and will open the door for the investigation of other preconditioning strategies as well (I like both multigrid and domain decomposition). Thanks for all the nice work.

Good to hear from you! Yes, so far everything looks like it's working well, both with sparse direct solvers (in serial) as well as AMS and ADS preconditioners for the LOR system. We are also currently looking into other preconditioning strategies, hopefully we'll have some updates soon!

crdohrm:
That's great news! Looking forward to hearing more as updates become available.

fem/bilinearform.hpp
```diff
@@ -176,6 +176,8 @@ class BilinearForm : public Matrix
    /// Returns the assembly level
    AssemblyLevel GetAssemblyLevel() const { return assembly; }
 
+   Hybridization *GetHybridization() const { return hybridization; }
```
pazner, line 179: This is not (currently) used in this PR, but it could be useful in the future

tzanio:

miniapps/solvers/lor_docs.md
```diff
@@ -0,0 +1,69 @@
...
+```c++
+LORSolver<HypreBoomerAMG> prec(a, ess_tdof_list);
+```
+
+### What does `LORSolver` do?
+
+In the above examples, `LORSolver<T>` does a couple of things:
+
+* Creates a LOR mesh and finite element space
+* Assembles a LOR version of the bilinear form `a`
```
barker29, line 25: Maybe most users won't care, but at least I am interested in how the Gerritsma basis functions are handled here. If the high-order form uses the standard MFEM basis, can you do anything with LOR? If so, how does it work?

fem/lor.hpp
```diff
@@ -0,0 +1,287 @@
...
+   bool RequiresDofPermutation() const;
+
+   /// Returns the low-order refined finite element space.
+   FiniteElementSpace &GetFESpace() const { return *fes; }
+
+   ~LORBase();
+};
+
+/// Create and assemble a low-order refined version of a BilinearForm.
+class LOR : public LORBase
```
barker29, line 128: I don't like the name `LOR` because I think it will be obscure to users and hard to google. I would prefer `LowOrderRefined`, or maybe `LORDiscretization` or `LORSystem`.

miniapps/solvers/README
```diff
@@ -61,3 +61,18 @@ The solvers for the above block system include:
 
              P^{-1} = [ diag(M)^{-1}     0    ]
                       [  0             AMG(S) ]
+
+# Low-Order Refined Solvers
+
+The miniapp `lor` (and its parallel counterpart `plor`) demonstrate the use of
```
barker29, line 67: Should we mention Clark Dohrmann's paper here (or in `lor_docs.md`)?

miniapps/solvers/lor_docs.md
```diff
@@ -0,0 +1,69 @@
...
+* Creates a preconditioner of type `<T>` using the LOR-assembled version
+* Creates a DOF permutation (if requires) to transition between the LOR and high-order DOF numbering
```
barker29, line 27:
```suggestion
* Creates a DOF permutation (if required) to transition between the LOR and high-order DOF numbering
```

miniapps/solvers/plor_solvers.cpp
```diff
@@ -0,0 +1,214 @@
...
+//
+// H(div):    grad-div problem, u - grad(div u) = f
+//
+// In each case, the high-order finite element problem is preconditioned using a
+// low-order finite element discretization defined on a Gauss-Lobatto refined
+// mesh. The low-order problem is solved using hypre's AMG preconditioners:
+// BoomerAMG is used for H1 and L2 problems, AMS is used for H(curl) and 2D
+// H(div) problems, and ADS is used for 3D H(div) problems.
+//
+// For vector finite element spaces, the special "Integrated" basis type is used
```
barker29, line 34: Forget my other comments, this is probably the place to cite Dohrmann's paper.

barker29 (commented):
Unfortunately I do not have the time to give this PR the review it deserves, but based on what I've seen I'm excited about it and think it looks very good. I do think a bit more documentation should make clear to users that they can't just expect `LORSolver` to work magic on H(curl) with the standard basis functions - the miniapp is clear but the classes themselves less so.

miniapps/solvers/lor_docs.md
pazner: This is a good point (which, incidentally, @v-dobrev also brought up today).

One option is to warn the user if the "wrong" basis type is being used. Another option is to apply a change of basis operation in the solver action the same way the DOF permutation is currently applied. I'll have to look more into both of these options to figure out the best way to go.

miniapps/solvers/plor_solvers.cpp
pazner: Good idea. Done.

fem/fe.cpp
```diff
@@ -1377,10 +1427,58 @@ void VectorFiniteElement::LocalInterpolation_RT(
...
+   }
+   DenseMatrixInverse fine_mass_inv(fine_mass);
+   fine_mass_inv.Mult(fine_coarse_mass, I);
+}
+
 void VectorFiniteElement::LocalInterpolation_ND(
    const VectorFiniteElement &cfe, const double *tk, const Array<int> &d2t,
    ElementTransformation &Trans, DenseMatrix &I) const
 {
+   if (!is_nodal) { return LocalL2Projection_ND(cfe, Trans, I); }
```
dylan-copeland, line 1480: In the nodal case, this function just does interpolation, so why introduce L2 projection? Alternatively, could we do something similar to the `ProjectIntegrated` functions, which do integrated interpolation? This would be specialized in classes such as `ND_QuadrilateralElement`, which would call its own function instead of this base class function. I am not saying we definitely should do this, but some questions are (i) why do L2 projection vs. interpolation; (ii) is it clear to the user whether projection or interpolation is done in each case; (iii) is L2 projection too expensive or unstable for very high order? My initial thoughts are (i) L2 projection is more accurate, but it is not clear why this choice is made here other than it is general for various derived classes; (ii) it is not clear to the user, and mixing projection and interpolation in different cases may cause confusion or problems; (iii) for moderate order (with PA we are limited to about order 5 or 6 for 3D Nedelec anyway), the dense matrix inversion should be fine, but for high enough order the inversion may get expensive and possibly inaccurate/unstable.

fem/lor.hpp
```diff
@@ -0,0 +1,287 @@
...
+   /// is required when applying the %LOR solver as a preconditioner for the
+   /// high-order problem. This permutation can be disabled (for example, in
+   /// order to precondition the low-order problem directly).
+   void UsePermutation(bool use_permutation_)
+   {
+      use_permutation = use_permutation_;
+   }
+
+   /// Support the use of -> to call methods of the underlying solver.
+   SolverType *operator->() const { return solver; }
```
dylan-copeland, line 277: If I understand this correctly, for a `LORSolver *s` pointer, `s->Mult` would call `LORSolver::Mult`; for a `LORSolver s` object, `s->Mult` would call `s.solver.Mult`. If that is correct, it seems perhaps convenient but also potentially confusing and likely to cause bugs (which may be hard to track down, i.e. it may run but not apply the permutation). Would it be clearer and safer to just have a `GetSolver()` function?

fem/fe.hpp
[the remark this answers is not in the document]
dylan-copeland: I agree that we should have a slightly more descriptive name, e.g. `IntegratedGLL`, `IntegratedTensor`, or `IntegratedGerritsma`.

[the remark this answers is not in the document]
dylan-copeland: I agree that we should have a slightly more descriptive name, e.g. `IntegratedGLL`, `IntegratedTensor`, or `IntegratedGerritsma`.

miniapps/solvers/lor_solvers.cpp
```diff
@@ -0,0 +1,200 @@
...
+//            (in L2 discretized using the symmetric interior penalty DG method)
+//
+// H(curl):   definite Maxwell problem, u + curl curl u = f
+//
+// H(div):    grad-div problem, u - grad(div u) = f
+//
+// In each case, the high-order finite element problem is preconditioned using a
+// low-order finite element discretization defined on a Gauss-Lobatto refined
+// mesh. The low-order problem is solved using a direct solver if MFEM is
+// compiled with Suite Sparse enabled, or with one iteration of symmetric
```
dylan-copeland, line 31: No space in SuiteSparse.

miniapps/solvers/lor_solvers.cpp
```diff
@@ -0,0 +1,200 @@
...
+//
+// H1 and L2: definite Helmholtz problem, u - Delta u = f
+//            (in L2 discretized using the symmetric interior penalty DG method)
+//
+// H(curl):   definite Maxwell problem, u + curl curl u = f
+//
+// H(div):    grad-div problem, u - grad(div u) = f
+//
+// In each case, the high-order finite element problem is preconditioned using a
+// low-order finite element discretization defined on a Gauss-Lobatto refined
```
dylan-copeland, line 29: The Dorhmann paper refers to the refined mesh as GLL. Is there a reason for the different name here?

miniapps/solvers/lor_solvers.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+   a.RecoverFEMSolution(X, b, x);
+
+   double er =
+      (H1 || L2) ? x.ComputeL2Error(u_coeff) : x.ComputeL2Error(u_vec_coeff);
+   cout << "L2 error: " << er << endl;
+
+   // Save the solution and mesh to disk. The output can be viewed using GLVis
+   // as follows: "glvis -m mesh.mesh -g sol.gf"
+   x.Save("sol.gf");
```
dylan-copeland, line 186: Perhaps add `if (visualization)`, so the user can decide whether to produce these files. Since ParaView is more rarely used than glvis (at least by me), it may also be convenient to have a ParaView output parameter, false by default, but I don't feel strongly about it.

miniapps/solvers/lor_solvers.cpp
pazner: I suppose Gauss-Lobatto, Gauss-Legendre-Lobatto (GLL), and Legendre-Gauss-Lobatto are all names for the same thing. Perhaps in abbreviations, people like to use GLL for Lobatto, so that GL can be used for Gauss-Legendre without ambiguity.

If we are spelling out the name in full, I think Gauss-Lobatto is fine. It's also consistent with MFEM's `BasisType`.

miniapps/solvers/README
```diff
@@ -61,3 +61,18 @@ The solvers for the above block system include:
...
+the `LOR` and `LORSolver` classes for the low-order refined preconditioning of
+high-order diffusion problems. Preconditioners for H1, H(curl), H(div), and L2
+finite element spaces are constructed.
+
+In serial, a direct solver is used for the LOR system if MFEM is compiled with
+SuiteSparse support (otherwise a simple Gauss-Seidel smoother is constructed
+using the low-order system).
+
+In paralle, hypre's scalable AMG solvers are used for the low-order systems:
```
dylan-copeland, line 76: Typo

miniapps/solvers/lor_mms.hpp
```diff
@@ -0,0 +1,95 @@
...
+   {
+      if (dim == 2)
+      {
+         f[0] = cos(x)*sin(y) + 2*pi2*cos(x)*sin(y);
+         f[1] = cos(y)*sin(x) + 2*pi2*cos(y)*sin(x);
+      }
+      else // dim == 3
+      {
+         double z = pi*xvec[2];
+         f[0] = cos(x)*sin(y)*sin(z) + 3*pi2*cos(x)*sin(y)*sin(z);
```
dylan-copeland, line 71: It would be more efficient to factor out `1 + 3*pi2`.

fem/lor.hpp
pazner: The purpose of overloading the `->` operator is to allow for setting solver-specific parameters, e.g.
```
LORSolver<HypreBoomerAMG> lor_solver(…);
lor_solver->SetRelaxType(…);
```

Of course, if you call `lor_solver->Mult` in this case, it will do the wrong thing, which may be a risk for confusion.

miniapps/solvers/lor_mms.hpp
```diff
@@ -0,0 +1,95 @@
...
+   int dim = xvec.Size();
+   double x = pi*xvec[0], y = pi*xvec[1];
+   if (dim == 2)
+   {
+      u[0] = cos(x)*sin(y);
+      u[1] = sin(x)*cos(y);
+   }
+   else // dim == 3
+   {
+      double z = pi*xvec[2];
```
dylan-copeland, line 50: Since we had to do some work on the projection to support inhomogeneous BC, it seems better to use an exact solution like that in `ex3p` to ensure something doesn't break in the future.

miniapps/solvers/plor_solvers.cpp
```diff
@@ -0,0 +1,220 @@
...
+//
+// The action of the high-order operator is computed using MFEM's partial
+// assembly/matrix-free algorithms (except in the case of L2, which remains
+// future work).
+//
+// Sample runs:
+//
+//    mpirun -np 4 plor_solvers -m ../../data/fichera.mesh -fe h
+//    mpirun -np 4 plor_solvers -m ../../data/fichera.mesh -fe n
+//    mpirun -np 4 plor_solvers -m ../../data/fichera.mesh -fe r
```
dylan-copeland, line 52: Since the exact solution is for the Maxwell problem, the H(div) test does not converge as the order is increased. This could be confusing to users. We could add an exact solution for the H(div) case, or just not output the error in that case.

fem/fe.cpp
```diff
@@ -1338,12 +1338,62 @@ void VectorFiniteElement::ProjectGrad_ND(
...
+            }
+            fine_coarse_mass(k,j) += (fabs(Mkj) < 1e-12) ? 0.0 : Mkj;
+         }
+      }
+   }
+   DenseMatrixInverse fine_mass_inv(fine_mass);
+   fine_mass_inv.Mult(fine_coarse_mass, I);
+}
+
+
```
dylan-copeland, line 1388: Extra line

fem/fe.hpp
```diff
@@ -1958,13 +1975,20 @@ class Poly_1D
...
+      mutable Vector u_aux, d_aux, d2_aux;
+      Basis *auxiliary_basis; // Non-NULL only for etype == Integrated
 
    public:
       /// Create a nodal or positive (Bernstein) basis
       Basis(const int p, const double *nodes, EvalType etype = Barycentric);
       void Eval(const double x, Vector &u) const;
       void Eval(const double x, Vector &u, Vector &d) const;
       void Eval(const double x, Vector &u, Vector &d, Vector &d2) const;
+      void EvalIntegrated(const Vector &d, Vector &i) const;
```
dylan-copeland, line 1989: The `Eval` functions could use some brief documentation, at least to describe the arguments.

dylan-copeland:
If the miniapps will be part of the autotests, then I think the new LOR solver miniapps are sufficient tests, so that we don't need comprehensive unit tests for individual components of the new LOR code. I don't think it is really necessary, but it would still be nice to have tests to verify spectral equivalence.

miniapps/solvers/plor_solvers.cpp
pazner: There should be a different forcing term in the case of H(div). Are you really not seeing convergence? What commands are you running and what errors do you get?

miniapps/solvers/lor_mms.hpp
pazner: Not sure what you mean by this.

miniapps/solvers/plor_solvers.cpp
dylan-copeland: I see now that there is a different forcing term for H(div), but I don't see convergence for this test. Also, the solution plot looks weird on some parts of the boundary.
mpirun -np 4 plor_solvers -m ../../data/fichera.mesh -fe r
-o 3
L2 error: 0.0964843 
-o 4
L2 error: 0.0929569

miniapps/solvers/lor_mms.hpp
pazner: You can also run with a different geometry, which should allow you to test nonzero tangential components.

fem/fe.cpp
dylan-copeland: Thanks, this makes sense. It seems we agree that possible future work (after this PR) may be to add special cases of integrated interpolation for some derived classes to replace the general L2 projection from the base class, and to improve the dense matrix inversion for high order.

miniapps/solvers/plor_solvers.cpp
pazner: Good catch, there was a bug in `RT_HexahedronElement::ProjectIntegrated`. Can you check if it's working properly for you now?

fem/lor.hpp
dylan-copeland: I would also prefer a more descriptive name like `LowOrderRefined`.

fem/lor.hpp
dylan-copeland: So if we replace this function with `GetSolver`, then `lor_solver->SetRelaxType(…);` becomes `lor_solver.GetSolver()->SetRelaxType(…);`. I think that would be safer.

fem/fe.hpp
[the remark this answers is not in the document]
pazner: My vote would be for `IntegratedGLL`.

miniapps/solvers/plor_solvers.cpp
dylan-copeland: Yes, it works now.

fem/lor.hpp
pazner: Ok, I removed the `->` operator, and added `GetSolver` (which returns a reference, rather than a pointer, since the underlying solver can never be NULL), so that now it will be `lor_solver.GetSolver().SetRelaxType(…);`

fem/fe.hpp
```diff
@@ -94,7 +96,8 @@ class BasisType
       static const char *name[] =
       {
          "Gauss-Legendre", "Gauss-Lobatto", "Positive (Bernstein)",
-         "Open uniform", "Closed uniform", "Open half uniform"
+         "Open uniform", "Closed uniform", "Open half uniform",
+         "Seredipity", "Closed Gauss-Legendre", "Integrated"
```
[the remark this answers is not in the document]
pazner: It now returns `"Integrated Gauss-Lobatto indicator"`

fem/fe.hpp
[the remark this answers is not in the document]
pazner: I changed it to `IntegratedGLL`. Let me know if that looks good to you.

fem/fe.hpp
pazner: I added a Doxygen comment here explaining what it is doing.

fem/lor.hpp
pazner: I changed it to `LORDiscretization` (which goes well with the related class, `LORSolver`). `LowOrderRefinedDiscretization` is probably too verbose.

miniapps/solvers/lor_docs.md
pazner: I think the best thing to do is to automatically apply a change of basis, but doing that properly and efficiently is somewhat involves, and will probably happen in a subsequent PR.

dylan-copeland (approved):
Thanks @pazner! This PR is not only very useful but also well written and user-friendly, which is challenging given the complexity of the new capabilities.

miniapps/solvers/lor_docs.md
barker29: It's fine for that to be a follow-on PR, but I think the documentation should be very clear about when a user can expect good performance from an LOR solver (IMO it's not quite clear enough right now).

pazner: I updated the documentation explaining which bases are spectrally equivalent, and I added code that checks this at runtime and emits a warning when non-spectrally equivalent bases are used.

crdohrm (commented):
Did you want to say anything about the BasisType::GaussLobatto requirement for L2 spectral equivalence?

pazner:
> Did you want to say anything about the BasisType::GaussLobatto requirement for L2 spectral equivalence?

The DG case is a bit more complicated. For the mass matrix, `BasisType::GaussLegendre` should be fine, and for the "stiffness matrix" one really wants to consider something like an IPDG discretization, which is a bit more involved.

barker29 (approved):
I'm very happy with how this turned out.

tzanio:

tzanio:
This deserves some coverage in `CHANGELOG` 😁

miniapps/solvers/lor_mms.hpp
```diff
@@ -0,0 +1,95 @@
+#ifndef __LOR_MMS_HPP__
```
tzanio, line 1: I know this is a simple file, but lets still add the COPYRIGHT header

miniapps/solvers/lor_solvers.cpp
```diff
@@ -0,0 +1,203 @@
...
+// compiled with SuiteSparse enabled, or with one iteration of symmetric
+// Gauss-Seidel smoothing otherwise.
+//
+// For vector finite element spaces, the special "Integrated" basis type is used
+// to obtain spectral equivalence between the high-order and low-order refined
+// discretizations. This basis is defined in reference [1] and spectral
+// equivalence is shown in [2]:
+//
+// [1]. M. Gerritsma. Edge functions for spectral element methods. Spectral and
+//      High Order Methods for Partial Differential Equations (2010).
```
tzanio, line 40:
```suggestion
//      High Order Methods for Partial Differential Equations. (2010)
```

miniapps/solvers/lor_solvers.cpp
```diff
@@ -0,0 +1,203 @@
...
+// [2]. C. Dohrmann. Spectral equivalence properties of higher-order tensor
+//      product finite elements and applications to preconditioning. (2021)
+//
+// The action of the high-order operator is computed using MFEM's partial
+// assembly/matrix-free algorithms (except in the case of L2, which remains
+// future work).
+//
+// Sample runs:
```
tzanio, line 48:
```suggestion
//
// Compile with: make lor-solvers
//
// Sample runs:
```

miniapps/solvers/lor_solvers.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+   Vector X, B;
+   OperatorHandle A;
+   a.FormLinearSystem(ess_dofs, x, b, A, X, B);
+
+#ifdef MFEM_USE_SUITESPARSE
+   LORSolver<UMFPackSolver> solv_lor(a, ess_dofs);
+#else
+   LORSolver<GSSmoother> solv_lor(a, ess_dofs);
+#endif
```
tzanio, line 167: This whole miniapp is very clean. I really like it!

miniapps/solvers/lor_solvers.cpp
```diff
@@ -0,0 +1,203 @@
...
+   }
+
+   if (ND) { a.AddDomainIntegrator(new CurlCurlIntegrator); }
+   else if (RT) { a.AddDomainIntegrator(new DivDivIntegrator); }
+   else if (L2)
+   {
+      a.AddInteriorFaceIntegrator(new DGDiffusionIntegrator(-1.0, kappa));
+      a.AddBdrFaceIntegrator(new DGDiffusionIntegrator(-1.0, kappa));
+   }
+   // TODO: L2 diffusion not implemented with partial assemble
```
tzanio, line 141:
```suggestion
   // TODO: L2 diffusion not implemented with partial assembly
```

miniapps/solvers/plor_solvers.cpp
```diff
@@ -0,0 +1,223 @@
...
+// at the Lawrence Livermore National Laboratory. All Rights reserved. See files
+// LICENSE and NOTICE for details. LLNL-CODE-806117.
+//
+// This file is part of the MFEM library. For more information and source code
+// availability visit https://mfem.org.
+//
+// MFEM is free software; you can redistribute it and/or modify it under the
+// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// CONTRIBUTING.md for details.
+//
```
tzanio, line 11: See the comments in the serial version

tzanio (approved):
Really awesome and powerful stuff 🎉 

Thanks @pazner and @crdohrm !

miniapps/solvers/makefile
```diff
@@ -21,34 +21,42 @@ CONFIG_MK = $(MFEM_BUILD_DIR)/config/config.mk
 MFEM_LIB_FILE = mfem_is_not_built
 -include $(CONFIG_MK)
 
-BLOCK_COMMON_SRC = div_free_solver.cpp
-BLOCK_COMMON_OBJ = $(BLOCK_COMMON_SRC:.cpp=.o)
+BLOCK_SOLVERS_SRC = div_free_solver.cpp block-solvers.cpp
+BLOCK_SOLVERS_OBJ = $(BLOCK_SOLVERS_SRC:.cpp=.o)
```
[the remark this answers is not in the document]
pazner: I think it's fine to keep it. The miniapp executable and source file are still named `block-solvers`.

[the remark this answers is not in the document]
tzanio: ok

v-dobrev:
I'm seeing a build error from this PR when using out-of-source build with GNU make -- `make all` fails due to this: https://github.com/mfem/mfem/blob/4fa50723b938945dc772bb4c95641ce2191a6e62/miniapps/solvers/makefile#L48-L49
For out-of-source builds to work, the dependency should be `$(SRC)lor_mms.hpp`.

I wonder why the autotests did not catch this. Maybe the Gitlab CI will catch this?

tzanio:
I don't know if the out-of-source GNU make builds are actually tested. We should check that and add it to https://github.com/mfem/mfem/issues/1255.

If the above is the only fix needed, lets just add it in the mfem-4.3-dev PR.

pazner:
Sorry about that! I'm pretty sure this is not caught by the GitLab tests. Is that the only change that is needed?

v-dobrev:
I think the change I mentioned above is the only thing needed -- we should double check of course. :-)

One option could be to switch the Mac builds in the CI to use out-of-source builds.

tzanio:
Addressed in 6bad77ca147a60d2c86d4fa4a16bd4ebbd440569

PR #2096, opened 2021-03-11
[WIP] Experimental triangle bisection with Sierpinski SFC [nc-tri-bisect-dev]

jakubcerveny:
I'm thinking how to improve our conforming refinement on triangles. Currently:
* A combination of red and green refinements is done in the conforming code. The green refinements stay in the mesh and may lead to a degradation of mesh quality.
* Conforming refinement does not support load balancing.
* Nonconforming triangle refinement is not ideal either, because a nice continuous SFC cannot be defined for red refinement. Also, red refinement may prefer certain directions in some cases (see the `shaper` miniapp on `inline-tri.mesh`).

This is a simple modification of NCMesh to allow a special refinement type for triangles. If `ref_type = 1`, a triangle is bisected. The newest vertex bisection is implemented, which naturally leads to the Sierpinski space-filling curve:

sierp1

(The coarse mesh was just two triangles.)

With local refienement we of course get hanging nodes. The following mesh is fully functional, but is a bit awkward:

sierp2

But by trivial generalization of the NC-limiting code (allow `nc_limit = 0`), we can get a fully conforming mesh:

sierp3

There are some limitations currently:
* ~~The first vertex of a triangle needs to be the one opposite to the longest edge (this could be circumvented, even without physically rotating the elements).~~
* The coarse mesh ordering as currently defined (sequence of face-neighbors) is not the right thing for this SFC. W. Mitchell has a paper on this (Hamiltonian paths in meshes?), maybe it could be adapted. Maybe Gecko could still be used, but on a different graph (primal instead of dual).
* It is unclear how derefinement would work while preserving conformity.

Is this something that would be worth pursuing?

### DIFF
--- a/data/inline-quad.mesh
+++ b/data/inline-quad.mesh
@@ -1,7 +1,7 @@
 MFEM INLINE mesh v1.0
 
 type = quad
-nx = 4
-ny = 4
+nx = 8
+ny = 8
 sx = 1.0
 sy = 1.0
--- a/data/inline-tri.mesh
+++ b/data/inline-tri.mesh
@@ -1,7 +1,7 @@
 MFEM INLINE mesh v1.0
 
 type = tri
-nx = 4
-ny = 4
+nx = 1
+ny = 1
 sx = 1.0
 sy = 1.0
--- a/examples/ex1.cpp
+++ b/examples/ex1.cpp
@@ -107,19 +107,32 @@ int main(int argc, char *argv[])
    //    the same code.
    Mesh mesh(mesh_file, 1, 1);
    int dim = mesh.Dimension();
+   mesh.EnsureNCMesh(true);
 
    // 4. Refine the mesh to increase the resolution. In this example we do
    //    'ref_levels' of uniform refinement. We choose 'ref_levels' to be the
    //    largest number that gives a final mesh with no more than 50,000
    //    elements.
    {
-      int ref_levels =
-         (int)floor(log(50000./mesh.GetNE())/log(2.)/dim);
+      int ref_levels = 12;
+         //(int)floor(log(50000./mesh.GetNE())/log(2.)/dim);
       for (int l = 0; l < ref_levels; l++)
       {
-         mesh.UniformRefinement();
+         //mesh.UniformRefinement();
+         Array<Refinement> refs;
+         for (int i = 0; i < mesh.GetNE(); i++)
+         {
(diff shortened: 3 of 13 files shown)


jakubcerveny:
Some more fun with the `shaper` miniapp:

shaper5
shaper6

Simulated partitioning (7 CPUs):
shaper-part

v-dobrev:
Looks great!

How does it look for `data/square-disc.mesh` (a little more unstructured triangle mesh)?

jakubcerveny:
> How does it look for data/square-disc.mesh (a little more unstructured triangle mesh)?

OK, so I switched to the longest edge bisection method to test this (didn't want to reorient the mesh). It seems to work correctly.

Uniform refinement test:
square-disc-2

Slightly tweaked `shaper` test, current conforming refinement:
sd-2-conf

"NC" longest edge bisection (it's hard to compare because the standard method progresses faster due to the red refinements):
sd-2-nc

You can download the images and compare by switching them. The test actually favors the current algorithm because the green refinements never need to be refined again here, so the angles don't degrade.

v-dobrev:
Sorry my question wasn't clear: I was asking about the SFC curve -- can that be defined on `data/square-disc.mesh` and how does it look? If not, I suppose you envision using Gecko for reordering the initial mesh and then the Sierpinski SFC will naturally arise with the NC bisection algorithm?

v-dobrev:
Sorry my question wasn't clear: I was asking about the SFC curve -- can that be defined on `data/square-disc.mesh` and how does it look? If not, I suppose you envision using Gecko for reordering the initial mesh and then the Sierpinski SFC will naturally arise with the NC bisection algorithm?

jakubcerveny:
Oh, OK, no problem. This is the coarse mesh ordered with Gecko:
gecko1

After uniform refinement, the Sierpinski curve emerges in regions where the original elements were, but globally the SFC typically doesn't connect smoothly from one coarse region to the next: 
gecko2

Still, there is a lot of locality, so the partitions aren't a disaster:
gecko4

The problem is that the curve within the coarse triangles cannot be rotated as in cubes and hexes: one is limited by having to split by the longest edge, so the ideal mesh ordering is a sequence of longest edges of the individual triangles, which is not easy to get. Cubes and hexes are more flexible, we can typically find an orientation of the root elements that will produce a globally continuous SFC (see `NCMesh::InitRootState`), if the coarse mesh is a sequence of face-neighbors.

PR #2114, opened 2021-03-22
Allow amgx solver to be used with Hypre BigInt

wcdawn:
If `--enable-bigint` is used with HYPRE, there is a type conflict between `HYPRE_Int` and `int` in the `AmgxSolver`. This PR allows the AmgX solver to be used when HYPRE is compiled with BigInt.

More generally, there seem to be many little places where conflicts like this occur when using HYPRE with BigInt. If there is a better way to fix all of them at once, please let me know.

Thanks!

| [#2114](https://github.com/mfem/mfem/pull/2114) | @wcdawn | @tzanio | @artv3 + @tzanio + @dylan-copeland | 03/23/21 | 04/08/21 | 04/10/21 | |


artv3:
> If `--enable-bigint` is used with HYPRE, there is a type conflict between `HYPRE_Int` and `int` in the `AmgxSolver`. This PR allows the AmgX solver to be used when HYPRE is compiled with BigInt.
> 
> More generally, there seem to be many little places where conflicts like this occur when using HYPRE with BigInt. If there is a better way to fix all of them at once, please let me know.
> 
> Thanks!

Hi @wcdawn just to double check, this still works if Hypre is not built with big int right?

wcdawn:
@artv3 Yes! This will work regardless. Typically, `HYPRE_Int` is an alias for `int` but if HYPRE is compiled with `--enable-bigint`, `HYPRE_Int` is fixed as a 64-bit integer which causes a type discrepancy. By using `HYPRE_Int` here instead of `int`, this ensures that the types will agree regardless of the underlying type of `HYPRE_Int`.

artv3:
I think we would want to propagate those changes to the following methods as well right? 
```
      SetMatrixMPIGPUExclusive(A, loc_A, loc_I, loc_J, update_mat);
      SetMatrixMPITeams(A, loc_A, loc_I, loc_J, update_mat);
      GatherArray(loc_I, all_I, devWorldSize, devWorld);
```

Otherwise we might get type narrowing when `HYPRE_Int` is a 64-bit integer right?

wcdawn:
I looked into this recommendation. At some point, the narrowing conversion is unavoidable because the AmgX library uses `int` throughout. For example, `AmgXSolver::SetMatrixMPIGPUExclusive()` calls `AMGX_matrix_upload_distributed()` which uses `int*` in the function signature. Additionally, the MPI calls in `AmgXSolver::SetMatrixMPITeams()` would need to check whether to use `MPI_INT` or `MPI_LONG_LONG`.

That being said, using a `static_cast` would be preferable to make the possibility for narrowing conversion more obvious.

As far as `GatherArray()`, I'm not clear. There are a lot of overloaded functions and I didn't make it to reading all of them.

artv3:
Gotcha, it probably is the case that we wouldn't have issues with casting `A_csr->i` down to an int since it corresponds to local row numbers. I think this change is reasonable, perhaps add some comments and use the `static_cast`?  Additionally, if someone a bit more familiar with Hypre internals could comment that would really help.

tzanio:

dylan-copeland (changes requested):
Thanks @wcdawn for fixing this. I tested `examples/amgx/ex1p` and confirmed that this PR fixes the `bigint` problem for amgx, but I also found that one of the unit tests, `tests/unit/linalg/test_constrainedsolver.cpp`, fails to build with `bigint`. It seems to need simple changes from `int` to `HYPRE_Int` type for some arrays. For some `SparseMatrix` arrays, we cannot change the type without affecting that class, so it may be best just to deep-copy those arrays to the `HYPRE_Int` type. I can do that for you if you are not interested in fixing that as well.

wcdawn:
@dylan-copeland  I've fixed `test_constrainedsolver` to compile with BigInt. Please let me know if you find any more tests failing.

Thanks!

dylan-copeland:
There is still one unit test that fails to build, `tests/unit/mesh/test_pmesh.cpp`. The problem is the `Min` and `Max` for `Array<HYPRE_Int>`. I added 
`#include "seq_mv.h"`
`template class Array<HYPRE_Int>;`
to `general/array.cpp` and then all the unit tests passed, as well as `amgx/ex1p`.

wcdawn:
Thanks for catching this!

I think your recommended patch may fail for builds without HYPRE (i.e. serial builds). We also cannot always instantiate `Array<HYPRE_Int>` because we cannot instantiate `Array<int>` more than once. One option is to modify `general/array.cpp` to
- `#include "./general/hypre.hpp"
- And guard using
```cpp
#if defined(MFEM_USE_MPI) && defined(HYPRE_BIGINT)
template class Array<Hypre_Int>;
#enidf
```

Alternatively, I propose that we simply instantiate `Array<long long>`. This avoids the dependency between `array.cpp` and `hypre.hpp` and does not appreciably increase the size of the library (<1MB for a 51MB library).

Please let me know if you would like a different method for solving this problem.

dylan-copeland (approved):

tzanio:
I am actually concerned about the `Array<long long>` solution.

We don't have control on `HYPRE_BIGINT` and while it may frequently be `long long` there could be exceptions which will lead to weird bugs.

Can't we just instantiate `Array<HYPRE_BigInt>` with the guards you suggested in `linalg/hypre.cpp`?

wcdawn:
@tzanio I don't know exactly why but that doesn't work. The linker cannot find `Array<HYPRE_Int>` if it's instantiated within `hypre.cpp`. If someone who knows more than me would like to take a look, it would be appreciated.

Alternatively, we could `static_assert` that `HYPRE_Int` is a `long long` (as specified in the HYPRE documentation). Something like the following in `hypre.cpp`.
```cpp
#ifdef HYPRE_BIGINT
namespace
{
   void hypre_bigint_size_check() __attribute__((unused));
   void hypre_bigint_size_check()
   {
      static_assert(sizeof(HYPRE_Int) == sizeof(long long),
                    "HYPRE_BIGINT is enabled and HYPRE_Int is not \"long long\". "
                    "An additional instantiation is necessary in general/array.cpp.");
   }
} // namespace unnamed
#endif
```

tzanio:
Ok, lets keep it like that for now.

@artv3, can you please review?

tzanio:
LGTM

artv3 (approved):

tzanio:

tzanio (approved):

PR #2130, opened 2021-03-24
Replace SaveWithMesh with separate Save functions

pazner:
(Alternative saving interface for #942)

What do you think about this interface for saving GridFunctions and Meshes @jakubcerveny @tzanio?

### DIFF
--- a/fem/gridfunc.hpp
+++ b/fem/gridfunc.hpp
@@ -671,10 +671,9 @@ public:
    /// Save the GridFunction to an output stream.
    virtual void Save(std::ostream &out) const;
 
-   /// Convenience function to save the GridFunction and associated Mesh to
-   /// files. The given @a precision will be used for ASCII output.
-   virtual void SaveWithMesh(const char *sol_f, const char *mesh_f,
-                             int precision=8) const;
+   /// Save the GridFunction to a file. The given @a precision will be used for
+   /// ASCII output.
+   virtual void Save(const char *fname, int precision=8) const;
 
 #ifdef MFEM_USE_ADIOS2
    /// Save the GridFunction to a binary output stream using adios2 bp format.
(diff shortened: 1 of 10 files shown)


pazner:
@adrienbernede, are the GitHub Actions supposed to be run on PRs whose base branch is not master? I don't see them here.

tzanio:
Either option is fine with me :)

tzanio:
> @adrienbernede, are the GitHub Actions supposed to be run on PRs whose base branch is not master? I don't see them here.

Based on https://github.com/mfem/mfem/tree/example0-save-gf-mesh/.github/workflows I believe the GH actions are run only for PRs toward `master`

pazner:
It might make sense to change the workflows so that they are run on all PRs.

fem/gridfunc.hpp
```diff
@@ -671,10 +671,9 @@ class GridFunction : public Vector
    /// Save the GridFunction to an output stream.
    virtual void Save(std::ostream &out) const;
```
jakubcerveny, line 672: We could add a `Print` method to mirror the `Mesh` interface, and make `Save(std::ostream&)` call it. I'm just not sure it's not too much :-)

jakubcerveny (approved):
I like this, thanks!

fem/gridfunc.hpp
tzanio: I'd prefer to keep it simple for the users and unify to `Save`.

The Save/Print has been pointed out as one of the interface inconsistencies in https://github.com/mfem/mfem/issues/1557

tzanio (approved):

barker29 (approved):
I think this is slightly better than what's currently in #942 but do not feel strongly about it, either way is fine.

PR #2132, opened 2021-03-25
Allow user to specify initial guess in AmgXSolver

wcdawn:
I'm using an `AmgXSolver` in a power iteration eigensolver algorithm so the linear solver is called many times. Eventually, I have a pretty good initial guess for the solution vector. I can cut my runtimes in half by specifying an initial guess instead of a zero guess.

| [#2132](https://github.com/mfem/mfem/pull/2132) | @wcdawn | @tzanio | @artv3 + @barker29 | 03/25/21 | 03/25/21 | 03/27/21 | |

### DIFF
--- a/linalg/amgxsolver.hpp
+++ b/linalg/amgxsolver.hpp
@@ -168,11 +168,6 @@ public:
    /// Add a check for convergence after applying Mult.
    void SetConvergenceCheck(bool setConvergenceCheck_=true);
 
-   /** Choice whether solver will use a zero initial guess (default/true) or the
-       value of the X vector (false).
-   */
-   void SetZeroGuess(bool val) { zero_guess = val; }
-
    ~AmgXSolver();
 
    void Finalize();
--- a/linalg/amgxsolver.hpp
+++ b/linalg/amgxsolver.hpp
@@ -313,8 +313,6 @@ private:
    int64_t mat_local_rows;
 
    std::string mpi_gpu_mode;
-
-   bool zero_guess = true;
 };
 }
 #endif // MFEM_USE_AMGX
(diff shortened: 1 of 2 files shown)


artv3 (approved):
Looks good to me, but I wonder if false should be the default?

wcdawn:
I set it this way to not change the default behavior of the code but I agree that setting `zero_guess=false` by default would probably be better. `zero_guess=false` would reflect the default behavior of, for example, `CGSolver`.

tzanio:
It really depends if we expect that its typical use is as a solver or as a preconditioner. In the former case, `zero_guess` should be `false`. In the latter case, it should be `true`.

tzanio:

linalg/amgxsolver.hpp
```diff
@@ -313,6 +318,8 @@ class AmgXSolver : public Solver
    int64_t mat_local_rows;
 
    std::string mpi_gpu_mode;
+
+   bool zero_guess = true;
```
barker29, line 322: Following @tzanio 's comment, can you set the `zero_guess` value based on `amgxMode` in `DefaultParameters()` method so it's `false` for a solver and `true` for a preconditioner?

barker29 (changes requested):
Other than my minor suggestion, this PR is a nice improvement.

wcdawn:
I've changed how this is implemented to avoid having to choose a default and be more consistent with the rest of MFEM. Now, the behavior is based on the `iterative_mode` boolean from the parent `Solver` class. 

I've also implemented @barker29 request to set the iterative mode based on whether the class is being used as a solver or preconditioner. As such, I've removed the setter function because it should be unnecessary now.

Thanks for the help!

barker29 (approved):
Looks good!

tzanio:
@artv3 can you confirm you are OK with the latest changes?

artv3:
Yup looks good to me!

tzanio:

tzanio (approved):
Thanks @wcdawn !

PR #2145, opened 2021-04-01
[WIP] Alternative solution to ReorientTetMesh [alt-tet-reorder-dev]

jakubcerveny:
This is an experiment to see if a less intrusive solution than #1046 is possible.

Somewhat in the spirit of `var-order-space-dev`, this PR introduces special faces with two sets of DOFs on them, called "double faces", which are used for Nedelec triangle faces of order >= 2 and orientation 1,2,3,4. The disjoint DOFs are then constrained with the P matrix.

The advantage is that handling of complex orientations is completely hidden inside the FiniteElementSpace.

The disadvantages that I see are 
* the Mesh needs to be NC,
* the concept of double faces is slightly ugly,
* boundary elements still need to be reoriented,
* there are extra DOFs (although they are added only when necessary, and their number can still be minimized with ReorientTetMesh),
* this is a new kind of space so GLVis currently doesn't handle this (I need to project to a vector L2 in ex3),
* if we decide to use this, the I/O of the space will need to be made rock-solid (in a new serialization format probably), or we will be entering a world of pain :-)
* the discontinuous faces might mess up ProjectCoefficient and/or something else.,
* right now this is serial only.

The modified `ex3` should now run without `ReorientTetMesh`, e.g. on `inline-tet.mesh`.

| [#2145](https://github.com/mfem/mfem/pull/2145) | @jakubcerveny | @tzanio | @psocratis + @mlstowell + @v-dobrev | 04/01/21 | ⌛due 06/03/21 | ⌛due 06/10/21 | |

### DIFF
--- a/examples/ex3.cpp
+++ b/examples/ex3.cpp
@@ -106,14 +106,15 @@ int main(int argc, char *argv[])
    //    largest number that gives a final mesh with no more than 50,000
    //    elements.
    {
-      int ref_levels =
-         (int)floor(log(50000./mesh->GetNE())/log(2.)/dim);
+      int ref_levels = 0;
+         //(int)floor(log(50000./mesh->GetNE())/log(2.)/dim);
       for (int l = 0; l < ref_levels; l++)
       {
          mesh->UniformRefinement();
       }
    }
-   mesh->ReorientTetMesh();
+   mesh->EnsureNCMesh(true);
+   //mesh->ReorientTetMesh();
 
    // 5. Define a finite element space on the mesh. Here we use the Nedelec
    //    finite elements of the specified order.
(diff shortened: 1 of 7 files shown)


tzanio:

mlstowell:
I'm impressed that this required so few changes to the core classes.

I think it's important to note that PR #1046 supports the standard `mfem::Mesh` and `mfem::ParMesh` objects but runs into trouble with the `mfem::NCMesh` and `mfem::ParNCMesh` objects. Whereas this PR supports the `mfem::NCMesh` object and makes no attempt to support the `mfem::Mesh` and `mfem::ParMesh` objects.

It may be best to blend these two solutions together so that we support both standard and NC mesh types. Ideally we could implement the necessary changes to `mfem::ParNCMesh` and remove the need for reorienting the boundary elements. This would place no special restrictions on the user relative to the other basis function types.

tzanio:
This PR is no longer under active review, so I will remove the `in-review` label. The work will be reviewed as part of a future PR for parallel hp support.

tzanio:
Closed in favor of #1046.

Issue #2152, opened 2021-04-03
 FiniteElementSpace::BuildDofToArrays(): Suspected memory issue.


woutertonnon:
Hello all,

I am using a finite-element space based on first-order `ND_FECollection` on a simple mesh in 2D. I noticed that `FiniteElementSpace::GetElementForDof()` returns `-1` for certain DoF's. I tracked this back through the code and noticed that in `FiniteElementSpace::BuildDofToArrays()`, there is a loop over all the elements in the mesh. Using the ElementToDofTable, the DoF's associated to each element are found and used to define the DofToElementTable, i.e. `dof_elem_array[dofs[j]]` is used. However, in the case of the above-specified FiniteElementSpace, ElementToDofTable can return negative values for the DoFs, which seems to be associated to the orientation of the edge with respect to the element. This means that the behaviour of `dof_elem_array[dofs[j]]` is undefined as I understand. An easy solution would be to replace all `dofs[j]` in `FiniteElementSpace::BuildDofToArrays()` by `abs(dofs[j])`. Let me know what you think.

mlstowell:
Hello, @woutertonnon,

This was a good catch. I don't know how this bug has persisted for so long. I suppose this method hasn't been tested with ND and RT spaces. The correct fix is to replace `dofs[j]` with `-dofs[j]-1` when `dofs[j]` is negative.

Thanks for pointing this out.
Best wishes,
Mark

tzanio:
Discussion continued in #2153

PR #2153, opened 2021-04-03
Flipping negative dofs in BuildDofToArrays [bugfix/build-dof-to-arrays]

mlstowell:
Quick PR to address issue #2152.

This seems like something that should have been caught with a unit test. I'm not sure of an appropriate use case so if anyone has suggestions please comment or add a suitable test.

| [#2153](https://github.com/mfem/mfem/pull/2153) | @mlstowell | @tzanio | @tzanio + @pazner | 04/04/21 | 04/07/21 | 04/10/21 | |

### DIFF
--- a/fem/fespace.cpp
+++ b/fem/fespace.cpp
@@ -357,10 +357,11 @@ void FiniteElementSpace::BuildDofToArrays()
       const int n = elem_dof -> RowSize(i);
       for (int j = 0; j < n; j++)
       {
-         if (dof_elem_array[dofs[j]] < 0)
+         int dof = (dofs[j] >= 0) ? dofs[j] : - dofs[j] - 1;
+         if (dof_elem_array[dof] < 0)
          {
-            dof_elem_array[dofs[j]] = i;
-            dof_ldof_array[dofs[j]] = j;
+            dof_elem_array[dof] = i;
+            dof_ldof_array[dof] = j;
          }
       }
    }
--- a/tests/unit/fem/test_build_dof_to_arrays.cpp
+++ b/tests/unit/fem/test_build_dof_to_arrays.cpp
@@ -15,7 +15,7 @@
 namespace mfem
 {
 
-namespace for_dof
+namespace build_dof_to_arrays
 {
 
 static double a_ = M_PI;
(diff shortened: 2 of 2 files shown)


tzanio:

woutertonnon (approved):

woutertonnon:
I find it difficult to come up with a unit test that covers all the functionality of this method. However, an idea for a unit test would be the following. We consider a simple mesh in 2D with edges on the boundary that are negatively oriented  with respect to the unique element they belong to. (I need to double-check this, but probably star.mesh satisfies this property). Next, we choose a finite-element space (e.g. Nedelec, Raviart-Thomas) and use `FiniteElementSpace::BuildDofsToArrays()` to build the necessary array. We then evaluate this array for every dof index using the method `int FiniteElementSpace::GetElementForDof(int dof_ind)`. If everything is working correctly, we expect the output to always be non-negative. In the case where negative dofs are not processed appropriately, the corresponding element index will be written to an arbitrary place in memory. This leaves the element in the array corresponding to the dof at the default value, i.e. `-1`, and we would catch the error. If you think this test is suitable, I can write the code tonight.

mlstowell:
Hi, @woutertonnon,

Unfortunately, I don't think that scheme will succeed. Negatively oriented boundary elements are caught, and "fixed", by during the mesh construction process.

Perhaps we just need a simple unit test which calls `GetElementForDof` and `GetLocalDofForDof` for every degree of freedom on a small mesh and confirms the results. Or perhaps not so simple. We would need to check various basis function types on various element types both in serial and parallel. I think I have a test with a good collection of meshes but it's not in master yet.  I'll try to put something together.

Thanks for the feedback!
Best wishes,
Mark

fem/fespace.cpp
```diff
@@ -357,10 +357,11 @@ void FiniteElementSpace::BuildDofToArrays()
       const int n = elem_dof -> RowSize(i);
       for (int j = 0; j < n; j++)
       {
-         if (dof_elem_array[dofs[j]] < 0)
+         int dof = (dofs[j] >= 0) ? dofs[j] : - dofs[j] - 1;
```
pazner, line 360:
```suggestion
         int dof = DecodeDof(dofs[j]);
```

A bit more self-documenting this way

pazner (approved):

tzanio:
LGTM

tzanio (approved):

tzanio:

PR #2160, opened 2021-04-08
Add ProjectCoefficient method to GridFunction restricted to certain element attributes

tuckerbabcock:
The PR adds a ProjectCoefficient method to the GridFunction class that restricts the projection to elements of the specified attribute. I felt I needed this to have more fine grained control over the order of projection with discontinuous coefficients compared to `ProjectDiscCoefficient`, where I don't want to average the values or have the element with the highest attribute "win".

I think it would make sense to offer this function with a `std::unordered_set` argument so that a subset of attributes could be efficiently projected to at once, but I didn't include that since I haven't seen any other APIs use STL containers. Is there a policy against not using STL containers in MFEM? The same idea could be conveyed using an `Array<int>` argument, but that signature is already taken by the version of ProjectCoefficient that specifies the dofs to project to.

Please let me know what you think, and if there is already a way to project a coefficient onto a subset of elements based on their attribute.

### DIFF
--- a/fem/gridfunc.cpp
+++ b/fem/gridfunc.cpp
@@ -2377,6 +2377,24 @@ void GridFunction::ProjectCoefficient(
    }
 }
 
+void GridFunction::ProjectCoefficient(VectorCoefficient &vcoeff, int attribute)
+{
+   int i;
+   Array<int> vdofs;
+   Vector vals;
+
+   for (i = 0; i < fes->GetNE(); i++)
+   {
+      if (fes->GetAttribute(i) != attribute)
+         continue;
+
+      fes->GetElementVDofs(i, vdofs);
+      vals.SetSize(vdofs.Size());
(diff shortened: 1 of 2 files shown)


tzanio:

acfisher:
Yes we typically avoid using stl containers on this project.  There are instances where we do use them, but we have to justify their use before we accept them.  For fine grained control of the projections we just use the Array<int> version that you mentioned and list the DOFs that we want to set.  It would be nice to have a few more convenience methods like you have added here though.

acfisher (approved):
It will be good to have this convenience method.

tuckerbabcock:
Thanks @acfisher. Just for my own curiosity, is the reasoning behind avoiding stl containers written somewhere?

mlstowell (approved):

tzanio:

PR #2206, opened 2021-04-29
[WIP] Utilities for matrix-free element-wise smoother with Fast Diagonalization

psocratis:
- [x] Lapack Eigensolver for general matrices
- [x] Kronecker Mult helper functions
- [x] Fast Diagonalization solver for systems where the matrix is given as a sum of Kronecker products
- [x] Unit tests
- [x] Valgrind


tests/unit/linalg/test_fdsolver.cpp
```diff
@@ -0,0 +1,127 @@
...
+                       };
+
+
+   SECTION("2D")
+   {
+      Array<DenseMatrix *> A(2), B(2);
+      A[0] = new DenseMatrix(A0Data, 3, 3);
+      A[1] = new DenseMatrix(A1Data, 4, 4);
+      B[0] = new DenseMatrix(B0Data, 3, 3);
+      B[1] = new DenseMatrix(B1Data, 4, 4);
```
pazner, line 54: Note you can use braced initializer lists to create a `DenseMatrix`, e.g.
```c++
   DenseMatrix A(
   {
      {6.0, 4.0, 2.0},
      {5.0, 3.0, 1.0}
   });
```
Each inner list is a row of the matrix.

tests/unit/linalg/test_fdsolver.cpp
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) 2010-2021, Lawrence Livermore National Security, LLC. Produced
```
pazner, line 1: Nice tests! 😄

linalg/fdsolver.hpp
```diff
@@ -0,0 +1,60 @@
...
+/// appropriate size
+void KronMultInvDiag(const Vector & a, const Vector & b, Vector & dinv);
+
+/// Computes the inverse diagonal dinv = (a⊗I⊗I + I⊗b⊗I + I⊗I⊗c)^-1
+/// where a, b, c are diagonal matrices and I is the identity of the
+/// appropriate size
+void KronMultInvDiag(const Vector & a, const Vector & b,
+                     const Vector & c, Vector & dinv);
+
+void KronMultInvDiag(const Array<Vector *> & X, Vector & diag);
```
pazner, line 32: Maybe `KronMult` is the wrong term here, since these construct the diagonal rather than multiply by it. Perhaps `KronProdInvDiag`?

linalg/fdsolver.cpp
```diff
@@ -0,0 +1,137 @@
...
+         for (int i = 0; i<n; i++)
+         {
+            dinv(i*m*l+j*l+k) = 1.0/(a(i) + b(j) + c(k));
+         }
+}
+
+void KronMultInvDiag(const Array<Vector *> & X, Vector & diag)
+{
+   int dim = X.Size();
+   if (dim == 2)
```
pazner, line 46: Maybe `dim == 1` makes sense as a trivial case here?

linalg/densemat.hpp
```diff
@@ -685,27 +701,44 @@ class DenseMatrixInverse : public MatrixInverse
...
    DenseMatrix EVect;
    Vector ev;
    int n;
 
 #ifdef MFEM_USE_LAPACK
+   bool sym;
    double *work;
    char jobz, uplo;
    int lwork, info;
 #endif
```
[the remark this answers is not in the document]
psocratis: Good point, indeed it should be. That was the current state already and so I left it as it was. I can fix it.

linalg/densemat.hpp
```diff
@@ -685,27 +701,44 @@ class DenseMatrixInverse : public MatrixInverse
...
 public:
 
-   DenseMatrixEigensystem(DenseMatrix &m);
+   DenseMatrixEigensystem(DenseMatrix &m, bool sym_ = false);
    DenseMatrixEigensystem(const DenseMatrixEigensystem &other);
    void Eval();
-   Vector &Eigenvalues() { return EVal; }
+   Vector &Eigenvalues(bool imag = false) { return imag ? EVali : EVal; }
    DenseMatrix &Eigenvectors() { return EVect; }
    double Eigenvalue(int i) { return EVal(i); }
```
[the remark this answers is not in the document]
psocratis: I am hesitant to start adding complex support here, and the reason is that so far we kept anything to do with complex numbers separately in `complex_operator.[hc]pp` and `complex_fem.[hc]pp`.  I can add `RealPartEigenvalues` and `ImagPartEigenvalues` as extra methods but keep the current `Vector &Eigenvalues` there, just to be backward compatible.

linalg/vector.cpp
```diff
@@ -216,6 +216,18 @@ Vector &Vector::operator+=(const Vector &v)
...
+{
+   MFEM_ASSERT(size == v.size, "incompatible Vectors!");
+
+   const bool use_dev = UseDevice() || v.UseDevice();
+   const int N = size;
+   auto y = ReadWrite(use_dev);
+   auto x = v.Read(use_dev);
+   MFEM_FORALL_SWITCH(use_dev, i, N, y[i] *= x[i];);
+   return *this;
+}
```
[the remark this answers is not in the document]
psocratis: I noticed that this is already added here: ee8891d, so it can be removed eventually.

tzanio:
@psocratis, can we delete this branch?

PR #2218, opened 2021-05-03
Add minimal memory manager failing tests with NewMemoryAndSize / Read…

camierjs:
Adds a few tests to capture the following behaviour which leads to a MemoryClassCheck "Unknown host pointer" error.
```
Vector x(1);
x.UseDevice(true);
{
   Vector X;
   X.NewMemoryAndSize(x.GetMemory(), x.Size(), false);
   X.Read();
   x.SyncMemory(X);
}
x.Read();
```

### DIFF
--- a/tests/unit/general/test_mem.cpp
+++ b/tests/unit/general/test_mem.cpp
@@ -111,4 +111,60 @@ TEST_CASE("MemoryManager", "[MemoryManager]")
    }
 }
 
+TEST_CASE("MemoryManager/Scopes",
+          "[MemoryManager]"
+          "[CUDA]")
+{
+   SECTION("WithNewMemoryAndSize")
+   {
+      Vector x(1);
+      x.UseDevice(true);
+      {
+         Vector X;
+         // from Operator::InitTVectors
+         X.NewMemoryAndSize(x.GetMemory(), x.Size(), false);
+         // from Vector::SetSubVectorComplement
+         X.Read();
+         // from Operator::RecoverFEMSolution
+         x.SyncMemory(X);
+      }
+      // Accessible Memory<double> to get the flags
+      struct MemoryDouble
+      {
+         double *h_ptr;
+         int capacity; ///< Size of the allocated memory
+         MemoryType h_mt; ///< Host memory type
+         mutable unsigned flags;
+      };
+      const MemoryDouble *mem = (MemoryDouble*) &x.GetMemory();
+      const double *h_x = mem->h_ptr;
+      REQUIRE(h_x == x.GetData());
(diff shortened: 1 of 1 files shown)


tzanio:
We do need a formal review for this PR (since it goes in a PR under review).

@v-dobrev and @YohannDudouit, can you just comment/approve?

tzanio (approved):
Thanks @camierjs, lets add more of these!

tests/unit/general/test_mem.cpp
```diff
@@ -12,11 +12,68 @@
...
+      x.UseDevice(true);
+      const double *x_data = x.GetData();
+      {
+         Vector X;
+         // from Operator::InitTVectors
+         X.MakeRef(x, 0, x.Size());
+         // from Vector::SetSubVectorComplement
+         X.Read();
+         // from Operator::RecoverFEMSolution
+         x.SyncMemory(X);
```
YohannDudouit, line 67: I think you're supposed to use the following since you used `MakeRef` (and not `NewMemoryAndSize(...,...,false)`):
```suggestion
         x.SyncAliasMemory(X);
```

camierjs: You're probably true. The idea here is to show what is called through `BilinearForm::RecoverFEMSolution`, which should open the discussion.

camierjs:
LGTM

v-dobrev (approved):

tests/unit/general/test_mem.cpp
```diff
@@ -12,11 +12,68 @@
...
+      const double *h_x = mem->h_ptr;
+      REQUIRE(h_x == x.GetData());
+      REQUIRE(mem->capacity == x.Size());
+      REQUIRE(mem->h_mt == Device::GetHostMemoryType());
+      constexpr unsigned REGISTERED = 1 << 0;
+      const bool registered = mem->flags & REGISTERED;
+      const bool registered_is_known = registered == mm.IsKnown(h_x);
+      // Failing CUDA test with NewMemoryAndSize/Read/SyncMemory
+      if (Device::IsEnabled()) { REQUIRE_FALSE(registered_is_known); }
+      else { REQUIRE(registered_is_known); }
```
YohannDudouit, line 52: This could probably be done in a cleaner way, but I guess that's good enough for now.

YohannDudouit (approved):

tests/unit/general/test_mem.cpp
camierjs: Agree, I dont know if we want to add a `GetFlags`.

PR #2230, opened 2021-05-07
Create periodic meshes [make-periodic-mesh]

pazner:
Mostly this is work of @mlstowell.

| [#2230](https://github.com/mfem/mfem/pull/2230) | @pazner | @tzanio | @mlstowell + @jakubcerveny + @rcarson3 | 05/08/21 | 05/16/21 | 05/18/21 | |

### DIFF
--- a/mesh/mesh.cpp
+++ b/mesh/mesh.cpp
@@ -4506,6 +4506,7 @@ std::vector<int> Mesh::CreatePeriodicVertexMapping(
    // list of masters.
    auto make_master_slave = [&slaves,&masters](int m1, int m2)
    {
+      if (m1 == m2) { return; }
       masters[m2].insert(m1);
       slaves[m1] = m2;
       for (int s : masters[m1])
@@ -4549,11 +4550,8 @@ std::vector<int> Mesh::CreatePeriodicVertexMapping(
             {
                // `master` is already a master and `slave` is already a slave
                int master_of_slave = slaves[slave];
-               if (master != master_of_slave)
-               {
-                  // Make `master` and its slaves slaves of `slave`'s master
-                  make_master_slave(master, master_of_slave);
-               }
+               // Make `master` and its slaves slaves of `slave`'s master
+               make_master_slave(master, master_of_slave);
             }
             else if (!mInM && sInM)
             {
@@ -4561,10 +4559,7 @@ std::vector<int> Mesh::CreatePeriodicVertexMapping(
                // `slave` is currently a master
                // Make `slave` and its slaves slaves of `master`'s master
                int master_of_master = slaves[master];
-               if (slave != master_of_master)
-               {
-                  make_master_slave(slave, master_of_master);
-               }
+               make_master_slave(slave, master_of_master);
             }
             else
             {
@@ -4573,10 +4568,7 @@ std::vector<int> Mesh::CreatePeriodicVertexMapping(
                int master_of_master = slaves[master];
                int master_of_slave = slaves[slave];
                // Move slave and its fellow slaves to master_of_master
-               if (master_of_master != master_of_slave)
-               {
-                  make_master_slave(master_of_slave, master_of_master);
-               }
+               make_master_slave(master_of_slave, master_of_master);
             }
             break;
          }
--- a/mesh/mesh.cpp
+++ b/mesh/mesh.cpp
@@ -4497,28 +4497,34 @@ std::vector<int> Mesh::CreatePeriodicVertexMapping(
    add(xMax, -1.0, xMin, xDiff);
    double dia = xDiff.Norml2(); // compute mesh diameter
 
-   // slaves[v] is the index of the master vertex of slaver vertex `v`
-   map<int, int> slaves;
-   // masters[v] is a set of indices of all slave vertices of master vertex `v`
-   map<int, set<int>> masters;
+   // We now identify coincident verties. Several originally distinct vertices
+   // may become coincident under the periodic mapping. One of these vertices
+   // will be identified as the "primary" vertex, and all other coincident
+   // vertices will be considered as "replicas".
 
-   // Make `m1` and all of `m1`'s slaves slaves of `m2`. Delete `m1` from the
-   // list of masters.
-   auto make_master_slave = [&slaves, &masters](int m1, int m2)
+   // replica2primary[v] is the index of the primary vertex of replica `v`
+   map<int, int> replica2primary;
+   // primary2replicas[v] is a set of indices of replicas of primary vertex `v`
+   map<int, set<int>> primary2replicas;
+
+   // We begin with the assumption that all vertices are primary, and that there
+   // are no replicas.
+   for (int v : bdr_v) { primary2replicas[v]; }
+
+   // Make `r` and all of `r`'s replicas replicas of `p`. Delete `r` from the
+   // list of primary vertices.
+   auto make_replica = [&replica2primary, &primary2replicas](int r, int p)
    {
-      if (m1 == m2) { return; }
-      masters[m2].insert(m1);
-      slaves[m1] = m2;
-      for (int s : masters[m1])
+      if (r == p) { return; }
+      primary2replicas[p].insert(r);
+      replica2primary[r] = p;
+      for (int s : primary2replicas[r])
       {
-         masters[m2].insert(s);
-         slaves[s] = m2;
+         primary2replicas[p].insert(s);
+         replica2primary[s] = p;
       }
-      masters.erase(m1);
+      primary2replicas.erase(r);
    };
-
-   // Assume for now that all vertices are masters
-   for (int v : bdr_v) { masters[v]; }
 
    for (unsigned int i = 0; i < translations.size(); i++)
    {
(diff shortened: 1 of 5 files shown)


rcarson3:
Glad to see this functionality finally included in the main branch. I've had to use this same logic in a project of mine with some non-trivial primary/secondary node relationships for the periodic boundaries, and I found Mark's implementation to just work.

tzanio:

mesh/mesh.hpp
```diff
@@ -579,8 +579,27 @@ class Mesh
...
+   /// @brief Creates a mapping @a v2v from the vertex indices of the mesh such
+   /// that coincident vertices under the given @a translations are identified.
+   /** Each Vector in @a translations should be of size @a sdim (the spatial
+       dimension of the mesh). Two vertices are considered coincident if the
+       translated coordinates of one vertex are within the given tolerance (@a
+       tol, relative to the mesh diameter) of the coordinates of the other
+       vertex.
+       @warning This algorithm does not scale well with the number of vertices
+       in the mesh, and is not recommended for very large meshes.  */
+   std::vector<int> CreatePeriodicVertexMapping(
```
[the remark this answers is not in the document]
pazner: At least for the return value, `Array<int>` should definitely be possible. Avoiding an extraneous copy of the vertex map in cases like
```c++
Mesh::MakePeriodic(mesh, mesh.CreatePeriodicVertexMapping(translations));
```
would then rely on (non-mandatory) NRVO, whereas `std::vector` has move semantics, so it is mandated by the standard.

On the other hand, for the translations, we can't use `Array<Vector>` because `Vector` is not a trivial type (cf. #2214). I would really prefer to avoid `Array<Vector*>`.

mesh/mesh.hpp
```diff
@@ -579,8 +579,27 @@ class Mesh
...
    ///@}
 
+   /// @brief Creates a mapping @a v2v from the vertex indices of the mesh such
+   /// that coincident vertices under the given @a translations are identified.
+   /** Each Vector in @a translations should be of size @a sdim (the spatial
+       dimension of the mesh). Two vertices are considered coincident if the
+       translated coordinates of one vertex are within the given tolerance (@a
+       tol, relative to the mesh diameter) of the coordinates of the other
+       vertex.
+       @warning This algorithm does not scale well with the number of vertices
```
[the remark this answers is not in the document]
pazner: cc: @mlstowell, do you suggest a change of the wording in this comment?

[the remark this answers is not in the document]
mlstowell: It may be best to say "This algorithm does not scale well with the number of boundary vertices". This is more precise and will allow the user to better gauge whether or not this is appropriate for their meshes.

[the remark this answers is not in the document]
mlstowell: Another important warning we should include involves the number of edges separating opposite boundary surfaces. I'm not sure how best to word this but MFEM periodic meshes require no fewer than three edges connecting periodic boundaries which ensures that each edge can be identified by a unique pair of vertices.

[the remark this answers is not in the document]
mlstowell: Yes, but it's more than that. Three vertices forming two connected edges e.g. `a--b--c`, are also not allowed because identifying the first and last vertices of the triplet will produce two edges, `a--b` and `b--a`, which cannot be distinguished by the vertices alone.

tests/unit/mesh/test_periodic_mesh.cpp
```diff
@@ -0,0 +1,102 @@
...
+            // Bottom row connected to top row
+            REQUIRE(IndicesAreConnected(e2e, i, i + n*(n-1)));
+            // Left column connected to right column
+            REQUIRE(IndicesAreConnected(e2e, i*n, n-1 + i*n));
+         }
+      }
+   }
+   SECTION("3D periodic mesh")
+   {
+      auto el = GENERATE(Element::TETRAHEDRON, Element::HEXAHEDRON);
```
mlstowell, line 71: I'd add the `Element::WEDGE` type here as well.

mlstowell (changes requested):
I think the changes @pazner made to the `CreatePeriodicVertexMapping` function are a great improvement. I just have two suggestions:

- [x] Add a note to the `MakePeriodic` member function stating the requirement that MFEM periodic meshes require no fewer than three edges connect periodic boundaries.
- [x] Add the `Element::WEDGE` to the new unit test.

tests/unit/mesh/test_periodic_mesh.cpp
pazner: Done

mesh/mesh.hpp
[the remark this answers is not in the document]
pazner: I added a comment to that effect in `Mesh::MakePeriodic`.

mesh/mesh.cpp
```diff
@@ -4430,6 +4432,161 @@ void Mesh::MakeSimplicial_(const Mesh &orig_mesh, int *vglobal)
...
+         {
+            xMax[j] = max(xMax[j], coord[j]);
+            xMin[j] = min(xMin[j], coord[j]);
+         }
+      }
+   }
+   add(xMax, -1.0, xMin, xDiff);
+   double dia = xDiff.Norml2(); // compute mesh diameter
+
+   // slaves[v] is the index of the master vertex of slaver vertex `v`
```
rcarson3, line 4500: So, I've started to see people move away from this master-slave terminology to describe these sort of periodic conditions. I think it wouldn't hurt for us to consider changing this here. I've seen a number of alternative descriptions such as A/B, first/secondary, or primary/replica. I'm sure there are others as well that could work.

mesh/mesh.hpp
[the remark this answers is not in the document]
rcarson3: So, I'm a fan of using `std::vector<T>` here based on what I've done on another project that used the `Mesh::MakePeriodic()`. It made creating the v2v vector fairly simple, since I was able to build up off a number of `std::algorithms`.

rcarson3 (commented):
Just a few small comments here and there. Overall, the PR looks fine to me. I've already been using `MakePeriodic` in an application of mine with very non-trivial periodic conditions, and at least for my test case with tet elements it was working just fine.

mesh/mesh.cpp
pazner: I think primary/replica would work well in this context. I can make the changes if people would like.

tzanio: If it is an exact copy, I like `primary/replica`

For more general interpolation, I'll prefer something like `primary/derived` or`true/constrained`.

rcarson3: Might want to avoid the `true/constrained` name with `true` being a keyword and all :P

pazner: I agree @tzanio, I think in this case replica is appropriate because the vertices really are (topologically) exact copies.

tzanio: > Might want to avoid the true/constrained name with true being a keyword and all :P


true that

tzanio: lets go with `primary`/`replica` here and maybe `primary`/`constrained` for the general case

mesh/mesh.hpp
```diff
@@ -583,7 +583,10 @@ class Mesh
    /** Each vertex @a i will be mapped to vertex @a v2v[i], such that all
        vertices that are coincident under the periodic mapping get mapped to
        the same index. The mapping @a v2v can be generated from translation
-       vectors using Mesh::CreatePeriodicVertexMapping. */
+       vectors using Mesh::CreatePeriodicVertexMapping.
+       @note MFEM requires that each edge of the resulting mesh be uniquely
+       identifiable by a pair of unique vertiecs. As a consequence, periodic
```
mlstowell, line 588: Nicely stated but there is a small typo "vertiecs".

pazner: Fixed, thanks.

mesh/mesh.cpp
pazner: I made the changes — @rcarson3 let me know if you think it looks good to you.

mlstowell (approved):

mesh/mesh.cpp
```diff
@@ -4430,6 +4432,165 @@ void Mesh::MakeSimplicial_(const Mesh &orig_mesh, int *vglobal)
...
+         {
+            xMax[j] = max(xMax[j], coord[j]);
+            xMin[j] = min(xMin[j], coord[j]);
+         }
+      }
+   }
+   add(xMax, -1.0, xMin, xDiff);
+   double dia = xDiff.Norml2(); // compute mesh diameter
+
+   // We now identify coincident verties. Several originally distinct vertices
```
rcarson3, line 4500: typo "verties" should be "vertices"

rcarson3 (approved):
@pazner looks good outside the one typo I saw.

tzanio:

tests/unit/mesh/test_periodic_mesh.cpp
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2010-2021, Lawrence Livermore National Security, LLC. Produced
```
tzanio, line 1: Thanks for adding a unit test!

mesh/mesh.hpp
```diff
@@ -579,8 +579,30 @@ class Mesh
...
    static Mesh MakeSimplicial(const Mesh &orig_mesh);
 
+   /// Create a periodic mesh by identifying vertices of @a orig_mesh.
+   /** Each vertex @a i will be mapped to vertex @a v2v[i], such that all
+       vertices that are coincident under the periodic mapping get mapped to
+       the same index. The mapping @a v2v can be generated from translation
+       vectors using Mesh::CreatePeriodicVertexMapping.
+       @note MFEM requires that each edge of the resulting mesh be uniquely
+       identifiable by a pair of distinct vertices. As a consequence, periodic
+       boundaries must be connected by at least three edges. */
```
tzanio, line 589: Excellent documentation

mesh/mesh.cpp
```diff
@@ -4430,6 +4432,165 @@ void Mesh::MakeSimplicial_(const Mesh &orig_mesh, int *vglobal)
...
+   double dia = xDiff.Norml2(); // compute mesh diameter
+
+   // We now identify coincident vertices. Several originally distinct vertices
+   // may become coincident under the periodic mapping. One of these vertices
+   // will be identified as the "primary" vertex, and all other coincident
+   // vertices will be considered as "replicas".
+
+   // replica2primary[v] is the index of the primary vertex of replica `v`
+   map<int, int> replica2primary;
+   // primary2replicas[v] is a set of indices of replicas of primary vertex `v`
```
tzanio, line 4507: I like the `primary` and `replica` terminology 👍

mesh/mesh.cpp
```diff
@@ -4430,6 +4432,165 @@ void Mesh::MakeSimplicial_(const Mesh &orig_mesh, int *vglobal)
...
+   map<int, set<int>> primary2replicas;
+
+   // We begin with the assumption that all vertices are primary, and that there
+   // are no replicas.
+   for (int v : bdr_v) { primary2replicas[v]; }
+
+   // Make `r` and all of `r`'s replicas replicas of `p`. Delete `r` from the
```
tzanio, line 4514: The double "replicas replicas"  here and elsewhere read a bit funny to me. How about: "Make `r` and all of its replicas be replicas of `p`"?

tzanio (approved):
Very nice and very useful work @mlstowell and @pazner.

Should we highlight it in `CHANGELOG`?

mesh/mesh.cpp
pazner: I agree it sounded a bit weird. I added the "be".

tzanio: thanks!

PR #2232, opened 2021-05-08
HO<->LOR Grid Transfer for H1 FE Spaces [lor-h1toh1-dev]

ebchin:
This PR builds on @pazner's grid transfer between high order and low order refined L2 FE spaces and implements it for H1 FE spaces.

| [#2232](https://github.com/mfem/mfem/pull/2232) | @ebchin | @tzanio | @pazner + @tzanio | 05/08/21 | 06/12/21 | 06/15/21 | |


tzanio:

tzanio:
LGTM

fem/transfer.cpp
```diff
@@ -522,6 +522,439 @@ const Operator &L2ProjectionGridTransfer::BackwardOperator()
...
+   }
+   ho2lor.MakeJ();
+   for (int ilor = 0; ilor < nel_lor; ++ilor)
+   {
+      int iho = cf_tr.embeddings[ilor].parent;
+      ho2lor.AddConnection(iho, ilor);
+   }
+   ho2lor.ShiftUpI();
+
+   // ML_inv contains the inverse lumped (row sum) mass matrix
```
srwopschall, line 571: I know the inversion of a low order lumped mass matrix works for our application. Do we anticipate another application where you would have to or want to invert the whole consistent low order mass matrix? Maybe extend the comment indicating that for now the implementation "chose" to use a lumped mass matrix inversion, unless we think this will extend across all use cases.

tzanio: good point

fem/transfer.cpp
```diff
@@ -522,6 +522,439 @@ const Operator &L2ProjectionGridTransfer::BackwardOperator()
...
+            ML_el += (shape_lor *= (el_tr->Weight() * ip_lor.weight));
+         }
+         fes_lor.GetElementDofs(ilor, dofs_lor);
+         ML_inv.AddElementVector(dofs_lor, ML_el);
+      }
+   }
+   // DOF by DOF inverse of non-zero entries
+   for (int i = 0; i < ndof_lor; ++i)
+   {
+      if (ML_inv[i] != 0.0)
```
srwopschall, line 614: Do we want to compare exactly to 0 here or some tolerance to prevent division by a very very small number?

fem/transfer.cpp
```diff
@@ -522,6 +522,439 @@ const Operator &L2ProjectionGridTransfer::BackwardOperator()
...
+      {
+         ML_inv[i] = 1.0 / ML_inv[i];
+      }
+   }
+
+   // Compute sparsity pattern for R and allocate
+   AllocR();
+   // Allocate M_LH (same sparsity pattern as R)
```
srwopschall, line 622: Elaborate comment to include what L and H refer to for clarity? (Obviously I know this notation :) )

fem/transfer.hpp
```diff
@@ -158,11 +158,11 @@ class InterpolationGridTransfer : public GridTransfer
...
-    coarse and fine FE spaces can have different orders, however, in order for
-    the backward operator to be well-defined, the number of the fine dofs (in a
-    coarse element) should not be smaller than the number of coarse dofs.
+    This class supports L2 finite element spaces and fine meshes that are a
+    uniform refinement of the coarse mesh. Generally, the coarse and fine FE
+    spaces can have different orders, however, in order for the backward
+    operator to be well-defined, the number of the fine dofs (in a coarse
+    element) should not be smaller than the number of coarse dofs.
 
     If used on H1 finite element spaces, the transfer will be performed locally,
```
srwopschall, line 167: Maybe I am not down far enough in this file, but there is still functionality that let's you use the L2 FE space L2_projection, but using H1 fields. Is there an instance where you want to do this? Should we add a message so that a user realizes they "may" be using the L2 discontinuous mapping option for their H1 field? Or if you try this it actually uses the H1 continuous mapping implementation that we are reviewing? @ebchin @tzanio please clarify (I'm curious).

fem/transfer.hpp
```diff
@@ -244,6 +244,103 @@ class L2ProjectionGridTransfer : public GridTransfer
...
+      mutable DenseMatrix* P;
+      const double p_rtol;
+      const double p_atol;
+
+      Table ho2lor;
+   public:
+      L2ProjectionH1(const FiniteElementSpace& fes_ho_,
+                     const FiniteElementSpace& fes_lor_);
+      virtual ~L2ProjectionH1();
+      /// Perform the L2 projection onto the LOR space
```
srwopschall, line 286: @ebchin It might be worth clarifying the intended use/purpose of the functions below in terms of the HO->LO or LO->HO for both primal and dual fields. I recall Mult() being a forward map (HO->LO) for primal fields, but MultTranspose() the backward (LO->HO) map for dual fields. I feel like this type of thing could lead to some confusion when trying to be exercised at a higher level. The fact that there is language like Forward, Backward, Restriction, Prolongation, and the use of these flips depending on primal or dual field transfer makes me think some clarifying "use" comments might go a long way. Let me know if this comment doesn't make sense.

fem/transfer.hpp
```diff
@@ -244,6 +244,103 @@ class L2ProjectionGridTransfer : public GridTransfer
...
+      /// Computes sparsity pattern and initializes R matrix.  Based on
+      /// BilinearForm::AllocMat() except maps between HO elements and LOR
+      /// elements.
+      void AllocR();
+      /// Computes prolongation matrix P.
+      void BuildP() const;
+   };
+
+   /** Mass-conservative prolongation operator going in the opposite direction
+       as L2Projection. This operator is a left inverse to the L2Projection. */
```
srwopschall, line 309: Consider elaborating on "opposite direction", even though I know what you mean.

srwopschall (approved):
@ebchin Looks good. I mainly had comments regarding clarification.

fem/transfer.hpp
ebchin: Fixed.

fem/transfer.hpp
ebchin: I agree, more extensive comments here would be beneficial.  I added a more complete description of the functions.  Hopefully these new comments clear it up?

fem/transfer.hpp
ebchin: I can see how this might cause confusion or lead a user to do something unintended.  I clarified the comments for the `L2ProjectionGridTransfer` and `L2ProjectionH1GridTransfer` classes so it's more clear which one is for L2 spaces and which one is for H1 spaces.  I'm not sure we need to explicitly disallow `L2ProjectionGridTransfer` on H1 spaces.  Maybe a user wants to avoid a global solve and doesn't need mass conservation?  Though the collocation approach in `InterpolationGridTransfer` might be more appropriate in that case...  Any thoughts @tzanio or @pazner ?

fem/transfer.cpp
ebchin: Sure.  Added some more descriptive comments.

fem/transfer.cpp
ebchin: Thanks for catching this.  This is a hold-over from a previous version which did the transfer on the surface and there were some mass matrix entries which were exactly zero.  All mass matrix entries should be non-zero, so this check has been removed.

fem/transfer.cpp
ebchin: I think the [comment in the header file](https://github.com/mfem/mfem/blob/be70f2a528c34a508b2c5a5bc4b1c983b201b8f4/fem/transfer.hpp#L253) covers this?  If not, I'd be happy to put more details in here.

tzanio:
thanks for the careful review @srwopschall !

fem/transfer.cpp
srwopschall: I think adding another comment in the .cpp file doesn't hurt. Just my opinion though.

fem/transfer.hpp
srwopschall: 👍

fem/transfer.cpp
tzanio: Comments never hurt :)

miniapps/tools/lor-transfer.cpp
```diff
@@ -95,6 +96,9 @@ int main(int argc, char *argv[])
    args.AddOption(&use_pointwise_transfer, "-t", "--use-pointwise-transfer",
                   "-no-t", "--dont-use-pointwise-transfer",
                   "Use pointwise transfer operators instead of L2 projection.");
+   args.AddOption(&forceL2transfer, "-fl2x", "--force-l2-transfer",
+                  "-no-fl2x", "--dont-force-l2-transfer",
+                  "Force L2 projection assuming L2 finite element space (even if H1).");
```
pazner, line 101: I'm not sure this option is necessary, I think it can be removed from the miniapp.

fem/transfer.cpp
```diff
@@ -522,6 +522,439 @@ const Operator &L2ProjectionGridTransfer::BackwardOperator()
...
+
+void L2ProjectionH1GridTransfer::L2ProjectionH1::BuildP() const
+{
+   int ndof_ho = fes_ho.GetNDofs();
+   int ndof_lor = fes_lor.GetNDofs();
+
+   // Compute P = (R^T M_LH)^(-1) M_LH^T column by column (CG solve)
+   SparseMatrix RTxM_LH = *TransposeMult(*R, M_LH);
+   DSmoother Ds(RTxM_LH);
+   P = new DenseMatrix(ndof_ho, ndof_lor);
```
pazner, line 915: I don't think the approach of having `P` as a globally dense matrix is advisable. This will severely limit the problem size for which this is applicable.

A more scalable approach would involve a (preconditioned) CG solve every time `P` or `P^T` is applied (rather than constructed).

fem/transfer.cpp
```diff
@@ -522,6 +522,439 @@ const Operator &L2ProjectionGridTransfer::BackwardOperator()
...
+      }
+      PCG(RTxM_LH, Ds, M_LH_col, P_col, 0, 1000, p_rtol * p_rtol, p_atol * p_atol);
+   }
+}
+
+
+L2ProjectionH1GridTransfer::~L2ProjectionH1GridTransfer()
+{
+   if (F) { delete F; }
+   if (B) { delete B; }
```
pazner, line 938: Can omit the non-NULL checks

fem/transfer.hpp
```diff
@@ -244,6 +245,126 @@ class L2ProjectionGridTransfer : public GridTransfer
...
+       lives on a mesh obtained by refining the mesh of the high-order space,
+       fes_ho. */
+   class L2ProjectionH1 : public Operator
+   {
+      const FiniteElementSpace& fes_ho;
+      const FiniteElementSpace& fes_lor;
+
+      SparseMatrix* R;
+      SparseMatrix M_LH;
+      mutable DenseMatrix* P;
```
pazner, line 280:
```suggestion
      SparseMatrix R;
      SparseMatrix M_LH;
      mutable DenseMatrix P;
```

These pointers can be replaced with objects, and when `P` is constructed, just use `SetSize`, and when `R` is allocated, you can use `SparseMatrix::Swap` to set up the proper sizes.

pazner (commented):
Thanks for your work on this @ebchin! I had some minor suggestions about code organization/factoring.

My larger concern is the dense `P` matrix — I don't think this approach will be viable beyond relatively small problems.

Also do you know what happens in the parallel case?

tzanio:
@ebchin, have you had a chance to look at @pazner's feedback above?

miniapps/tools/lor-transfer.cpp
ebchin: Agreed.  Fixed

fem/transfer.hpp
```diff
@@ -158,18 +158,19 @@ class InterpolationGridTransfer : public GridTransfer
...
-    numbering in the finite element space). As a consequence, the mass
-    conservation properties for this operator from the L2 case do not carry over
-    to H1 spaces. */
+    This class supports L2 finite element spaces and fine meshes that are a
+    uniform refinement of the coarse mesh. Generally, the coarse and fine FE
+    spaces can have different orders, however, in order for the backward
+    operator to be well-defined, the number of the fine dofs (in a coarse
+    element) should not be smaller than the number of coarse dofs.
+
+    For H1 finite element spaces, see the L2ProjectionH1GridTransfer class. This
```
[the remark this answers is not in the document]
ebchin: Agreed.  I think it makes more sense to diverge at the `L2Projection` level.  See updated source.

fem/transfer.hpp
ebchin: Fixed in latest commit.

fem/transfer.cpp
ebchin: Fixed in latest commit.

fem/transfer.hpp
```diff
@@ -244,6 +245,126 @@ class L2ProjectionGridTransfer : public GridTransfer
...
+    is a left inverse of the forward transfer operator, F, i.e. B F = I. Both F
+    and B are defined in physical space.
+
+    This class supports H1 finite element spaces and fine meshes that are a
+    uniform refinement of the coarse mesh.  Generally, the coarse and fine FE
+    spaces can have different orders, however, in order for the backward
+    operator to be well-defined, the number of fine dofs (in a coarse element)
+    should not be smaller than the number of coarse dofs.
+
+    For L2 finite element spaces, see L2ProjectionGridTransfer. */
```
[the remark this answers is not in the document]
ebchin: Good idea.  Updated in the latest commit.

fem/transfer.cpp
```diff
@@ -522,6 +522,439 @@ const Operator &L2ProjectionGridTransfer::BackwardOperator()
...
+      int iho = cf_tr.embeddings[ilor].parent;
+      ho2lor.AddAColumnInRow(iho);
+   }
+   ho2lor.MakeJ();
+   for (int ilor = 0; ilor < nel_lor; ++ilor)
+   {
+      int iho = cf_tr.embeddings[ilor].parent;
+      ho2lor.AddConnection(iho, ilor);
+   }
+   ho2lor.ShiftUpI();
```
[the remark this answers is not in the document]
ebchin: Agreed.  Taken care of in latest commit.

fem/transfer.cpp
```diff
@@ -522,6 +522,439 @@ const Operator &L2ProjectionGridTransfer::BackwardOperator()
...
 
+
+L2ProjectionH1GridTransfer::L2ProjectionH1::L2ProjectionH1(
+   const FiniteElementSpace& fes_ho_, const FiniteElementSpace& fes_lor_)
+   : Operator(fes_lor_.GetVSize(), fes_ho_.GetVSize()),
+     fes_ho(fes_ho_),
+     fes_lor(fes_lor_),
+     P(NULL),
+     p_rtol(1e-13),
+     p_atol(1e-13)
```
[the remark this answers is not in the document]
ebchin: I added a `SetTolerance()` method, but this might not be the best way to expose this to a user?  It would be nice to set this upon construction of `L2ProjectionGridTransfer::L2ProjectionH1Space`, but that seems like it would require changing the interface for `ForwardOperator()` and `BackwardOperator()`, which also isn't ideal.

fem/transfer.cpp
```diff
@@ -522,6 +522,439 @@ const Operator &L2ProjectionGridTransfer::BackwardOperator()
...
+   Vector x_dim(ndof_ho);
+   Vector y_dim(ndof_lor);
+
+   for (int d = 0; d < vdim; ++d)
+   {
+      for (int i = 0; i < ndof_ho; ++i)
+      {
+         dofs_ho[i] = i;
+      }
+      fes_ho.DofsToVDofs(d, dofs_ho);
```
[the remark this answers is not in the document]
ebchin: I added `FiniteElementSpace::GetVDofs()` in the latest commit.

fem/transfer.cpp
```diff
@@ -522,6 +522,439 @@ const Operator &L2ProjectionGridTransfer::BackwardOperator()
...
+L2ProjectionH1GridTransfer::L2ProjectionH1::~L2ProjectionH1()
+{
+   delete R;
+   if (P) { delete P; }
+}
+
+void L2ProjectionH1GridTransfer::L2ProjectionH1::Mult(
+   const Vector& x, Vector& y) const
+{
+   int vdim = fes_ho.GetVDim();
```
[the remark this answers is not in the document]
ebchin: I thought about ways to do this (either virtual functions or a policy-based design).  I agree it's nice to reduce code reproduction, but I'm worried in this case, since the code is relatively short and the number of functions that would utilize this abstraction is small (4), that introducing more general code might not be worth the increased complexity and loss of readability the abstracted code would introduce.  However, I'd be happy to implement something to reduce some of the code redundancy.  Did you have a particular implementation in mind?

fem/transfer.cpp
ebchin: Agreed.  This has been implemented in the latest commit.

fem/transfer.cpp
ebchin: Fixed in latest commit.

fem/transfer.hpp
ebchin: Agreed.  Fixed in the latest commit.

fem/transfer.cpp
```diff
@@ -522,6 +522,439 @@ const Operator &L2ProjectionGridTransfer::BackwardOperator()
...
+            ip_tr.Transform(ip_lor, ip_ho);
+            fe_lor.CalcShape(ip_lor, shape_lor);
+            fe_ho.CalcShape(ip_ho, shape_ho);
+            el_tr->SetIntPoint(&ip_lor);
+            // For now we use the geometry information from the LOR space
+            // which means we won't be mass conservative if the mesh is curved
+            double w = el_tr->Weight() * ip_lor.weight;
+            shape_lor *= w;
+            AddMultVWt(shape_lor, shape_ho, M_LH_el);
+         }
```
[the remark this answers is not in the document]
ebchin: Agreed.  Fixed in latest commit.

ebchin:
> @ebchin, have you had a chance to look at @pazner's feedback above?

Hi @tzanio, thanks for the reminder!  Thanks @pazner for the helpful review.

tzanio:
thanks @ebchin !

fem/fespace.hpp
```diff
@@ -647,6 +647,8 @@ class FiniteElementSpace
 
    void GetEdgeInteriorDofs(int i, Array<int> &dofs) const;
 
+   void GetVDofs(int vd, Array<int> &dofs, int ndofs = -1) const;
```
pazner, line 650: Docstring would be good here (e.g. explaining what `ndofs = -1` signifies)

fem/transfer.cpp
```diff
@@ -226,12 +226,63 @@ const Operator &InterpolationGridTransfer::BackwardOperator()
    return *B.Ptr();
 }
 
-
```
pazner, line 229: I think the extra space here was intentional (two newlines between different classes)

fem/transfer.cpp
```diff
@@ -226,12 +226,63 @@ const Operator &InterpolationGridTransfer::BackwardOperator()
...
 L2ProjectionGridTransfer::L2Projection::L2Projection(
    const FiniteElementSpace &fes_ho_, const FiniteElementSpace &fes_lor_)
    : Operator(fes_lor_.GetVSize(), fes_ho_.GetVSize()),
      fes_ho(fes_ho_),
      fes_lor(fes_lor_)
+{ }
+
+void L2ProjectionGridTransfer::L2Projection::BuildHo2Lor(
```
pazner, line 236: Nice! I like how you structured `L2Projection::BuildHo2Lor` and `L2Projection::ElemMixedMass`.

fem/transfer.cpp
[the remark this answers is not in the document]
pazner: I think it already looks a lot better now that you've added `FiniteElementSpace::GetVDofs`.

fem/transfer.cpp
```diff
@@ -659,44 +639,35 @@ L2ProjectionH1GridTransfer::L2ProjectionH1::L2ProjectionH1(
...
       }
    }
-}
 
-L2ProjectionH1GridTransfer::L2ProjectionH1::~L2ProjectionH1()
-{
-   delete R;
-   if (P) { delete P; }
+   // Create PCG solver
+   RTxM_LH = *TransposeMult(R, M_LH);
```
pazner, line 660: This will leak memory

fem/transfer.hpp
```diff
@@ -307,62 +289,62 @@ class L2ProjectionH1GridTransfer : public GridTransfer
...
       /// coefficients defined on a refined mesh with a low order H1 finite
       /// element space.  Refined mesh should be a uniform refinement of the coarse
       /// mesh. Coefficients are computed from the transpose of the mass
       /// conservative left-inverse prolongation operation.  This functionality is
-      /// also provided as an Operator by L2ProlongationH1. Builds the prolongation
-      /// operator if P is NULL.
-      void ProlongateTranspose(const Vector& x, Vector& y) const;
+      /// also provided as an Operator by L2Prolongation.
+      virtual void ProlongateTranspose(const Vector& x, Vector& y) const;
+      virtual void SetTolerance(double p_rtol_, double p_atol_);
```
pazner, line 302: It might be better to split this into `SetRelTol` and `SetAbsTol` just like in MFEM's iterative solvers.

miniapps/tools/lor-transfer.cpp
```diff
@@ -125,13 +121,13 @@ int main(int argc, char *argv[])
...
          cerr << "Switching the H1 LOR space order from 0 to 1\n";
       }
-      fec = new H1_FECollection(order-1, dim);
-      fec_lor = new L2_FECollection(lorder, dim);
+      fec = new H1_FECollection(order, dim);
+      fec_lor = new H1_FECollection(lorder, dim);
    }
    else
    {
       space = "L2";
```
pazner, line 129: Running `./lor-transfer --use-l2` fails with error message `AddMultVWt(...): dimension mismatch`.

pazner (commented):
This is looking really close to being ready! There are just a few minor comments.

fem/transfer.cpp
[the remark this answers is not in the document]
ebchin: Great!  I'll leave it as is.

fem/transfer.cpp
ebchin: Ahh, I'll add it back in.  I noticed there are other places in `transfer.cpp` that don't have two newlines between classes.  Should I add an extra newline there as well?  (I haven't looked at other source files)

fem/transfer.cpp
ebchin: Thanks!

fem/transfer.cpp
pazner: Sure, it would be good if they were there in `transfer.cpp` for consistency.

fem/transfer.cpp
ebchin: Thanks for catching this!  Fixed in the latest commit.

fem/fespace.hpp
ebchin: Agreed.  Added in the latest commit.

fem/transfer.hpp
ebchin: Good idea.  I put this in the latest commit.

miniapps/tools/lor-transfer.cpp
ebchin: Thanks for catching this!  I forgot to check l2 spaces after I implemented `ElemMixedMass`.  `ElemMixedMass` was being called with `M_mixed` instead of `M_mixed_el`.  Fixed in the latest commit.

fem/transfer.cpp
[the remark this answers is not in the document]
ebchin: Resolved.  Added `SetRelTol()` and `SetAbsTol()` in latest commit.

fem/transfer.cpp
ebchin: OK, fixed.

fem/transfer.cpp
pazner: I don't think your fix will work. I would suggest making `RTxM_LH` a pointer. You could also do something like
```c++
SparseMatrix *tmp = TransposeMult(R, M_LH);
RTxM_LH = *tmp;
delete tmp;
```
but that causes an unnecessary copy.

ebchin: Sounds good.  I made `RTxM_LH` a pointer.

pazner (approved):
Looks good to me now. Thanks @ebchin!

tzanio (approved):
Thanks for the nice contribution @ebchin 👍 

@pazner -- thanks for the careful review!

tzanio:

PR #2240, opened 2021-05-12
Fix gmsh periodic meshes [fix-gmsh-periodic]

### LINKED ISSUE #2216: GMSH Support
Uploaded a gmsh file(.msh) and got the following error:
MFEM abort: Gmsh file : vertices indices are not unique
 ... in function: void mfem::Mesh::ReadGmshMesh(std::istream&, int&, int&)
 ... in file: mesh/mesh_readers.cpp:1551

with errorcode 1.

Is there already an mfem periodic tet mesh for a cube?

pazner:
Loading of periodic gmsh meshes was broken in a regression because [this check](https://github.com/mfem/mfem/blob/4ec8ef2fb46aea0922bd6edf1b0f9f7fcaa76f78/fem/fespace.cpp#L1908-L1909) was failing when reading the mesh [here](https://github.com/mfem/mfem/blob/4ec8ef2fb46aea0922bd6edf1b0f9f7fcaa76f78/mesh/mesh_readers.cpp#L2640).

I added the line `this->SetMeshGen();`, which seems to fix the problem, but the conditions that the partially constructed mesh must satisfy in order to create a `FiniteElementSpace` are still a bit mysterious to me. This should ideally be clarified in documentation.

I added two sample straight-sided periodic gmsh meshes, and added them as sample runs to example 9.

Resolves #2216.

| [#2240](https://github.com/mfem/mfem/pull/2240) | @pazner | @tzanio | @jakubcerveny + @mlstowell + @bslazarov | 05/13/21 | 05/16/21 | 05/18/21 | |


pazner:
> I added the line `this->SetMeshGen();`, which seems to fix the problem, but the conditions that the partially constructed mesh must satisfy in order to create a `FiniteElementSpace` are still a bit mysterious to me. This should ideally be clarified in documentation.

@jakubcerveny, do you think you could clarify the conditions on the mesh? It took me quite a long time to realize `SetMeshGen` was the right thing to do to fix this issue, and I'm not sure if this is a general solution or one that just happens to work in this case. It seems like there are some implicit/unstated assumptions, and it took some guesswork to figure out a fix.

The error message `"Mesh was not correctly finalized"` also took me down the wrong path of looking into `Mesh::Finalize` and `Mesh::FinalizeTopology`.

pazner:
@goinea, I think this PR should fix your issue with reading periodic gmsh meshes. It should work with the following small modification of your geo file:
```
SetFactory("OpenCASCADE");

Box(1) = {0, 0, 0, 1, 1, 1};
MeshSize {:} = 0.2;

Periodic Surface {2} = {1} Translate {1, 0, 0};
Periodic Surface {6} = {5} Translate {0, 0, 1};
Periodic Surface {4} = {3} Translate {0, 1, 0};

Physical Volume(1) = {1};
Physical Surface(1) = {1,2,3,4,5,6};

Mesh 3;
Mesh.MshFileVersion = 2.2;
```

jakubcerveny:
Thanks @pazner, I think this should work.

> do you think you could clarify the conditions on the mesh? It took me quite a long time to realize SetMeshGen was the right thing to do to fix this issue, and I'm not sure if this is a general solution or one that just happens to work in this case. It seems like there are some implicit/unstated assumptions, and it took some guesswork to figure out a fix.

The idea is that it's not OK to construct a FiniteElementSpace over a "partially constructed" Mesh, which unfortunately is a state the mesh can exist in the current design, until `Finalize...` is called (I'm actually not sure which one the "finalization" methods is the right one). The reason is that the space absolutely needs to know if there are mixed faces in the mesh, for example, which is the kind of information that is not initialized until the mesh is finalized. 

I'm sorry that the message misled you. It was indended for users who most likely forgot to call Finalize after constructing the mesh procedurally. Internally, in mesh readers, we are probably forced to use the SetMeshGen hack to be able to initialize the nodal space before the mesh can be finalized.

pazner:
Thanks for the clarification @jakubcerveny. For now I think this fix is fine.

It would be good at some point to address some of these larger issues to avoid similar bugs coming up in the future, but that is likely not a simple task.

goinea:
Thanks for the update and bug fix! @pazner

tzanio:

bslazarov (commented):
Nice examples of periodic meshes.

It is not clear to me what this->SetMeshGen(); does for the mesh? Is it possible for someone who knows how the mesh class is working to explain it?

mlstowell (approved):
I ran the new sample runs without incident. I also processed the Gmsh files using Gmsh 4.5.6 (April 2020) which showed syntax errors related to the `{:}` in `MeshSize {:} = 0.25` but produced correct mesh files in any case. Using Gmsh 4.8.4 (April 2021) the files processed cleanly.

pazner:
> It is not clear to me what this->SetMeshGen(); does for the mesh? Is it possible for someone who knows how the mesh class is working to explain it?

+1. It's pretty cryptic, this needs better documentation

bslazarov (approved):

pazner:
> I ran the new sample runs without incident. I also processed the Gmsh files using Gmsh 4.5.6 (April 2020) which showed syntax errors related to the `{:}` in `MeshSize {:} = 0.25` but produced correct mesh files in any case. Using Gmsh 4.8.4 (April 2021) the files processed cleanly.

I pushed some changes that should address the syntax errors in older versions of gmsh.

tzanio (approved):
Thanks @pazner !

tzanio:

examples/ex9.cpp
```diff
@@ -16,6 +16,8 @@
 //    ex9 -m ../data/disc-nurbs.mesh -p 2 -r 3 -dt 0.005 -tf 9
 //    ex9 -m ../data/periodic-square.mesh -p 3 -r 4 -dt 0.0025 -tf 9 -vs 20
 //    ex9 -m ../data/periodic-cube.mesh -p 0 -r 2 -o 2 -dt 0.02 -tf 8
+//    ex9 -m ../data/periodic-square.msh -p 0 -r 2 -dt 0.005 -tf 2
+//    ex9 -m ../data/periodic-cube.msh -p 0 -r 1 -o 2 -tf 2 -no-vis -paraview
```
tzanio, line 20: The `-no-vis` option causes problems in autotest (because it is added there automatically). Can we remove it?

```suggestion
//    ex9 -m ../data/periodic-cube.msh -p 0 -r 1 -o 2 -tf 2 -paraview
```

examples/ex9p.cpp
```diff
@@ -16,6 +16,8 @@
 //    mpirun -np 4 ex9p -m ../data/disc-nurbs.mesh -p 2 -rp 1 -dt 0.005 -tf 9
 //    mpirun -np 4 ex9p -m ../data/periodic-square.mesh -p 3 -rp 2 -dt 0.0025 -tf 9 -vs 20
 //    mpirun -np 4 ex9p -m ../data/periodic-cube.mesh -p 0 -o 2 -rp 1 -dt 0.01 -tf 8
+//    mpirun -np 4 ex9p -m ../data/periodic-square.msh -p 0 -rs 2 -dt 0.005 -tf 2
+//    mpirun -np 4 ex9p -m ../data/periodic-cube.msh -p 0 -rs 1 -o 2 -tf 2 -no-vis -paraview
```
tzanio, line 20: Another instance...

```suggestion
//    mpirun -np 4 ex9p -m ../data/periodic-cube.msh -p 0 -rs 1 -o 2 -tf 2 -paraview
```

examples/ex9.cpp
pazner: Yes, we can remove it. The reason `no-vis` is specified there is because none of the periodic surfaces are visualized in GLVis — it could be potentially confusing for a user to run one of the sample runs, and then have nothing show up in the visualization window.

Perhaps the periodic boundaries need to be added as "fake" boundary elements for visualization? I don't remember how that works.

tzanio: I'd prefer to "fix" the GLVis visualization if possible, maybe @mlstowell can help with this?

The alternative is for OptionParser not to complain if you have the same option twice, the current autotest error is `Option --visualization or --no-visualization provided multiple times`.

examples/ex9p.cpp
mlstowell: I'll take a look at it.

mlstowell: I think the trick is to add boundary elements on both periodic boundaries with opposite orientations. I don't know if we can convince Gmsh to do this but we can probably add one surface or the other which would be better than nothing.

mlstowell: OK, Gmsh will be happy to create the missing surfaces. Unfortunately, it's more complicated than that. Our Gmsh reader automatically removes internal surface from periodic meshes. This was done because the other periodic meshes in the `data` directory do not contain boundary elements.

We could switch this behavior in the Gmsh reader but this would break some of our existing sample runs involving periodic Gmsh meshes in `ex1` and `ex11p` (and perhaps more) because these examples apply homogeneous Dirichlet BCs on any boundary surfaces they find in the mesh which would include these periodic surfaces.

We might be able to control the Gmsh reader behavior with respect to internal boundaries but I'm not sure how we would pass this option to it through the `Mesh` constructor.

pazner: A longer term solution might involve decoupling of boundary elements for discretization and "fake boundary elements" used for visualization of internal or periodic surfaces. These two things seem pretty separate to me, and it probably makes sense to clearly distinguish between them in the code.

pazner: I think issue #2231 is also related to "fake boundary elements".

tzanio: Is it possible to solve this alternatively in GLVis?

mlstowell: I think it would be possible but perhaps expensive. GLVis might need to visualize volume elements when boundary elements are not present. This might involve determining which elements are hidden by others but this would of course change as the mesh rotates.

tzanio:

goinea:
@pazner  @tzanio  Do you think any changes need to be made to the miniapps with these new updates?

pazner:
> @pazner @tzanio Do you think any changes need to be made to the miniapps with these new updates?

I think the miniapps should be fine. We test the miniapp sample runs automatically, so any hopefully that will catch any potential issues.

Issue #2238, opened 2021-05-11
Calculate x_e^T*A_e*x_e?


nghiakvnvsd:
Hi community of MFEM! MFEM is a good application to solve my problem parallely but when I try to solve the problems after the FEM step, I am having trouble calculating some math related to the element displacement vector and element stiffness matrix. I hope you will give me good advice.

1) I want to calculate : 
![image](https://user-images.githubusercontent.com/68639058/117865068-77560580-b2d0-11eb-9bf9-49e635552e50.png)

where N is a number of element, x_e and A_e are element vector displacement and element stiffness matrix, respectively

2) I want to calculate vector d satisfy:
![image](https://user-images.githubusercontent.com/68639058/117865171-9ce30f00-b2d0-11eb-8213-d6bc4f040135.png)

Is there any way to do so? If you have a sample code that does this for ex2p (https://github.com/mfem/mfem/blob/master/examples/ex2p.cpp), that is very great for me! Thank you very much!

mlstowell:
Hi, @nghiakvnvsd,

We're glad MFEM is able to help with your problem. MFEM doesn't currently have this type of post-processing although it would be a valuable addition to the library. The closest thing we have at the moment is `GridFunction::ComputeElementLpErrors` and related member functions. These work by looping over the elements, extracting the local element vector from the `GridFunction`, evaluating the basis functions (stored as a vector `B_e`) at each integration point  within the element and forming the product `x_e^T B_e^T B_e x_e`. You could do something similar by replacing the basis function evaluation with an evaluation of their gradients (stored as a dense matrix `D_e`) and form `x_e^T D_e^T C D_e x_e` where `C` is your material coefficient. For elasticity problems this can be tricky because we use multiple copies of the scalar-valued basis functions to represent the vector-valued displacement (not to mention the rank four tensor-valued material coefficient).

An alternative approach would be to use the `ElasticityIntegrator` to form the `D_e^T C D_e` product for you. This would involve looping over the elements and extracting the local element vector (just as is done in `GridFunction::ComputeElementLpErrors`). And computing the local element matrix as `elas_integ->AssembleElementMatrix(*fe, *T, A_e);`. Once you have `A_e` and `x_e` you can use `d_e = A_e.InnerProduct(x_e, x_e);`.

I hope this will get you started. If you run into any trouble just followup here and we'll try to talk you through it.

Best wishes,
Mark

nghiakvnvsd:
> Hi, @nghiakvnvsd,
[...]

PR #2244, opened 2021-05-14
Adding `BilinearForm::ElementWiseInnerProduct` [elem-inner-prod-dev]

mlstowell:
This code was written in response to issue #2238.

Is there any interest in including this in `master`? Should we add any similar functions? Are there any examples or miniapps that might benefit from  such a function?


artv3:
I think it is a neat feature!

tzanio:
This is slightly non-standard, but I don't see any harm in including it. The only thing is that with the current implementation we should clearly indicate in the comments that this is a relatively expensive operation that will re-compute the element matrices.

This will of course be a very natural operation with ` AssemblyLevel::ELEMENT`, cc: @YohannDudouit

mlstowell:
As implemented this will use the precomputed element matrices when using `AssemblyLevel::ELEMENT`.

PR #2260, opened 2021-05-20
Serial KellyErrorEstimator

dylan-copeland:
This PR adds the option to use the Kelly error estimator in the serial example ex15, meaning that `KellyErrorEstimator` can now generally be constructed and used for serial or parallel `{Par}FiniteElementSpace`, `{Par}GridFunction`, `{Par}Mesh`. Also, `ex15` works with a serial or parallel build of MFEM.

The parallel unit tests in `tests/unit/fem/test_estimator.cpp` could easily be copied to make serial unit tests, but I don't think it is necessary to test the serial case, as the code differences between serial and parallel are insignificant. The sample runs in `ex15` should be adequate for testing.

There has been some discussion (@YohannDudouit) about unifying the serial and parallel classes in MFEM. There may be pros and cons to this, but this PR is an example of the problems caused by having them separate.

| [#2260](https://github.com/mfem/mfem/pull/2260) | @dylan-copeland | @tzanio | @brendankeith + @jandrej | 05/23/21 | 06/08/21 | 06/10/21 | |

### DIFF
--- a/examples/ex15.cpp
+++ b/examples/ex15.cpp
@@ -19,6 +19,14 @@
 //               ex15 -m ../data/square-disc.mesh
 //               ex15 -m ../data/escher.mesh -r 2 -tf 0.3
 //
+//               Kelly estimator:
+//
+//               ex15 -est 1
+//               ex15 -est 1 -o 1 -y 0.4
+//               ex15 -est 1 -o 4 -y 0.1
+//               ex15 -est 1 -n 5
+//               ex15 -est 1 -p 1 -n 3
+//
 // Description:  Building on Example 6, this example demonstrates dynamic AMR.
 //               The mesh is adapted to a time-dependent solution by refinement
 //               as well as by derefinement. For simplicity, the solution is
@@ -28,10 +36,10 @@
 //               At each outer iteration the right hand side function is changed
 //               to mimic a time dependent problem.  Within each inner iteration
 //               the problem is solved on a sequence of meshes which are locally
-//               refined according to a simple ZZ error estimator.  At the end
-//               of the inner iteration the error estimates are also used to
-//               identify any elements which may be over-refined and a single
-//               derefinement step is performed.
+//               refined according to a simple ZZ or Kelly error estimator.  At
+//               the end of the inner iteration the error estimates are also
+//               used to identify any elements which may be over-refined and a
+//               single derefinement step is performed.
 //
 //               The example demonstrates MFEM's capability to refine and
 //               derefine nonconforming meshes, in 2D and 3D, and on linear,
--- a/examples/ex15.cpp
+++ b/examples/ex15.cpp
@@ -78,6 +78,7 @@ int main(int argc, char *argv[])
    int nc_limit = 3;         // maximum level of hanging nodes
    bool visualization = true;
    bool visit = false;
+   int which_estimator = 0;
 
    OptionsParser args(argc, argv);
    args.AddOption(&mesh_file, "-m", "--mesh",
@@ -98,6 +99,9 @@ int main(int argc, char *argv[])
                   "Maximum level of hanging nodes.");
    args.AddOption(&t_final, "-tf", "--t-final",
                   "Final time; start time is 0.");
+   args.AddOption(&which_estimator, "-est", "--estimator",
+                  "Which estimator to use: "
+                  "0 = ZZ, 1 = Kelly. Defaults to ZZ.");
    args.AddOption(&visualization, "-vis", "--visualization", "-no-vis",
                   "--no-visualization",
                   "Enable or disable GLVis visualization.");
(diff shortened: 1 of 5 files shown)


dylan-copeland:
Unrelated to this PR: `ex15 -n 5` makes visually stunning plots in glvis.

psocratis (approved):
Other than that small valgrind compaint, everything looks good to me!

tzanio:

tzanio:
@brendankeith and @jandrej, the review for this PR are due soon 😁

examples/ex15.cpp
```diff
@@ -19,6 +19,14 @@
 //               ex15 -m ../data/square-disc.mesh
 //               ex15 -m ../data/escher.mesh -r 2 -tf 0.3
 //
+//               Kelly estimator:
+//
+//               ex15 -est 1
```
brendankeith, line 24: This runs a lot faster than the ZZ estimator. There seems to be about an order of magnitude difference in the errors being predicted.

I found that the command `./ex15 -est 1 -e 0.0001` gave more similar results

fem/estimators.hpp
```diff
@@ -426,13 +425,13 @@ class LpErrorEstimator : public ErrorEstimator
         ||∇(u-uₕ)||ₑ ≦ √( C hₑ ∑ₖ (hₖ ∫ |J[∇uₕ]|²) dS )
```
brendankeith, line 425: Typically, we want both an upper bound and a lower bound.

fem/estimators.hpp
```diff
@@ -426,13 +425,13 @@ class LpErrorEstimator : public ErrorEstimator
...
     corresponding faces. u is the analytic solution and uₕ the discretized
-    solution. hₖ and hₑ are factors dependend on the face and element geometry.
+    solution. hₖ and hₑ are factors dependent on the face and element geometry.
     J is the jump function, i.e. the difference between the limits at each point
     for each side of the face. A custom method to compute hₖ can be provided. It
     is also possible to estimate the error only on a subspace by feeding this
     class an attribute array describing the subspace.
 
-    @note This algorithm is only for Poisson problems a proper error esimator.
+    @note This algorithm is a proper error esimator only for Poisson problems.
```
brendankeith, line 434: What is meant by "proper" here? I think it that you can only get the upper bound with lowest order FEM for Poisson.

brendankeith (approved):
I provisionally approve this PR, conditioned on a reference being added to support the statements made about the error estimator which are made in `fem/estimators.hpp`.

This is a nice feature to have and I'm sure it will be well-used and appreciated.

fem/estimators.hpp
brendankeith: I suggest the following:

```
 It can be roughly described by:
        ||∇(u-uₕ)||ₑ ≅ √( C hₑ ∑ₖ (hₖ ∫ |J[∇uₕ]|²) dS )
    where "e" denotes an element, ||⋅||ₑ the corresponding local norm and k the
    corresponding faces. u is the analytic solution and uₕ the discretized
    solution. hₖ and hₑ are factors dependent on the face and element geometry.
    J is the jump function, i.e. the difference between the limits at each point
    for each side of the face. A custom method to compute hₖ can be provided. It
    is also possible to estimate the error only on a subspace by feeding this
    class an attribute array describing the subspace.

    @note This algorithm is only appropriate for problems with scalar diffusion 
coefficients (e.g., Poisson problems) because it measures only the flux of the
gradients of the discrete solution. The current implementation also does not
include the volume term present in Equation 75 of Kelly et al. Instead, it includes
only the flux term recommended in Equation 82.  The current implementation also 
does not include the constant factor "C".
    It furthermore assumes that the approximation error at the boundary is small
    enough, as the implementation ignores boundary faces.



```

dylan-copeland: Thanks for the suggested documentation. It is now committed.

examples/ex15.cpp
dylan-copeland: The sample run has been updated.

brendankeith (approved):

fem/estimators.hpp
```diff
@@ -423,31 +422,34 @@ class LpErrorEstimator : public ErrorEstimator
...
 
-    @note This algorithm is only for Poisson problems a proper error esimator.
-    The current implementation does not reflect this, because the "C" factor is
-    not included.
-    It further assumes that the approximation error at the boundary is small
-    enough, as the implementation ignores boundary faces.
+    @note This algorithm is appropriate only for problems with scalar diffusion
+    coefficients (e.g. Poisson problems), because it measures only the flux of
+    the gradient of the discrete solution. The current implementation also does
+    not include the volume term present in Equation 75 of Kelly et al. Instead,
```
jandrej, line 437: Can you reference Kelly et al. specifically with a ref footnote like
```
Equation 75 of [1] Instead ...
...
[1] Kelley reference
```

jandrej (approved):
I left a minor comment. Nicely done.

tzanio:
Can you also please update the `CHANGELOG` entry.

tzanio:

tzanio:
Can you check if the new Example 15 results in [tonight's test on tux426](https://mybitbucket.llnl.gov/projects/MFEM/repos/autotest/browse/tux426/2021-06-09-llnl-gh-next/sample-runs-baseline-tux426.diff) look OK?

tzanio (approved):
Thanks for adding this @dylan-copeland!

@brendankeith, @psocratis and @jandrej -- thank you for reviewing!

PR #2267, opened 2021-05-23
Expanding copyright statement checks [repo-check-dev]

mlstowell:
This PR expands the checks done in `repo-check.yml` to include the outdated release number and GPL license type. It also appeared that the check was only looking at lines starting with '#' i.e. comments in shell scripts so I expanded this to include "//" i.e. comments in C++ files.

Addresses issue #2265. 

| [#2267](https://github.com/mfem/mfem/pull/2267) | @mlstowell | @tzanio | @tzanio + @adrienbernede + @v-dobrev | 05/23/21 | 05/26/21 | 05/26/21 | |

### DIFF
--- a/.github/workflows/repo-check.yml
+++ b/.github/workflows/repo-check.yml
@@ -31,7 +31,7 @@ jobs:
     - name: copyright check
       run: |
         cd mfem
-        if git grep -l "^#.*\-2020" > matches.txt
+        if git grep -l "^\(#\|//\).*\(\-2020\|\ 2010,\)" > matches.txt
         then
           echo "Please update the following files to Copyright (c) 2010-2021:"
           cat matches.txt
--- a/.github/workflows/repo-check.yml
+++ b/.github/workflows/repo-check.yml
@@ -19,7 +19,7 @@ on:
   pull_request:
 
 jobs:
-  repo-check:
+  file-headers-check:
     runs-on: ubuntu-18.04
 
     steps:
--- a/.github/workflows/repo-check.yml
+++ b/.github/workflows/repo-check.yml
@@ -40,6 +40,48 @@ jobs:
           echo "No outdated copyright found."
         fi
 
+  license-check:
+    runs-on: ubuntu-18.04 # needed for astyle 2.05.1
+
+    steps:
+    - name: checkout mfem
+      uses: actions/checkout@v2
+      with:
+        path: mfem
+
+    - name: license check
+      run: |
+        cd mfem
+        if git grep -li "^\(#\|//\).*GNU\ Lesser\ General\ Public\ License" > matches.txt
+        then
+          echo "Please update the following files to the BSD-3 license:"
+          cat matches.txt
+          exit 1
+        else
+          echo "No GNU GPL license found."
+        fi
+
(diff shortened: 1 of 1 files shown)


tzanio:

tzanio (approved):

adrienbernede (changes requested):
I think those new jobs should be added as steps of a same job. This would prevent from having many very small jobs.
Minor: the comment "needed for astyle" should be removed.

adrienbernede:
Otherwise I think this is a great addition. Maybe we could add this to the git hooks.

mlstowell:
> I think those new jobs should be added as steps of a same job. This would prevent from having many very small jobs.
> Minor: the comment "needed for astyle" should be removed.

I tried this at first but I could not make it work the way I wanted. The desired outcome was to check for any of these outdated statements and show the user which files needed to be updated. However, since each check contains an `exit 1` the first failure would prevent the subsequent checks from running. I looked for a way to remove the `exit 1` and replace it with a status variable. Then after all checks had been run we could exit if the status variable indicated a failure had occurred. Do you know of a simple way to define such a status variable?

tzanio (approved):

tzanio:
@adrienbernede, can you please review

.github/workflows/repo-check.yml
```diff
@@ -19,8 +19,8 @@ on:
   pull_request:
 
 jobs:
-  copyright-check:
-    runs-on: ubuntu-18.04 # needed for astyle 2.05.1
+  repo-check:
```
adrienbernede, line 22: Maybe "headers-check" is a better name? The workflow is already called repo-check.
```suggestion
  headers-check:
```

adrienbernede (approved):
A last change? (correcting myself in fact).

.github/workflows/repo-check.yml
mlstowell: Except that it's checking every file not just headers, isn't it? I don't feel strongly about the name. We can certainly change it.

adrienbernede: I see. I meant "file header", the beginning of a file. But there is certainly a confusion with C++ header files...

mlstowell: That makes sense. Would `file-headers-check` be too long?

adrienbernede: No. If you think that’s enough to avoid the confusion, that works for me.
It seems that the generic term is "intellectual property", and `intellectual-property-check` seems too much, doesn’t it?

adrienbernede:
```suggestion
  file-headers-check:
```

mlstowell: I think `file-headers-check` will work well. Technically these checks are scanning the entire file looking for lines starting with comment strings so there's probably no concise way to state exactly what they are looking for. As you say `intellectual-property-check` may be the most accurate but it does seem a bit much.

tzanio: Another name some people use is `banner`. But I think `file-headers-check` is probably sufficient :)

tzanio:

PR #2271, opened 2021-05-25
tcew/occa/quad2d/matvec: added streamlined OKL PA diffusion kernels

tcew:
This is a proof of concept to show that reasonably well tuned quad and hex OCCA OKL kernels can match or exceed the performance of the cuda-gen backend kernels, at least on a single NVIDIA Titan V. 

**Notes**: 

- The hex kernel is designed to only use a single cube of shared memory and one 1D array of register memory in addition to variables created by the compiler and some other small shared memory arrays. This is intentional to reduce register and shared memory pressure enabling use for higher degree and quadratures.

- The hex kernel first interpolates (applies B) in all three directions before differentiating at the quadrature nodes and then tests in all three directions. This is a different approach to the existing kernels in mfem and is designed to reduce shared memory and register pressure.

-  The hex kernel needs a GL quadrature differentiation matrix for each quadrature size. I added these manually since it was not obvious how to do this via available mfem infrastructure. 

-  I did not explicitly write code to handle multiple fields in the quad and hex OKL kernels.

- The OKL kernels are not completely optimized and have only so far been tweaked and tested on the Titan V Volta class GPU.

- These kernels have not been tested on AMD MI class GPUs. To conform with the existing infrastructure in the cuda-gen the kernels use ```atomicAdd``` instructions in the gather phase. This will likely not perform well on the AMD MI GPUs.

The OKL kernels in this commit are based on modified versions of kernels from [libParanumal](https://github.com/paranumal/libparanumal) and [benchParanumal](https://github.com/paranumal/benchparanumal) as documented in Swirydowicz et al [arXiv version](https://arxiv.org/abs/1711.00903). Thanks and credit are due to Kasia Swirydowicz and Noel Chalmers for their contributions to the original kernel designs. 

Tim Warburton


tzanio:
Thanks for opening this PR, @tcew!

We will review and post comments/questions below.

PR #2280, opened 2021-05-28
Oops

adrienbernede:
Test branch history check on a PR.

Because actions checks out the merge commit (not the branch head) I want to make sure this does not invalidate the branch-history check.

I will remove the branch afterward.


adrienbernede:
That’s good news.

PR #2315, opened 2021-06-09
Implement tangential derivative evaluation for `Mesh::GetFaceGeometricFactors`

samuelpmishLLNL:
This is a feature serac/smith needs to support our tools for integrating over boundary regions.

in 2D (star.mesh) : 

in 3D (toroid-hex.mesh) :

I would have expected the tangents in 2D to be consistent (clockwise or anticlockwise), I'm not sure if I should be taking into a sign on the tangential derivatives or if that should only apply to the normal.

### DIFF
--- a/fem/quadinterpolator_face.cpp
+++ b/fem/quadinterpolator_face.cpp
@@ -91,7 +91,7 @@ void FaceQuadratureInterpolator::Eval2D(
    const int vdim,
    const DofToQuad &maps,
    const Array<bool> &signs,
-   const Vector &r_vec,
+   const Vector &f_vec,
    Vector &q_val,
    Vector &q_der,
    Vector &q_det,
--- a/fem/quadinterpolator_face.cpp
+++ b/fem/quadinterpolator_face.cpp
@@ -91,7 +91,7 @@ void FaceQuadratureInterpolator::Eval2D(
    const int vdim,
    const DofToQuad &maps,
    const Array<bool> &signs,
-   const Vector &f_vec,
+   const Vector &r_vec,
    Vector &q_val,
    Vector &q_der,
    Vector &q_det,
@@ -111,11 +111,9 @@ void FaceQuadratureInterpolator::Eval2D(
    auto F = Reshape(f_vec.Read(), ND1D, VDIM, NF);
    auto sign = signs.Read();
    auto val = Reshape(q_val.Write(), NQ1D, VDIM, NF);
-   // auto der = Reshape(q_der.Write(), NQ1D, VDIM, NF); // only tangential der
+   auto tangent = Reshape(q_der.Write(), NQ1D, VDIM, NF);
    auto det = Reshape(q_det.Write(), NQ1D, NF);
    auto n   = Reshape(q_nor.Write(), NQ1D, VDIM, NF);
-   MFEM_VERIFY(eval_flags | DERIVATIVES,
-               "Derivatives on the faces are not yet supported.");
    // If Gauss-Lobatto
    MFEM_FORALL(f, NF,
    {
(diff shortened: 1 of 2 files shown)


tzanio:
Not directly related, but how did you make these pictures?

samuelpmishLLNL:
I write the data to file in NDJSON format in this source file:
https://pastebin.com/WWW5SzTu

then read the data into mathematica and plot
[draw_facegeomfactors.pdf](https://github.com/mfem/mfem/files/6620580/draw_facegeomfactors.pdf)

tzanio:
Very cool!

v-dobrev:
> I would have expected the tangents in 2D to be consistent (clockwise or anticlockwise), I'm not sure if I should be taking into a sign on the tangential derivatives or if that should only apply to the normal.

This is a good point, I think the current direction of the derivative is the result of how it is computed. Maybe the direction of the tangent is not that important, however, the normal vector should probably point outside the domain.

samuelpmishLLNL:
Are there any existing unit tests for GetFaceGeometricFactors that I might modify to verify the tangential derivative calculations as well?

samuelpmishLLNL:
Applying the sign to the tangents
```cpp
if (eval_flags & DERIVATIVES)
{
  const double s = sign[f] ? -1.0 : 1.0;
  tangent(q,0,f) = s * D[0];
  tangent(q,1,f) = s * D[1];
}
```
does seem to make their orientations consistent:



It seems like it would be easier to directly encode an element's orientation through its node numbering, as opposed to maintaining a separate list of which elements are "backwards". 

Is there any preference for whether or not the sign should be applied, or if the coordinate system defined by {dx_dxi, n} (in 2D) and {dx_dxi, dx_deta, n} (in 3D) be left- or right-handed?

v-dobrev:
I now see that these input  `sign`a are applied to the normals, so for consistency, I think we should apply them to the tangential derivatives as well.

v-dobrev:
@samuelpmishLLNL, is the PR ready for review? If so, please add the "ready-for-review" label.

samuelpmishLLNL:
Can you help me better understand what the "sign" represents in this context (in 2D and 3D), to ensure that the tangential derivatives we provide will be consistent with the rest of the mfem? Off the top of my head, there are few different possible interpretations of what the "sign" might mean:

1. The sign of a face _only_ applies to the normal vector.
2. The sign of a face implies a local reordering of the nodes (e.g. from CW to CCW).
3. The sign of a face keeps the node ordering the same, but flips the orientation of parent element coordinates.
4. Something else entirely

The treatment of the tangential derivatives for these cases will depend on this interpretation, so I'd like to make sure we're all on the same page, and I'm not just sprinkling factors of -1 here and there to make one example look correct! I also want to be sure that we get the handedness of the Jacobian matrix to be in agreement with mfem conventions, too.

v-dobrev:
When the tangential derivatives are computed initially (in the variable `D` in `FaceQuadratureInterpolator::Eval2D` and the variables `BGu` and `GBu` in `FaceQuadratureInterpolator::Eval3D`), these derivatives are with respect to a reference coordinate system on the face that is not necessarily oriented in a way that will give normal pointing from element 1 to element 2 (the two elements adjacent to the current face). That is why the "signs" are computed with the function `GetSigns` in `quadinterpolator_face.cpp` -- the "sign" is set to `false` if the normal is oriented to point outside element 1, and `true` otherwise.

As I suggested above, for consistency with the normals, it makes sense to also apply -1 to the original tangential derivatives when the "sign" is `true`. This introduces an inconsistency between the ordering of the face quadrature points and the reference coordinate system for the tangential derivatives, however, this should not be a problem -- at this level each quadrature point is independent of the other quadrature points.

PR #2326, opened 2021-06-11
Githooks: share checks with github actions

adrienbernede:
@tzanio @pazner 

Per Tzanio request this is a suggestion about how to have the checks coded in only one place.

This is true for "copyright", "license", "release". For "style" and "history" this is an extra layer on top of "runtest".

That's why it is a suggestion. Advantage: Github Actions really run the the same script as the pre-push githook. (documentation check is separate).

### DIFF
--- a/.github/workflows/repo-check.yml
+++ b/.github/workflows/repo-check.yml
@@ -32,49 +32,24 @@ jobs:
         access_token: ${{ github.token }}
     - name: checkout mfem
       uses: actions/checkout@v2
-      with:
-        path: mfem
 
     - name: copyright check
       id: copyright
       run: |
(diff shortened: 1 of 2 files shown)


adrienbernede:
And I should have tested it locally... To Be Continued.

tzanio:
reminder 😁

adrienbernede:
Fixed!
The strategy here is to ensure that git pre-push hooks does the same as github actions repo-check.
Put another way: use the githook and you are sure the repo-check action will pass.

tzanio:
Thanks! Lets review as part of https://github.com/mfem/mfem/tree/bernede1/githooks.

PR #2335, opened 2021-06-15
Splitting fe.[ch]pp

mlstowell:
This is a draft PR to test the splitting proposed in discussion #2322.

| [#2335](https://github.com/mfem/mfem/pull/2335) | @mlstowell | @tzanio | @pazner + @tzanio + @v-dobrev + @psocratis | 07/08/21 | 09/26/21 | 09/28/21 | |


tzanio:
I like this a lot -- it brings clarity and useful organization to the code. Thank you @mlstowell !

Some additional suggestions:

- I personally prefer `fe_base.?pp` to be just `fe.?pp`, partially because we don't use `base` it elsewhere, but I don't feel strongly about that.

- I like `fe_fixed_order`, it will be great if there is a shorter name for it, but I couldn't think of anything.

- I like `fe_pos`, `fe_h1`, `fe_nd`, `fe_rt` and `fe_l2` these are all very clean and useful.

- I am not sure if `fe_nurbs` and `fe_ser` deserve their own file, but I don't see any big harm in having them either

- Suggestion: do we want to put all `fe_*?pp` files in a `fem/fe/` subdirectory? I got a feedback today that "mfem has a lot of files"...

mlstowell:
> * I personally prefer `fe_base.?pp` to be just `fe.?pp`, partially because we don't use `base` it elsewhere, but I don't feel strongly about that.

That was the original intent but we need to declare the base classes before any derived classes. This means that we would need the base class declarations in `fe.hpp` before including the other `fe_*.hpp` headers which would put these `include` directives towards the end of `fe.hpp`. We could do this but I feel it would be easier to follow if all of the `includes` are at the top.

> * I like `fe_fixed_order`, it will be great if there is a shorter name for it, but I couldn't think of anything.

I agree completely but I couldn't think of a better name either. Suggestions are welcome.

> * I am not sure if fe_nurbs and fe_ser deserve their own file, but I don't see any big harm in having them either

Agreed. We could combine them into `fe_misc.?pp` but I don't feel strongly either way. Having separate files does highlight them somewhat which I do like.

> * Suggestion: do we want to put all fe_*?pp files in a fem/fe/ subdirectory?

I have no strong feelings about this. It might be cleaner to have them in a subdirectory.

mlstowell:
After merging with the current master I made sure to update `ND_SegmentElement` so that it inherits from `VectorTensorFiniteElement`.

tzanio:

fem/fe.cpp
```diff
@@ -13,7 +13,6 @@
 
 #include "fe.hpp"
 #include "fe_coll.hpp"
-#include "../mesh/nurbs.hpp"
 #include "bilininteg.hpp"
 #include <cmath>
```
psocratis, line 17: Are these includes (besides fe.hpp) needed here?

psocratis (approved):
Looks good to me!

fem/fe.cpp
mlstowell: These were carried over from the original fe.cpp but you make a good point. I'll double check them and make sure they're still needed here. I may even add comments explaining why they're needed if it isn't obvious...

fem/fe/fe_h1.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+#ifndef MFEM_FE_H1
+#define MFEM_FE_H1
+
+#include "fe_base.hpp"
+
+namespace mfem
+{
+
+/// Arbitrary H1 elements in 1D
```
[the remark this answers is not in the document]
mlstowell: Good idea!

fem/fe/fe_fixed_order.hpp
```diff
@@ -0,0 +1,976 @@
...
+   /** virtual function which evaluates the values of all
+       partial derivatives of all shape functions at a given
+       point ip and stores them in the matrix dshape (Dof x Dim) (9 x 2)
+       so that each row contains the derivatives of one shape function */
+   virtual void CalcDShape(const IntegrationPoint &ip,
+                           DenseMatrix &dshape) const;
+};
+
+
+/*
```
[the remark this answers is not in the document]
mlstowell: Good catch. These classes are not used anywhere and they created an unresolved dependency while I was rearranging things. We can resolve this now by adding a couple more include directives.

fem/fe.cpp
mlstowell: Apparently they were not needed. I have removed them. Thanks for catching this.

mlstowell:
I wrote a pair of Perl scripts to recombine the split files into one source and one header. After using `diff` to compare these recombined files with those in master a handful of small changes were made to the split files to bring them up to date with master.

The only remaining differences were white space or one outdated comment.

Unfortunately these Perl scripts are not fully automated so we cannot easily use them to track changes going forward.

pazner (approved):
Wow, this is so much easier to read and navigate! Huge improvement!

@mlstowell, could you share your Perl script (even if it's not fully automated)?

mlstowell:
> Wow, this is so much easier to read and navigate! Huge improvement!
> 
> @mlstowell, could you share your Perl script (even if it's not fully automated)?

Thanks, @pazner. I sent you the scripts on Slack.

tzanio:

tzanio (approved):
Thanks @mlstowell, this really is _much_ cleaner!

PR #2339, opened 2021-06-17
Update ParaViewDataCollection::Save() method to preserve metadata for previous timesteps

koomie:
#### Background

The current `ParaViewDataCollection::Save` method looks to nicely save time-series data as a collection so that multiple time steps can be interrogated after loading in Paraview.  Metadata for the timestep values is included in an XML file  with a suffix of `.pvd`. There are no issues encountered for a single run of a transient analysis. However, if a user considers using this funcionality in conjunction with a restart method for their application, one has to be careful to save the `.pvd` file in advance as it is overwritten every run and the exact timestep values associate with a given iteration will be lost; however, the raw solution files (`.vtu`) will still exist from previous time steps.

This PR proposes an update to enable preserving previous timesteps (if the `.pvd` file exists) so that end users can more easily maintain a single output directory with Paraview output in longer-running simulations that require restart capability.

#### Example of current behavior

Consider a time-stepping CFD example that is run for 20 timesteps that calls  the `ParaViewDataCollection::Save()` method every 5th timestep.  In that case, we have four solutions saved with the timestep values and a `.pvd` file as follows:

```
$ ls output
Cycle000005  Cycle000010  Cycle000015  Cycle000020   output.pvd

$ cat output/output.pvd
<?xml version="1.0"?>
<VTKFile type="Collection" version="0.1" byte_order="LittleEndian">
<Collection>
<DataSet timestep="1.22855e-05" group="" part="0" file="Cycle000005/data.pvtu"/>
<DataSet timestep="2.40118e-05" group="" part="0" file="Cycle000010/data.pvtu"/>
<DataSet timestep="3.5577e-05" group="" part="0" file="Cycle000015/data.pvtu"/>
<DataSet timestep="4.70662e-05" group="" part="0" file="Cycle000020/data.pvtu"/>
</Collection>
</VTKFile>
```
If we restart this case and run another 10 iterations, we will add two additional solutions to the output directory, but because the `.pvd` file is currently overwritten every time, we lose the time values associated with the previous outputs.

```
$ ls output/
Cycle000005  Cycle000010  Cycle000015  Cycle000020  Cycle000025  Cycle000030  output.pvd

$ cat output/output.pvd 
<?xml version="1.0"?>
<VTKFile type="Collection" version="0.1" byte_order="LittleEndian">
<Collection>
<DataSet timestep="5.85358e-05" group="" part="0" file="Cycle000025/data.pvtu"/>
<DataSet timestep="6.99925e-05" group="" part="0" file="Cycle000030/data.pvtu"/>
</Collection>
</VTKFile>
```

#### Example of proposed behavior

With this PR, the `Save` method is updated slightly to check if the `.pvd` file exists, and if so, it is scanned for any previous timesteps and the stream `pos` is updated accordingly so subsequent timesteps are appended.  The resulting file after completing the same restart process highlighted above is then:

```
 $ ls output/
Cycle000005  Cycle000010  Cycle000015  Cycle000020  Cycle000025  Cycle000030  output.pvd

$ cat output/output.pvd 
<?xml version="1.0"?>
<VTKFile type="Collection" version="0.1" byte_order="LittleEndian">
<Collection>
<DataSet timestep="1.22855e-05" group="" part="0" file="Cycle000005/data.pvtu"/>
<DataSet timestep="2.40118e-05" group="" part="0" file="Cycle000010/data.pvtu"/>
<DataSet timestep="3.5577e-05" group="" part="0" file="Cycle000015/data.pvtu"/>
<DataSet timestep="4.70662e-05" group="" part="0" file="Cycle000020/data.pvtu"/>
<DataSet timestep="5.85358e-05" group="" part="0" file="Cycle000025/data.pvtu"/>
<DataSet timestep="6.99925e-05" group="" part="0" file="Cycle000030/data.pvtu"/>
</Collection>
</VTKFile>
```

Note: regardless of the current implementation or proposed update in this PR, the end application will need to take care of what to do when a simulation is starting from scratch in the same directory where they may have saved previous solution data.  In the case of this PR, a user can additional logic to remove an existing `.pvd` file if they do not want to preserve any previous data, but would gain the benefit of being able to do restarts and have their Paraview output all in one directory.

| [#2339](https://github.com/mfem/mfem/pull/2339) | @koomie | @tzanio | @pazner + @jandrej | 06/18/21 | 06/28/21 | 06/29/21 | |

### DIFF
--- a/fem/datacollection.cpp
+++ b/fem/datacollection.cpp
@@ -867,16 +867,16 @@ void ParaViewDataCollection::Save()
          std::smatch match;
          while (getline(pvd_stream,line))
          {
-            if(regex_search(line,match,regexp))
+            if (regex_search(line,match,regexp))
             {
                MFEM_ASSERT(match.size() == 2,
-			   "Unable to ascertain time value from existing pvd file: " << pvdname);
+                           "Unable to ascertain time value from existing pvd file: " << pvdname);
                double tvalue = std::stod(match[1]);
                if (tvalue < GetTime())
                {
                   pos = pvd_stream.tellg();
                }
-             }
+            }
          }
          pvd_stream.clear();
          pvd_stream.seekp(pos);
--- a/fem/datacollection.cpp
+++ b/fem/datacollection.cpp
@@ -869,7 +869,7 @@ void ParaViewDataCollection::Save()
          {
             if(regex_search(line,match,regexp))
             {
-               MFEM_ASSERT(match.size == 2,
+               MFEM_ASSERT(match.size() == 2,
 			   "Unable to ascertain time value from existing pvd file: " << pvdname);
                double tvalue = std::stod(match[1]);
                if (tvalue < GetTime())
--- a/fem/datacollection.cpp
+++ b/fem/datacollection.cpp
@@ -892,8 +892,10 @@ void ParaViewDataCollection::Save()
             // remove timesteps newer than current time
             pvd_stream.close();
             int status = truncate(pvdname.c_str(), pos);
-            MFEM_ASSERT(status == 0,
-                        "Unable to truncate newer time values from pvd file: " << pvdname);
+            if (status != 0)
+            {
+               MFEM_ABORT("Unable to truncate newer time values from pvd file:: " << pvdname);
+            }
             pvd_stream.open(pvdname.c_str(),std::ios::in|std::ios::out);
          }
 #endif
--- a/fem/datacollection.cpp
+++ b/fem/datacollection.cpp
@@ -17,6 +17,7 @@
 
 #include <cerrno>      // errno
 #include <sstream>
+#include <regex>
 
 #ifndef _WIN32
 #include <sys/stat.h>  // mkdir
@@ -764,7 +765,8 @@ ParaViewDataCollection::ParaViewDataCollection(const std::string&
    : DataCollection(collection_name, mesh_),
      levels_of_detail(1),
      pv_data_format(VTKFormat::BINARY),
-     high_order_output(false)
+     high_order_output(false),
+     restart_mode(false)
 {
 #ifdef MFEM_USE_ZLIB
    compression = -1; // default zlib compression level, equivalent to 6
@@ -842,7 +844,10 @@ void ParaViewDataCollection::Save()
    }
    // the directory is created
 
-   // create pvd file if needed
+   // create pvd file if needed. If we are not in restart mode, a new pvd file
+   // is always created. In restart mode, we keep any previously defined
+   // timestep values as long as they are less than the currently defined time.
+
    if (myid == 0 && !pvd_stream.is_open())
    {
       std::string dpath=GenerateCollectionPath();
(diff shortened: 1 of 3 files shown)


pazner:
Hi @koomie, this looks like a nice contribution, thank you. If it is ready, and you would like it to be reviewed for inclusion in MFEM, you can tag the PR as "ready-for-review".

koomie:
> Hi @koomie, this looks like a nice contribution, thank you. If it is ready, and you would like it to be reviewed for inclusion in MFEM, you can tag the PR as "ready-for-review".

I don't believe I have the ability to add issue labels, but if you think it would be useful, the PR is `ready-for-review` from my perspective.

tzanio:
@koomie, you are now part of our `Contributors` team and should be able to add labels in the future

tzanio:

koomie:
> @koomie, you are now part of our `Contributors` team and should be able to add labels in the future

Sweet. Thank you - I promise to use any future labeling powers for good....

jandrej (approved):

pazner:
This PR will always append to the end of the existing PVD file, independent of the `timestep` value. I think we might want to be a little bit more sophisticated here.

A common use-case for us (e.g. in example 5) is to output a single timestep for a steady problem. If you run the example multiple times, this will duplicate the `DataSet` entry:
```xml
<DataSet timestep="0" group="" part="0" file="Cycle000000/data.pvtu"/>
<DataSet timestep="0" group="" part="0" file="Cycle000000/data.pvtu"/>
<DataSet timestep="0" group="" part="0" file="Cycle000000/data.pvtu"/>
```
This is pretty harmless (ParaView will still properly read the file), but I think we want to avoid this.

Similarly, if we restart a time-dependent run *without checkpointing*, the new entries will simply be appended:
```xml
<DataSet timestep="0" group="" part="0" file="Cycle000000/data.pvtu"/>
<DataSet timestep="0.05" group="" part="0" file="Cycle000005/data.pvtu"/>
<DataSet timestep="0.1" group="" part="0" file="Cycle000010/data.pvtu"/>
<DataSet timestep="0" group="" part="0" file="Cycle000000/data.pvtu"/>
<DataSet timestep="0.05" group="" part="0" file="Cycle000005/data.pvtu"/>
```
In this case, the user would want the old results to be completely overwritten rather than appended to. In situations where simulation parameters change (e.g. different time step), this could get very confusing.

I think there are two points:
* We probably want a flag to enable or disable checkpointing. Sometimes a user would want to append to an existing file, and sometimes they would want to replace it. This could be added to the interface of `ParaViewDataCollection`.
* Rather than always appending to the *end* of the file, it would make sense to find the largest time `t0` that is strictly less than the current time `t`, and append after `t0`, deleting everything else. The restart point used for checkpointing may be older than the latest visualization step.

koomie:
> I think there are two points:
> 
> * We probably want a flag to enable or disable checkpointing. Sometimes a user would want to append to an existing file, and sometimes they would want to replace it. This could be added to the interface of `ParaViewDataCollection`.
> * Rather than always appending to the _end_ of the file, it would make sense to find the largest time `t0` that is strictly less than the current time `t`, and append after `t0`, deleting everything else. The restart point used for checkpointing may be older than the latest visualization step.

I like both of these suggestions and am happy to take a crack at an updated PR to try and address these points.  Would you like a fresh PR when ready, or is it ok to force push this branch in this case?

koomie:
PR has been updated to add a `SetRestartMode` method for the `ParaviewDataCollection` that a user can enable to trigger saving older timesteps in an existing PVD file. Should hopefully address the two points above.

koomie:
> PR has been updated to add a `SetRestartMode` method for the `ParaviewDataCollection` that a user can enable to trigger saving older timesteps in an existing PVD file. Should hopefully address the two points above.

Small addendum. I realized I did not delete timesteps that might exist that after the current `t` so tweaked again to do that.  This does use `truncate` which I believe would not be available on windows so I disabled this new `SetRestartMode()` method for windows builds. 

I believe this is now ready for a second look.

tzanio:
LGTM

pazner:
@koomie, thanks for the update. I pushed some changes to your branch.

In particular, I changed the implementation slightly so that it doesn't depend on `truncate`. Also, I added a check that the current cycle will not overwrite any of the existing cycles that will be retained. I also renamed `SetRestartMode` to `UseRestartMode` and added a small unit test to try to test this behavior.

Let me know how this looks to you.

@jandrej, could you take a look at these latest changes?

koomie:
@pazner, the updates look good to me (and nice to add a unit test).

pazner (approved):

tzanio:
@jandrej, can you re-approve if you are OK with the new changes?

jandrej (approved):

tzanio:

tzanio (approved):
Thanks for adding this useful functionality @koomie 👍 

@pazner and @jandrej -- _Thank you for reviewing!_

koomie:
My pleasure; thanks for adding and making it better.

PR #2356, opened 2021-06-24
Change the destructor of `Memory<T>` to delete.

YohannDudouit:
The current destructor of the `Memory<T>` class currently doesn't delete the memory, even when owned, this PR proposes to change this behavior. This should also help catching issues by enforcing a more strict memory ownership.


YohannDudouit:
It looks like this PR reveals some multiple ownership issues.

PR #2360, opened 2021-06-26
Solver const correctness [solver-const-correctness]

pazner:
Some simple const correctness fixes for `OperatorChebyshevSmoother` and `BlockILU`.

| [#2360](https://github.com/mfem/mfem/pull/2360) | @pazner | @tzanio | @mlstowell + @tzanio | 06/26/21 | 06/28/21 | 06/29/21 | |

### DIFF
--- a/linalg/solvers.cpp
+++ b/linalg/solvers.cpp
@@ -232,7 +232,7 @@ void OperatorJacobiSmoother::Mult(const Vector &x, Vector &y) const
    MFEM_FORALL(i, height, Y[i] += DI[i] * R[i]; );
 }
 
-OperatorChebyshevSmoother::OperatorChebyshevSmoother(Operator* oper_,
+OperatorChebyshevSmoother::OperatorChebyshevSmoother(const Operator* oper_,
                                                      const Vector &d,
                                                      const Array<int>& ess_tdofs,
                                                      int order_, double max_eig_estimate_)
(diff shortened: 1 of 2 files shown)


tzanio:

tzanio (approved):

linalg/solvers.cpp
```diff
@@ -232,7 +232,7 @@ void OperatorJacobiSmoother::Mult(const Vector &x, Vector &y) const
    MFEM_FORALL(i, height, Y[i] += DI[i] * R[i]; );
 }
 
-OperatorChebyshevSmoother::OperatorChebyshevSmoother(Operator* oper_,
+OperatorChebyshevSmoother::OperatorChebyshevSmoother(const Operator* oper_,
```
mlstowell, line 235: Is there a compelling reason to keep this argument a pointer? The other similar methods use references and internally there is an abort if `oper` is NULL.

pazner: No, I think you're right that the correct interface here would be to pass `oper` by reference since it should never be NULL. However, I think we're stuck with the current version for backwards compatibility (we could add alternatives that accept `const Operator &` but I'm not sure it would be worth the effort).

Perhaps we can keep track of this and other things that would be nice to change but break backwards compatibility so that we could make this change at some point (e.g. next major version release).

mlstowell: Little things like this are so easily forgotten. @tzanio, do we have something like a todo list for the next release?

I'd be happy to put together a small PR to deprecate these constructors and add new ones with const references.

mlstowell (approved):

linalg/solvers.cpp
tzanio: Good point @mlstowell. We don't have a specific list yet... how about starting an mfem-4.4 todo list [discussion](https://github.com/mfem/mfem/discussions)?

tzanio: see https://github.com/mfem/mfem/discussions/2365

tzanio:

PR #2370, opened 2021-06-29
Changing OperatorChebyshevSmoother to pass-by-reference [op-by-ref-dev]

mlstowell:
The reason for this change in two-fold. First, all other classes declared in `solvers.hpp` use pass-by-reference for their operators so `OperatorChebyshevSmoother` may cause confusion. Second, this class will abort if the operator is NULL so pass-by-reference will ensure that a valid operator object has been passed.

Note: this PR includes deprecation warnings on the original constructors, however, these don't seem to work. I'm not sure why...

| [#2370](https://github.com/mfem/mfem/pull/2370) | @mlstowell | @tzanio | @pazner + @barker29 | 06/30/21 | 07/06/21 | 07/09/21 | |

### DIFF
--- a/examples/ex26.cpp
+++ b/examples/ex26.cpp
@@ -105,7 +105,7 @@ private:
       Vector diag(fespace.GetTrueVSize());
       bfs.Last()->AssembleDiagonal(diag);
 
-      Solver* smoother = new OperatorChebyshevSmoother(opr.Ptr(), diag,
+      Solver* smoother = new OperatorChebyshevSmoother(*opr, diag,
                                                        *essentialTrueDofs.Last(), 2);
       AddLevel(opr.Ptr(), smoother, true, true);
    }
--- a/examples/ex26p.cpp
+++ b/examples/ex26p.cpp
@@ -115,7 +115,7 @@ private:
       Vector diag(fespace.GetTrueVSize());
       bfs.Last()->AssembleDiagonal(diag);
 
-      Solver* smoother = new OperatorChebyshevSmoother(opr.Ptr(), diag,
+      Solver* smoother = new OperatorChebyshevSmoother(*opr, diag,
                                                        *essentialTrueDofs.Last(), 2, fespace.GetParMesh()->GetComm());
 
       AddLevel(opr.Ptr(), smoother, true, true);
--- a/fem/ceed/algebraic.cpp
+++ b/fem/ceed/algebraic.cpp
@@ -143,7 +143,7 @@ Solver *BuildSmootherFromCeed(ConstrainedOperator &op, bool chebyshev)
    if (chebyshev)
    {
       const int cheb_order = 3;
-      out = new OperatorChebyshevSmoother(&op, t_diag, ess_tdofs, cheb_order);
+      out = new OperatorChebyshevSmoother(op, t_diag, ess_tdofs, cheb_order);
    }
    else
    {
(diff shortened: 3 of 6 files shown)


tzanio:

pazner (commented):
Looks great, except I find this a bit concerning:
> Note: this PR includes deprecation warnings on the original constructors, however, these don't seem to work. I'm not sure why...

mlstowell:
> Looks great, except I find this a bit concerning:
> 
> > Note: this PR includes deprecation warnings on the original constructors, however, these don't seem to work. I'm not sure why...

Yes, I do as well. These constructors are used in `ex26`, `ex26p`,  and a dedicated unit test (also in the ceed subdir). In each case, except ceed, an `OperatorHandle` object is used. I was initially concerned that this class might have a copy constructor which was avoiding the use of the deprecated `OperatorChebyshevSmoother` constructor. I was able to confirm that this was not the case and the deprecated constructor was indeed being called even though no warning was being displayed.

It may just be an issue with my compiler (clang 11.0.0 on Mac OS 10.14.6). I should try the same tests on an LC linux machine and see what happens.

mlstowell:
> > Looks great, except I find this a bit concerning:
> > > Note: this PR includes deprecation warnings on the original constructors, however, these don't seem to work. I'm not sure why...
> 
> Yes, I do as well. These constructors are used in `ex26`, `ex26p`, and a dedicated unit test (also in the ceed subdir). In each case, except ceed, an `OperatorHandle` object is used. I was initially concerned that this class might have a copy constructor which was avoiding the use of the deprecated `OperatorChebyshevSmoother` constructor. I was able to confirm that this was not the case and the deprecated constructor was indeed being called even though no warning was being displayed.
> 
> It may just be an issue with my compiler (clang 11.0.0 on Mac OS 10.14.6). I should try the same tests on an LC linux machine and see what happens.

The new deprecation warning does work on borax using gcc. Perhaps its an Apple specific issue...

pazner:
> The new deprecation warning does work on borax using gcc. Perhaps its an Apple specific issue...

Good to know.

I would suggest changing the pointer versions of the constructors to simply call the reference versions to avoid code duplication.

barker29 (approved):
This looks fine to me.

tzanio:
> This looks fine to me.

thanks Andrew!

drzisga (approved):
I wonder why I didn't do it like this in the first place when creating the ChebyshevSmoother, but thanks for fixing it now! This looks good to me.

tzanio:
thanks @dridzal !

mlstowell:
> > The new deprecation warning does work on borax using gcc. Perhaps its an Apple specific issue...
> 
> Good to know.
> 
> I would suggest changing the pointer versions of the constructors to simply call the reference versions to avoid code duplication.

To be honest I'm not sure how to do that with constructors. We could move the member data initializations into an `Init` member function and the power method initialization into another member function and call each of these as appropriate. Frankly this seems like a bit much if we're just going to remove the deprecated constructors in a forthcoming version.

pazner:
> To be honest I'm not sure how to do that with constructors. We could move the member data initializations into an `Init` member function and the power method initialization into another member function and call each of these as appropriate. Frankly this seems like a bit much if we're just going to remove the deprecated constructors in a forthcoming version.

I just pushed a change that calls the reference constructors from the pointer constructors. Let me know if it looks good to you.

mlstowell:
> > To be honest I'm not sure how to do that with constructors. We could move the member data initializations into an `Init` member function and the power method initialization into another member function and call each of these as appropriate. Frankly this seems like a bit much if we're just going to remove the deprecated constructors in a forthcoming version.
> 
> I just pushed a change that calls the reference constructors from the pointer constructors. Let me know if it looks good to you.

Thanks, @pazner! For some reason I never realized that would work. I do like this change. Thanks again.

pazner (approved):

tzanio:

PR #2424, opened 2021-07-23
ParELAG miniapps: AMGe for H(curl) and H(div) [parelag-miniapps-dev]

delyank:
This introduces miniapps that demonstrate ParELAG in the context of solving H(curl) and H(div) finite element problems. ParELAG itself demonstrates the utility of MFEM in the context of a complex finite element and AMGe code, involving the entire de Rham sequence of Sobolev spaces and the ability to work on mesh subentities.

Please, see the `README` file, including the references therein.

| [#2424](https://github.com/mfem/mfem/pull/2424) | @delyank | @tzanio | @dylan-copeland + @mlstowell + @tzanio | 07/29/21 | 09/16/21 | 09/26/21 | |


chakshinglee:
@delyank I haven't tried to link with ParELAG and play with the miniapps yet (I will), but I did go through the files, and the contents look fine to me. I do have some quick comments on the high-level arrangement/structure. Not sure if I misunderstood, but I view this as demonstration of using the solvers in ParELAG to solve FE problems produced by MFEM. At least I think this is what the MFEM community will be interested in. With this in mind:
1. I think the details of the internal construction in ParELAG (like looping over coarsening of topology, de Rham sequences) may be wrapped into one or two functions. Ideally, it will be nice if we have a function/class that takes the mesh, coefficient, boundary attributes, coarsening parameters, and solver parameters, etc. as inputs, and returns the solver. Or make a class `parelag::AMGe`. In fact, such a function may be more suitable to reside in `ParELAG`. 
2. The two miniapps are solving a more general version of problems in `ex3p` and `ex4p`. It is probably easier for the MFEM community to follow if the discrete problems are assembled using the same way as in `ex3p` and `ex4p`. That is, call `FormLinearSystem` through `ParBilinearForm` instead of extracting matrices from ParELAG. This way we can also hide/avoid the usage of `SharingMap`. Essentially, I am picturing (without the PA and static condensation stuff), in `ex3p` for example, after assembling the discrete problem, we just replace AMS by the AMGe solver constructed in ParELAG.

delyank:
@chakshinglee, it is a demonstration of ParELAG, where building certain solvers using ParELAG is a utility to demonstrate how the components of the library are invoked and used. ParELAG is a user library for MFEM so implicitly it demonstrates the usability of MFEM across the board.

No, ParELAG is not a solver library per se. It is a numerical up-scaling library and that determines a lot of things. The core idea of the library is not to provide solvers for MFEM-generated problems and be invoked as a solver after MFEM is invoked to build a problem, but rather it provides tools that can be used for building solvers and upscaled solutions for finite element problems. In fact, ParELAG is more of a discretization library that wraps around MFEM, not something that is side-by-side with MFEM (like HYPRE/AMS/ADS for example).

1. My understanding is that ParELAG's interface is decided already and I don't think it's the place here to discuss the interface to another library apart from MFEM itself. Here, we should discuss the miniapps. They use ParELAG correctly according to the available interface that was decided by Umberto, Andrew and probably yourself. It's not up to me to determine ParELAG's interface. This we can discuss in the ParELAG's repo and you feel free to submit a pull request there with the suggested interface improvement. There's no resource to work on ParELAG's interface currently. We can discuss this and decide on this, but I don't think this is the place. The current interface has flexibility though, since it allows the choice of partitioners, etc., directly by the library user (without calling a function with tens of arguments that can be also confusing), and demonstrates the separate components explicitly, which is part of the goal of these miniapps.

2. Again, this falls in the category of ParELAG being a wrapper around MFEM on the finest level. Since it has its own finest level constructs (that internally invoke MFEM), it actually makes sense to extract the matrices from it, since it is a multilevel discretization library and not simply a solver library that you would typically call with your extarnal matrix. In principle, what you suggest can be done, but it would be an incorrect way of using ParELAG. I also have grievances with the way things are, but the alternative that you suggest would mean computing things twice: once within ParELAG, which computes element matrices (from MFEM bilinear forms) and assembles internally, and another time within the miniapp itself invoking MFEM directly and using `FormLinearSystem`. Such a suboptimal and not fully right demonstration fails the purpose of exhibiting ParELAG. We need to at least use it correctly in accordance with its philosophy (whether good or bad).

So, Chak, I don't know what to do with your suggestions while being faithful to the goal of using ParELAG correctly as provided. I don't think the particularity of ParELAG's interface is MFEM's concern. Somebody decided for their own good reasons to wrap MFEM within ParELAG classes and they thought this is the best thing to do for that context.

chakshinglee:
If the purpose is not focus on solvers, then that's fine. Dont worry about my previous comments then.

tzanio:

miniapps/parelag/MultilevelHcurlSolver.cpp
```diff
@@ -0,0 +1,739 @@
...
+// at the Lawrence Livermore National Laboratory. All Rights reserved. See files
+// LICENSE and NOTICE for details. LLNL-CODE-806117.
+//
+// This file is part of the MFEM library. For more information and source code
+// availability visit https://mfem.org.
+//
+// MFEM is free software; you can redistribute it and/or modify it under the
+// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// CONTRIBUTING.md for details.
+
```
dylan-copeland, line 11: As done in some other miniapps, it may be helpful to comment at the beginning something like: "// We recommend viewing example 3 before viewing this miniapp."

miniapps/parelag/MultilevelHcurlSolver.cpp
```diff
@@ -0,0 +1,739 @@
...
+   mpi_session session(argc, argv);
+   MPI_Comm comm = MPI_COMM_WORLD;
+   int num_ranks, myid;
+   MPI_Comm_size(comm, &num_ranks);
+   MPI_Comm_rank(comm, &myid);
+
+   Timer total_timer = TimeManager::AddTimer("Program Execution -- Total");
+   Timer init_timer = TimeManager::AddTimer("Initial Setup");
+
+   if (!myid)
```
dylan-copeland, line 43: There are few differences between this file and MultilevelHdivSolver.cpp, so it may not be too awkward to combine them into one file. The advantage of keeping them separate is a little more clarity, but merging them may be more manageable in the future.

miniapps/parelag/MultilevelHcurlSolver.cpp
```diff
@@ -0,0 +1,739 @@
...
+      Vector X(A->Width()), x(sequence[0]->GetNumberOfDofs(jform));
+      X=0.0;
+
+      {
+         Vector tmp(A->Height());
+         A->Mult(X, tmp);
+         tmp *= -1.0;
+         tmp += B;
+
+         double local_norm = tmp.Norml2() * tmp.Norml2();
```
dylan-copeland, line 650: I'm sure this has no significant impact on performance, but it is generally better not to compute `Norml2` twice.

miniapps/parelag/MultilevelHcurlSolver.cpp
```diff
@@ -0,0 +1,739 @@
...
+
+   return EXIT_SUCCESS;
+}
+
+// A vector field, used for setting boundary conditions.
+void bdrfunc(const Vector &p, Vector &F)
+{
+   const int dim = p.Size();
+
+   F(0) = 0.0;
```
dylan-copeland, line 719: A few lines of code could be saved by just setting `F=0.0;` here and the RHS function.

miniapps/parelag/MultilevelHcurlSolver.cpp
```diff
@@ -0,0 +1,739 @@
...
+   const int nLevels = amge_levels <= 0 ? par_ref_levels + 1 : amge_levels;
+   MFEM_ASSERT(nLevels <= par_ref_levels + 1,
+               "Number of AMGe levels too high relative to parallel"
+               " refinements.");
+   vector<int> level_nElements(nLevels);
+   for (int l = 0; l < par_ref_levels; ++l)
+   {
+      if (!myid)
+      {
+         cout << "Parallelly refining mesh: " << l + 1
```
dylan-copeland, line 293: "Refining mesh in parallel"

miniapps/parelag/MultilevelHcurlSolver.cpp
```diff
@@ -0,0 +1,739 @@
...
+   if (!myid)
+      cout << "-- This is an example of using a geometric-like multilevel "
+           "hierarchy, constructed by ParElag,\n"
+           "-- to solve a finite element H(curl) form: "
+           "(alpha curl u, curl v) + (beta u, v).\n\n";
+
+   // Get basic parameters from command line.
+   const char *xml_file_c = "MultilevelHcurlSolver_cube_example_parameters.xml";
+   bool visualize = false;
+   double tolSVD = 1e-3;
```
dylan-copeland, line 52: It would be helpful to briefly document what this SVD tolerance is for.

miniapps/parelag/MultilevelHcurlSolver.cpp
```diff
@@ -0,0 +1,739 @@
...
+              << ".\n";
+         mesh_msg << "*    Serial refinements: " << ser_ref_levels << '\n'
+                  << "*      Coarse mesh size: " << mesh->GetNE() << "\n*\n";
+      }
+
+      pmesh = make_shared<ParMesh>(comm, *mesh);
+   }
+
+   // Mark essential boundary attributes.
+   MFEM_ASSERT(par_ess_attr.size() <= 1 ||
```
dylan-copeland, line 208: Since `MFEM_ASSERT` is checked only in debug mode, I would prefer to use `MFEM_VERIFY` in all instances in this file. When verifying input like this, the cost of the check is negligible, and checking in optimized mode can help the user find input errors.

miniapps/parelag/MultilevelHcurlSolver.cpp
```diff
@@ -0,0 +1,739 @@
...
+   Timer derham0_timer = TimeManager::AddTimer("DeRhamSequence Construction -- "
+                                               "Level 0");
+   if (!myid)
+   {
+      cout << "Building the fine-level de Rham sequence...\n";
+   }
+
+   vector<shared_ptr<DeRhamSequence>> sequence(topology.size());
+
+   const int jform = 1; // This is the H(curl) form.
```
dylan-copeland, line 420: Would it be better to use an `enum` rather than hard-coding the value 1 here?

miniapps/parelag/MultilevelHcurlSolver.cpp
```diff
@@ -0,0 +1,739 @@
...
+                                                 append(solver_name).
+                                                 append("\" -- Total"));
+      // Get the solver factory.
+      auto solver_factory = lib->GetSolverFactory(solver_name);
+      auto solver_state = solver_factory->GetDefaultState();
+      solver_state->SetDeRhamSequence(sequence[0]);
+      solver_state->SetBoundaryLabels(ess_attr);
+      solver_state->SetForms({jform});
+
+      // Build the silver.
```
dylan-copeland, line 619: solver

miniapps/parelag/MultilevelHcurlSolver.cpp
delyank: I'll add that both to the README and the .cpp.

miniapps/parelag/MultilevelHcurlSolver.cpp
delyank: Yes, I was on the fence which way to go with this. Both cases have advantages and disadvantages, but I think from the perspective of a demonstrative miniapp, it might be better to have it as a single entity. I'll combine them in one and test them.

miniapps/parelag/MultilevelHcurlSolver.cpp
delyank: I'll add more information in the respective `args.AddOption()`

delyank:
> If the purpose is not focus on solvers, then that's fine. Dont worry about my previous comments then.

@chakshinglee: Well, the demonstration is in the context of solvers, but ParELAG is supposed to be more broad. Otherwise, in principle, you are right that a more convenient interface in the case of solvers would be good, but it's a nontrivial task.

miniapps/parelag/MultilevelHcurlHdivSolver.cpp
```diff
@@ -0,0 +1,774 @@
...
+      Timer post_timer = TimeManager::AddTimer(std::string("Solver \"").
+                                               append(solver_name).
+                                               append("\" -- Post-solve"));
+      {
+         Vector tmp(A->Height());
+         A->Mult(X, tmp);
+         tmp *= -1.0;
+         tmp += B;
+
+         double local_norm = tmp.Norml2() * tmp.Norml2();
```
dylan-copeland, line 727: Not a big deal, but there are some places with `local_norm *= local_norm` and some places with two calls to `Norml2()`.

delyank:
I'm sorry for the delay. I finally managed to look at this and address the comments. The main change is that now the two miniapps are merged into one. I've built and tested again to verify that the expected behavior is preserved in this configuration. Note that a small change to ParELAG was introduced to address one of the comments. So if anyone wants to run it, they would need to pull ParELAG again.

@dylan-copeland, I think I've addressed your comments. Please, have a look at it.

dylan-copeland (approved):
Thanks @delyank!

miniapps/parelag/MultilevelHcurlHdivSolver.cpp
delyank: Ops. I forgot this one. It should be fixed now.

dylan-copeland:
This may be a question for @tzanio. I think miniapps are included in our automated regression testing, but will that be practical for this miniapp, since it will require building ParELAG?

delyank:
@dylan-copeland, good point. Such a regression test would almost resemble a ParELAG regression test and that's just a very unnecessary burden on MFEM's workflow.

tzanio:
> This may be a question for @tzanio. I think miniapps are included in our automated regression testing, but will that be practical for this miniapp, since it will require building ParELAG?

I am missing some context here, but we do have other miniapps that depend on third-party packages like `miniapps/gslib`. Those are built and tested only if MFEM has been configured to use the external package

delyank:
> @tzanio The context I had in mind is that ParELAG is tricky to build (at least for me), so having it build on the various machines used for autotesting may be tricky but should be feasible. Also, building and testing may take a significant amount of extra time. It may be fine, but these are some things to keep in mind, and part of my question, implicitly, was whether something extra needs to be done to the autotest system to set up these tests for the new miniapp (something beyond what is in the PR).

It's a bit tricky to build mostly because of dependencies, a lot of which also MFEM can depend on (like SuiteSparse). The most nonstandard part is the presence of a circular dependence: ParELAG depends on MFEM, and the miniapp depends on MFEM and ParELAG. That may not fit a standard workflow, where you have all the libraries built and you regularly rebuild only MFEM with its respective dependencies activated.

tzanio:
With the caveat that I have not look at the code -- in MFEM we assume that if `MFEM_USE_PARELAG` is set, then the user has already installed `ParELAG` and we can just use its headers and link with its library. We neve build third-party libraries as part of our build or testing process.

delyank:
OK. Do we need any additional work on this PR?

tzanio:
> OK. Do we need any additional work on this PR?

We need one more approval. @mlstowell, have you had a chance to take a look?

chakshinglee:
@delyank Thanks again for helping with the build issues that I encountered. I have tried to run the miniapp, with both libraries compiled with hypre 2.18.2. I got some seg fault error, here is what I got from gdb:
```
Thread 1 "MultilevelHcurl" received signal SIGSEGV, Segmentation fault.
0x000055555622a21e in hypre_CSRMatrixSubtract (A=0x5555568675f0, B=0x5555572fbe00) at /home/chak/Codes/parelag/src/hypreExtension/hypre_ParCSRMatrixCompare.c:56
56	            jcol = A_j[ia];
```

I didn't check carefully, but I think it is the bigint issue with new hypre (A_j is null pointer) 
I also tried with and older version of hypre (2.15.1),  the miniapp in this case runs fine.

You may want to double check this. If it's true, it will be good to specify (in e.g. README) which hypre is assumed, or how to configure hypre in the case of recent versions.

delyank:
Thank you, @chakshinglee ! The README has been edited to signify that.

> @delyank Thanks again for helping with the build issues that I encountered. I have tried to run the miniapp, with both libraries compiled with hypre 2.18.2. I got some seg fault error, here is what I got from gdb:
> 
> ```
> Thread 1 "MultilevelHcurl" received signal SIGSEGV, Segmentation fault.
> 0x000055555622a21e in hypre_CSRMatrixSubtract (A=0x5555568675f0, B=0x5555572fbe00) at /home/chak/Codes/parelag/src/hypreExtension/hypre_ParCSRMatrixCompare.c:56
> 56	            jcol = A_j[ia];
> ```
> 
> I didn't check carefully, but I think it is the bigint issue with new hypre (A_j is null pointer)
> I also tried with an older version of hypre (2.15.1), the miniapp in this case runs fine.
> 
> You may want to double check this. If it's true, it will be good to specify (in e.g. README) which hypre is assumed, or how to configure hypre in the case of recent versions.

chakshinglee (approved):

delyank:
Does this need more review?

tzanio:
I'll merge it for testing and take a closer look if everything passes.

I will let you know if have comments/suggestions when I do that. 

@mlstowell, do you also want to review?

tzanio:

mlstowell:
> I'll merge it for testing and take a closer look if everything passes.
> 
> I will let you know if have comments/suggestions when I do that.
> 
> @mlstowell, do you also want to review?

I have looked at the code but haven't had a chance to build ParELAG and test it. I'll go ahead and approve based on the other reviewers comments.

mlstowell (approved):

tzanio (approved):
Thanks @delyank. Please consider adding documentation on the website, see e.g. https://github.com/mfem/web/pull/103

PR #2430, opened 2021-07-27
libCEED 0.9.0

jeremylt:
libCEED has released v0.9.0 - this is a minor change, so there shouldn't be any issues here. This has the update to silence those pedantic warnings from Clang.


tzanio:
I am on the fence about this -- the version in `INSTALL` should be the minimal required version for the integration to work (the users are of course free to use newer libCEED if it brings new features to them, but they don't need to upgrade if they don't want to). 

How important is to change this from your pespective?

jeremylt:
I'm similarly on the fence. Not a lot changed in libCEED, but v0.9.0 is the minimum version with the JOSS citation.

tzanio:
Lets postpone this then.

PR #2433, opened 2021-07-28
[WIP] Data oscillation refinement / mesh pre-processing 

brendankeith:


tzanio:

examples/osc.cpp
```diff
@@ -0,0 +1,199 @@
...
+                  "Finite element order (polynomial degree).");
+   args.AddOption(&nc_limit, "-l", "--nc-limit",
+                  "Maximum level of hanging nodes.");
+   args.AddOption(&double_max_elems, "-me", "--max-elems",
+                  "Stop after reaching this many elements.");
+   args.AddOption(&visualization, "-vis", "--visualization", "-no-vis",
+                  "--no-visualization",
+                  "Enable or disable GLVis visualization.");
+   args.AddOption(&osc_threshold, "-e", "--error",
+                  "relative data oscillation threshold");
```
mlstowell, line 109: For conformity this description should be capitalized and end with a period.

mesh/mesh_operators.hpp
```diff
@@ -307,6 +307,105 @@ class ThresholdDerefiner : public MeshOperator
...
+       marked for refinement; REFINED + CONTINUE otherwise. */
+   virtual int PreprocessMesh(Mesh &mesh, int max_it);
+
+   int PreprocessMesh(Mesh &mesh)
+   {
+      int max_it = 10;
+      return PreprocessMesh(mesh, max_it);
+   }
+
+   /// Set the refinement threshold. The default value is 1.0e-3.
```
mlstowell, line 363: It looks like the default value set above is 1e-2.

mesh/mesh_operators.hpp
```diff
@@ -307,6 +307,105 @@ class ThresholdDerefiner : public MeshOperator
...
+   {
+      MFEM_ASSERT(nc_limit_ >= 0, "Invalid NC limit");
+      nc_limit = nc_limit_;
+   }
+
+   // Set a custom integration rule
+   void SetIntRule(const IntegrationRule *irs_[]) { irs = irs_; }
+
+   // Return the value of the global relative data oscillation
+   double GetOsc() { return global_osc; }
```
mlstowell, line 396: This should be a "const" member function.

mesh/mesh_operators.hpp
```diff
@@ -307,6 +307,105 @@ class ThresholdDerefiner : public MeshOperator
...
+
+   // Return the local relative data oscillation errors
+   Vector GetLocalOscs()
```
mlstowell, line 399: Unless the Vector class now has some internal magic to handle data copies it might be better to declare this as:
`const Vector & GetLocalOscs() const`

mesh/mesh_operators.hpp
```diff
@@ -307,6 +307,105 @@ class ThresholdDerefiner : public MeshOperator
...
+   int nc_limit = 1;
+   int nonconforming = -1;
+   int order;
+   long max_elements = std::numeric_limits<long>::max();
+   double threshold = 1.0e-2;
+   double global_osc = 0.0;
+   Array<int> mesh_refinements;
+   Vector element_oscs;
+   Coefficient *coeff = NULL;
+   GridFunction *gf;
```
mlstowell, line 338: This should also be initialized.

mesh/mesh_operators.hpp
```diff
@@ -307,6 +307,105 @@ class ThresholdDerefiner : public MeshOperator
...
+   /// Set the refinement threshold. The default value is 1.0e-3.
+   void SetThreshold(double threshold_) { threshold = threshold_; }
+
+   /** @brief Set the maximum number of elements stopping criterion: stop when
+       the input mesh has num_elements >= max_elem. The default value is
+       LONG_MAX. */
+   void SetMaxElements(long max_elements_) { max_elements = max_elements_; }
+
+   /// Set the function f
+   void SetCoefficient(Coefficient &coeff_)
```
mlstowell, line 372: The coefficient seems to be required. At at minimum this requirement should be mentioned in the comments. Alternatively a `Coefficient &` argument could be added to the constructor. Yet another alternative would be to remove the requirement by using a constant value of 1 when the user doesn't provide a coefficient. I'd like to see more than just a change in the comment but I'm not certain what functional change would be most appropriate.

mesh/mesh_operators.cpp
```diff
@@ -157,6 +157,142 @@ int ThresholdDerefiner::ApplyImpl(Mesh &mesh)
...
+
+   bool par = false;
+
+#ifdef MFEM_USE_MPI
+   ParMesh* pmesh = dynamic_cast<ParMesh*>(&mesh);
+   if (pmesh && pmesh->Nonconforming())
+   {
+      par = true;
+      l2fes = new ParFiniteElementSpace(pmesh, &l2fec);
+      gf = new ParGridFunction(static_cast<ParFiniteElementSpace*>(l2fes));
```
mlstowell, line 184: Since the `[Par]GridFunction` will be deleted before exiting this function perhaps it should only be declared locally rather than being member data.

examples/oscp.cpp
```diff
@@ -0,0 +1,245 @@
...
+      return 1.0;
+   }
+}
+
+// Piecewise-constant function which is never mesh-conforming
+double jump_function(const Vector &p)
+{
+   if (p.Normlp(2.0) > 0.4 && p.Normlp(2.0) < 0.6) { return 1; }
+   if (p.Normlp(2.0) < 0.4 || p.Normlp(2.0) > 0.6) { return 5; }
+   return 0;
```
jakubcerveny, line 65: Is zero really the intended third value of the function (at radius == 0.4 and 0.6)? Shouldn't the second line be just "return 5"?

jakubcerveny (approved):
This looks fine!

jakubcerveny:
One question, when I run `osc` with just one refinement for the jump coefficient, I get this:
osc1
The full three refinements look like this:
osc2
I'm wondering why does the double circle look different in the second case, shouldn't this be just a superposition of three independent refinement patterns?

brendankeith:
> One question, when I run `osc` with just one refinement for the jump coefficient, I get this:
> osc1
> The full three refinements look like this:
> osc2
> I'm wondering why does the double circle look different in the second case, shouldn't this be just a superposition of three independent refinement patterns?

I think that this is mostly an artifact of using a global stopping condition. In the second figure, the mesh started out finer in the vicinity of the straight line jump function. It many scenarios, one would just have a single grid function and the results would be more symmetric, like the first figure.

I could add an optional minimum local error threshold. Thoughts?

examples/oscp.cpp
```diff
@@ -62,7 +62,7 @@ double jump_function(const Vector &p)
 {
    if (p.Normlp(2.0) > 0.4 && p.Normlp(2.0) < 0.6) { return 1; }
    if (p.Normlp(2.0) < 0.4 || p.Normlp(2.0) > 0.6) { return 5; }
```
jakubcerveny, line 64: This line (64) is now superfluous :-)

jakubcerveny:
> I think that this is mostly an artifact of using a global stopping condition. In the second figure, the mesh started out finer in the vicinity of the straight line jump function. It many scenarios, one would just have a single grid function and the results would be more symmetric, like the first figure.
> 
> I could add an optional minimum local error threshold. Thoughts?

That's OK, I was trying to see if there is an explanation (just in case this was a bug), thanks.

examples/osc.cpp
```diff
@@ -0,0 +1,197 @@
...
+   const IntegrationRule *irs[Geometry::NumGeom];
+   int order_quad = 2*order + enriched_order;
+   for (int i = 0; i < Geometry::NumGeom; ++i)
+   {
+      irs[i] = &(IntRules.Get(i, order_quad));
+   }
+
+   // 6. Apply custom refiner settings.
+   coeffrefiner.SetIntRule(irs);
+   coeffrefiner.SetMaxElements( (long) max_elems);
```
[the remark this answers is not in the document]
brendankeith: I guess not...

tzanio:
@psocratis, @rw-anderson and @mlstowell -- can you please complete your reviews? We need one more approval before we can merge in `next`.

examples/osc.cpp
```diff
@@ -0,0 +1,197 @@
...
+//               which lowers the data oscillation [1] to a user-defined
+//               relative threshold. There is no PDE being solved.
+//
+//               MFEM's capability to work with both conforming and
+//               nonconforming meshes is demonstrated in example 6. In some
+//               problems, the material data or loading data is not sufficiently
+//               resolved on the initial mesh. This missing fine scale data
+//               reduces the accuracy of the solution as well as the accuracy
+//               of some local error estimators. By preprocessing the mesh
+//               before the solving the PDE, many issues can be avoided.
```
psocratis, line 27: "before solving the PDE"

examples/oscp.cpp
```diff
@@ -0,0 +1,243 @@
...
+//               which lowers the data oscillation [1] to a user-defined
+//               relative threshold. There is no PDE being solved.
+//
+//               MFEM's capability to work with both conforming and
+//               nonconforming meshes is demonstrated in example 6. In some
+//               problems, the material data or loading data is not sufficiently
+//               resolved on the initial mesh. This missing fine scale data
+//               reduces the accuracy of the solution as well as the accuracy
+//               of some local error estimators. By preprocessing the mesh
+//               before the solving the PDE, many issues can be avoided.
```
psocratis, line 27: "before solving the PDE"

mesh/mesh_operators.cpp
```diff
@@ -157,6 +157,142 @@ int ThresholdDerefiner::ApplyImpl(Mesh &mesh)
...
+      for (int i=0; i < Geometry::NumGeom; ++i)
+      {
+         ir_default[i] = &(IntRules.Get(i, order_quad));
+      }
+      irs = ir_default;
+   }
+
+   for (int i = 0; i < max_it; i++)
+   {
+
```
psocratis, line 207: empty line

psocratis (approved):
Very nice PR. There are a couple of sample runs that take some time to complete. I am not sure if we want to keep them there. For example `mpirun -np 4 ./oscp -m ../data/ball-nurbs.mesh -o 2 -eo 3 -e 1e-2` takes about 1.5 minutes on my mac with a debug built. I am also wondering if we want to modify the shaper mini-app to use the newly added `CoeffiecientRefiner` class. What do @tzanio or @v-dobrev  think?

tzanio:
@mlstowell -- can you take another look please?

mesh/mesh_operators.hpp
mlstowell: Is there a reason for not making this "const"?

mesh/mesh_operators.cpp
```diff
@@ -157,6 +157,141 @@ int ThresholdDerefiner::ApplyImpl(Mesh &mesh)
...
+         MPI_Allreduce(MPI_IN_PLACE, &global_osc, 1, MPI_DOUBLE, MPI_SUM, comm);
+         MPI_Comm_rank(comm, &rank);
+      }
+#endif
+      global_osc = sqrt(global_osc)/(norm_of_coeff + 1e-10);
+
+      // Exit if the global threshold or maximum number of elements is reached.
+      if (global_osc < threshold || globalNE > max_elements)
+      {
+         if (global_osc > threshold && globalNE > max_elements && rank == 0)
```
mlstowell, line 264: One small point. I would suggest adding a flag to disable this warning similar to the `print_level` used in our iterative solvers. Some people have a strong aversion to warnings and they may want to be able to disable them in some situations. For example if MFEM's `example` codes produce warnings some test runs might lead to questions e.g. "was this warning expected or should we be concerned?" .

mesh/mesh_operators.hpp
brendankeith: My bad

mlstowell (approved):

tzanio:

tzanio:
My apologies for the delayed review. I have several comments:

- Overall this looks really good. Nice work @brendankeith!
- Rename `osc.cpp` to `ex30.cpp` and `oscp.cpp` to `ex30p.cpp`
- Update `.gitignore` and `examples/makefile` accordingly
- Mention in `CHANGELOG`
- Check that all items in the [PR checklist](https://github.com/mfem/mfem/blob/master/CONTRIBUTING.md#pull-request-checklist) have been addressed

PR #2435, opened 2021-07-29
HyperParVector Erase{X,Y}

tomstitt:
- add `EraseX` and `EraseY` methods to `HyperParVector`
- check for `nullptr` and ctor `X` and `Y` separately

This is needed for an app that uses temporary memory for the `Y` argument in `HyperParVector::Mult`. `Mult` makes an alias to `Y` (and `X`) and when we release the temporary memory that `Y` is aliased to we get a runtime error; we need a way to explicitly tell the `HyperParVector` to erase its reference.


tomstitt:
No longer needed

PR #2454, opened 2021-08-05
Fix total error definition in KellyErrorEstimator

brendankeith:
Small bug fix. Total error should be defined as the sum of the squares of the local errors.


brendankeith:
Accidentally began this branch inside another branch I was working on. Most of these commits  are not necessary.

PR #2458, opened 2021-08-06
Support hypre's Gershgorin estimate for eigenvalues

AerysL:
A recent merge in hypre changes the prototype of `hypre_ParCSRMaxEigEstimateHost` as it now includes a min eigenvalue estimate (and includes better support for non spd matrices). This PR makes the relevant changes on mfem's side to support it.

This branch includes `hypre_owners` branch, to allow it to compile with hypre's master. See #2452 for more info.


jandrej:
Depends on  #2452

v-dobrev:
The starting point for new branches should be the `master` branch. Does this PR really require the commits in #2452?

AerysL:
@v-dobrev Because the changes to hypre's internal API were made in sequence, all the associated changes must be made on mfem's side to compile with hypre master. I have provided a pull request for each of the merges/commits on the hypre side, so one can diff them and see the changes while still being able to compile the branches (for example, if this branch did not include the ownership code, it would not compile with hypre master). It may be most clean for a single merge with all the associated changes to be done once hypre releases it's next version, but I wanted to provide some temporary solutions (or at least reference code for them).

jwang125:
MFEM currently still returns HYPRE version: 22200 to me for hypre master. This seems to cause `HYPRE_USE_GERSHGORIN` to be 0 which causes `hypre_ParCSRMaxEigEstimate` to choose the branch with three input, and thus a compiling error. Setting `HYPRE_USE_GERSHGORIN=1` manually solves the issue. So perhaps the version number could be `>=22200`? Or perhaps something I did was wrong.

AerysL:
Correct, there has not yet been a hypre release since the internal API
changed, and so a manual override must be used. There was a good idea
discussed about using an ifdef to remove the requirement of the manual
override, but I have not had a chance to implement it yet. I believe all
three variables will require a manual override until hypre has a new
release. You should also be able to check out a tagged version of hypre
rather than master if you so desire.

tzanio:
> MFEM currently still returns HYPRE version: 22200 to me for hypre master. 

In MFEM we have a special version number during development between official releases, e.g. the version of `mfem:master` is currently `4.3.1`. See https://github.com/mfem/mfem/blob/master/CONTRIBUTING.md#releases.

Hypre doesn't do that, so the version returned by `hypre:master` is the same as the last release.

v-dobrev:
> > MFEM currently still returns HYPRE version: 22200 to me for hypre master.
> 
> In MFEM we have a special version number during development between official releases, e.g. the version of `mfem:master` is currently `4.3.1`. See https://github.com/mfem/mfem/blob/master/CONTRIBUTING.md#releases.
> 
> Hypre doesn't do that, so the version returned by `hypre:master` is the same as the last release.

@rfalgout, is it possible to consider adding to hypre some way of distinguishing between releases and other non-release commits in hypre? Something that will let us do something like this:
```c
#if (HYPRE_RELEASE_NUMBER > 22200) || (HYPRE_RELEASE_NUMBER == 22200 && HYPRE_DEVELOPMENT == 1)
// support use of hypre's master branch after the 2.22.0 release or later hypre versions, > 2.22.0
#else
// support hypre version <= 2.22.0
#endif
```
This will help us support the latest hypre master branch along with previous release versions.

There are other options to support something like this. In MFEM, after a release, we simply increase the last version number to be odd and that implies this is a "development" version that will not have a corresponding release. For example, after the v4.3 release we set the version in the repo to v4.3.1 and that will stay the same during development before the next release.

rfalgout:
Hi @v-dobrev .  It would not be hard to implement something like this, but I am a little bit hesitant to go down this path.  If the master branch of hypre is being used, then I think it's safe to say that Git is also being used to acquire the source.  With Git you can get all of the information you are asking for (and more) with commands such as the following:

    git describe
    git log v2.22.0..HEAD --oneline | wc

The first command currently produces the following output for hypre: `v2.22.0-25-gae8bbf619`.  The prefix of this string is the release tag and `25` is the number of commits since that release.  With some simple scripting, this number could be used to set a variable like `MFEM_HYPRE_DEVNUM`, which serves the same purpose as your `HYPRE_DEVELOPMENT` variable, but with the actual number of commits since the last release.

The main question I am wondering is "who acquires the release?"  If MFEM or spack is grabbing it, then this scripting is easy to implement.  If the user grabs it, then they would need a mechanism for giving it to the MFEM build system, e.g., through some `configure` or `CMake` option.

Related to this, I think it might be useful to do something in the xSDK project to address the question of how to use development branches of code interchangeably, and to me it makes sense to utilize the tools available in the various resource control systems like Git to do that.

What are your thoughts?

Also, note that we could easily do a `v2.22.1` release to help with this specific issue.

tzanio:
Closed in favor of https://github.com/mfem/mfem/pull/2476

PR #867, opened 2019-04-17
[WIP] DG Advection diffusion example [adv-diff-ex-dev]

mlstowell:
This PR proposes a new example of advection-diffusion using discontinuous Galerkin.  It is a simple merging of `ex9p` and `ex14p` to get `ex23p` (how can we pass up that opportunity?)

The new `ex23p` runs without error as it is written but I want to be sure to get the boundary conditions correct.  The true goal of this example, for me at least, is to better understand the application of boundary conditions when using DG.  For example if we run:
`mpirun -np 4 ex23p -m ../data/disc-nurbs.mesh -p 3 -rp 1 -dt 0.005 -tf 9 -d 0.05`
The solution remains fixed at the boundary so I believe an inhomogeneous Dirichlet BC is being applied correctly.  I suspect that the `BoundaryFlowIntegrator` is not needed in this case but I don't really know (it's velocity is zero so I suspect it is having no effect).  I'd like to also support inhomogeneous Neumann BCs but I don't know how to accomplish that.  Would that involve the `BoundarFlowIntegrator` or perhaps something else entirely?

Another question involves the parameters used with the DG integrators.  The `DGDiffusionIntegrator` uses `sigma` and `kappa` to define the DG formulation being used.  The `DGTraceIntegrator` uses `alpha` and `beta` but I'm not sure what these are.  Should these two pairs of parameters be related somehow? Are all of these face integrators even necessary in this situation or do we require new
integrators that are specialized to this modified PDE?

I appreciate any guidance that might be offered.

PR #2474, opened 2021-08-18
Added framework for IMEX BDF and RK time integration

bensworth:
I added two files, imex.hpp and imex.cpp that introduce a framework for IMEX time integration. This includes many of the standard IMEX RK methods, IMEX BDF methods, and some new polynomial BDF currently methods being developed that allow for tuneable stability or accuracy. I have a DG adv-diff script that I am cleaning up and can include as an example when the imex files are linked to MFEM correctly (see point 2 below).

There are two main questions/issues:
1. On a high level, the question is how to handle the `ImplicitMult`, `ExplicitMult`, and `ImplicitSolve` functions. As far as I understand, MFEM uses the EXPLICIT/IMPLICIT type definition *pre* time discretization, `F*du/dt = G(u)`. I've never encountered applications where F is not M, so I don't really "get" this functionality. From an IMEX perspective, `ImplicitMult` and `ExplicitMult` make a lot of sense to apply the two partitions of `M*du/dt = N_E(u,t) + N_I(u,t)`, where I denote N_E the explicit splitting and N_I the implicit splitting. Also, for BDF schemes it is best to not include the M^{-1} in the action of these operators, so I added separate functions `MassMult` and `MassInv`. Last, for RK and the way MFEM poses time integration, `ImplicitSolve` solves `k = f(x+dt*k)` for stage k. For BDF methods, you solve `M*x - dtf(x, t) = b` for solution x. I called this `ImplicitSolve2`. To keep this all separate from existing MFEM infrastructure, I defined a new class `IMEXTimeDependentOperator`; the inconsistency with function meaning in `TimeDependentOperator` is not ideal, but the implementation is fairly clean I think. Thoughts and/or ideas on this structure are the most important question for the MFEM team.

2. The code compiles and runs fine if I keep the imex files separate from MFEM (with slight modifications, like including mfem.hpp instead of the specific header files), link to MFEM, and compile my adv-diff script. Here I tried to put the imex files into the linalg directory of MFEM, but don't have things compiling and linking correctly. MFEM compiles fine, but when I try to compile the adv-diff script that uses imex.hpp, I get undefined symbols errors (below). Probably this is an easy fix and I just didn't add the files to MFEM correctly.
```
Undefined symbols for architecture x86_64:
  "__ZN4mfem10IMEXRKData7SetDataEv", referenced from:
      __ZN4mfem10IMEXRKDataC1ENS0_4TypeE in nl-imex-dg-adv-diff.o
  "__ZN4mfem6IMEXRK4InitERNS_25IMEXTimeDependentOperatorE", referenced from:
      __Z12run_adv_diffiPPc in nl-imex-dg-adv-diff.o
  "__ZN4mfem6IMEXRKD1Ev", referenced from:
      __Z12run_adv_diffiPPc in nl-imex-dg-adv-diff.o
  "__ZN4mfem7BDFData7SetDataEv", referenced from:
      __ZN4mfem7BDFData5SetIDENS0_4TypeEd in nl-imex-dg-adv-diff.o
  "__ZN4mfem7IMEXBDF4InitERNS_25IMEXTimeDependentOperatorE", referenced from:
      __Z12run_adv_diffiPPc in nl-imex-dg-adv-diff.o
  "__ZN4mfem7IMEXBDFD1Ev", referenced from:
      __Z12run_adv_diffiPPc in nl-imex-dg-adv-diff.o
  "__ZTVN4mfem6IMEXRKE", referenced from:
      __ZN4mfem6IMEXRKC1ENS_10IMEXRKDataE in nl-imex-dg-adv-diff.o
  NOTE: a missing vtable usually means the first non-inline virtual member function has no definition.
  "__ZTVN4mfem7IMEXBDFE", referenced from:
      __ZN4mfem7IMEXBDFC1ENS_7BDFData4TypeEd in nl-imex-dg-adv-diff.o
  NOTE: a missing vtable usually means the first non-inline virtual member function has no definition.
ld: symbol(s) not found for architecture x86_64
```

### DIFF
--- a/linalg/imex.cpp
+++ b/linalg/imex.cpp
@@ -1,0 +1,1164 @@
+#include "imex.hpp"
+
+// Lagrange interpolating polynomials
+// q=2
+// (1-1/2(2+a))y1 + 1/2(2+a)y2
+// q=3
+// (-1-a+1/2 (1+a) (2+a)) y1 + (2+a+(-1-a) (2+a)) y2 + 1/2 (1+a) (2+a) y3
+// q=4
+// (1-(3 (2+a))/2+9/8 (4/3+a) (2+a)-9/16 (2/3+a) (4/3+a) (2+a)) y1 + 
+//     ((3 (2+a))/2-9/4 (4/3+a) (2+a)+27/16 (2/3+a) (4/3+a) (2+a)) y2 +
+//     (9/8 (4/3+a) (2+a)-27/16 (2/3+a) (4/3+a) (2+a)) y3+9/16 (2/3+a) (4/3+a) (2+a) y4
+void InterpolateBDF(int q, double alpha, std::vector< Vector*> sols, Vector &x)
+{
+    double &a = alpha;
+    if (q==2)
+    {
+        std::cout << "Not implemented for linear\n";
+    }
+    if (q==2)
+    {
+        Vector &y1 = *sols[0];
+        Vector &y2 = *sols[1];
+        double c1 = 1.-0.5*(2+a);
+        double c2 = 0.5*(2+a);
+        for (int i=0; i<x.Size(); i++)
+        {
+            x(i) = c1*y1(i) + c2*y2(i);
+        }
+    }
+    else if (q==3)
+    {
+        Vector &y1 = *sols[0];
+        Vector &y2 = *sols[1];
+        Vector &y3 = *sols[2];
+        double c1 = 0.5*(1+a)*(2+a) - (1+a);
+        double c2 = (2+a)-(1+a)*(2+a);
+        double c3 = 0.5*(1+a)*(2+a);
+        for (int i=0; i<x.Size(); i++)
+        {
+            x(i) = c1*y1(i) + c2*y2(i) + c3*y3(i);
+        }
+    }
+    else if (q==4)
+    {
+        Vector &y1 = *sols[0];
+        Vector &y2 = *sols[1];
+        Vector &y3 = *sols[2];
+        Vector &y4 = *sols[3];
+        double d43a = a + 4.0/3.0;
+        double d23a = a + 2.0/3.0;
+        double ap2 = a + 2;
+        double c1 = (1.0 - 1.5*ap2 + 1.125*d43a*ap2 - 0.5625*d23a*d43a*ap2);
+        double c2 = (1.5*ap2 - 2.25*d43a*ap2 + 1.6875*d23a*d43a*ap2);
+        double c3 = (1.125*d43a*ap2 - 1.6875*d23a*d43a*ap2);
+        double c4 = 0.5625*d23a*d43a*ap2;
+        for (int i=0; i<x.Size(); i++)
+        {
+            x(i) = c1*y1(i) + c2*y2(i) + c3*y3(i) + c4*y4(i);
+        }
+    }
+    else
+    {
+        mfem_warning("Only implemented up to 4th order.\n");
+    }
+}
+
+IMEXRK::~IMEXRK()
+{
+    for (int i = 0; i < tableaux.s; i++)
+    {
+        if (imp_stages[i])
+        {
+            delete imp_stages[i];
+        }
+        if (exp_stages[i])
+        {
+            delete exp_stages[i];
+        }
+    }
+}
+
+void IMEXRK::Init(IMEXTimeDependentOperator &_imex)
+{
+    ODESolver::Init(_imex);
+    imex = &_imex;
+    exp_stages.resize(tableaux.s);
+    imp_stages.resize(tableaux.s);
+
+    // Only allocate first implicit stage vector if ESDIRK
+    if (tableaux.esdirk)
+    {
+        imp_stages[0] = new Vector(imex->Width());
+        *imp_stages[0] = 0.0;
+    }
+    else
+    {
+        imp_stages[0] = NULL;
+    }
+    exp_stages[0] = new Vector(imex->Width());
+    *exp_stages[0] = 0.0;
+
+    // Allocate interior implicit/explicit stages
+    for (int i=1; i<(tableaux.s-1); i++)
+    {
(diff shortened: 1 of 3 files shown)


pazner:
Potentially related:
* #867

bensworth:
> Potentially related:
> 
>     * [[WIP] DG Advection diffusion example [adv-diff-ex-dev] #867](https://github.com/mfem/mfem/pull/867)

Indeed, I think I used that branch as a model for the adv-diff discretization I am using; if the IMEX schemes are merged, may be a good combined PR with the adv-diff example

v-dobrev:
This PR is probably not exactly ready for review (for merging in `master`), however, in order to get some feedback to @bensworth, we should add the label `ready-for-review` and start the review process with the understanding that it may require a second round of reviews. @tzanio, does this sound reasonable to you too?

tzanio:
> @tzanio, does this sound reasonable to you too?

Yes it does.

v-dobrev:
Hi @bensworth, can you please try to address the compilation errors? For example the errors from the `gitignore` test start here: https://github.com/mfem/mfem/pull/2474/checks?check_run_id=3362944018#step:9:89

cjvogl:
It would be preferable to be able to define one `TimeDependentOperator` class for an additively split problem M*du/dt = N_E(u,t) + N_I(u,t) and use either the ARKODE methods or these IMEX solvers.  I believe at the moment, one would need to define a separate `IMEXTimeDependentOperator` to use these IMEX solvers.... maybe I'm not reading this correctly, but If I am, is there a way to reconcile the ARKODE/SUNDIALS functions in `TimeDependentOperator` with the ones defined in `IMEXTimeDependentOperator`?

bensworth:
Hi Veselin, I was actually hoping for some input from somebody that works
on MFEM in that regard. The code compiles and runs fine if I link it
outside of MFEM. But when I try to include it as a new hpp/cpp file in
linalg/ or add it to ode.hpp/cpp, I get a huge number of errors. It seems
related to the new(er) memory structure in MFEM, which I am not familiar
with. Is it possible it has to do with my use of std::vector? I looked at
this for a little bit, but didn't really understand why it would work
separately but not with the library.

Chris, probably, at least to some extent. I believe the BDF methods need
different functionality than RK, so that makes compatibility easy in some
sense. Can you comment the main functions used by Sundials/how they work?
Another option is that I can add a modified function for the BDF class of
methods called an iterator that constructs high-order initial conditions.
It is similar to an IMEX-RK method for each step, but notationally and
functionality-wise will be consistent with the BDF methods and
implementation. Then we could leave IMEX-RK to ARKODE and just add the BDF
methods to MFEM. I actually like this approach "better" in some sense, as
everything is self-consistent, but have not gotten around to the iterator
implementation and testing yet.

cjvogl:
The SUNDIALS methods in the Time-Dependent operator are for the nonlinear solve of a Runge-Kutta stage,
Mz_i - alpha_ii * f(z_i) = ...
where f can either be the RHS of the ODE (e.g., a DIRK method) or the implicit RHS (e.g., an IMEX-RK method).

The solve is generalized to solving the linear system Ax=b, where A is I - g*J or M-g*J, depending on how you want to handle the mass matrix, and J is the Jacobian of f (or some approximation thereof).

The methods are, at a surface level,
- SUNImplicitSolve - solve Ax=b for a given b to a specified tolerance
- SUNImplicitSetup - perform any set up needed to solve Ax=b in the function above, given where A should be evaluated, the value of g, and the current value of f
- SUNMassMult - compute Mx for a given x
- SUNMassSolve - solve for x in Mx=b for a given b
- SUNMassSetup - perform any set up needed to solve Mx=b in the function above

cjvogl:
As I'm currently watching @jandrej's seminar, I wonder if he would have any input on BDF implementations... given my understanding that he implemented BDF methods in the Navier miniapp?

pazner:
> As I'm currently watching @jandrej's seminar, I wonder if he would have any input on BDF implementations... given my understanding that he implemented BDF methods in the Navier miniapp?

The time integration used in Navier is implemented here: https://github.com/mfem/mfem/blob/9a73c2d94c5ecff6dd64c9f7f575e55da0b250cf/miniapps/navier/navier_solver.cpp#L344

It's not a generic BDF method implemented as an MFEM `ODESolver` because the time integration for incompressible N-S is somewhat "special" (it's more than just a vanilla BDF method).

cjvogl:
Thanks @pazner!

bensworth:
> The SUNDIALS methods in the Time-Dependent operator are for the nonlinear solve of a Runge-Kutta stage,
> Mz_i - alpha_ii * f(z_i) = ...
> where f can either be the RHS of the ODE (e.g., a DIRK method) or the implicit RHS (e.g., an IMEX-RK method).
> 
> The nonlinear solver iteration is generalized to solving the linear system Ax=b, where A is I - g_J or M-g_J, depending on how you want to handle the mass matrix, and J is the Jacobian of f (or some approximation thereof).
> 
> The methods are, at a surface level,
> 
>     * SUNImplicitSolve - solve Ax=b for a given b to a specified tolerance
> 
>     * SUNImplicitSetup - perform any set up needed to solve Ax=b in the function above, given where A should be evaluated, the value of g, and the current value of f
> 
>     * SUNMassMult - compute Mx for a given x
> 
>     * SUNMassSolve - solve for x in Mx=b for a given b
> 
>     * SUNMassSetup - perform any set up needed to solve Mx=b in the function above

@cjvogl I'm a bit confused on the implicit solve. RK stages in the more common context of `u_{n+1} = u_n + \sum b_ik_i` require solving systems `k = f(x+dt*k, t+ci)` for stage k, where f() is the implicit part of the operator. What you wrote is what we need for BDF type methods, `M*x - dtf(x, t) = b` for solution x. I think that RK methods can be posed in a similar context, but that's solving for the solution as RK abscissa rather than the stages.

On a higher level, so for ARKCode you just let ImplicitSolve denote a solve of the implicit part of the operator?

PR #2489, opened 2021-08-25
Allow reading Gmsh meshes where all elements have attribute zero [gmsh-zero-attributes]

pazner:
We [commonly encounter](https://github.com/mfem/mfem/issues?q=is%3Aissue+gmsh+attribute+physical) the issue where users try to load Gmsh meshes where all the elements have physical tag zero (since if no physical groups are explicitly created in the geometry file, this is what Gmsh will do by default). However, MFEM requires strictly positive element attributes.

This is a small change to the Gmsh reader that will check to see if **all** elements of the Gmsh mesh have physical tag zero. If so, then it will load the mesh and give the elements attribute 1. If there are some elements with attribute zero, and others with positive attributes, it will throw an error, since in this case it is not clear what attribute we would want to assign to the zero elements.

I suspect that this will resolve the vast majority of user issues loading Gmsh meshes with non-positive physical tags.

| [#2489](https://github.com/mfem/mfem/pull/2489) | @pazner | @tzanio | @tzanio + @mlstowell + @bslazarov | 08/25/21 | 08/30/21 | 09/02/21 | |

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,9 @@ Version 4.3.1 (development)
   functions on wedges and pyramids which are not amenable to reordering. The
   ReorientTetMesh method of the Mesh and ParMesh classes has been deprecated.
 
+- Gmsh meshes where all elements have zero physical tag (the default Gmsh
+  output format if no physical groups are defined) are now successfully loaded,
+  and elements are reassigned attribute number 1.
 
 Version 4.3, released on July 29, 2021
 ======================================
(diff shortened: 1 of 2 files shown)


mlstowell (approved):

tzanio:
Should this be mentioned in `CHANGELOG`?

tzanio:

pazner:
> Merged in `next` for testing...

@tzanio it looks like this PR didn't make it into `next`, does the merge need to be pushed?

tzanio:
My bad sorry, should be merged now.

PR #2496, opened 2021-08-29
WIP: 1d ncmesh dev

rw-anderson:
The idea behind this is to allow "ncmesh" refinement of 1D meshes, even though 1D is a degenerate case that has no hanging nodes. The advantage of this is that derefinement is only supported in non-conforming refinement, and we have a desire to do 1D refinement/derefinement in an ML project for the cheap evaluation. 

Also, it seems so far like the modifications/additions needed to support this are small.

I'd like to add a 1D run of example 15 to test this, but haven't added that yet.

@jakubcerveny I thought you might have some thoughts about viability of this path and whether these changes make sense so far.

### DIFF
--- a/mesh/mesh.cpp
+++ b/mesh/mesh.cpp
@@ -3495,8 +3495,6 @@ void Mesh::SetMeshGen()
             mesh_geoms |= (1 << Geometry::TETRAHEDRON);
          case Element::TRIANGLE:
             mesh_geoms |= (1 << Geometry::TRIANGLE);
-         case Element::SEGMENT:
-            mesh_geoms |= (1 << Geometry::SEGMENT);
          case Element::POINT:
             mesh_geoms |= (1 << Geometry::POINT);
             meshgen |= 1;
@@ -3508,6 +3506,8 @@ void Mesh::SetMeshGen()
             mesh_geoms |= (1 << Geometry::SQUARE);
             mesh_geoms |= (1 << Geometry::SEGMENT);
             mesh_geoms |= (1 << Geometry::POINT);
+         case Element::SEGMENT:
+            mesh_geoms |= (1 << Geometry::SEGMENT);
             meshgen |= 2;
             break;
 
--- a/mesh/ncmesh.cpp
+++ b/mesh/ncmesh.cpp
@@ -4253,6 +4253,19 @@ void NCMesh::GetPointMatrix(Geometry::Type geom, const char* ref_path,
             pm = PointMatrix(mid12, mid20, mid01);
          }
       }
+      else if (geom == Geometry::SEGMENT)
+      {
+         Point mid01(pm(0), pm(1));
+
+         if (child == 0)
+         {
+            pm = PointMatrix(pm(0), mid01);
+         }
+         else if (child == 1)
+         {
+            pm = PointMatrix(mid01, pm(1));
+         }
+      }
    }
 
    // write the points to the matrix


jakubcerveny:
Hi Bob, I thought 1D NCMesh has already been done some time ago (#2038), is this different from that PR?

rw-anderson:
Currently, when you use GeneralRefinement with a 1D mesh, even if you have called EnsureNCMesh(), it uses conforming refinement. This would be fine and make sense, but it does not allow derefinement.

rw-anderson:
It appeared that in this conditional the NCMesh is suppressed from being created:

https://github.com/mfem/mfem/blob/b1e84b81275f2f1bb23a2bbd1c62a72f265fdd8b/mesh/mesh.cpp#L8819-L8829

because Element::Segment is not given the flags that are tested for there. That was the change I made in the first commit.

The second thing I saw was that child elements were not being split, there were two child elements that had the full extent of the parent element upon refinement. This looked to be the result of there being no code to do the PointMatrix for the segment elements. That was the second commit.

If this worked before, I am not sure how.

rw-anderson:
Maybe there is something new going on here in that I am using periodic-segment.mesh, which uses an FESpace for the nodes. Is this why the PointMatrix clause for Segment may not have been needed before?

mesh/ncmesh.cpp
```diff
@@ -4253,6 +4253,19 @@ void NCMesh::GetPointMatrix(Geometry::Type geom, const char* ref_path,
...
+
+         if (child == 0)
+         {
+            pm = PointMatrix(pm(0), mid01);
+         }
+         else if (child == 1)
+         {
+            pm = PointMatrix(mid01, pm(1));
+         }
+      }
```
jakubcerveny, line 4268: This is clear, Geometry::SEGMENT needs to be here. I would also add a final `else` branch with an `MFEM_ABORT("unsupported geometry " << geom);`, like everywhere else (this should probably have been here all along).

mesh/mesh.cpp
```diff
@@ -3508,6 +3506,8 @@ void Mesh::SetMeshGen()
             mesh_geoms |= (1 << Geometry::SQUARE);
             mesh_geoms |= (1 << Geometry::SEGMENT);
             mesh_geoms |= (1 << Geometry::POINT);
+         case Element::SEGMENT:
+            mesh_geoms |= (1 << Geometry::SEGMENT);
```
jakubcerveny, line 3510: I am slightly concerned about this change. It's clear that SEGMENT can be regarded as both "simplex" and "tensor product", depending on context. I am not 100% sure this won't break something else, but let's have this tested...

jakubcerveny (approved):
I think these changes make sense, and it's something that was probably missing in #2038.

mesh/mesh.cpp
rwa: Yes, I agree. I am not sure I fully understand the consequences or intent of the meshgen variable. Testing should tell us something.

rw-anderson: Yes, I agree. I am not sure I fully understand the consequences or intent of the meshgen variable. Testing should tell us something about it anyway.

I guess another avenue here would be to modify the conditional I pointed to above. But I'm not entirely sure what modification should be made.

rw-anderson: I guess I didn't prefer that option because I though it should be default to treat 1D meshes this way, since it has more functionality. The other reason was that it seemed non-intuitive to be setting a flag with that name for a 1D mesh. So I wanted to pick up the behavior by default.
But certainly that would eliminate these other issues.

jakubcerveny: I like Veselin's suggestion, if that solves the problem it's probably better to keep the SEGMENT meshgen as it is...

PR #2504, opened 2021-09-03
Fix compilation errors on Visual Studio 2019 with CUDA support

antonhudym:
1. Remove "static" keyword from functions used as lambdas on "device"
2. Add _USE_MATH_DEFINES to CmakeLists.txt (without this change VS doesn't "see" this preprocessor in config.hpp)
3. Remove asm definition - asm is not allowed in VS anymore
4. Convert "const double Epsilon" to constexpr (fixes CUDA compilation error)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,6 +175,10 @@ else()
   set(MFEM_DEBUG OFF)
 endif()
 
+if (WIN32)
+	add_definitions(-D_USE_MATH_DEFINES)
+endif()
+
 # MPI -> hypre; PETSc (optional)
 if (MFEM_USE_MPI)
   find_package(MPI REQUIRED)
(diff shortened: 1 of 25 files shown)


tzanio:
@camierjs, @v-dobrev and @YohannDudouit -- is removing the static qualifier acceptable or should we add a `#define` to have it on/off depending on the architecture? What will be the implications on non-Windows systems if we remove it?

camierjs:
I'm ok removing them - it will make them visible to the compilation unit, but I guess that is not a real issue.

antonhudym:
Hi @tzanio,

I downloaded Astyle but I have a hard time running that makefile on Windows. In case this is just running makefile and committing changes - could you please do it on your side, if you have a chance? I guess when formatting is fixed - all tests will pass fine.

Thank you,
Anton

tzanio:
> could you please do it on your side, if you have a chance

done

PR #2510, opened 2021-09-04
Adding DofTransformations NonlinearForm methods [nonlinearform-doftrans-dev]

tuckerbabcock:
Hi all,

I've run into a problem using a `NonlinearForm` with a high-order ND finite element space and a tet mesh. Based on other issues ([specifically this comment](https://github.com/mfem/mfem/issues/2492#issuecomment-907559593)), I think I've been able to implement a fix (proposed here), but I'd love to get some feedback.

I've only applied the fix for domain integrators, since that's where the `BilinearForm` and `LinearForm` methods use the `DofTransformation` object, but I'm wondering where the interior/boundary face intergrators don't get the same treatment? I'm also wondering if the `BlockNonlinearForm` class should have a similar fix?

### DIFF
--- a/fem/nonlinearform.cpp
+++ b/fem/nonlinearform.cpp
@@ -96,6 +96,8 @@ double NonlinearForm::GetGridFunctionEnergy(const Vector &x) const
    Vector el_x;
    const FiniteElement *fe;
    ElementTransformation *T;
+   DofTransformation *doftrans;
+
    double energy = 0.0;
 
    if (dnfi.Size())
@@ -103,9 +105,10 @@ double NonlinearForm::GetGridFunctionEnergy(const Vector &x) const
       for (int i = 0; i < fes->GetNE(); i++)
       {
          fe = fes->GetFE(i);
-         fes->GetElementVDofs(i, vdofs);
+         doftrans = fes->GetElementVDofs(i, vdofs);
          T = fes->GetElementTransformation(i);
          x.GetSubVector(vdofs, el_x);
+         if (doftrans) {doftrans->InvTransformPrimal(el_x); }
          for (int k = 0; k < dnfi.Size(); k++)
          {
             energy += dnfi[k]->GetElementEnergy(*fe, *T, el_x);
@@ -166,6 +169,7 @@ void NonlinearForm::Mult(const Vector &x, Vector &y) const
    Vector el_x, el_y;
    const FiniteElement *fe;
    ElementTransformation *T;
+   DofTransformation *doftrans;
    Mesh *mesh = fes->GetMesh();
 
    py = 0.0;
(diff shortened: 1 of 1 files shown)


tzanio:

mlstowell:
Hi, @tuckerbabcock,

Thanks for catching this. Yes, the `BlockNonlinearForm` will need the same treatment. We could apply the `DofTransformation` on the interior/boundary faces as well but at this time we have no way to test the resulting algorithms. This transformation is only currently needed for high order Nedelec spaces on meshes with triangular interfaces between elements. When such elements impose different orientations on their shared triangular face we sometimes need to transform the degrees of freedom from one element to match those of the other. On boundary faces this should never be required but we could apply the transformation, which should be the identity, without any harm. On interior faces this transformation would certainly be necessary but the interior face integrators are only used with Discontinuous Galerkin methods and we don't currently have any code that I'm aware of which uses Nedelec basis functions in such a discontinuous formulation.

Thanks again,
Mark

tuckerbabcock:
Hi @mlstowell, thanks for the clarification. I guess I was thinking that the transformation would also be used for supporting finite element spaces that use variable p, but I'm not sure where I got that idea. 

I'll go ahead and give the same fix to the `BlockNonlinearForm`.

tuckerbabcock:
I added the `DofTransformations` to the `BlockedNonlinearForm` methods, storing them in `Array`s for `Mult` and `ComputeGradientBlocked` since they need to be reused to `TransformDual`.

mlstowell:
@tuckerbabcock, matching dofs for elements using different basis function orders did come up during the development of `DofTransformation` but we decided to handle those in a different fashion (more similar to nonconforming dofs on AMR meshes). As far as I know that is still in the planning or possibly the experimental phase.

mlstowell:
@tuckerbabcock and @jandrej, this PR looks good to me but I haven't been able to test it. Do either of you know of a suitable example or mini app that might exercise this?

tuckerbabcock:
@mlstowell I don't know of any, I don't think there are any examples or miniapps that use a `NonlinearForm` and an ND space. I've tested it with some of my own code that solves a nonlinear magnetostatic problem, and I can report that that works. I've got no ideas for the `BlockNonlinearForm` though.

mlstowell (approved):

jandrej:
The changes do make sense, but I'd like to see them tested somewhere.

mlstowell:
> The changes do make sense, but I'd like to see them tested somewhere.

I agree. There is a unit test for the `BlockNonlinearForm`. I haven't looked at it but we might be able to extend it to exercise this code as well.

jandrej:
We could simply test the `::Mult` part. Then the test doesn't need a solve or has to be nonlinear - would `VectorFEDomainLFCurlIntegrator` work for that? E.g.
```
- Add VectorFEDomainLFCurlIntegrator to NonlinearForm
- Define function F where we know curl(F)
- NonlinearForm.Mult(F, curlF)
- Check error norm of curlF against analytical solution
```
I've never done that with Vector finite elements though.

tuckerbabcock:
@jandrej I think that might work, but since the integrator does not actually depend on the state passed to it this would only end up testing the `TransformDual` part of the changes.

mlstowell:
It would be nice to have but it is not strictly necessary. I'll leave the final decision to @tzanio and @jandrej. Whatever they decide is fine with me.

tzanio:

PR #2526, opened 2021-09-10
Add a check to prevent bad parallel results on Nedelec NC tets of order >= 2 [nedelec-nc-tet-warning]

jakubcerveny:
As a result of #1046, triangular shared faces are broken in Nedelec spaces on NC meshes. This PR adds a check that warns the user of this, until the problem is fixed.

The fix will require looping over ranges of DOFs in `NeighborRowMessage::Decode` (those with the same `id.index`), so that internal face DOFs are processed together, if necessary. The problem only occurs on conforming shared triangles, whose DOFs should all be present in a message from the owner.

We will need to have access to a face-only DOF transformation, specifically one that only works with the internal face DOFs only (i.e. no edges).

| [#2526](https://github.com/mfem/mfem/pull/2526) | @jakubcerveny | @tzanio | @mlstowell + @tzanio | 09/10/21 | 09/17/21 | 09/18/21 | |

### DIFF
--- a/fem/pfespace.cpp
+++ b/fem/pfespace.cpp
@@ -2294,6 +2294,12 @@ int ParFiniteElementSpace
                                        Array<int> *dof_tdof,
                                        bool partial) const
 {
+   // TODO: general face DOF transformations in NeighborRowMessage::Decode()
+   MFEM_VERIFY(!(fec->GetOrder() >= 2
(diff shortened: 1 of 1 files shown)


tzanio:

tzanio (approved):
I think it is good to warn users of limitations like this.

mlstowell (approved):

tzanio:

PR #2572, opened 2021-09-27
Fix deprecation warnings for IterativeSolver::print_level

pazner:
I think it's better to deprecate `print_level` rather than just the setter, so that custom solvers that depend on this member variable can change to use `print_options`, and then we will be able to remove the legacy one altogether.

### DIFF
--- a/fem/tmop_tools.cpp
+++ b/fem/tmop_tools.cpp
@@ -444,14 +444,14 @@ double TMOPNewtonSolver::ComputeScalingFactor(const Vector &x,
       if (untangling == false && min_detT_out < 0.0)
       {
          // No untangling, and detJ got negative -- no good.
-         if (print_level >= 0)
+         if (print_options & PrintLevel::WARNINGS)
          { mfem::out << "Scale = " << scale << " Neg det(J) found.\n"; }
          scale *= detJ_factor; continue;
       }
       if (untangling == true && min_detT_out < *min_det_ptr)
       {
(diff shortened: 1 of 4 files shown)


termi-official:
Thanks for the elaboration on this @pazner . I think we are on the same page that the deprecation should be added - however we also should keep in mind that this pragma is not portable, which is why I have not added it before. However, I would even go one step futher and also deprecate the old `SetPrintLevel` method. What do you think? If you agree I will merge this and also deprecate the old `SetPrintLevel`.

pazner:
My only concern with deprecating the old version is verbosity. I think replacing
```c++
cg.SetPrintLevel(1);
```
with
```c++
cg.SetPrintLevel(IterativeSolver::PrintLevel::ERRORS | IterativeSolver::PrintLevel::WARNINGS |
                 IterativeSolver::PrintLevel::ITERATION_DETAILS);
```
is not very pleasant.

(One easy way to get around this would be to just have predefined print level constants, e.g.
```c++
cg.SetPrintLevel(IterativeSolver::PRINT_LEVEL_1);
```
or similar, but it still doesn't seem optimal.

termi-official:
I could prefix the members of the enum to ` PRINT_* ` , which seems to be compromise. It is slightly less verbose:
```cpp 
cg.SetPrintLevel(IterativeSolver::PRINT_ERRORS | IterativeSolver::PRINT_WARNINGS |
                 IterativeSolver::PRINT_ITERATION_DETAILS);
```

PR #2582, opened 2021-09-29
[WIP] Get Face Adjacent Elements for an NCMesh

psocratis:


psocratis:
Thanks @pazner, this is indeed much simpler.

PR #2613, opened 2021-10-18
Implement Local Discontinuous Galerkin method for Poisson 

smsolivier:
The Local Discontinuous Galerkin method for Poisson requires "lifting operators" that involve mixed DG face integrators. Currently, `MixedBilinearForm` only supports trace integrators where the trial space is assumed to be continuous. I have added interior and boundary face integrators to `MixedBilinearForm` that allow both the test and trial spaces to be discontinuous. This required adding an `AssembleFaceMatrix` to `BilinearFormIntegrator` that accepts four elements: 2 trial elements from each side of the face and 2 test elements from each side of the face. I also added functionality to build the resulting rectangular parallel matrix. 

The `LDGTraceIntegrator` implements the consistent term from integrating by parts as well as an optional upwinding term (the "switches"). `DGPenaltyIntegrator` is just the penalty term from `DGDiffusionIntegrator`. With the switches, penalization on interior faces is not required for stability.  

I added `ex30.cpp` and `ex30p.cpp`. These solve the Poisson equation with LDG in first-order form by forming the Schur complement. 

@pazner what do you think? @bensworth this branch should be good to go for your purposes (see `ex30.cpp`).


pazner:
@smsolivier, sorry for the delay. This looks great! I will try to take a closer look at it soon, and review it more thoroughly.

smsolivier:
@pazner no worries! This is still a draft. I can help fix things. 

It's failing tests because I use `KLUSolver` in `ex30.cpp` without checking if that is available. I just wanted to show how LDG would work. It might make more sense to add this as an option in example 14.

tzanio:

examples/ex30p.cpp
```diff
@@ -0,0 +1,103 @@
...
+	HypreParMatrix *P = Pform.ParallelAssemble(); 
+
+	HypreParMatrix *DMi = ParMult(D, Mi, true); 
+	HypreParMatrix *DMiDt = ParMult(DMi, DT, true); 
+	HypreParMatrix *S = ParAdd(DMiDt, P); 
+
+	HypreBoomerAMG amg; 
+	amg.SetPrintLevel(0); 
+
+	BiCGSTABSolver cg(MPI_COMM_WORLD);
```
bensworth, line 84: @smsolivier Why are you using BiCGStab? Wouldn't just CG make sense here?

examples/ex33.cpp
```diff
@@ -0,0 +1,90 @@
...
+   Pform.Assemble();
+   Pform.Finalize();
+   SparseMatrix P(Pform.SpMat());
+
+   SparseMatrix *tmp = RAP(Mi, D);
+   SparseMatrix *S = Add(*tmp, P);
+
+   KLUSolver klu(*S);
+   GridFunction T(&sfes);
+   klu.Mult(fform, T);
```
pazner, line 81: Could you add some `#ifdefs` so that this example will still compile and run without SuiteSparse?

bensworth:
Hi @pazner, this disappeared from everybodys radar, and now we rely on a branch of mfem thats 4000 commits out of date. Do you think we can just copy the integrators to a fresh branch and PR that w/o an example? I think there are lots of other functions that are not used in examples, and this may make an easier pr.

PR #2611, opened 2021-10-15
Add HypreParVector::Read

pazner:
Resolves #2608.

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge| 
 | --- | --- | --- | --- | --- | --- | --- | 
| [#2611](https://github.com/mfem/mfem/pull/2611) | @pazner | @tzanio | @dylan-copeland + @tzanio | 10/22/21 | 10/31/21 | 11/02/21 | |

PR #2639, opened 2021-11-04
Add move constructors/assignment for Vector and HypreParVector [hypreparvec-move-dev]

joshessman-llnl:
This PR adds move constructors and assignment operators to `Vector` and `HypreParVector` to avoid unnecessary copies and to allow for defaulted move constructors/assignment of types that are composed of `Vector` or `HypreParVector` 

| [#2639](https://github.com/mfem/mfem/pull/2639) | @joshessman-llnl | @pazner | @pazner + @mlstowell | 11/05/21 | 11/22/21 | 11/30/21 | |

### DIFF
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -147,10 +147,7 @@ HypreParVector::HypreParVector(const HypreParVector &y) : Vector()
 HypreParVector::HypreParVector(HypreParVector &&y) : HypreParVector(
       static_cast<HYPRE_ParVector>(y))
 {
-   // If the argument vector owns its data, then the constructed vector will as well
-   own_ParVector = y.own_ParVector;
-   // Either way, the argument vector will not own its data
-   y.own_ParVector = 0;
+   *this = std::move(y);
 }
 
 HypreParVector::HypreParVector(const HypreParMatrix &A,
@@ -234,6 +231,9 @@ HypreParVector& HypreParVector::operator=(HypreParVector &&y)
    WrapHypreParVector(static_cast<hypre_ParVector*>(y), y.own_ParVector);
    // Either way the argument vector will no longer own its data
    y.own_ParVector = 0;
+   y.x = nullptr;
+   y.data.Reset();
+   y.size = 0;
    return *this;
 }
 
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -141,12 +141,13 @@ HypreParVector::HypreParVector(const HypreParVector &y) : Vector()
    own_ParVector = 1;
 }
 
-HypreParVector::HypreParVector(HypreParVector &&other) : HypreParVector(static_cast<HYPRE_ParVector>(other))
+HypreParVector::HypreParVector(HypreParVector &&y) : HypreParVector(
+      static_cast<HYPRE_ParVector>(y))
 {
    // If the argument vector owns its data, then the constructed vector will as well
-   own_ParVector = other.own_ParVector;
+   own_ParVector = y.own_ParVector;
    // Either way, the argument vector will not own its data
-   other.own_ParVector = 0;
+   y.own_ParVector = 0;
 }
 
 HypreParVector::HypreParVector(const HypreParMatrix &A,
@@ -221,6 +222,21 @@ HypreParVector& HypreParVector::operator=(const HypreParVector &y)
 #endif
 
    Vector::operator=(y);
+   return *this;
+}
+
+HypreParVector& HypreParVector::operator=(HypreParVector &&y)
+{
+#ifdef MFEM_DEBUG
+   if (size != y.Size())
+   {
+      mfem_error("HypreParVector::operator=");
+   }
+#endif
+   // If the argument vector owns its data, then the calling vector will as well
+   WrapHypreParVector(static_cast<hypre_ParVector*>(y), y.own_ParVector);
+   // Either way the argument vector will no longer own its data
+   y.own_ParVector = 0;
    return *this;
 }
 
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -144,8 +144,7 @@ HypreParVector::HypreParVector(const HypreParVector &y) : Vector()
    own_ParVector = 1;
 }
 
-HypreParVector::HypreParVector(HypreParVector &&y) : HypreParVector(
-      static_cast<HYPRE_ParVector>(y))
+HypreParVector::HypreParVector(HypreParVector &&y)
 {
    *this = std::move(y);
 }
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -227,12 +227,6 @@ HypreParVector& HypreParVector::operator=(const HypreParVector &y)
 
 HypreParVector& HypreParVector::operator=(HypreParVector &&y)
 {
-#ifdef MFEM_DEBUG
-   if (size != y.Size())
-   {
-      mfem_error("HypreParVector::operator=");
-   }
-#endif
    // If the argument vector owns its data, then the calling vector will as well
    WrapHypreParVector(static_cast<hypre_ParVector*>(y), y.own_ParVector);
    // Either way the argument vector will no longer own its data
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -141,6 +141,14 @@ HypreParVector::HypreParVector(const HypreParVector &y) : Vector()
    own_ParVector = 1;
 }
 
+HypreParVector::HypreParVector(HypreParVector &&other) : HypreParVector(static_cast<HYPRE_ParVector>(other))
+{
+   // If the argument vector owns its data, then the constructed vector will as well
+   own_ParVector = other.own_ParVector;
+   // Either way, the argument vector will not own its data
+   other.own_ParVector = 0;
+}
+
 HypreParVector::HypreParVector(const HypreParMatrix &A,
                                int transpose) : Vector()
 {
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -127,25 +127,18 @@ HypreParVector::HypreParVector(MPI_Comm comm, HYPRE_BigInt glob_size,
    own_ParVector = 1;
 }
 
-HypreParVector::HypreParVector(const HypreParVector &y) : Vector()
+// Call the move constructor on the "compatible" temp vector
+HypreParVector::HypreParVector(const HypreParVector &y) : HypreParVector(
+      y.CreateCompatibleVector())
 {
-   x = hypre_ParVectorCreate(y.x -> comm, y.x -> global_size,
-                             y.x -> partitioning);
-   hypre_ParVectorInitialize(x);
-#if MFEM_HYPRE_VERSION <= 22200
-   hypre_ParVectorSetPartitioningOwner(x,0);
-#endif
-   hypre_ParVectorSetDataOwner(x,1);
-   hypre_SeqVectorSetDataOwner(hypre_ParVectorLocalVector(x),1);
    // Deep copy the local data
    hypre_SeqVectorCopy(hypre_ParVectorLocalVector(y.x),
                        hypre_ParVectorLocalVector(x));
-   _SetDataAndSize_();
-   own_ParVector = 1;
 }
 
 HypreParVector::HypreParVector(HypreParVector &&y)
 {
+   own_ParVector = 0;
    *this = std::move(y);
 }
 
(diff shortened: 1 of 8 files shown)


joshessman-llnl:
I also noticed that the test added in https://github.com/mfem/mfem/pull/2611 was not actually being run and also may have needed some changes to the partitioning logic.

v-dobrev:
> I also noticed that the test added in #2611 was not actually being run and also may have needed some changes to the partitioning logic.

I think you are right -- it looks like it was not added to the `CMakeLists.txt`.

pazner:

tests/unit/linalg/test_hypre_vector.cpp
```diff
@@ -38,11 +32,30 @@ TEST_CASE("HypreParVector I/O", "[Parallel], [HypreParVector]")
    else
    {
       col.resize(world_size+1);
-      for (int i=0; i<world_size; ++i)
+      for (int i=1; i<=world_size; ++i)
```
pazner, line 35: Good catch on the `<=`. Shouldn't it still start at 0 though?

pazner:
> > I also noticed that the test added in #2611 was not actually being run and also may have needed some changes to the partitioning logic.
> 
> I think you are right -- it looks like it was not added to the `CMakeLists.txt`.

My bad! Thanks for fixing it.

linalg/vector.cpp
```diff
@@ -54,6 +54,13 @@ Vector::Vector(const Vector &v)
    UseDevice(v.UseDevice());
 }
 
+Vector::Vector(Vector &&v) : data(std::move(v.data)), size(v.size)
+{
+   data.SetHostPtrOwner(v.data.OwnsHostPtr());
+   v.data.Reset();
+   v.size = 0;
+}
+
```
pazner, line 63: Can avoid duplicating this logic
```suggestion
Vector::Vector(Vector &&v)
{
   *this = std::move(v);
}
```

tests/unit/linalg/test_hypre_vector.cpp
```diff
@@ -65,6 +78,101 @@ TEST_CASE("HypreParVector I/O", "[Parallel], [HypreParVector]")
...
+   // Initialize vector
+   HypreParVector v1(MPI_COMM_WORLD, glob_size, col.data());
+   for (int i=0; i<size_per_rank; ++i)
+   {
+      v1(i) = rank*size_per_rank + i;
+   }
+
+   // Make a copy so we can compare it later
+   HypreParVector v1_copy(v1);
+   Vector v1_base_copy(v1);
```
pazner, line 101: I don't think you need these "`base`" vectors, you can index directly into the `HypreParVector`.

tests/unit/linalg/test_hypre_vector.cpp
```diff
@@ -17,18 +17,12 @@ namespace mfem
 
 #ifdef MFEM_USE_MPI
 
-TEST_CASE("HypreParVector I/O", "[Parallel], [HypreParVector]")
+namespace detail
```
v-dobrev, line 20: We use `internal` to hide some implementation details in Doxygen, see https://github.com/mfem/mfem/blob/167ff72edccc4e5b0a642ae3392ccaf50f455b9e/doc/CodeDocumentation.conf.in#L857
You can also add `mfem::detail` there if you want to keep the `detail` namespace here and exclude it from Doxygen.

tests/unit/linalg/test_hypre_vector.cpp
```diff
@@ -65,6 +78,101 @@ TEST_CASE("HypreParVector I/O", "[Parallel], [HypreParVector]")
...
+   REQUIRE(v1_move.Size() == v1_copy.Size());
+   REQUIRE(v1_move.GlobalSize() == v1_copy.GlobalSize());
+
+   REQUIRE(v1_move.GetData() == v1_data);
+
+   // The local subvectors should be the same
+   Vector v1_base_move(v1_move);
+   for (int i = 0; i < v1_base_move.Size(); i++)
+   {
+      REQUIRE((v1_base_move(i) - v1_base_copy(i)) == MFEM_Approx(0.0));
```
pazner, line 168: Ditto re: "base" vectors.

linalg/hypre.cpp
```diff
@@ -216,6 +228,15 @@ HypreParVector& HypreParVector::operator=(const HypreParVector &y)
    return *this;
 }
 
+HypreParVector& HypreParVector::operator=(HypreParVector &&y)
+{
+   // If the argument vector owns its data, then the calling vector will as well
+   WrapHypreParVector(static_cast<hypre_ParVector*>(y), y.own_ParVector);
+   // Either way the argument vector will no longer own its data
+   y.own_ParVector = 0;
```
pazner, line 236: Would it make sense to set `y`'s `hypre_ParVector` member data to `nullptr` (the same way you zero-out the data and size in the `Vector` move assignment operator)?

linalg/hypre.cpp
```diff
@@ -137,10 +137,22 @@ HypreParVector::HypreParVector(const HypreParVector &y) : Vector()
...
 
+HypreParVector::HypreParVector(HypreParVector &&y) : HypreParVector(
+      static_cast<HYPRE_ParVector>(y))
+{
+   // If the argument vector owns its data, then the constructed vector will as well
+   own_ParVector = y.own_ParVector;
+   // Either way, the argument vector will not own its data
+   y.own_ParVector = 0;
+}
+
```
pazner, line 155:
```suggestion
HypreParVector::HypreParVector(HypreParVector &&y)
{
   *this = std::move(y);
}
```

linalg/hypre.hpp
```diff
@@ -141,8 +141,10 @@ class HypreParVector : public Vector
        allocated in the memory location HYPRE_MEMORY_DEVICE. */
    HypreParVector(MPI_Comm comm, HYPRE_BigInt glob_size, double *data_,
                   HYPRE_BigInt *col, bool is_device_ptr = false);
-   /// Creates vector compatible with y
+   /// Creates vector compatible with y containing a deep copy of the local data
```
pazner, line 144: Maybe just this is clearer?
```suggestion
/// Creates a deep copy of @a y
```

pazner (commented):
Looks like a good addition! Just some minor comments.

tests/unit/linalg/test_hypre_vector.cpp
joshessman-llnl: I believe it shouldn't make a difference since that first element will always be zero anyways, but for clarity I agree that the loop should start at zero.

linalg/hypre.cpp
```diff
@@ -147,10 +147,7 @@ HypreParVector::HypreParVector(const HypreParVector &y) : Vector()
 HypreParVector::HypreParVector(HypreParVector &&y) : HypreParVector(
       static_cast<HYPRE_ParVector>(y))
```
pazner: Could we do this?
```suggestion
HypreParVector::HypreParVector(HypreParVector &&y)
```

pazner (approved):

linalg/vector.cpp
```diff
@@ -54,6 +54,13 @@ Vector::Vector(const Vector &v)
    UseDevice(v.UseDevice());
 }
 
+Vector::Vector(Vector &&v) : data(std::move(v.data)), size(v.size)
+{
+   data.SetHostPtrOwner(v.data.OwnsHostPtr());
```
v-dobrev: This is already done by invoking the move ctor for `data`.

v-dobrev:
Do you mind adding an optional parameter to the ctor: https://github.com/mfem/mfem/blob/12cc340afb472a95e941513f611aedf0e5a8803a/linalg/hypre.hpp#L144-L145 like this:
```c++
   /// Creates a deep copy of @a y, optionally skipping the copy of the data
   HypreParVector(const HypreParVector &y, bool copy_data = true); 
```

pazner:
Another option that might be worth considering is instead of adding a boolean flag to the constructor (whose purpose might not be obvious at the call-site) would be to define
```c++
HypreParVector HypreParVector::CreateCompatibleVector() const
```
which would create and return a compatible vector, and the use could then take advantage of the move constructor/assignment operator.

linalg/hypre.cpp
```diff
@@ -137,10 +137,18 @@ HypreParVector::HypreParVector(const HypreParVector &y) : Vector()
...
    hypre_ParVectorSetDataOwner(x,1);
    hypre_SeqVectorSetDataOwner(hypre_ParVectorLocalVector(x),1);
+   // Deep copy the local data
+   hypre_SeqVectorCopy(hypre_ParVectorLocalVector(y.x),
+                       hypre_ParVectorLocalVector(x));
    _SetDataAndSize_();
    own_ParVector = 1;
 }
 
+HypreParVector::HypreParVector(HypreParVector &&y)
```
pazner, line 147:
```suggestion
HypreParVector::HypreParVector(HypreParVector &&y) : HypreParVector()
```
My previous suggestion caused the CI failure on Mac (`WrapHypreParVector` was trying to destroy `x` before it initialized). Sorry about that! This should be better now.

v-dobrev:
I think it is better for `HypreParVector::CreateCompatibleVector` to be *in addition*, not *instead of* the proposed optional parameter. Clarity at the call site is easy to achieve:
```c++
   const bool copy_data = false;
   HypreParVector tmp(x, copy_data);
```
Also, the implementation of `HypreParVector::CreateCompatibleVector` is one line instead of needing to copy-paste the ctor and delete the line that copies the data.

linalg/hypre.cpp
joshessman-llnl: Yep - I noticed this as well and just set `own_parVector` to false before the move assignment operator call

pazner: Yeah, that will fix it too. Go with whichever option you think is clearer.

samuelpmishLLNL:
> Do you mind adding an optional parameter to the ctor:
> 
> https://github.com/mfem/mfem/blob/12cc340afb472a95e941513f611aedf0e5a8803a/linalg/hypre.hpp#L144-L145
> 
> like this:
> ```c++
>    /// Creates a deep copy of @a y, optionally skipping the copy of the data
>    HypreParVector(const HypreParVector &y, bool copy_data = true); 
> ```

I agree with @pazner that a named function is preferable to this implementation. The comment "Creates a deep copy of @a y, optionally skipping the copy of the data" doesn't make sense to me, as the entire point of a deep copy is to not skip copying the data. 

The code snippet 
```cpp
   const bool copy_data = false;
   HypreParVector tmp(x, copy_data);
```
is only as clear as the name of the temporary, in which case it's just as easy for someone to accidentally write
```cpp
   const bool copy_partitioning = false;
   HypreParVector tmp(x, copy_partitioning);
```
which introduces a much more dangerous bug that will likely go unnoticed by the compiler and developer.

A named function like `HypreParVector::CreateCompatibleVector` can't be accidentally misunderstood in the same way (although I think the word "Compatible" is ambiguous). Another option is to encapsulate the partitioning of a HypreParVector in an object, and make a separate constructor that takes a partitioning, e.g.

```cpp
mfem::HypreParVector foo = ... ;

// copy ctor makes a deep copy of foo, semantically consistent with mfem::Vector
mfem::HypreParVector bar(foo) ;

// HypreParVector(mfem::Partitioning p) makes a new HypreParVector with the specified partitioning
mfem::HypreParVector baz(foo.partitioning);
```

v-dobrev:
Okay, @joshessman-llnl's solution in the last commit looks good.

Since you both don't like it, I don't insist on adding the optional `bool copy_data = true` parameter to the copy ctor.

pazner:
@mlstowell, could you take a look?

tests/unit/linalg/test_hypre_vector.cpp
```diff
@@ -65,6 +78,85 @@ TEST_CASE("HypreParVector I/O", "[Parallel], [HypreParVector]")
...
+   HypreParVector v1_copy(v1);
+
+   // Also save off the data pointer
+   double* v1_data = v1.GetData();
+
+   // Now move from the original and compare
+   HypreParVector v1_move;
+   v1_move = std::move(v1);
+
+   REQUIRE(v1.GetOwnership() == 0);
```
mlstowell, line 148: Could this also check the size and global size of v1? This would be similar to what happens in the `test_vector.cpp`. If this is possible the same checks should be added to the move constructor test above.

mlstowell (approved):
With the small changes to the unit tests, if possible, I think this looks good.

tests/unit/linalg/test_hypre_vector.cpp
joshessman-llnl: Because the underlying vector is set to `nullptr` per https://github.com/mfem/mfem/pull/2639#discussion_r744037793 the call to `GlobalSize` will segfault, but I added checks for `size() == 0` and `GetData() == nullptr`.

pazner:

PR #2656, opened 2021-11-10
GridFunction coefficients that work with LOR preconditioning

pazner:
This PR:
- Adds `mesh` data member to `ElementTransformation`
- Allows `GridFunctionCoefficient`, `VectorGridFunctionCoefficient`, `GradientGridFunctionCoefficient`, `CurlGridFunctionCoefficient`, and `DivergenceGridFunctionCoefficient` to work on both coarse and refined meshes
    - The same coefficient will work on the high-order mesh and the low-order refined mesh)
    - Note, however, the versions of the vector coefficients that evaluate an entire `IntegrationRule` at once fall back on slower versions. These could be optimized.

<details>
<summary>See here for a patch that tests this new functionality using the LOR solvers miniapps</summary>

```diff
diff --git a/miniapps/solvers/lor_solvers.cpp b/miniapps/solvers/lor_solvers.cpp
index dac300971..e6b914f7d 100644
--- a/miniapps/solvers/lor_solvers.cpp
+++ b/miniapps/solvers/lor_solvers.cpp
@@ -70,6 +70,11 @@ using namespace mfem;
 
 bool grad_div_problem = false;
 
+double eps_fn(const Vector &xvec)
+{
+   return 0.01 + xvec[0]*xvec[0] + xvec[1]*xvec[1];
+}
+
 int main(int argc, char *argv[])
 {
    const char *mesh_file = "../../data/star.mesh";
@@ -122,11 +127,16 @@ int main(int argc, char *argv[])
    // In DG, boundary conditions are enforced weakly, so no essential DOFs.
    if (!L2) { fes.GetBoundaryTrueDofs(ess_dofs); }
 
+   FunctionCoefficient eps_coeff(eps_fn);
+   GridFunction coeff_gf(&fes);
+   coeff_gf.ProjectCoefficient(eps_coeff);
+   GridFunctionCoefficient coeff(&coeff_gf);
+
    BilinearForm a(&fes);
    if (H1 || L2)
    {
       a.AddDomainIntegrator(new MassIntegrator);
-      a.AddDomainIntegrator(new DiffusionIntegrator);
+      a.AddDomainIntegrator(new DiffusionIntegrator(coeff));
    }
    else
    {
diff --git a/miniapps/solvers/plor_solvers.cpp b/miniapps/solvers/plor_solvers.cpp
index 0df6715d8..b8f747aef 100644
--- a/miniapps/solvers/plor_solvers.cpp
+++ b/miniapps/solvers/plor_solvers.cpp
@@ -68,6 +68,11 @@ using namespace mfem;
 
 bool grad_div_problem = false;
 
+double eps_fn(const Vector &xvec)
+{
+   return 0.01 + xvec[0]*xvec[0] + xvec[1]*xvec[1];
+}
+
 int main(int argc, char *argv[])
 {
    MPI_Session mpi;
@@ -137,11 +142,16 @@ int main(int argc, char *argv[])
    // In DG, boundary conditions are enforced weakly, so no essential DOFs.
    if (!L2) { fes.GetBoundaryTrueDofs(ess_dofs); }
 
+   FunctionCoefficient eps_coeff(eps_fn);
+   ParGridFunction coeff_gf(&fes);
+   coeff_gf.ProjectCoefficient(eps_coeff);
+   GridFunctionCoefficient coeff(&coeff_gf);
+
    ParBilinearForm a(&fes);
    if (H1 || L2)
    {
       a.AddDomainIntegrator(new MassIntegrator);
-      a.AddDomainIntegrator(new DiffusionIntegrator);
+      a.AddDomainIntegrator(new DiffusionIntegrator(coeff));
    }
    else
    {
```
</details>

| [#2656](https://github.com/mfem/mfem/pull/2656) | @pazner | @tzanio | @dylan-copeland + @mlstowell + @psocratis | 11/23/21 | 12/12/21 | 12/16/21 | |


pazner:
I added a unit test that tests all the variations. It should be ready for review now.

tests/unit/fem/test_lor.cpp
```diff
@@ -0,0 +1,120 @@
...
+using namespace mfem;
+
+// Tests the use of refined/LOR grid function coefficients.
+//
+// Given a space fes, and a refined space fes_refined (either vector or scalar
+// spaces), projects coeff_1 onto a grid function in fes, and then creates the
+// corresponding grid function coefficient (could be scalar, vector, grad, div,
+// or curl grid function coefficients). Then, this grid function coefficient is
+// projected onto fes_refined, and compared with the result of projecting
+// coeff_2. onto fes_refined.
```
dylan-copeland, line 26: typo (period)

dylan-copeland (approved):
Thanks @pazner! It would be nice to have parallel unit tests as well.

mlstowell:
I have a concern about this design because the `Coefficient` classes are one of the most common user-implemented classes in MFEM. Granted, these `GridFunction` dependent coefficients are not the most common type of coefficient. However, I'm concerned that users may look at these and wonder if they need to do something similar in their implementations.

Also, the `GridFunction::Get*` methods will have the same issue with refined meshes when called in other contexts.

Would it be possible and/or advisable to move the use of `RefinedToCoarse` inside the `GridFunction::Get*` methods? It looks like the same information would be available inside a call to `GridFunction::GetVectorValue` for example.

tests/unit/fem/test_lor.cpp
pazner: Fixed, thanks.

pazner:
> Thanks @pazner! It would be nice to have parallel unit tests as well.

I thought about this — the parallel tests would essentially be duplicates of the current tests, but with `Par` objects. I think this is probably unnecessary. The new code doesn't do anything in parallel.

pazner:
> I have a concern about this design because the `Coefficient` classes are one of the most common user-implemented classes in MFEM. Granted, these `GridFunction` dependent coefficients are not the most common type of coefficient. However, I'm concerned that users may look at these and wonder if they need to do something similar in their implementations.
> 
> Also, the `GridFunction::Get*` methods will have the same issue with refined meshes when called in other contexts.
> 
> Would it be possible and/or advisable to move the use of `RefinedToCoarse` inside the `GridFunction::Get*` methods? It looks like the same information would be available inside a call to `GridFunction::GetVectorValue` for example.

These are good points. From an implementation point of view, these changes can definitely happen at the GridFunction rather than Coefficient level.

A few comments on weighing the different options:

* I think changing `GridFunction` would be more invasive. It would also likely necessitate changes in `ParGridFunction`. `GridFunction::GetValue` and related functions (`GetVectorValue`, `GetGradient`, etc.) are already quite complicated, and this would add another layer of complication.
* This functionality would only work with `GridFunction::GetValue(ElementTransformation &, […])` and not with `GridFunction::GetValue(int , […])`, for which it doesn't really make sense.
* `GridFunction` objects are clearly associated with a finite element space (and hence mesh), but `Coefficient` objects are not. If handed a `Coefficient`, the user may rightly expect to be able to evaluate it given an element transformation and integration point. On the other hand, a `GridFunction` is explicitly tied to the mesh. The case of `GridFunctionCoefficient` adds a limitation on `Coefficient` that the element must come from the mesh corresponding to the grid function's space, but if the user only has a reference to the base class `Coefficient`, this limitation is not evident. That's why I think it makes sense to make `GridFunctionCoefficient` behave a little bit more like e.g. `FunctionCoefficient`, which is not tied to any mesh. One could imagine going even further, and falling back on something like `findpts` if there is no coarse-fine relationship between the meshes, but that's beyond the scope here.

If I understand what you wrote correctly, an advantage of adding this functionality to GridFunction rather than the specific Coefficient classes is that if users implement their own custom Coefficients that use the `GridFunction::GetValue` (and related functions), they will get this functionality "for free". I'm not sure how useful this would be to users, and if it is worth the more complicated implementation.

tzanio:

psocratis:
I understand that both approaches have their advantages and disadvantages, but I would vote for the current implementation mostly because of @pazner's 3rd bullet point above. That is, a `GridFunction` is an entity associated with a `FiniteElementSpace` hence a `Mesh`. As far as confusion for users, I think that this can be clearly documented, though it's pretty clear that this only applies to `GridFunctionCoefficient`s and not to the generic `Coefficient/FunctionCoefficient` 
I also like that idea that this can be naturally be extended using FindPts.

mlstowell (approved):

psocratis (approved):

tzanio:
Please mention in `CHANGELOG`

tzanio:

PR #2657, opened 2021-11-10
Formalize the names for QuadratureSpace [qspace-new-dev]

joshessman-llnl:
In order to save and restore `QuadratureFunction` data within Axom's `MFEMSidreDataCollection`, "basis" metadata is saved to disk - roughly equivalent to `FiniteElementCollection::Name()` and `FiniteElementCollection::New`.

A naming convention for `QuadratureSpace` is already implemented unilaterally by Axom, so this PR "formalizes" this naming convention within MFEM.  This convention was discussed previously on Slack and takes into account potential extensions to `QuadratureSpace` and `QuadratureFunction`.

See https://github.com/LLNL/axom/compare/refactor/essman/qfunc_new?expand=1 for an overview of the corresponding changes in Axom.  Note that because the vdim is not associated with the QSpace that information has to be added "manually".  Possibly a better solution exists?

ping @rcarson3


fem/fespace.hpp
```diff
@@ -975,6 +977,25 @@ class QuadratureSpace
...
+       @param[in] mesh The mesh to use when constructing the QuadratureSpace */
+   /**
+
+   | QSpace Basis Name | Notes |
+   | :------: | :---: |
+   | QF_Default_[ORDER] | Default quadrature type |
+
+   | Tag | Description |
+   | :------: | :--------: |
+   | [ORDER]  | Approximation order of the elements (P0, P1, P2, ...) |
```
rcarson3, line 996: Order here isn't the approximate order of the element. Rather, it is the order of the integration rule being used for the `QuadratureSpace`.

rcarson3:
@joshessman-llnl so I would say the vdim info could be fed into a QF initialization using this same naming scheme. It would be able to parse the name:

1.   See if it matches up with a provided QSpace and fail if they don't
2.  Make use of the vdim term from the name to set the vdim for the QF

If a mesh is supplied to a QF initialization with the name, one could also create an owned QSpace on the fly based on the name as well. However, I don't know if that is a desirable feature to have.

tzanio:

jamiebramwell:
I will follow up with the Axom team to see how we want to proceed with this.

PR #2689, opened 2021-12-03
Iterable DataCollection

termi-official:
For post-processing purposes it can be useful to load a DataCollection at the stored cycles. Assuming these cycles are not contiguous, this PR presents a very simple take on designing an iterable DataCollection (called MFEMDataCollection in the first version) along a simple miniapp to replay parallel simulations with GLVis.

Currently the `\Delta t` in the is the time step length used at the corresponding time point in the simulation, and not the time difference between two consecutive calls to `Save`. AMR may be not functional at the current state.

**NOTE!** This PR primarily exists for the purpose of discussing possible interfaces (see #2574 , ping @v-dobrev).

### DIFF
--- a/config/defaults.mk
+++ b/config/defaults.mk
@@ -24,7 +24,7 @@ EGREP_BIN = $(shell command -v egrep 2> /dev/null)
 CXX = g++
 MPICXX = mpicxx
 
-BASE_FLAGS  = -std=c++11
+BASE_FLAGS  = -std=c++17
 OPTIM_FLAGS = -O3 $(BASE_FLAGS)
 DEBUG_FLAGS = -g $(XCOMPILER)-Wall $(BASE_FLAGS)
 
--- a/examples/ex9p.cpp
+++ b/examples/ex9p.cpp
@@ -250,6 +250,7 @@ int main(int argc, char *argv[])
    double t_final = 10.0;
    double dt = 0.01;
    bool visualization = true;
+   bool internal = false;
    bool visit = false;
    bool paraview = false;
    bool adios2 = false;
@@ -298,6 +299,9 @@ int main(int argc, char *argv[])
    args.AddOption(&visualization, "-vis", "--visualization", "-no-vis",
                   "--no-visualization",
                   "Enable or disable GLVis visualization.");
+   args.AddOption(&internal, "-idf", "--mfem-datafiles", "-no-idf",
+                  "--no-mfem-datafiles",
+                  "Save data files in the internal formal.");
    args.AddOption(&visit, "-visit", "--visit-datafiles", "-no-visit",
                   "--no-visit-datafiles",
                   "Save data files for VisIt (visit.llnl.gov) visualization.");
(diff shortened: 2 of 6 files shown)


termi-official:
I could need some feedback on this, whether this idea is generally useful and if it is, then in which direction I should head with the software design.

mlstowell:

termi-official:
Great, thanks!

termi-official:
If I have to clarify something feel free to just ask.

mlstowell:
Thanks for your patience, @termi-official. The last few weeks have been unusually busy for us.

config/defaults.mk
```diff
@@ -24,7 +24,7 @@ EGREP_BIN = $(shell command -v egrep 2> /dev/null)
 CXX = g++
 MPICXX = mpicxx
 
-BASE_FLAGS  = -std=c++11
+BASE_FLAGS  = -std=c++17
```
mlstowell, line 27: This type of change will require a lengthy internal discussion. Is this something needed by this PR or can we revert it?

mlstowell:
I've run `ex9p` and compared the VisIt and ParaView output files to the new "internal" output files. As you mentioned in issue #2574 the existing formats provide the same information but there are drawbacks to extracting it reliably. For example, if you have output multiple VisIt data files from different simulations in the same directory you cannot distinguish output files from different simulations if you're only looking at lists of file names. The VisIt tool itself suffers from this issue.

Your solution to write a meta file containing cycle and time data for a particular simulation is a good idea with small overhead. While it won't fix the VisIt issue it certainly will help while loading MFEM's `DataCollection` objects. It looks like ParaView uses a similar idea. I haven't looked a Sidre.

I think @v-dobrev suggested adding a virtual method to the DataCollection base class so that this type of information could be more easily accessed. In the case of the `VisItDataCollection` this would likely require the addition of meta data file as you suggest.

While it doesn't hurt to have a generic `MFEMDataCollection` designed for the purposes you mention I think this should be done in addition to augmenting the existing `DataCollection` classes to support this type of iteration.

mlstowell:
One final note on the `MFEMDataCollection`. I like your idea of writing the mesh only once if it won't be changing. However, as I'm sure you're aware, AMR is not the only way a mesh can change. We could move the nodes of the mesh without changing its connectivity. It would be nice to capture this as well. Perhaps writing the connectivity once while writing multiple `GridFunctions` for the nodes. Probably beyond the scope of this PR but I thought I'd mention it.

config/defaults.mk
termi-official: I can remove this one. I just did it for convenience to use an optional instead of a maybe-nullptr.

termi-official:
The main issue with the other data collections which I have faced in the past was loading parallel NC meshes and fields on it. To the best of my knowledge Sidre and the other standardized formats do not have support for non-conforming mesh formats (and hence also no support for forest mesh formats).

I an happy for suggestions regarding a method in the base class, but I think in general such "metadata files" as in this PR are too different to give a nice common interface. We can discuss this matter further. The format which I propose is actually quite similar to the `VisItDataCollection`. Maybe we can put another abstract class in-between the hierarchy like for example `MFEMDataCollection <: LoadableDataCollection <: DataCollection`?

Also you made a good catch regarding the dynamic meshes, I totally forgot this case when writing this, sorry. Do we have a cheap heuristic to check if the geometry of a mesh changed? It should not be too hard to add this - at least for the case when the nodes of a mesh are given by a grid function (which is not always the case IIRC).

acfisher:
@termi-official Is this something you are still working on?  It is marked both in-review and WIP?

termi-official:
Currently not as I do not have much spare time. This was an attempt to upstream what we use internally in our packages. There is still quite a bit to do I think, as this was opened for review to discuss the design.

PR #2731, opened 2021-12-22
Fixing shadow warnings about "out/err" globals [shadow-out-dev]

mlstowell:

To resolve these warnings we can rename any local variables named "out" or "err" to avoid reusing the names assigned to the global output methods defined in globals.hpp. The compiler has no trouble determining which variable is intended but the choice is not always obvious to developers who may not be aware of one or the other of the possibilities. 

| [#2731](https://github.com/mfem/mfem/pull/2731) | @mlstowell | @tzanio | @tzanio + @v-dobrev + @pazner | 12/22/21 | 2/19/22 | 2/23/22 | |


tzanio:

pazner:
Suggested edit:
```diff
diff --git a/fem/datacollection.cpp b/fem/datacollection.cpp
index ddd0c5f64..a08fafb8d 100644
--- a/fem/datacollection.cpp
+++ b/fem/datacollection.cpp
@@ -787,44 +787,36 @@ void ParaViewDataCollection::Load(int )
 
 std::string ParaViewDataCollection::GenerateCollectionPath()
 {
-   std::string sout = "";
-   sout = prefix_path + DataCollection::GetCollectionName();
-   return sout;
+   return prefix_path + DataCollection::GetCollectionName();
 }
 
 std::string ParaViewDataCollection::GeneratePVTUPath()
 {
-   std::string sout = "Cycle" + to_padded_string(cycle,pad_digits_cycle);
-   return sout;
+   return "Cycle" + to_padded_string(cycle,pad_digits_cycle);
 }
 
 std::string ParaViewDataCollection::GenerateVTUPath()
 {
-   std::string sout = GeneratePVTUPath();
-   return sout;
+   return GeneratePVTUPath();
 }
 
 std::string ParaViewDataCollection::GeneratePVDFileName()
 {
-   std::string sout = GetCollectionName()+".pvd";
-   return sout;
+   return GetCollectionName() + ".pvd";
 }
 
 std::string ParaViewDataCollection::GeneratePVTUFileName()
 {
-   std::string sout = "data.pvtu";
-   return sout;
+   return "data.pvtu";
 }
 
 std::string ParaViewDataCollection::GenerateVTUFileName()
 {
-   std::string sout = "proc" + to_padded_string(myid,pad_digits_rank)+".vtu";
-   return sout;
+   return "proc" + to_padded_string(myid,pad_digits_rank) + ".vtu";
 }
 std::string ParaViewDataCollection::GenerateVTUFileName(int crank)
 {
-   std::string sout = "proc" + to_padded_string(crank,pad_digits_rank)+".vtu";
-   return sout;
+   return "proc" + to_padded_string(crank,pad_digits_rank) + ".vtu";
 }
 
 void ParaViewDataCollection::Save()

```

pazner (commented):
In general I like the idea of avoiding shadowing global and member variables, but I have some concerns:

1. `out` and `err` are such common variable names (especially `err` in a finite element code), this becomes an intrusive change.
2. I'm not convinced by the merits of having `sout`, `pout`, `wout`, `oout` variables. The letter prefix is not so meaningful, and the code is a bit cryptic. If anything, we should settle on a convention, e.g. for any output stream, call it `out_stream` or something.
3. If we are going to make these changes, we should definitely enable `-Wshadow` in our builds, otherwise it's too easy to introduce new shadowing variables and then only catch them later.

mlstowell:
> Suggested edit:
> 
> ```diff
> diff --git a/fem/datacollection.cpp b/fem/datacollection.cpp
> index ddd0c5f64..a08fafb8d 100644
> --- a/fem/datacollection.cpp
> +++ b/fem/datacollection.cpp
> @@ -787,44 +787,36 @@ void ParaViewDataCollection::Load(int )
>  
>  std::string ParaViewDataCollection::GenerateCollectionPath()
>  {
> -   std::string sout = "";
> -   sout = prefix_path + DataCollection::GetCollectionName();
> -   return sout;
> +   return prefix_path + DataCollection::GetCollectionName();
>  }
>  
>  std::string ParaViewDataCollection::GeneratePVTUPath()
>  {
> -   std::string sout = "Cycle" + to_padded_string(cycle,pad_digits_cycle);
> -   return sout;
> +   return "Cycle" + to_padded_string(cycle,pad_digits_cycle);
>  }
>  
>  std::string ParaViewDataCollection::GenerateVTUPath()
>  {
> -   std::string sout = GeneratePVTUPath();
> -   return sout;
> +   return GeneratePVTUPath();
>  }
>  
>  std::string ParaViewDataCollection::GeneratePVDFileName()
>  {
> -   std::string sout = GetCollectionName()+".pvd";
> -   return sout;
> +   return GetCollectionName() + ".pvd";
>  }
>  
>  std::string ParaViewDataCollection::GeneratePVTUFileName()
>  {
> -   std::string sout = "data.pvtu";
> -   return sout;
> +   return "data.pvtu";
>  }
>  
>  std::string ParaViewDataCollection::GenerateVTUFileName()
>  {
> -   std::string sout = "proc" + to_padded_string(myid,pad_digits_rank)+".vtu";
> -   return sout;
> +   return "proc" + to_padded_string(myid,pad_digits_rank) + ".vtu";
>  }
>  std::string ParaViewDataCollection::GenerateVTUFileName(int crank)
>  {
> -   std::string sout = "proc" + to_padded_string(crank,pad_digits_rank)+".vtu";
> -   return sout;
> +   return "proc" + to_padded_string(crank,pad_digits_rank) + ".vtu";
>  }
>  
>  void ParaViewDataCollection::Save()
> ```
> 
> These temporary `out` (or `sout`) variables don't really add any value. Let's just get rid of them.

I don't have strong feelings about this but I felt it was beyond the scope of this PR. In my experience temporaries like these are very helpful when debugging so I didn't want to remove them. Presumably this code performs as the author intended so this type of debugging is no longer necessary.

mlstowell:
> In general I like the idea of avoiding shadowing global and member variables, but I have some concerns:
> 
> 1. `out` and `err` are such common variable names (especially `err` in a finite element code), this becomes an intrusive change.
>
This problem was created when we added `out` and `err` to the `mfem` namespace. Unfortunately, I don't think we can undo that decision. Another option might be to move `mfem::out` and `mfem::err` to a specialized namespace such as `mfem::output::out` (not a great suggestion for a name but you get the idea). I don't know if this would satisfy the original requirement for these global variables but it would solve this issue.

> 2. I'm not convinced by the merits of having `sout`, `pout`, `wout`, `oout` variables. The letter prefix is not so meaningful, and the code is a bit cryptic. If anything, we should settle on a convention, e.g. for any output stream, call it `out_stream` or something.
>
I'm in favor of standardization in theory but it probably won't be enforced as new functions are created. However, if the reviewers can agree on a standard I'll be happy to make the necessary changes. I only hesitate because many output lines will need to be reformatted to stay within 80 characters if we use a longer name for the output stream variable.

> 3. If we are going to make these changes, we should definitely enable `-Wshadow` in our builds, otherwise it's too easy to introduce new shadowing variables and then only catch them later.
>
I completely agree.

tzanio:
@mlstowell, can you please resolve the conflicts?

mlstowell:
The conflicts should be resolved.

Also, I think I have addressed the requested changes including changing `[sout|pout|wout|dout] -> os`. If anyone finds anything I may have missed just let me know.

tzanio (approved):

linalg/hypre.cpp
```diff
@@ -2130,7 +2130,7 @@ void HypreParMatrix::Threshold(double threshold)
 
 void HypreParMatrix::DropSmallEntries(double tol)
 {
-   HYPRE_Int err = 0, old_err = hypre_error_flag;
```
pazner, line 2133: Small nitpick, but I think the following is better:
```diff
diff --git a/linalg/hypre.cpp b/linalg/hypre.cpp
index e23d2417e..729171d27 100644
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -2130,7 +2130,7 @@ void HypreParMatrix::Threshold(double threshold)
 
 void HypreParMatrix::DropSmallEntries(double tol)
 {
-   HYPRE_Int new_err = 0, old_err = hypre_error_flag;
+   HYPRE_Int old_err = hypre_error_flag;
    hypre_error_flag = 0;
 
 #if MFEM_HYPRE_VERSION < 21400
@@ -2161,15 +2161,15 @@ void HypreParMatrix::DropSmallEntries(double tol)
 
 #elif MFEM_HYPRE_VERSION < 21800
 
-   new_err = hypre_ParCSRMatrixDropSmallEntries(A, tol);
+   HYPRE_Int err_flag = hypre_ParCSRMatrixDropSmallEntries(A, tol);
 
 #else
 
-   new_err = hypre_ParCSRMatrixDropSmallEntries(A, tol, 2);
+   HYPRE_Int err_flag = hypre_ParCSRMatrixDropSmallEntries(A, tol, 2);
 
 #endif
 
-   MFEM_VERIFY(!new_err, "error encountered: error code = " << new_err);
+   MFEM_VERIFY(!err_flag, "error encountered: error code = " << err_flag);
 
    hypre_error_flag = old_err;
 }
```

linalg/sparsemat.cpp
```diff
@@ -338,26 +338,26 @@ int SparseMatrix::RowSize(const int i) const
 
 int SparseMatrix::MaxRowSize() const
 {
-   int out=0;
+   int rout=0;
```
pazner, line 341: How about `max_row_size`?

linalg/sparsemat.cpp
```diff
@@ -3718,9 +3718,9 @@ SparseMatrix *RAP(const SparseMatrix &Rt, const SparseMatrix &A,
    SparseMatrix * R = Transpose(Rt);
    SparseMatrix * RA = Mult(*R,A);
    delete R;
-   SparseMatrix * out = Mult(*RA, P);
+   SparseMatrix * mout = Mult(*RA, P);
    delete RA;
-   return out;
+   return mout;
```
pazner, line 3723: Personally I am not a fan of the prefix + `out` pattern. Will the reader understand that `mout` stands for "matrix output" even if all traces of `out` are gone?

Other options could be `rap`, `RAP_matrix`, `result`, `output`, etc.

linalg/constraints.cpp
```diff
@@ -92,50 +93,50 @@ EliminationProjection::EliminationProjection(const Operator& A,
...
-      out.SetSubVector(elim->SecondaryDofs(), 0.0);
+      vout.AddElementVector(elim->PrimaryDofs(), subvec_out);
+      vout.SetSubVector(elim->SecondaryDofs(), 0.0);
    }
 }
 
 SparseMatrix * EliminationProjection::AssembleExact() const
 {
-   SparseMatrix * out = new SparseMatrix(height, width);
+   SparseMatrix * mout = new SparseMatrix(height, width);
```
pazner, line 135: Is there a better name for this variable? What is the projection matrix typically called in the literature?

linalg/hypre.cpp
mlstowell: Sounds good to me. Thanks @pazner !

linalg/sparsemat.cpp
mlstowell: Done

linalg/sparsemat.cpp
mlstowell: The previous function uses `RAP_` so I suggest we reuse that.

linalg/constraints.cpp
mlstowell: Perhaps we can free up the name `mat` by renaming the eliminator specific matrix as `mat_k` where `k` is the loop iterator over eliminators.

pazner (approved):
Looks like all of my concerns have been addressed. Thanks @mlstowell!

tzanio:
@mlstowell and @pazner, can you help me resolve the conflicts, e.g. by merging `master` in this branch?

pazner:
> @mlstowell and @pazner, can you help me resolve the conflicts, e.g. by merging `master` in this branch?

I tried to do it, let's see if it works.

tzanio:
Thanks @pazner !

tzanio:

tzanio:

PR #2733, opened 2021-12-23
Fixing shadowed variables in fem/fe classes [shadow-fe-dev]

mlstowell:

The main difficulty in this PR was the `J` and `Jinv` dense matrices in the `VectorFiniteElement` abstract base class. These were not being used consistently by the derived classes. Most of the newer classes didn't use them at all because these matrices are cached in the `ElementTransformation` object. Rather than adjust the variable names in this case it seemed more logical to also cache the one thing that `Jinv` was still being used for namely the transpose of the adjugate Jacobian.

Hopefully, splitting these into small PRs will assist the original authors and/or reviewers in assessing the changes.

The compiler has no trouble determining which variable is intended but the choice is not always obvious to developers who may not be aware of one or the other of the possibilities.

| [#2733](https://github.com/mfem/mfem/pull/2733) | @mlstowell | @tzanio | @tzanio + @v-dobrev | 1/4/22 | 2/24/22 | 2/25/22 | |


fem/fe/fe_fixed_order.cpp
```diff
@@ -1809,15 +1809,12 @@ void RT0TriangleFiniteElement::Project (
 {
    double vk[2];
    Vector xk (vk, 2);
-#ifdef MFEM_THREAD_SAFE
-   DenseMatrix Jinv(dim);
-#endif
```
tzanio, line 1812: I am not sure about all of these `Jinv` changes in `MFEM_THREAD_SAFE`... @v-dobrev, are they okay?

fem/fe/fe_base.cpp
```diff
@@ -1833,15 +1832,15 @@ void Poly_1D::Basis::Eval(const double y, Vector &u, Vector &d,
    }
 }
 
-void Poly_1D::Basis::EvalIntegrated(const Vector &d_aux, Vector &u) const
+void Poly_1D::Basis::EvalIntegrated(const Vector &_d_aux, Vector &u) const
```
tzanio, line 1835: Rename `_d_aux` -> `d_aux_`?

fem/fe/fe_base.hpp
```diff
@@ -789,7 +789,6 @@ class VectorFiniteElement : public FiniteElement
 protected:
    bool is_nodal;
 #ifndef MFEM_THREAD_SAFE
-   mutable DenseMatrix J, Jinv;
```
tzanio, line 792: This feels wrong... don't we need these in the class for thread safety?

tzanio:

fem/fe/fe_fixed_order.cpp
mlstowell: I may be missing some subtlety but the thought was that we are calculating the transposed adjugate of the Jacobian which is itself cached in the element transformation object. If the cached Jacobian is "thread safe" then the new cached adjugate Jacobian will also be thread safe. If not, then we have a larger problem because we are trying to compute a thread safe object from something that is not thread safe.

tzanio: Ok, this sounds good. I'm just wary of thread safety -- it often requires local variables, that will likely shadow the ones on the class.

tzanio (approved):

fem/fe/fe_base.hpp
mlstowell: I still believe my previous comment regarding `Jinv`, however, you're probably correct about `J`. This is used as a local variable to store the value of `J^T * J` so I was probably wrong to remove it.

I've put `J` back but renamed it `JtJ`. It seems to be used in only one place, `ProjectCulr_ND`, where I've used the usual pattern of declaring a new variable or resizing a member variable based on the thread safety flag.

v-dobrev (approved):

tzanio:

PR #2736, opened 2021-12-24
Fixing shadow variables in general classes [shadow-gen-dev]

mlstowell:

I tried to make minimal changes and keep the spirit of the original variable names. Hopefully, splitting these into small PRs will assist the original authors and/or reviewers in assessing the changes.

The compiler has no trouble determining which variable is intended but the choice is not always obvious to developers who may not be aware of one or the other of the possibilities.

| [#2736](https://github.com/mfem/mfem/pull/2736) | @mlstowell | @tzanio | @tzanio + @camierjs | 12/28/21 | 01/13/22 | 01/17/22 | |

### DIFF
--- a/general/device.cpp
+++ b/general/device.cpp
@@ -177,7 +177,7 @@ Device::~Device()
    Get().device_mem_class = MemoryClass::HOST;
 }
 
-void Device::Configure(const std::string &device, const int dev)
+void Device::Configure(const std::string &device, const int dev_)
 {
    // If a device was configured via the environment, skip the configuration,
    // and avoid the 'singleton_device' to destroy the mm.
@@ -240,7 +240,7 @@ void Device::Configure(const std::string &device, const int dev)
 #endif
 
    // Perform setup.
-   Get().Setup(dev);
+   Get().Setup(dev_);
 
    // Enable the device
    Enable();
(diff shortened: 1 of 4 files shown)


tzanio:

jakubcerveny (commented):
I'm not sure if this is the right fix. I think it's good to use simple names like `dev`, `name`, `size` for unimportant local variables, it makes the code concise and easier to read IMO. However, higher level names with a broader scope, like member variables, not to mention globals, should have less generic and more descriptive names, e.g. `deviceId` instead of `dev`. In other words, the higher level objects should be renamed instead of the locals I think.

mlstowell:
> I'm not sure if this is the right fix. I think it's good to use simple names like `dev`, `name`, `size` for unimportant local variables, it makes the code concise and easier to read IMO. However, higher level names with a broader scope, like member variables, not to mention globals, should have less generic and more descriptive names, e.g. `deviceId` instead of `dev`. In other words, the higher level objects should be renamed instead of the locals I think.

I agree. On a project I worked on earlier all member variables ended in an underscore. This was enormously helpful because there was never any doubt wether you were using a local variable or a member variable. I doubt the team would agree to a similar convention. Without such a convention I'm afraid we're stuck with a hodgepodge of different solutions in different snippets of code.

In this case I felt that this solution would have minimal impact. However, I'll be happy to alter this PR to reflect whatever strategy the reviewers can agree on.

tzanio (approved):

camierjs:
LGTM

camierjs (approved):

tzanio:

PR #2754, opened 2022-01-05
Shifted bug fix

atallah727:
Fixed a bug in sbm_solver.cpp file. I have to reset the temp_elvect vector to zero at the beginning of the quadrature point loop. 
I also fixed a small typo in the comments pertaining to the sign of the penalty term in variational form.

| [#2754](https://github.com/mfem/mfem/pull/2754) | @atallah727 | @tzanio | @vladotomov + @kmittal2 | 01/05/22 | 01/14/22 | 01/17/22 | |

### DIFF
--- a/miniapps/shifted/CMakeLists.txt
+++ b/miniapps/shifted/CMakeLists.txt
@@ -17,6 +17,7 @@ if (MFEM_USE_MPI)
   list(APPEND DIST_COMMON_HEADERS
         dist_solver.hpp
         sbm_solver.hpp
+	sbm_aux.hpp
         marking.hpp)
 
 convert_filenames_to_full_paths(DIST_COMMON_SOURCES)
--- a/miniapps/shifted/makefile
+++ b/miniapps/shifted/makefile
@@ -53,15 +53,15 @@ COMMON_LIB += $(if $(MFEM_SHARED:YES=),,\
 %: %.cpp
 %.o: %.cpp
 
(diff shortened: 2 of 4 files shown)


tzanio:

vladotomov (approved):
Thanks Nabil, good catch.

kmittal2:
Hi @atallah727, 
Earlier we were calculating the integral in `temp_elvect` and setting `elvect = temp_elvect` at the end (I was accidentally doing it inside the loop over quadrature points instead of outside.. it should not change the final `elvect` but I was doing more FLOPS than needed). 
What we are doing now is that we are setting `temp_elvect = 0` for each quadrature point and adding the contribution as `elvect += temp_elvect` inside the loop. 
Seems like the final result should be the same in both cases? I did a quick check but let me know what you find out as well.
If you agree, I think it might be more computationally efficient to keep the original approach and move the `elvect = temp_elvect` part outside the loop over quadrature points. 
p.s. We should fix this in the BilinearFormIntegrator as well (including the ones for Neumann boundary condition). I can help you with that once we agree on which approach to use.

atallah727:
Hello Ketan,

We can talk tomorrow to discuss your proposition. Let me know when would be a good time.

Thank you.

Best

Nabil
________________________________

kmittal2:
Had offline discussions with @atallah727 and made changes accordingly.
@vladotomov, can you please take a look at the PR again?

kmittal2 (approved):

vladotomov (approved):

tzanio:

tzanio (approved):
Thanks for the fix @atallah727

PR #2781, opened 2022-01-23
Fix hash table resize on NCMesh load [hash-resize-fix]

jakubcerveny:
The `HashTable` normally resizes itself when it becomes overfull, but this was missing in the special case of the `Alloc()` method, which is used when loading an NC mesh.

The fix consists of adding the line `CheckRehash()` in `HashTable<T>::Alloc`.

Also added a debugging method to dump a histogram if bin sizes.

### DIFF
--- a/general/hash.hpp
+++ b/general/hash.hpp
@@ -591,6 +591,7 @@ void HashTable<T>::Alloc(int id, int p1, int p2)
       item.p2 = p2;
 
       Insert(Hash(p1, p2), id, item);
+      CheckRehash();
    }
 }
 
--- a/general/hash.hpp
+++ b/general/hash.hpp
@@ -136,6 +136,9 @@ public:
    /// Write details of the memory usage to the mfem output stream.
    void PrintMemoryDetail() const;
 
+   /// Print hash table statistic for debugging purposes.
+   void PrintStats() const;
+
    class iterator : public Base::iterator
    {
    protected:
@@ -210,6 +213,8 @@ protected:
    /// Check table load factor and resize if necessary
    inline void CheckRehash();
    void DoRehash();
+
+   int BinSize(int id) const;
 };
 
 
@@ -648,6 +653,47 @@ void HashTable<T>::PrintMemoryDetail() const
 {
    mfem::out << Base::MemoryUsage() << " + " << (mask+1) * sizeof(int)
              << " + " << unused.MemoryUsage();
+}
+
+template<typename T>
+int HashTable<T>::BinSize(int id) const
+{
+   int count = 0;
+   while (id >= 0)
+   {
+      const T& item = Base::At(id);
+      id = item.next;
+      count++;
+   }
+   return count;
+}
+
+template<typename T>
+void HashTable<T>::PrintStats() const
+{
+   int table_size = mask+1;
+   mfem::out << "Hash table size: " << table_size << "\n";
+   mfem::out << "Item count: " << Size() << "\n";
+   mfem::out << "BlockArray size: " << Base::Size() << "\n";
+
+   const int H = 16;
+   int hist[H];
+
+   for (int i = 0; i < H; i++) { hist[i] = 0; }
+
+   for (int i = 0; i < table_size; i++)
+   {
+      int bs = BinSize(i);
+      if (bs >= H) { bs = H-1; }
+      hist[bs]++;
+   }
+
+   mfem::out << "Bin size histogram:\n";
+   for (int i = 0; i < H; i++)
+   {
+      mfem::out << "  size " << i << ": "
+                << hist[i] << " bins" << std::endl;
+   }
 }
 
 


tzanio:

tzanio (approved):

Maatinski (approved):
This is awesome! Thanks! This bug-fix makes a huge difference; some quick tests showed a speed-up of a factor of 13 for a rather big mesh. I posted more details in https://github.com/mfem/mfem/issues/2759

general/hash.hpp
```diff
@@ -136,6 +136,9 @@ class HashTable : public BlockArray<T>
    /// Write details of the memory usage to the mfem output stream.
    void PrintMemoryDetail() const;
 
+   /// Print hash table statistic for debugging purposes.
```
YohannDudouit, line 139: This documentation is not clear, what are these statistics?

general/hash.hpp
```diff
@@ -210,6 +213,8 @@ class HashTable : public BlockArray<T>
    /// Check table load factor and resize if necessary
    inline void CheckRehash();
    void DoRehash();
+
+   int BinSize(int id) const;
```
YohannDudouit, line 217: Documentation please.

general/hash.hpp
```diff
@@ -591,6 +596,7 @@ void HashTable<T>::Alloc(int id, int p1, int p2)
       item.p2 = p2;
 
       Insert(Hash(p1, p2), id, item);
+      CheckRehash();
```
YohannDudouit, line 599: Shouldn't this line rather be inside the `Insert` method?

general/hash.hpp
jakubcerveny: This is not a public method and besides that, what else would the docs say, other than "Returns the size of bin 'id'"?

general/hash.hpp
jakubcerveny: No, Insert is needed elsewhere without the rehash.

general/hash.hpp
jakubcerveny: Improved.

general/hash.hpp
YohannDudouit: From the [developer guideline](https://github.com/mfem/mfem/blob/master/CONTRIBUTING.md#developer-guidelines):
- All new public, protected, and private classes, methods, data members, and functions have Doxygen-style documentation in source comments.
- In addition to arguments and functionality, documentation should include the current limitations of the code, any background information that is implicitly assumed in the implementation, and the ownership and lifetime of data."

general/hash.hpp
YohannDudouit: Can you describe in the documentation why, and also add documentation to `Insert`, `Hash`, and `CheckRehash`?

general/hash.hpp
jakubcerveny: OK...

general/hash.hpp
jakubcerveny: Sorry, no.

jakubcerveny: Sure: Insert simply adds an item into a bin, it shouldn't do anything else. Also it is used in situations where an item is deleted and reinserted elsewhere (see Reparent), so enlarging the hash table is not needed in that case.

general/hash.hpp
jakubcerveny: I believe that rule is fairly new in the guidelines. I'm all for useful comments. However, a formalistic rule that absolutely every method needs documentation does not guarantee good documentation. You will get a lot of useless comments like: 
```
/// Constructor.
ClassName() { ... }

/// Destructor.
~ClassName() { ... }

/// Returns element size.
GetElementSize() { ... }
```
Regarding other comments, please understand that I don't mean to be rude, I'm just trying to spend as little time on this as possible, since this was supposed to be a one-line bug fix, and also because I'm just a volunteer at the moment.

YohannDudouit: I understand your point, though I don't think it applies to this case since I still have no idea what "bin" is referring to, even after reading the code. @tzanio since you agree with @jakubcerveny , do you want to write the documentation for him?

jakubcerveny: Yohann, I don't think we need to document fundamental concepts of computer science. Please refer to https://en.wikipedia.org/wiki/Hash_table for a refresher on hash tables (they do call it "buckets" here, but other sources use "bin" for the same thing).

general/hash.hpp
jakubcerveny: No, the logic is "do one thing", "don't mix functionality". Also rehash is expensive, it simply doesn't belong in Insert. And it's not the interface, it's a private method.

general/hash.hpp
jakubcerveny: Yes, it was a bug, but I did retest the experiment and the results still hold.

general/hash.hpp
YohannDudouit: If I understand correctly the code in `CheckRehash`, it only rehashes when needed, but it's difficult to tell with this undocumented `mask`. Also how comes you `CheckRehash` after you `Insert`, this seems counter-intuitive to me, I would expect an algorithm to check if there is space before inserting, not after.

The idea to put `CheckRehash` inside `Insert` is actually to have `Insert` always do one thing: insert. Currently `Insert` is doing two things: inserting when there is space, creating a bug when there isn't enough space (the reason for this PR).
If you want the unsafe capability I think it would be better to consider a design like the following in the future:
```c++
/// @brief Inserts without checking if there is the necessary space.
template<typename T>
inline void HashTable<T>::UnsafeInsert(int idx, int id, T &item)
{
   // add item at the beginning of the linked list
   item.next = table[idx];
   table[idx] = id;
}

template<typename T>
inline void HashTable<T>::Insert(int idx, int id, T &item)
{
   CheckRehash();
   UnsafeInsert(idx, id, item);
}
```

acfisher: I threw a comment in there to clarify the Insert method for future developers.

YohannDudouit: @acfisher @jakubcerveny @tzanio I pushed documentation for the `HashTable` class, tell me if you think it is clear.

YohannDudouit (approved):

tzanio:

tzanio (approved):
Thank you for this important fix @jakubcerveny and for the nice documentation @YohannDudouit and @acfisher -- it looks great! 👍

Issue #2779, opened 2022-01-20
Bug in GetHpRestrictionMatrix


dylan-copeland:
The minimal example in this Gist produces the attached plot. The example does the following:

1. Make a serial mesh with 2 elements.
2. h-refine element 1.
3. Create a `FiniteElementSpace` with order 2, then increase order to 3 just on element 0.
4. Project (locally interpolate) the constant coefficient 1 to a `GridFunction`.
5. Unsuccessfully enforce the conforming constraints to the interpolated `GridFunction` by multiplying by `GetHpRestrictionMatrix` and `GetProlongationMatrix`.
6. Visualize the result after prolongating to maximum order (for glvis, which cannot handle variable order).

For orders 1 and 2, the result is exactly 1 everywhere (correct), but for orders 2 and 3 (defaults in the Gist example), the attached plot shows the result is wrong. Also, if the multiplication by `GetHpRestrictionMatrix` and `GetProlongationMatrix` is omitted, the plot looks correct (even for orders 2 and 3). However, it seems necessary in general to enforce constraints in this way.

The example also prints out `GetHpRestrictionMatrix`. In particular, there is 
[row 12]
 (13,1)
meaning that true DOF 12 is set to full DOF 13. I suspect this is wrong, because DOF 13 does not belong to any element. There is also some code in the example to find the missing DOFs from `fespace.GetElementToDofTable()`, and it prints

DOF missing: 13

In this example, the projected `GridFunction x` has a 0 in entry 13, as that DOF is not set during projection (not belonging to any element). Multiplying by `R` produces the true DOF vector `y` with 0 in entry 12.

The comments for `GetHpRestrictionMatrix` explain that this matrix takes care of the hp case where some DOFs do not belong to any element, but it seems wrong to have an entry for a DOF that does not belong to any element, and I suppose this causes the incorrect enforcement of the constraints, using uninitialized values.
<img width="412" alt="proj" src="https://user-images.githubusercontent.com/12700975/150390197-edcd30c5-3bb4-4eb1-95e1-c58d4a6e697c.png">

https://gist.github.com/dylan-copeland/085e98184e029f3fbfb4bf6200589171

dylan-copeland:
https://github.com/mfem/mfem/blob/bba2c08025c0700ead5965d27d7ed0f8564a0018/fem/fespace.cpp#L899

[...]

PR #2787, opened 2022-01-25
Bug fix for hp restriction matrix

dylan-copeland:
This PR fixes the bug described in https://github.com/mfem/mfem/issues/2779.

| [#2787](https://github.com/mfem/mfem/pull/2787) | @dylan-copeland | @tzanio | @jakubcerveny + @psocratis | 01/26/22 | 01/27/22 | 01/30/22 | |

### DIFF
--- a/fem/fespace.cpp
+++ b/fem/fespace.cpp
@@ -894,9 +894,10 @@ int FiniteElementSpace::GetDegenerateFaceDofs(int index, Array<int> &dofs,
 int FiniteElementSpace::GetNumBorderDofs(Geometry::Type geom, int order) const
 {
    // return the number of vertex and edge DOFs that precede inner DOFs
-   int nv = fec->GetNumDof(Geometry::POINT, order);
-   int ne = fec->GetNumDof(Geometry::SEGMENT, order);
-   return Geometry::NumVerts[geom] * (nv + ne);
+   const int nv = fec->GetNumDof(Geometry::POINT, order);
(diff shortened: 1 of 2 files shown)


psocratis:
I can confirm that with this all the unit tests for the variable-order  true-transfer operator in the case of strict conformity are successful!

I do have one question here, maybe for @jakubcerveny. Shouldn't `GetConfomingRestriction()` always return the `hpRestriction` for a variable order `FiniteElementSpace`. In other words, will `R` (and not Rhp) be useful anywhere in the  code in this case?

jakubcerveny:
Thanks for fixing this! 

As for the restriction matrix, my hopes were initially also that R would simply be R_hp in the case of variable orders, but it turns out the classic R matrix is still needed for essential DOF identification/elimination, which gets spoiled by the extra elements when the R_hp matrix is treated as a Boolean matrix :-|

jakubcerveny (approved):

tzanio:
Thanks @dylan-copeland !

Can we mark it as `ready-for-review`?

psocratis:
> As for the restriction matrix, my hopes were initially also that R would simply be R_hp in the case of variable orders, but it turns out the classic R matrix is still needed for essential DOF identification/elimination, which gets spoiled by the extra elements when the R_hp matrix is treated as a Boolean matrix :-|

Thanks @jakubcerveny

tzanio:

psocratis (approved):

tzanio:

PR #2797, opened 2022-02-01
Second proposal for a human readable face information

YohannDudouit:
Second proposal attempting to represent the face information as simply as possible.
The first `enum` type describes the topology of the mesh's face:
```c++
   enum class FaceTopology { Boundary,
                             Conforming,
                             NonConforming,
                             NA
                           };
```
The second `enum` type describes the location of the two elements sharing the mesh's face:
```c++
   enum class ElementLocation { Local, Shared, NA };
```
The third `enum` type describes the topology of the element's face relative to the mesh's face:
```c++
   enum class ElementConformity { Conforming, Coarse, Fine, NA };
```
The fourth `enum` type describes the `FaceInfo` encryption case, and is used to cast from `FaceInformation` to `FaceInfo`: (can be removed if judged confusing)
```c++
   enum class FaceInfoTag { LocalConformingInterior,
                            LocalNonConformingInterior,
                            TrueBoundary,
                            SharedConformingInterior,
                            MasterNonConforming,
                            SlaveNonConforming,
                            GhostMaster,
                            GhostSlave,
                            Invalid
                          };
```
I dropped the `FineCoarse` and `CoarseFine` in the methods' name as I find them confusing: I think non-conforming coarse refers to the fact the mesh's face contains several refined element's faces; whereas non-conforming fine refers to the fact the mesh's face correspond exactly to one refined element's face. (Please chime in about this, I would prefer to find something clear and transparent to everyone)


YohannDudouit:
@pazner I know you don't like the terminology coarse/fine for non-conforming faces... I think coarse-fine/fine-coarse is inappropriate since it's not describing the right information... If you have a better proposition...

Also, most of the methods in `FaceInformation` are now unused, so it may make sense to remove the ones that conflates properties, e.g., `IsLocalNonConformingFine`, `IsSharedNonConformingCoarse`, etc...

pazner:
Very minor point, I think we should prefer `Nonconforming` over `NonConforming` (more consistent with e.g. `Mesh::Nonconforming()`)

For `ElementLocation`
```c++
   enum class ElementLocation { Local, Shared, NA };
```
`Shared` should probably be replaced by `Neighbor` or `FaceNbr` or something (the element is not "shared" in the MFEM terminology, it's what is referred to elsewhere as a "face neighbor element").

For `ElementConformity`
```c++
   enum class ElementConformity { Conforming, Coarse, Fine, NA };
```
I think it's a bit confusing. I don't think "conforming" is a good way to describe the situation where the "mesh face" and the "element face" coincide… it still is a nonconforming interface, and I think attaching the word conforming to it will be confusing (even though I do understand your logic). Also, for the case of `Fine` it's a little weird, because "element 2" is not well-defined in that case.

Another way of talking about these would be something like `Coincident`, `Subset`, `Superset`. Not sure this is any better.

Finally, for `FaceInfoTag`, my personal opinion is that we shouldn't use the "master" terminology, we should switch in a unified way to the coarse/fine terminology (or whatever else we settle on), rather than keep both naming schemes with a "dictionary" that translates in between them.

YohannDudouit:
> Very minor point, I think we should prefer `Nonconforming` over `NonConforming` (more consistent with e.g. `Mesh::Nonconforming()`)
> 
> For `ElementLocation`
> 
> ```c++
>    enum class ElementLocation { Local, Shared, NA };
> ```
> 
> `Shared` should probably be replaced by `Neighbor` or `FaceNbr` or something (the element is not "shared" in the MFEM terminology, it's what is referred to elsewhere as a "face neighbor element").

Not a big fan of this, but if you think this is more clear, why not.
> 
> For `ElementConformity`
> 
> ```c++
>    enum class ElementConformity { Conforming, Coarse, Fine, NA };
> ```
> 
> I think it's a bit confusing. I don't think "conforming" is a good way to describe the situation where the "mesh face" and the "element face" coincide… it still is a nonconforming interface, and I think attaching the word conforming to it will be confusing (even though I do understand your logic). Also, for the case of `Fine` it's a little weird, because "element 2" is not well-defined in that case.

I agree that it is confusing to use conforming for the element's face conformity to the mesh's face, though this is actually what is happening topologically...
> 
> Another way of talking about these would be something like `Coincident`, `Subset`, `Superset`. Not sure that is any better.

This may be more accurate, I'm just worried that it may be less clear, but it may make the overall design more clear...
> 
> Finally, for `FaceInfoTag`, my personal opinion is that we shouldn't use the "master" terminology, we should switch in a unified way to the coarse/fine terminology (or whatever else we settle on), rather than keep both naming schemes with a "dictionary" that translates in between them.

Regarding the `FaceInfoTag` my reasoning was that we want to reflect the terminology used in `FaceInfo`. Are you suggesting that we should replace the terminology `slave`,  `master`, and `ghost` everywhere in the code, deprecate functions and replace them with the new terminology?

pazner:
> > `Shared` should probably be replaced by `Neighbor` or `FaceNbr` or something (the element is not "shared" in the MFEM terminology, it's what is referred to elsewhere as a "face neighbor element").
> 
> Not a big fan of this, but if you think this is more clear, why not.

I just think the consistency is worth it — we should stick with one set of terminology rather than mixing.

> > Another way of talking about these would be something like `Coincident`, `Subset`, `Superset`. Not sure that is any better.
> 
> This may be more accurate, I'm just worried that it may be less clear, but it may make the overall design more clear...

Yeah, I agree it's not great, let's see if others have better ideas.

> > Finally, for `FaceInfoTag`, my personal opinion is that we shouldn't use the "master" terminology, we should switch in a unified way to the coarse/fine terminology (or whatever else we settle on), rather than keep both naming schemes with a "dictionary" that translates in between them.
> 
> Regarding the `FaceInfoTag` my reasoning was that we want to reflect the terminology used in `FaceInfo`. Are you suggesting that we should replace the terminology `slave`, `master`, and `ghost` everywhere in the code, deprecate functions and replace them with the new terminology?

Ideally, the "user-facing interface" should use the new terminology. I don't know if this is feasible though, it might be too many changes. Changing it everywhere in the internals will definitely be too much work.

samuelpmishLLNL:
It seems like the enums are not "orthogonal" in the sense that properties like conformity are repeated in multiple places: `FaceTopology` (which, I don't really think of conformity as a topological property), `ElementConformity`, and `FaceInfoTag`.

To me, the main ideas are:
1. Is this face on the inside of the region or on the boundary? `FaceTopology` essentially answers this, but some of the information is communicated _implicitly_ rather than explicitly. That is, `FaceTopology::Conforming` _implies_ that the face is interior, rather than stating it directly. Something like `enum class FaceTopology { Boundary, Interior, NA }` seems truer to the mathematics, and separates concerns (treating conformity and topology independently).
2. Is the discretization of this face conforming or not? This is a yes or no question, so it could be a `bool` rather than an enum. I'm not sure how to interpret `enum class ElementConformity { Coincident, Superset, Subset, NA };`. Does `face.elem_1_conformity == ElementConformity::Superset` imply that element 1's side is a superset of the face, or the other way around?
3. How do I access information about the elements that claim this face? This design does a good job providing this information in the members (although tautological comments bother me!)
```cpp
int elem_1_index, elem_2_index; // Elem1 and Elem2 indices
int elem_1_local_face, elem_2_local_face; // Elem1 and Elem2 local faces
int elem_1_orientation, elem_2_orientation; // Elem1 and Elem2 orientations
```

Together, I think something like the code below is a little bit more explicit:

```cpp
enum class FaceTopology { Boundary, PeriodicBoundary, Interior, NA };

enum class ElementLocation : uint64_t { Local, Shared, Ghost, NA };

struct FaceInformation {

    // where in the mesh does this face appear?
    FaceTopology topology; 

    // is there a refinement mismatch?
    // if so, we'll adopt the convention of putting the "fine" face first in the array below
    bool is_conforming;

    // upward connectivities, packed to fit in a 64-bit word
    struct {
        uint64_t index : 55;          // index of the parent element
        uint64_t local_face : 4;      // which of the parent's faces is it
        uint64_t orientation : 3;     // does the face need to be reoriented
        ElementLocation location : 2; // is this information on my processor
    } element[2]; 

    int ncface_index; // used to access additional info in NCMesh (?)

    const DenseMatrix* point_matrix;

};
```

YohannDudouit:
@samuelpmishLLNL We cannot have the convention that elem1 is the one that contains the "fine" face, there is already the convention that elem1 is always "local". Therefore you need to explicitly give this information, and thus the problem of finding appropriate terminology. Also you have to be careful when swapping elem1 and elem2 as this flips the direction of the normal (MFEM convention is that the normal goes from elem1 to elem2), so for instance you would need to know which faces have been flipped to get the correct direction of their normal.

However, exposing the concept of Ghost to the users may be important, not because the enums don't describe these faces correctly, but because specific methods should be called when the face is "Ghost", for instance `ParMesh::GetGhostFaceTransformation` should be called instead of `GetFaceTransformation`. Though, I think this was a design mistake, and would favor having `GetFaceTransformation` do the right thing in all cases without exposing the concept of ghost faces, like this for instance:
```c++
void ParMesh::GetFaceTransformation(...)
{
   if( is_ghost_face ) // Currently a user has to guess this...
   {
      GetGhostFaceTransformation(...);
   } else {
      GetLocalFaceTransformation(...);
   }
}
```
The reason why I don't really want to expose the "ghost" concept is also because, even if it is reasonably defined in `ParNCMesh`, this definition does not transpose well to its use in the `Mesh`. Indeed, in `ParNCMesh` there is a standard design with a ghost layer, independently of conforming or non-conforming faces. Whereas, in the `Mesh`, ghost is exclusive to faces that couldn't be represented by previously defined non-ghost faces. As a result ghost faces are only a subset of the shared non-conforming faces; some shared non-conforming faces being ghost other not. I think this touches implementation details that we really don't want to have to explain in a high-level interface. As a first step I can maybe add `MFEM_ASSERT`s that check that correct methods have been called based on the face index.

YohannDudouit:
@pazner @psocratis @dylan-copeland @samuelpmishLLNL @v-dobrev @tzanio I'm mostly happy about the current content of this PR, I've tried to take everyone's opinion into account, I am personally not a fan of:
```c++
enum class ElementConformity { Coincident, Superset, Subset, NA };
```
as I feel it is using a terminology that we never use verbally and will never use, it may be more accurate than "coarse" and "fine", but I am not convinced it is more clear.

The remaining points to agree on are:
- Approval for enum naming conventions,
- Approval for methods/methods naming (most methods are currently unused),
- Keep or drop `FaceInfoTag` and cast operator from `FaceInformation` to `FaceInfo` (We could add this when the need arises).

dylan-copeland (approved):
Thanks @YohannDudouit, this looks good to me. As for `ElementConformity { Coincident, Superset, Subset, NA }`, the new terminology is not what we typically use, but I think it is more precise, and I would be fine with anything that is reasonable and documented. The burden of learning these new terms is not bad, compared to potential confusion that they prevent. I also agree with @pazner about consistent use of `Nonconforming` or `nonconforming`. The code currently has a mixture of `NonConforming`, `Nonconforming`, and `non-conforming` (latter in comments). The English dictionary contains `nonconforming`, so there is no need for the hyphen or the capital `C`. Consistent usage of `{N,n}onconforming` will make it easier to search the code.

YohannDudouit:
I am not against replacing "non-conforming" with "nonconforming", but this is used a bit everywhere in MFEM.

pazner (approved):
This looks like a good change to me! I'm not completely sure about the `FaceInfoTag`, @v-dobrev, do you think this is beneficial (explicitly listing the possible states and converting between them)?

Otherwise I think it's ready to go.

tzanio (approved):

PR #2804, opened 2022-02-02
Fix for HypreParMatrixFromBlocks for square matrices

sebastiangrimberg:
If using `HypreParMatrixFromBlocks` to form a square matrix, we should call the `HypreParMatrix` constructor with `row_starts = col_starts` so that `hypre_CSRMatrixReorder` gets properly called. Without this, methods like `HypreParMatrix::GetDiag` will fail on the merged matrix.

| [#2804](https://github.com/mfem/mfem/pull/2804) | @sebastiangrimberg | @pazner | @dylan-copeland + @jandrej | 2/4/22 | 3/22/22 | ⌛due 2/25/22 | |

### DIFF
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -2976,10 +2976,8 @@ HypreParMatrix * HypreParMatrixFromBlocks(Array2D<HypreParMatrix*> &blocks,
    rowStarts2[0] = first_loc_row;
    rowStarts2[1] = first_loc_row + all_num_loc_rows[rank];
 
-   int square = (first_loc_row == first_loc_col &&
-                 all_num_loc_rows[rank] == all_num_loc_cols[rank]);
-   MPI_Allreduce(MPI_IN_PLACE, &square, 1, MPI_INT, MPI_MAX, comm);
-
+   int square = std::equal(all_num_loc_rows.begin(), all_num_loc_rows.end(),
+                           all_num_loc_cols.begin());
    if (square)
    {
       return new HypreParMatrix(comm, num_loc_rows, glob_nrows, glob_ncols,
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -2976,8 +2976,11 @@ HypreParMatrix * HypreParMatrixFromBlocks(Array2D<HypreParMatrix*> &blocks,
    rowStarts2[0] = first_loc_row;
    rowStarts2[1] = first_loc_row + all_num_loc_rows[rank];
 
-   if (first_loc_row == first_loc_col &&
-       all_num_loc_rows[rank] == all_num_loc_cols[rank])
+   int square = (first_loc_row == first_loc_col &&
+                 all_num_loc_rows[rank] == all_num_loc_cols[rank]);
+   MPI_Allreduce(MPI_IN_PLACE, &square, 1, MPI_INT, MPI_MAX, comm);
+
+   if (square)
    {
       return new HypreParMatrix(comm, num_loc_rows, glob_nrows, glob_ncols,
                                 opI.data(), opJ.data(),
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -2969,22 +2969,34 @@ HypreParMatrix * HypreParMatrixFromBlocks(Array2D<HypreParMatrix*> &blocks,
       }
    }
 
+   MFEM_VERIFY(HYPRE_AssumedPartitionCheck(),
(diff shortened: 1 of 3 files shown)


pazner:

linalg/hypre.cpp
```diff
@@ -2969,22 +2969,34 @@ HypreParMatrix * HypreParMatrixFromBlocks(Array2D<HypreParMatrix*> &blocks,
...
    rowStarts2[1] = first_loc_row + all_num_loc_rows[rank];
 
-   std::vector<HYPRE_BigInt> colStarts2(2);
-   colStarts2[0] = first_loc_col;
-   colStarts2[1] = first_loc_col + all_num_loc_cols[rank];
-
-   MFEM_VERIFY(HYPRE_AssumedPartitionCheck(),
-               "only 'assumed partition' mode is supported");
+   if (first_loc_row == first_loc_col &&
+       all_num_loc_rows[rank] == all_num_loc_cols[rank])
```
dylan-copeland, line 2980: It seems very unlikely, but possible, that a rectangular matrix could have `all_num_loc_rows[rank] == all_num_loc_cols[rank]` for a particular rank but not for other ranks. Would it be better to loop over all ranks and check this condition, to set a flag `bool square`?

dylan-copeland:
There was a failing test, `mpirun -np 2 ex0p`, on macos-10.15-make-opt-par-int32, with a seg fault. The test works on my mac. @pazner could this have been an issue with the testing process, so we should restart the tests?

pazner:
> There was a failing test, `mpirun -np 2 ex0p`, on macos-10.15-make-opt-par-int32, with a seg fault. The test works on my mac. @pazner could this have been an issue with the testing process, so we should restart the tests?

I restarted the tests and they are passing now, hopefully it was some kind of CI glitch.

linalg/hypre.cpp
sebastiangrimberg: That's right, I guess my point was just that even if the global matrix isn't square, doing the diagonal reordering locally doesn't seem like it would cause any problem and avoids a global communication to check that every rank is square.

dylan-copeland: That may be a valid point about the local diagonal reordering, but another issue is that passing the same pointer for `rows` and `cols` to the 9-argument constructor will change the behavior of the constructor. These arrays are already of size equal to the number of ranks, and the communication is already done by `GatherBlockOffsetData`, so the cost to check all ranks is negligible. More important is to prevent going into the square case of the constructor on some ranks and the rectangular case on other ranks, which seems sure to cause problems.

sebastiangrimberg: Yeah, that sounds fair enough. I’ll add the check in ASAP.

linalg/hypre.cpp
```diff
@@ -2969,22 +2969,37 @@ HypreParMatrix * HypreParMatrixFromBlocks(Array2D<HypreParMatrix*> &blocks,
...
    std::vector<HYPRE_BigInt> rowStarts2(2);
    rowStarts2[0] = first_loc_row;
    rowStarts2[1] = first_loc_row + all_num_loc_rows[rank];
 
-   std::vector<HYPRE_BigInt> colStarts2(2);
-   colStarts2[0] = first_loc_col;
-   colStarts2[1] = first_loc_col + all_num_loc_cols[rank];
+   int square = (first_loc_row == first_loc_col &&
+                 all_num_loc_rows[rank] == all_num_loc_cols[rank]);
+   MPI_Allreduce(MPI_IN_PLACE, &square, 1, MPI_INT, MPI_MAX, comm);
```
dylan-copeland, line 2981: This looks correct, but I think you can avoid the `MPI_Allreduce` since the sizes are already gathered by `GatherBlockOffsetData` into `all_num_loc_rows` and `all_num_loc_cols`, which are arrays with sizes for all ranks. Also, `first_loc_row` and `first_loc_col` are computed from these arrays, so there is no need to compare them. You can just loop over all ranks `r` and check `all_num_loc_rows[r] == all_num_loc_cols[r]`.

sebastiangrimberg: Yup, duh. Should be fixed now.

dylan-copeland (approved):
Thanks for the fix, @sebastiangrimberg!

pazner:
@jandrej, could you take a look when you get a moment?

pazner:
@jandrej, could you take a look when you have the time? It would be good to merge this once it's approved.

jandrej (approved):

pazner:

PR #2814, opened 2022-02-08
Nonlinear elasticity example with PA and AD

jandrej:
This PR adds a nonlinear example that showcases MFEM's interface to the [Fundamental finite element operator decomposition](https://www.researchgate.net/figure/Fundamental-finite-element-operator-decomposition-This-algebraically-factored-form-is-a_fig2_333339156). In this specific case, it implements a matrix-free method to solve linear and nonlinear elasticity with user selected (or defined) material models. It showcases

* Catering bits and pieces of MFEM's tools to a specific problem. Contrary to existing examples trying a one size fits all approach.
* Providing automatic differentiation for functions on the quadrature point with
  - Dual numbers
  - Leveraging Enzyme through the LLVM intermediate representation
  - Finite differences
* Simple matrix free preconditioners
  - Jacobi
  - Block Jacobi (inverted element blocks on the diagonal)
* It shows very clearly how to manually apply essential boundary conditions in a matrix free setting and being consistent with it in the preconditioned application. Even in a nonlinear setting.

Internally (_not_ user facing) the PR adds a static tensor implementation from [Serac](https://github.com/LLNL/serac) written by @samuelpmishLLNL. This is accompanied by a dual number implementation.

- [x] Commit history has to be rewired properly in the end
- [x] build system flags for enzyme

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#2814](https://github.com/mfem/mfem/pull/2814) | @jandrej | @tzanio | @pazner + @jamiebramwell + @bslazarov | 3/22/22 | 6/28/22 | 7/5/22 | |

<details>
<summary>PR Checklist</summary>
    
- [x] Code builds.
- [x] Code passes `make style`.
- [ ] Update `CHANGELOG`:
- [x] Update `.gitignore`:
- [ ] New capability:
   - [ ] All new public, protected, and private classes, methods, data members, and functions have full Doxygen-style documentation in source comments. Documentation should include descriptions of member data, function arguments and return values, template parameters, and prerequisites for calling new functions.
   - [ ] Pointer arguments and return values must specify whether ownership is being transferred or lent with the call.
   - [ ] Any new functions should include descriptions of their intended use e.g. for internal use only, user-facing, etc., along with references to example code whenever possible/appropriate.
   - [ ] Consider adding new sample runs in existing examples to highlight the new capability.
[...]

PR #2820, opened 2022-02-09
Proof of concept that #2814 can use `StaticTensor`.

YohannDudouit:
This PR tries to be a proof of concept that #2814 can use `StaticTensor` with minimal changes.


YohannDudouit:
@jandrej @samuelpmishLLNL You may be interested to take a quick look at the difference in `ex42.cpp` using my `StaticTensor` class.

PR #2836, opened 2022-02-17
Make SymmetricMatrixCoefficient inherit from MatrixCoefficient

pazner:
A `SymmetricMatrixCoefficient` is conceptually a special type of `MatrixCoefficient`, and it could be useful to use a `SymmetricMatrixCoefficient` in most places where `MatrixCoefficient` is being used.

We shouldn't require writing new code to opt-in to allowing `SymmetricMatrixCoefficient` (though if they want to optimize for lower memory requirements, they should be able to opt-in to these optimizations). Ideally, we shouldn't need to create special constructors to just accept symmetric coefficients.

For example, `SymmetricMatrixCoefficient` is allowed for some integrators, and will work with partial assembly enabled, but will cause a runtime failure when the same integrators are used with partial assembly disabled.

This PR makes `SymmetricMatrixCoefficient` inherit from `MatrixCoefficient`. Consequently, it can be used anywhere where `MatrixCoefficient` can be used, and should immediately work with all integrators. Opting in to optimizations that take advantage of symmetry is possible using `dynamic_cast`.

| [#2836](https://github.com/mfem/mfem/pull/2836) | @pazner | @mlstowell | @dylan-copeland + @mlstowell | 2/18/22 | 3/4/22 | 4/5/22 | |


mlstowell:

tests/unit/fem/test_assemblediagonalpa.cpp
```diff
@@ -209,10 +208,8 @@ TEST_CASE("diffusiondiag")
                }
                else if (coeffType == 3)
                {
-                  mcoeff = new MatrixFunctionCoefficient(dimension,
```
dylan-copeland, line 212: It looks like you removed the testing of `fullSymmetricMatrixCoeffFunction`, which is not really necessary. In that case, the function `fullSymmetricMatrixCoeffFunction` could be removed.

pazner: Good point, yes they are not used now because the legacy integrator can directly use the symmetric coefficients.

I removed those functions now.

dylan-copeland (approved):

pazner:
@mlstowell, can you take a look when you get a chance?

mlstowell (approved):

fem/bilininteg_hcurl.cpp
```diff
@@ -1002,43 +1004,43 @@ void CurlCurlIntegrator::AssemblePA(const FiniteElementSpace &fes)
...
-         MFEM_VERIFY(MQ->GetHeight() == dimc && MQ->GetWidth() == dimc, "");
-      }
       if (SMQ)
       {
          SM.SetSize(dimc);
          MFEM_VERIFY(SMQ->GetSize() == dimc, "");
       }
+      else if (MQ)
+      {
+         M.SetSize(dimc);
```
mlstowell, line 1014: @pazner, it looks like we introduced a problem here. The `DenseMatrix` is named `GM` rather than `M` so this fails to compile. I'm not sure how we could have missed this. Perhaps a merge chose the wrong name...

pazner: There is a member variable `M` that exists whenever `MFEM_THREAD_SAFE` is not defined.

Yes, this was caused by a merge with the shadow variables PR.

PR #2837, opened 2022-02-17
Reduce some unit test runtimes

pazner:
Reduces the runtimes of some of the longer unit tests:

* In "Assembly Levels" there were some completely redundant tests. I also got rid of order 4 in 2D.
    * Runtime reduction from 49.3s to ~13.2s~ now down to 5.9s.
* In "Hcurl/Hdiv diagonal PA" I got rid of order 3 in 3D.
    * Runtime reduction from 15.1s to 1.5s.

Overall reduction in debug mode on my machine from 1:57 to ~1:15~ 1:05.

Personally I don't think the extra tests are worth the longer runtime (they are just running the same code at higher orders), but in the "Assembly Levels", some of the tests are actually duplicated and can be removed without any consequences.

| [#2837](https://github.com/mfem/mfem/pull/2837) | @pazner | @tzanio | @dylan-copeland + @YohannDudouit + @v-dobrev | 2/18/22 | 2/22/22 | 2/23/22 | |

### DIFF
--- a/tests/unit/fem/test_assembly_levels.cpp
+++ b/tests/unit/fem/test_assembly_levels.cpp
@@ -111,13 +111,13 @@ TEST_CASE("Assembly Levels", "[AssemblyLevel], [PartialAssembly]")
 {
    auto assembly = GENERATE(AssemblyLevel::PARTIAL, AssemblyLevel::ELEMENT,
                             AssemblyLevel::FULL);
-   auto pb = GENERATE(0, 1, 2);
-   auto dg = GENERATE(true, false);
-   auto order_2d = GENERATE(2, 3, 4);
+   auto order_2d = GENERATE(2, 3);
    auto order_3d = GENERATE(2);
 
    SECTION("2D")
    {
+      auto pb = GENERATE(0, 1, 2);
+      auto dg = GENERATE(true, false);
       test_assembly_level("../../data/periodic-square.mesh",
                           order_2d, dg, pb, assembly);
       test_assembly_level("../../data/periodic-hexagon.mesh",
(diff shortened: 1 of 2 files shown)


tests/unit/fem/test_assembly_levels.cpp
```diff
@@ -111,13 +111,13 @@ TEST_CASE("Assembly Levels", "[AssemblyLevel], [PartialAssembly]")
...
                             AssemblyLevel::FULL);
-   auto pb = GENERATE(0, 1, 2);
-   auto dg = GENERATE(true, false);
-   auto order_2d = GENERATE(2, 3, 4);
+   auto order_2d = GENERATE(2, 3);
    auto order_3d = GENERATE(2);
 
    SECTION("2D")
    {
+      auto pb = GENERATE(0, 1, 2);
```
dylan-copeland, line 119: It looks like `pb` and `dg` are now in the 2D and 3D sections with the same values as before. Did you intend for them to be different in these two cases?

dylan-copeland: So `GENERATE` makes nested loops, even if the variable is not used for a test? Does that mean `order_2d` and `order_3d` are also nested loops, which would make the 3D case test run twice (since `order_2d` has size 2)?

pazner: Good catch! I didn't notice that, yes that is exactly what that means.

I fixed it now, runtime is even shorter.

YohannDudouit: Though, we may want to test dg now, as the PR has just been merged for AMR.

YohannDudouit (approved):

tzanio:

PR #2849, opened 2022-02-22
Partial Essential Boundary Conditions in NonlinearForm

rcarson3:
This adds support for per boundary to set what component(s) of the DOFs as being essential DOFs. This is particularly useful for certain types of problems such as those in commonly found in solid mechanics where it's useful to have varying # of the DOF components constrained per surface/boundary.

I've taken the code here from what we've been using within the exaconstit-dev branch for the past few years. A limitation of the current implementation presented in this PR is that it has not been generalized to nonconforming meshes. Although, it might be trivial to extend it to that case given the code for this use case is largely shared with the code already in MFEM for getting essential vdofs. However, I don't really understand that use case well enough to know how the nonconforming case differs.

| [#2849](https://github.com/mfem/mfem/pull/2849) | @rcarson3 | @tzanio | @mlstowell + @vladotomov + @jandrej | 2/24/22 | ⌛due 4/4/22 | ⌛due 4/11/22 | |


tzanio:

fem/fespace.cpp
```diff
@@ -561,6 +561,68 @@ void FiniteElementSpace::GetEssentialVDofs(const Array<int> &bdr_attr_is_ess,
...
+      if (bdr_attr_is_ess[id])
+      {
+         cmpID_rows = true;
+         //Checking to see if all the values in the row are greater than -1 if so then
+         //all components on a boundary are said to be marked.
+         for (int j = 0; j < component.NumCols(); ++j)
+         {
+            cmpID_rows = cmpID_rows && ( component(id, j) > -1);
+         }
+         if (cmpID_rows) // same as srw component id system
```
jandrej, line 590: what is srw?

fem/nonlinearform.cpp
```diff
@@ -53,6 +53,22 @@ void NonlinearForm::SetEssentialBC(const Array<int> &bdr_attr_is_ess,
    }
 }
 
+void NonlinearForm::SetEssentialBCPartial(const Array<int> &bdr_attr_is_ess,
```
jandrej, line 56: What does `Partial` mean in this case? The term is ambiguous. I would opt for keeping the same name `NonlinearForm::SetEssentialBC`. The added argument `bdr_component` and class method comment make it clear that this will work different from the non component version.

fem/nonlinearform.hpp
```diff
@@ -146,6 +146,19 @@ class NonlinearForm : public Operator
...
+   /// assigning the component ID to its location in bdr_component array
+   /// The bdr_component has dimensions number boundary attributes x vdim
+   /** This method calls FiniteElementSpace::GetEssentialTrueDofs() and stores
+       the result internally for use by other methods. If the @a rhs pointer is
+       not NULL, its essential true dofs will be set to zero. This makes it
+       "compatible" with the output vectors from the Mult() method which also
+       have zero entries at the essential true dofs. */
+   void SetEssentialBCPartial(const Array<int> &bdr_attr_is_ess,
+                              const Array2D<int> &bdr_component,
+                              Vector *rhs);
```
jandrej, line 160: OK for me only because the existing class already implements this the same way. For naming, see above.

jandrej (commented):
I'm wondering what these additions bring to the table. Can't I just loop over boundary attributes separately with the existing `FiniteElementSpace::GetEssentialTrueDofs` and 3 more lines of code instead of adding a whole new interface?

For me this is a pass.

fem/fespace.cpp
rcarson3: Sorry I missed that comment. I thought I'd gotten most of them. It's the initials of one the initial contributors to ExaConstit.

fem/nonlinearform.cpp
rcarson3: `Partial` was just the name we'd had picked back in the day. It was supposed to just signify that you're not constrained to either constraining all dofs or only 1 dof. Instead, you could constrain multiple DOFs at a time for a boundary. I really don't have strong preferences to keeping the name though.

tzanio:
Mention in `CHANGELOG`?

rcarson3:
> I'm wondering what these additions bring to the table. Can't I just loop over boundary attributes separately with the existing `FiniteElementSpace::GetEssentialTrueDofs` and 3 more lines of code instead of adding a whole new interface?
> 
> For me this is a pass.

@jandrej so using your way of doing it, you would need to call the existing `FiniteElementSpace::GetEssentialTrueDofs`  vdim x number of boundary attributes and combine the resulting matrices multiple times to achieve the same results. Alternatively, you could reduce the number of calls by calling it only vdim times but that route would require you to have some book-keeping about which boundary attribute is constrained to a given component. I would say that this method does simplify things for the user as it's much easier to keep track of what boundaries are constrained and for what DOFs.

It might be that we could make use of the existing code to do this rather than duplicating things, where we could do the book-keeping internally for the user when they call the `FiniteElementSpace` methods added in this PR. However, it might require a few small changes to `FiniteElementSpace::GetEssentialVDofs` aka we might want to add a flag which allows the inputted `ess_vdofs` to not be overwritten.

fem/pfespace.hpp
```diff
@@ -355,12 +355,36 @@ class ParFiniteElementSpace : public FiniteElementSpace
...
                                      int component = -1);
 
+   /** @brief Get a list of essential true dofs, ess_tdof_list, corresponding to the
+       boundary attributes marked in the array bdr_attr_is_ess.
+       For spaces with 'vdim' > 1, the 'component' array can be used
+       to restricts the marked tDOFs per boundary to the specified components.
+       If vdim > 1 then one can specify per boundary attribute which components
+       on a boundary are essential by assigning the component ID to its location
+       in component array.
+       The component has dimensions number of boundary attributes x vdim. */
```
pazner, line 383: Ditto about the "brief" docs.

pazner (commented):
Some minor Doxygen formatting comments…

vladotomov:
I'm not convinced that a new interface is needed. I achieve this in Laghos in very few lines:
```
Array<int> ess_tdofs, ess_bdr(pmesh->bdr_attributes.Max()), dofs_list;
for (int d = 0; d < pmesh->Dimension(); d++)
{
   // Attributes 1/2/3 correspond to fixed-x/y/z boundaries,
   // i.e., we must enforce v_x/y/z = 0 for the velocity components.
   ess_bdr = 0; ess_bdr[d] = 1;
   H1FESpace.GetEssentialTrueDofs(ess_bdr, dofs_list, d);
   ess_tdofs.Append(dofs_list);
}
```
Sure the proposed would be slightly more flexible, but useful for very specific cases, and can be achieved with minimal bookkeeping from the user with the existing interface.

fem/fespace.cpp
```diff
@@ -561,6 +561,68 @@ void FiniteElementSpace::GetEssentialVDofs(const Array<int> &bdr_attr_is_ess,
...
+         {
+            GetBdrElementDofs(i, dofs);
+            for (int d = 0; d < dofs.Size(); d++)
+            {
+               // loop over actively constrained components
+               for (int k = 0; k < component.NumCols(); ++k)
+               {
+                  if (component(id, k) != -1)   // -1 means inactive component
+                  {
+                     // component(id, k) should be equal to k here
```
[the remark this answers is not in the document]
rcarson3: We'd just been trying to keep things consistent with how the original function works as it took an `int` argument earlier. A bool would probably work just as well.

rcarson3:
> I'm not convinced that a new interface is needed. I achieve this in Laghos in very few lines: 
```c++
Array<int> ess_tdofs, ess_bdr(pmesh->bdr_attributes.Max()), dofs_list;
for (int d = 0; d < pmesh->Dimension(); d++)
{
   // Attributes 1/2/3 correspond to fixed-x/y/z boundaries,
   // i.e., we must enforce v_x/y/z = 0 for the velocity components.
   ess_bdr = 0; ess_bdr[d] = 1;
   H1FESpace.GetEssentialTrueDofs(ess_bdr, dofs_list, d);
   ess_tdofs.Append(dofs_list);
}
```
> Sure the proposed would be slightly more flexible, but useful for very specific cases, and can be achieved with minimal bookkeeping from the user with the existing interface.

@vladotomov it would probably be worth having a webex discussion about this with the other reviewers as I would say the book keeping required for problems with non-trivial BCs or those with evolving BCs using the existing interface is not user friendly. I can set that up a webex discussing if @jandrej , @v-dobrev or @pazner are interested in one as well.

tzanio:
Update on the current status: after internal discussion we decided to add the new API functions, but minimize the code duplication in the implementation. @rcarson3 will update the PR accordingly.

rcarson3:
@vladotomov ,  @jandrej , @mlstowell , and @pazner  I believe this should be good to review again. I believe I've addressed the earlier reviewer comments by eliminating the code duplication related to `FiniteElementSpace::GetEssentialVDofs`. The new `FiniteElementSpace::GetEssentialTrueDofs` variation now does booking keeping on its own end of things and makes use of the previous version of `FiniteElementSpace::GetEssentialVDofs`. I will mention that for `FiniteElementSpace::GetEssentialVDofs` I did add an additional function argument that tells the function whether or not it should overwrite the incoming `ess_vdofs` argument. This new argument has a default value such that the old behavior of the function is retained. So, it should not breaking any existing code.

Also the doxygen comments should be updated to address earlier comments regarding them.

jandrej (changes requested):

fem/fespace.hpp
```diff
@@ -775,11 +775,16 @@ class FiniteElementSpace
...
-       to restricts the marked vDOFs to the specified component. */
+       to restricts the marked vDOFs to the specified component.
+       If overwrite is set to false then values in ess_vdofs are preserved
+       and not reset. However, the assumption here is that ess_vdofs is set to
+       the correct size already.*/
    virtual void GetEssentialVDofs(const Array<int> &bdr_attr_is_ess,
                                   Array<int> &ess_vdofs,
-                                  int component = -1) const;
+                                  int component = -1,
+                                  bool overwrite = true) const;
```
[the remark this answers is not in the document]
rcarson3: Yes, it's just away to avoid having `ess_vdofs` being recreated every time this function is called and the need for another `Array` to keep track of the accumulated values. If you think improved documentation would help here to make that more clear I can do that.

fem/nonlinearform.hpp
```diff
@@ -146,6 +146,24 @@ class NonlinearForm : public Operator
...
+       on a boundary are essential by assigning a value of true to its location
+       in the bdr_component array.
+       The bdr_component has dimensions number of boundary attributes x vdim
+
+       This method calls FiniteElementSpace::GetEssentialTrueDofs() and stores
+       the result internally for use by other methods. If the @a rhs pointer is
+       not NULL, its essential true dofs will be set to zero. This makes it
+       "compatible" with the output vectors from the Mult() method which also
+       have zero entries at the essential true dofs. */
+   void SetEssentialBC(const Array<int> &bdr_attr_is_ess,
```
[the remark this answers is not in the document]
rcarson3: I could mention that here and in the other `SetEssentialBC` function right above.

fem/pfespace.cpp
```diff
@@ -1059,6 +1061,58 @@ void ParFiniteElementSpace::GetEssentialTrueDofs(const Array<int>
...
+      bdr_attr_is_ess_single_comp = 0;
+      for (int j = 0; j < bdr_attr_is_ess.Size(); j++)
+      {
+         if (bdr_attr_is_ess[j] && component(j, i))
+         {
+            bdr_attr_is_ess_single_comp[j] = bdr_attr_is_ess[j];
+         }
+      }
+      GetEssentialVDofs(bdr_attr_is_ess_single_comp, ess_dofs, i, overwrite);
+   }
```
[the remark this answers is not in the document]
rcarson3: I'll try and see in what ways I can't refactor this to simplify things further.

tzanio:
ping: @rcarson3

vladotomov:
What is the state of this? Do you still want to merge it @rcarson3?

rcarson3:
> What is the state of this? Do you still want to merge it @rcarson3?

@vladotomov sorry for a lack of movement on this. The project I was working on this for shifted gears shortly around my last set of comments. So, I just never got back to addressing the last set of comments. I'll see if I can't address all of Julian's comments next week when my schedule frees up just a little bit.

rcarson3:
@vladotomov , @jandrej , @tzanio and @mlstowell sorry for the super long delay in responding to things... I just brought the branch back up to date with the main branch, added some additional comments to the doxygen to hopefully address additional concerns, and then did a small refactoring of some code to address another comment made. Let me know if there's anything else needed as I at least temporarily have the bandwidth to address things.

mlstowell (approved):

rcarson3:
@vladotomov, @tzanio  or @jandrej is there anything else you want to see done in this PR?

jandrej (commented):
See https://github.com/mfem/mfem/pull/2849#pullrequestreview-893963762

PR #2857, opened 2022-02-28
[WIP] Enabling shadow variable warnings

mlstowell:


mlstowell:
It looks like the c++ compiler used for the Ubuntu tests does not regard constructor arguments as exceptions to the shadowing rules.

YohannDudouit:
> It looks like the c++ compiler used for the Ubuntu tests does not regard constructor arguments as exceptions to the shadowing rules.

I think it's because the arguments with same names have different types, I did a PR that solves the issue in `estimator.hpp/cpp`: https://github.com/mfem/mfem/pull/2862
Should we check if this is the issue?

pazner:
> > It looks like the c++ compiler used for the Ubuntu tests does not regard constructor arguments as exceptions to the shadowing rules.
> 
> For the constructors, I think it's because the arguments with same names have different types, I did a PR that solves the issue in `estimator.hpp/cpp`: #2862 Should we check if this is the issue?

Some experimenting suggestions that this is just a `gcc` vs. `clang` difference. While using the same name for members and arguments in constructor initializer lists is unambiguous, `gcc` still warns about it (whereas `clang` does not).

PR #2859, opened 2022-03-01
Fix `#ifdef` name for libCEED's non-linear convection.

YohannDudouit:
Change the name of the `hpp` `ifdef` guard.


tzanio:
Moved to https://github.com/mfem/mfem/pull/2839

PR #2864, opened 2022-03-02
Conduit Blueprint Bugfix

aaronskinner:
This fixes a bug in `ConduitDataCollection::MeshToBlueprintMesh()` wherein boundary topology is missing if a rank happens not to have any boundary elements (i.e., is purely internal).  First, I cribbed a function from `SidreDataCollection` to determine whether any rank has any boundary elements (an all-periodic mesh might have none), then I initialized a few geometry-related fields to unimportant data in case there are no boundary elements to use.  This technically gives the wrong answer for some ranks, and it assumes all boundary elements are of the same type, but that's probably okay for now since it's never used.

| [#2864](https://github.com/mfem/mfem/pull/2864) | @aaronskinner | @tzanio | @tzanio + @v-dobrev + @camierjs | 3/2/22 | 3/7/22 | 3/7/22 | |

### DIFF
--- a/fem/conduitdatacollection.cpp
+++ b/fem/conduitdatacollection.cpp
@@ -765,7 +765,7 @@ ConduitDataCollection::MeshToBlueprintMesh(Mesh *mesh,
    ////////////////////////////////////////////
 
    // guard vs if we have boundary elements
-   if (mesh->GetNBE() > 0)
+   if (HasBoundaryElements(mesh))
    {
       n_topo["boundary_topology"] = boundary_topology_name;
 
@@ -774,15 +774,14 @@ ConduitDataCollection::MeshToBlueprintMesh(Mesh *mesh,
       n_bndry_topo["type"]     = "unstructured";
       n_bndry_topo["coordset"] = coordset_name;
 
-      Element::Type bndry_ele_type = mesh->GetBdrElementType(0);
+      int num_bndry_ele = mesh->GetNBE();
 
-      std::string bndry_ele_shape = ElementTypeToShapeName(bndry_ele_type);
-
+      // must initialize this to something
+      Element::Type bndry_ele_type   = (num_bndry_ele > 0) ? mesh->GetBdrElementType(0) : mfem::Element::POINT;
+      std::string bndry_ele_shape    = ElementTypeToShapeName(bndry_ele_type);
       n_bndry_topo["elements/shape"] = bndry_ele_shape;
 
-
-      int num_bndry_ele = mesh->GetNBE();
-      int bndry_geom    = mesh->GetBdrElementBaseGeometry(0);
+      int bndry_geom          = (num_bndry_ele > 0) ? mesh->GetBdrElementBaseGeometry(0) : mfem::Element::POINT;
       int bndry_idxs_per_ele  = Geometry::NumVerts[bndry_geom];
       int num_bndry_conn_idxs =  num_bndry_ele * bndry_idxs_per_ele;
 
@@ -937,6 +936,28 @@ ConduitDataCollection::GridFunctionToBlueprintField(mfem::GridFunction *gf,
       }
    }
 
+}
+
+//---------------------------------------------------------------------------//
+bool
+ConduitDataCollection::HasBoundaryElements(mfem::Mesh* mesh)
+{
+   // check if this rank has any boundary elements
+   int hasBndElts = mesh->GetNBE() > 0 ? 1 : 0;
+
+#ifdef MFEM_USE_MPI
+   // check if any rank has boundary elements
+   ParMesh *pmesh = dynamic_cast<ParMesh*>(mesh);
+   if (pmesh)
+   {
+      int hasBndElts_g;
+      MPI_Allreduce(&hasBndElts, &hasBndElts_g, 1,
+         MPI_INT, MPI_MAX,pmesh->GetComm());
+      hasBndElts = hasBndElts_g;
+   }
+#endif
+
+   return hasBndElts > 0 ? true : false;
 }
 
 //------------------------------
--- a/fem/conduitdatacollection.cpp
+++ b/fem/conduitdatacollection.cpp
@@ -765,7 +765,7 @@ ConduitDataCollection::MeshToBlueprintMesh(Mesh *mesh,
    ////////////////////////////////////////////
 
    // guard vs if we have boundary elements
-   if (HasBoundaryElements(mesh))
+   if (mesh->HasBoundaryElements())
    {
       n_topo["boundary_topology"] = boundary_topology_name;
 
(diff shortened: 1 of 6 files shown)


tzanio:
Hi @aaronskinner, can you add the `ready-for-review` label when this is ready?

tzanio:

tzanio:
@cyrush, can you please help with this review?

fem/conduitdatacollection.cpp
```diff
@@ -939,6 +940,28 @@ ConduitDataCollection::GridFunctionToBlueprintField(mfem::GridFunction *gf,
 
 }
 
+//---------------------------------------------------------------------------//
+bool
+ConduitDataCollection::HasBoundaryElements(mfem::Mesh* mesh)
```
tzanio, line 945: No strong opinion, but this seems useful in general, so it could be moved as a method of `ParMesh`.

cyrush:
This looks like a good solution to the issue of assuming the boundary is always there. 

The only question I have:

How does this impact the corresponding: `BlueprintMeshToMesh` reconstruction. 
(that takes a Blueprint Mesh and creates MFEM Mesh Objects from ti)

It would be good to have a test case to make sure we don't have problems when the boundary is presented - but has empty arrays.  I can help with the test - but can't start this week.

fem/conduitdatacollection.cpp
kennyweiss: Should it be a virtual method of `mfem::Mesh` with an override in `ParMesh`?

fem/conduitdatacollection.hpp
```diff
@@ -209,6 +209,9 @@ class ConduitDataCollection : public DataCollection
                                                      const conduit::Node &n_field,
                                                      bool zero_copy = false);
 
+   /** Checks if any rank in the mesh has boundary elements */
+   static bool HasBoundaryElements(mfem::Mesh *mesh);
```
kennyweiss, line 213: Minor: Since the `ConduitDataCollection` is over a specific mesh, perhaps this should be a non-static member function?

kennyweiss, line 213: [The following comment is only relevant if `HasBoundaryElements()` does not move to `mfem::Mesh` / `mfem::ParMesh`]

The `HasBoundaryElements` function was useful for `SidreDataCollection`. Since it's being used in `ConduitDataCollection` as well, it seems like it should be upgraded to the `DataCollection` base class.

fem/conduitdatacollection.cpp
```diff
@@ -774,15 +774,16 @@ ConduitDataCollection::MeshToBlueprintMesh(Mesh *mesh,
...
+      Element::Type bndry_ele_type   = (num_bndry_ele > 0) ? mesh->GetBdrElementType(
+                                          0) : mfem::Element::POINT;
+      std::string bndry_ele_shape    = ElementTypeToShapeName(bndry_ele_type);
       n_bndry_topo["elements/shape"] = bndry_ele_shape;
 
-
-      int num_bndry_ele = mesh->GetNBE();
-      int bndry_geom    = mesh->GetBdrElementBaseGeometry(0);
+      int bndry_geom          = (num_bndry_ele > 0) ? mesh->GetBdrElementBaseGeometry(
+                                   0) : mfem::Element::POINT;
```
kennyweiss, line 786: I'm concerned that arbitrarily choosing `Element::POINT` might be problematic in some cases.

Possible alternatives for the empty boundary case on a given rank:
* Use an `MPI_Allreduce` to find the value
* deduce the boundary element type based on the volumetric mesh elements ? E.g. if the volume mesh is `Element::HEXAHEDRON`, the boundary type should be `Element::QUADRILATERAL`.

My concern is about the case where rank 0 has no boundary elements, but the incorrect value is used for the blueprint index. VisIt might interpret the overall boundary elements incorrectly.

This might also be a useful utility function to promote to the `DataCollection` base class.

kennyweiss (approved):
Thanks @aaronskinner!

fem/conduitdatacollection.cpp
cyrush: @kennyweiss the blueprint index does not have any details about the element types, however it will provide overall topology type metadata (like `unstructured` vs `rectilinear`). I recommend trying to see if there are problems (and even if there are -- we can adapt to mitigate them)

cyrush: (I am pretty sure this wont trip things up, but we should check to avoid surprises )

tzanio:
@camierjs, I added you as a reviewer just for https://github.com/mfem/mfem/pull/2864/commits/0ab2a5be53ee94959f28a8a9656743504414a729

fem/conduitdatacollection.cpp
tzanio: @aaronskinner, I can make the move to an `Mesh`/`ParMesh` method for you if you prefer

fem/conduitdatacollection.cpp
aaronskinner: I originally had an `MPI_Allreduce` but I was concerned about that for a few reasons:  1) it's more expensive, although this is a startup cost, 2) what if it's different among ranks? what's the right one to pick?, 3) the rest of the code sort of assumes the value is the same everywhere, 4) it's never actually used as far as I can tell, 5) I did what was done in `MeshBlueprintToMesh`, which also makes this assumption.  I'm happy making an assumption based on the volumetric type.  I'm also happy putting useful functions in the base class.

fem/conduitdatacollection.cpp
aaronskinner: Yes, I think it should move; these seems like a useful public method for that class.  I'm less sure about the other functions that get the boundary element type, because here we're assuming the type is the same for all boundary elements on the mesh.  I'm now sure how the blueprint even looks if they're different...  I'm also not sure it matters based on Cyrus' later comment.

general/forall.hpp
```diff
@@ -25,8 +25,8 @@ namespace mfem
 
 // Maximum size of dofs and quads in 1D.
 #ifdef MFEM_USE_HIP
-const int MAX_D1D = 11;
-const int MAX_Q1D = 11;
+const int MAX_D1D = 10;
+const int MAX_Q1D = 10;
```
tzanio, line 29: @camierjs and @v-dobrev can you click "approve" just for this change?

tzanio:
Merging in `next` just so we get some more test over the weekend...

kennyweiss (approved):
Thanks for the updates @tzanio !

v-dobrev (approved):

cyrush (approved):

tzanio:

mesh/pmesh.cpp
```diff
@@ -1594,6 +1594,15 @@ void ParMesh::SetAttributes()
    }
 }
 
+bool ParMesh::HasBoundaryElements() const
+{
+   // maximum number of boundary elements over all ranks
+   int maxNumOfBdrElements;
+   MPI_Allreduce(&NumOfBdrElements, &maxNumOfBdrElements, 1,
+                 MPI_C_BOOL, MPI_MAX, MyComm);
```
tomstitt, line 1602: It turns out `MPI_MAX` is not supported for `MPI_C_BOOL`, we will need to change the type:

```
MPI_Allreduce: the reduction operation MPI_MAX is not defined on the MPI_C_BOOL datatype
```

tzanio: That's a bug, it should have been `MPI_INT`

tzanio: I added the fix to this branch and `master`

PR #2874, opened 2022-03-08
Enzyme nonlinear device

camierjs:
Couple of device optimizations using the provided `tensor` class.

### DIFF
--- a/examples/ex42.cpp
+++ b/examples/ex42.cpp
@@ -3,18 +3,10 @@
 // Compile with: make ex42p
 //
 
-#define MFEM_NVTX_COLOR Orange
-#include "../general/nvtx.hpp"
-
-#define MFEM_DEBUG_COLOR 227
-#include "general/debug.hpp"
-
 #include "../general/forall.hpp"
-#include "../fem/kernels.hpp"
 #include "../linalg/kernels.hpp"
 #include "../linalg/tensor.hpp"
 #include "mfem.hpp"
-#include <cassert>
 
 using namespace std;
 using namespace mfem;
@@ -44,7 +36,7 @@ return_type __enzyme_fwddiff(Args...);
 #endif
 
 /// Define the identity tensor in three dimensions.
-//MFEM_HOST_DEVICE static constexpr auto I = Identity<dimension>();
+//static constexpr auto I = Identity<dimension>();
 
 /** @brief Linear elastic material.
  *
@@ -470,7 +462,7 @@ public:
     * @tparam material_type
     * @param material
     */
-   template <bool opt, typename material_type>
+   template <typename material_type>
    void SetMaterial(const material_type &material);
 
    /**
@@ -519,7 +511,6 @@ public:
    void AssembleGradientDiagonal(Vector &Ke_diag, Vector &K_diag_local,
                                  Vector &K_diag) const
    {
-      MFEM_NVTX;
       elasticity_op_.AssembleGradientDiagonal(Ke_diag, K_diag_local, K_diag);
    }
 
@@ -550,7 +541,6 @@ public:
 
    void SetOperator(const Operator &op) override
    {
-      MFEM_NVTX;
       gradient_operator_ = dynamic_cast<const ElasticityGradientOperator *>(&op);
       MFEM_ASSERT(gradient_operator_ != nullptr,
                   "Operator is not ElasticityGradientOperator");
@@ -567,46 +557,40 @@ public:
 
    void Mult(const Vector &x, Vector &y) const override
    {
-      MFEM_NVTX;
+      const int ns = num_submats_, sh = submat_height_, nsh = ns * sh;
+
+      const auto K_diag_submats = Reshape(K_diag_.Read(), ns, sh, sh);
+      const auto X = Reshape(x.Read(), ns, dim);
+
+      auto Y = Reshape(y.Write(), ns, dim);
+
       if (type_ == Type::Diagonal)
       {
-         auto Y = y.Write();
-         const auto X = x.Read();
-         const int ns = num_submats_, sh = submat_height_, nsh = ns * sh;
-         const auto K_diag_submats = Reshape(K_diag_.Read(), ns, sh, sh);
          // Assuming Y and X are ordered byNODES. K_diag is ordered byVDIM.
          MFEM_FORALL(si, nsh,
          {
             const int s = si / sh;
             const int i = si % sh;
-            const int idx = s + i * ns;
-            Y[idx] = X[idx] / K_diag_submats(s,i,i);
+            Y(s,i) = X(s,i) / K_diag_submats(s, i, i);
          });
       }
       else if (type_ == Type::BlockDiagonal)
       {
-         MFEM_ABORT("Using host here!");
-         auto K_diag_submats =
-            Reshape(K_diag_.Read(), num_submats_, submat_height_, submat_height_);
-
-         for (int s = 0; s < num_submats_; s++)
+         MFEM_FORALL(s, ns,
          {
-            auto submat_inv = inv(make_tensor<dim, dim>(
-            [&](int i, int j) { return K_diag_submats(s, i, j); }));
-
-            auto x_block =
-            make_tensor<dim>([&](int i) { return x(s + i * num_submats_); });
-
-            tensor<double, dim> y_block;
-
-            y_block = submat_inv * x_block;
-
-            for (int i = 0; i < dim; i++)
+            const auto submat = make_tensor<dim, dim>([&](int i, int j)
             {
-               int idx = s + i * num_submats_;
-               y(idx) = y_block(i);
-            }
-         }
+               return K_diag_submats(s, i, j);
+            });
+
+            const auto submat_inv = inv(submat);
+
+            const auto x_block = make_tensor<dim>([&](int i) { return X(s,i); });
+
+            tensor<double, dim> y_block = submat_inv * x_block;
+
+            for (int i = 0; i < dim; i++) { Y(s,i) = y_block(i); }
+         });
       }
       else
       {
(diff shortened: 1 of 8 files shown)


examples/ex42.cpp
```diff
@@ -1034,18 +1119,44 @@ ApplyGradient3D(const int ne, const Array<double> &B_, const Array<double> &G_,
...
+               {
+                  for (int m = 0; m < dim; m++)
+                  {
+                     dudxi(qx, qy, qz, l, m) = 0.0;
+                     ddudxi(qx, qy, qz, l, m) = 0.0;
+                  }
+               }
+            }
+         }
+      }
```
jandrej, line 1154: Ugh... I really don't like this. `tensor` is supposed to make things easier.

linalg/dual.hpp
```diff
@@ -310,50 +314,48 @@ dual<value_type, gradient_type> tan(dual<value_type, gradient_type> a)
...
 }
 
 /** @brief implementation of `a` (non-dual) raised to the `b` (dual) power */
-template <typename value_type, typename gradient_type>
+template <typename value_type, typename gradient_type> MFEM_HOST_DEVICE
 dual<value_type, gradient_type> pow(double a, dual<value_type, gradient_type> b)
 {
-   using std::pow;
-   value_type value = pow(a, b.value);
+   value_type value = std::pow(a, b.value);
```
jandrej, line 350: This is not equivalent and might actually introduce errors. Why the switch to `std::pow`?

linalg/dual.hpp
```diff
@@ -310,50 +314,48 @@ dual<value_type, gradient_type> tan(dual<value_type, gradient_type> a)
...
 }
 
 /** @brief implementation of `a` (dual) raised to the `b` (non-dual) power */
-template <typename value_type, typename gradient_type>
+template <typename value_type, typename gradient_type> MFEM_HOST_DEVICE
 dual<value_type, gradient_type> pow(dual<value_type, gradient_type> a, double b)
 {
-   using std::pow;
-   value_type value = pow(a.value, b);
+   value_type value = std::pow(a.value, b);
```
jandrej, line 358: see other comment

linalg/tensor.hpp
```diff
@@ -124,18 +124,18 @@ struct tensor<T, n0, n1, n2, n3, n4>
...
+   MFEM_HOST_DEVICE tensor< T, n1, n2, n3, n4 >& operator()(int i) { return values[i]; }
+   MFEM_HOST_DEVICE tensor< T, n1, n2, n3, n4 >  operator()(int i) const { return values[i]; }
+   MFEM_HOST_DEVICE tensor< T, n2, n3, n4 >& operator()(int i, int j) { return values[i][j]; }
+   MFEM_HOST_DEVICE tensor< T, n2, n3, n4 >  operator()(int i, int j) const { return values[i][j]; }
+   MFEM_HOST_DEVICE tensor< T, n3, n4>& operator()(int i, int j, int k) { return values[i][j][k]; }
+   MFEM_HOST_DEVICE tensor< T, n3, n4>  operator()(int i, int j, int k) const { return values[i][j][k]; }
+   MFEM_HOST_DEVICE tensor< T, n4 >& operator()(int i, int j, int k, int l) { return values[i][j][k][l]; }
+   MFEM_HOST_DEVICE tensor< T, n4 >  operator()(int i, int j, int k, int l) const { return values[i][j][k][l]; }
+   MFEM_HOST_DEVICE T& operator()(int i, int j, int k, int l, int m) { return values[i][j][k][l][m]; }
+   MFEM_HOST_DEVICE T  operator()(int i, int j, int k, int l, int m) const { return values[i][j][k][l][m]; }
```
jandrej, line 138: fingers crossed for c++17 :D

linalg/dual.hpp
camierjs: I'm hitting with `nvcc`: 

- `error: no instance of overloaded function "mfem::internal::pow" matches the argument list`
- `argument types are: (double, double)`

I'll revert all `std::` and will keep the only one I needed, commented.

jandrej: maybe it needs a specialized version of pow that uses `std::pow` @samuelpmishLLNL?

camierjs: I added such specialized version - looks better.

examples/ex42.cpp
YohannDudouit: This is exactly what I was referring to when I was saying that we want to abstract the parallelization strategy for a `tensor`, this is precisely to avoid this kind of issues. But it should be fairly easy to hide for now.

samuelpmishLLNL: > Ugh... I really don't like this. `tensor` is supposed to make things easier.

It looks like the verbose implementation is a byproduct of trying to use macros to define the CPU and GPU kernels at the same time. e.g. a CPU-only implementation would just be the one-liner:
```cpp
tensor<double, q1d, q1d, q1d, dim, dim> dudxi{};
```
and a GPU-only implementation would be (assuming kernel launched with appropriate block dimensions)
```cpp
__shared__ tensor<double, q1d, q1d, q1d, dim, dim> dudxi;
du_dxi(threadIdx.x, threadIdx.y, threadIdx.z) = tensor<double, dim, dim>{};
```

examples/ex42.cpp
```diff
@@ -996,8 +1080,9 @@ Apply3D(const int ne, const Array<double> &B_, const Array<double> &G_,
          }
       }
       MFEM_SYNC_THREAD;
+
       auto F = Reshape(&force(0, 0, 0, 0, e), d1d, d1d, d1d, dim);
-      KernelHelpers::CalcGradTSum<dim, d1d, q1d>(B, G, invJ_sigma_detJw, F);
+      KernelHelpers::CalcGradTSum<dim, d1d, q1d>(B, G, smem, invJ_sigma_detJw, F);
```
samuelpmishLLNL, line 1085: Is it necessary to explicitly specify the template arguments here? I would have guessed the compiler would deduce those values if instead we wrote:

`KernelHelpers::CalcGradTSum(B, G, smem, invJ_sigma_detJw, F);`

https://godbolt.org/z/53arKb1xo

examples/ex42.cpp
```diff
@@ -1034,18 +1119,44 @@ ApplyGradient3D(const int ne, const Array<double> &B_, const Array<double> &G_,
...
+      MFEM_FOREACH_THREAD(qx, x, q1d)
+      {
+         MFEM_FOREACH_THREAD(qy, y, q1d)
+         {
+            if (MFEM_THREAD_ID(z) == 0 && qy < d1d)
+            {
+               const int dy = qy;
+               B(qx, dy) = b(qx, dy);
+               G(qx, dy) = g(qx, dy);
+            }
```
samuelpmishLLNL, line 1141: this seems like it it might be leaving part of `B`, `G` uninitialized if d1d > q1d

examples/ex42.cpp
```diff
@@ -1101,31 +1217,64 @@ static inline void AssembleGradientDiagonal3D(
...
+               const bool inside_d1d_cube = qx < d1d && qy < d1d && qz < d1d;
+               for (int l = 0; l < dim; l++)
+               {
+                  for (int m = 0; m < dim; m++)
+                  {
+                     dudxi(qx, qy, qz, l, m) = 0.0;
+                     if (inside_d1d_cube) { Ke_diag(qx, qy, qz, l,m) = 0.0; }
+                  }
+               }
+            }
```
samuelpmishLLNL, line 1252: same comment as https://github.com/mfem/mfem/pull/2874/files#r822201282

is there an implicit assumption being made that q1d will never be less than d1d? I understand you don't attain the maximum-rank element matrices by such integration rules, but it still is pretty common.

samuelpmishLLNL: edit: @camierjs showed me that there is a check to ensure this assumption holds:

```
   MFEM_VERIFY(d1d <= q1d, "d1d <= q1d");
```

but it's in a different part of the code, so I missed it!

examples/ex42.cpp
camierjs: `done`

linalg/dual.hpp
camierjs: `done`

examples/ex42.cpp
```diff
@@ -1101,31 +1217,64 @@ static inline void AssembleGradientDiagonal3D(
...
+         {
+            if (MFEM_THREAD_ID(z) == 0 && qy < d1d)
+            {
+               const int dy = qy;
+               B(qx, dy) = b(qx, dy);
+               G(qx, dy) = g(qx, dy);
+            }
+            MFEM_FOREACH_THREAD(qz, z, q1d)
+            {
+               const bool ddd = (qx < d1d) && (qy < d1d) && (qz < d1d);
```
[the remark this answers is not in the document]
camierjs: `removed`

examples/ex42.cpp
camierjs: the `d1d <= q1d` assumption should hold

examples/ex42.cpp
```diff
@@ -1376,6 +1302,15 @@ ElasticityOperator::ElasticityOperator(ParMesh &mesh, const int order)
...
    gradient_ = new ElasticityGradientOperator(*this);
+
+   BG2_[0] = make_tensor<2,2>([&](int i, int j) { return maps_->B[i+2*j]; });
+   BG2_[1] = make_tensor<2,2>([&](int i, int j) { return maps_->G[i+2*j]; });
+
+   BG3_[0] = make_tensor<3,3>([&](int i, int j) { return maps_->B[i+3*j]; });
+   BG3_[1] = make_tensor<3,3>([&](int i, int j) { return maps_->G[i+3*j]; });
+
+   BG4_[0] = make_tensor<4,4>([&](int i, int j) { return maps_->B[i+4*j]; });
+   BG4_[1] = make_tensor<4,4>([&](int i, int j) { return maps_->G[i+4*j]; });
```
jandrej, line 1313: Why are those kernel dependent in the dispatch but not here? Why is this needed?

PR #2876, opened 2022-03-08
[WIP] Proof-of-concept singleton session objects

pazner:
This is a proof-of-concept of a backwards-compatible way to use singletons for the MPI and hypre sessions.

It changes the implementation of `MPI_Session` to use a singleton behind the scenes.

The only difference in behavior compared with the current implementation is that the MPI session is finalized at application exit, rather than when the `MPI_Session` goes out of scope.

It is my understanding that if the MPI session and hypre session are created in the same translation unit, then their destruction order will be deterministic.


pazner:
Closing since this idea is integrated in #2873.

PR #2903, opened 2022-03-18
[TMOP] RAJA & HIP out-of-registers-during-register-allocation error workaround

camierjs:
Reuse local arrays, to help register allocation


rcarson3 (approved):
Thanks @camierjs ! I can confirm this fixes the build issues I was running into.

v-dobrev (approved):

PR #2908, opened 2022-03-21
Fix memory leak in FA assembly

pazner:
The `FULL` assembly mode has a memory leak since the initial pointers for the `I`, `J`, and `A` arrays are overwritten (`J` and `A` are allocated but are actually zero-length, `I` has nonzero length).

This PR adds a new "expert function" `SparseMatrix::OverrideSize` that lets one change the `width` and `height` of a `SparseMatrix` after it has already been constructed (e.g. with the default constructor).

The use-case here is creating an "empty" `SparseMatrix` (but with the correct dimensions), and letting the user handle allocating and filling the CSR arrays.

### DIFF
--- a/fem/bilinearform_ext.cpp
+++ b/fem/bilinearform_ext.cpp
@@ -903,7 +903,8 @@ void FABilinearFormExtension::Assemble()
    }
    else // We create, compute the sparsity, and fill the sparse matrix
    {
-      mat = new SparseMatrix(height, width, 0);
+      mat = new SparseMatrix;
+      mat->OverrideSize(height, width);
       if (fes.IsDGSpace())
       {
          const L2ElementRestriction *restE =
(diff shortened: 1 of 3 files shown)


YohannDudouit (approved):
I remember struggling finding a constructor that wouldn't do or allocate things I don't want, ideally the `SparseMatrix::SparseMatrix(int nrows, int ncols)` constructor should just do that I think. I think this issue is arising because this `SparseMatrix` class conflates two sparse matrix representations into one. I think the code would gain in clarity and simplicity by separating the `SparseMatrix` class in two, one based on linked list (`LLSparseMatrix` for instance), and one based on the CSR format (`CSRSparseMatrix` for instance). By conflating these two sparse matrix representations into one, it is hard to follow what is actually happening inside the current `SparseMatrix` class implementation, and the interface is hard to use correctly because impacting both these representations internally at the same time.

rcarson3 (approved):
I guess this seems fine with me after looking over how the class does things.

tzanio:

PR #2916, opened 2022-03-23
Fix bug in ParDiscreteLinearOperator::ParallelAssemble

pazner:
When computing RAP for a `DiscreteLinearOperator`, `R` should be the restriction matrix, not the transpose of the prolongation matrix.

### DIFF
--- a/fem/pbilinearform.cpp
+++ b/fem/pbilinearform.cpp
@@ -628,13 +628,21 @@ void ParDiscreteLinearOperator::ParallelAssemble(OperatorHandle &A)
                                domain_fes->GetDofOffsets(),
                                mat);
 
-   OperatorHandle R_test_transpose(A.Type()), P_trial(A.Type());
+   SparseMatrix *Rt = Transpose(*range_fes->GetRestrictionMatrix());
+   OperatorHandle R_test_transpose(A.Type());
+   R_test_transpose.MakeRectangularBlockDiag(range_fes->GetComm(),
+                                             range_fes->GlobalVSize(),
+                                             range_fes->GlobalTrueVSize(),
+                                             range_fes->GetDofOffsets(),
+                                             range_fes->GetTrueDofOffsets(),
+                                             Rt);
 
    // TODO - construct the Dof_TrueDof_Matrix directly in the required format.
-   R_test_transpose.ConvertFrom(range_fes->Dof_TrueDof_Matrix());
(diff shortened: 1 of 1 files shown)


mlstowell:
@pazner, this looks important. Is this distinction explained someplace? If not, we should add some documentation.

Also, I noticed `ParFiniteElementSpace::GetRestrictionOperator()` forms `Rt` in different ways in different contexts. We should make sure we have documentation covering the different contexts and what limitations might apply.

pazner:
There is some documentation here 
https://github.com/mfem/mfem/blob/ec5ce4b525067f3a44c3343513888ce67f48c6f7/fem/bilinearform.hpp#L956-L985
and here
https://github.com/mfem/mfem/blob/ec5ce4b525067f3a44c3343513888ce67f48c6f7/fem/pbilinearform.hpp#L325-L327

In particular, the Doxygen comment for `ParDiscreteLinearOperator::ParallelAssemble` describes the correct procedure, only the implementation was wrong.

Note that `ParDiscreteLinearOperator::ParallelAssemble()` was already correct, only `ParDiscreteLinearOperator::ParallelAssemble(OperatorHandle&)` was wrong.

tzanio (approved):
I am not sure why this was implemented differently from `ParallelAssemble()`, but you are absolutely right -- it should be `R A P`, not `P^T A P`

tzanio:
To add a bit more context:

For bilinear forms, `P^T A P` comes just from the fact that we are restricting the form to a subspace (the range of `P`) and so `(A P x, P y)` leads to `P^T A P`.

For discrete operators, we want instead a commuting diagram between `A` the version of `A` on truedofs, lets call it `X` and the interpolation operators in the domain, `P_d, and range, `P_r` spaces. This means `P_r A = X P_d` implying `A = R_r X P_d`, since `R_r P_r = I`. See also (31)-(33) in https://doi.org/10.1190/geo2015-0013.1.

pazner:
> I am not sure why this was implemented differently from `ParallelAssemble()`, but you are absolutely right -- it should be `R A P`, not `P^T A P`

I believe this implementation is a little bit different because it is supposed to support more general operator types, e.g. PETSc matrices rather than just HypreParMatrix. In practice I don't know how well-supported other operator types are.

mlstowell (approved):

tzanio:

PR #2951, opened 2022-03-31
MeshConnections object [WIP]

acfisher:


acfisher:
@jamiebramwell Do you think the:
bool IsXInY() methods have a use for your team?  Any other mesh id number info that you think would be useful here?

acfisher:
@pazner Are the dimension independent methods that I added about what you had in mind (at least for the interface)?

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,95 @@
...
+   bool IsEdgeInFace(int face, int edge);
+   bool IsVertInFace(int face, int vert);
+   bool IsVertInEdge(int edge, int vert);
+
+   //Neighbor mappings
+   void NDNeighbors(int nd, int id, Array<int> &neighbors);
+   void NeighborsToElem(int elem, Array<int> &elems);
+   void NeighborsToFace(int face, Array<int> &faces);
+   void NeighborsToEdge(int edge, Array<int> &edges);
+   void NeighborsToVert(int vert, Array<int> &verts);
```
pazner, line 24: What is meant by neighbors? Are two elements neighbors if they share a face? Edge? Vertex?

The more general version of this is a query like: give me all of the `dim`-dimensional entities that share a `sdim`-dimensional entity with the given entity.

pazner:
> @pazner Are the dimension independent methods that I added about what you had in mind (at least for the interface)?

Yes, something like that, though the names could probably be refined a little bit.

mesh/mesh_connections.hpp
acfisher: Yeah what I am driving at with these is entities that share a dim-1 with the given entity which is typically what folks think of as neighbors (at least element neighbors).  Although your generalization would be valuable
NDNeighborsSharingMDParts?

acfisher: I'm definitively open to suggestions on the names of these things.  Parts->Entities?

pazner: My suggestion would be to rename `NDPartsInMDPart` simply `Children` and `NDPartsWithMDPart` to `Parents`.

A mesh entity is uniquely determined by the pair `(dim, id)` (i.e., in a 3D mesh, the pair `(3, i)` would identify element `i`, `(2, i`) would identify face `i`, and so on). We could either have an entity be `std::pair<int,int>` or just a simple struct.

Then the signature would be something like this (and the "multiple" queries could just be overloads):

```c++
/// Return all indices of dim-dimensional children of the given entity
void Children(Entity entity, int dim, Array<int> &ids);
void Children(const Array<Entity> &entities, int dim, Array<int> &ids);

/// Return all indices of dim-dimensional parents of the given entity
void Parents(Entity entity, int dim, Array<int> &ids);
void Parents(const Array<Entity> &entities, int dim, Array<int> &ids);
```

So to get the vertices of the `i`th element, you would just call
```c++
Children({dim, i}, 0, vertex_list);
```
and to get the list of all elements containing edge `j`, you would call
```c++
Parents({1, j}, dim, element_list);
```

For the neighbors, the signature could look like
```c++
/// Return the indices of all dim-dimensional neighbors of the given
/// entity, sharing a sharing_dim dimensional interface
Neighbors(Entity entity, int dim, int sharing_dim, Array<int> &ids);
```

This way, the face, edge, and vertex neighbors of an element would be obtained by
```c++
Neighbors({dim, i}, dim, dim-1, face_neighbors);
Neighbors({dim, i}, dim, 1, edge_neighbors);
Neighbors({dim, i}, dim, 0, vertex_neighbors);
```
It also enables more general queries, like give me all edges sharing a vertex with the given element
```c++
Neighbors({dim, i}, 1, 0, edge_list);
```

As an aside, another reason why I tend to prefer an approach like the above is that it avoids the ambiguity around the meaning of "faces" that is already present in MFEM: `GetNFaces` returns the number of 2D entities in a 3D mesh, and `GetNumFaces returns the number of `dim-1` entities in a `dim` mesh. When simply referring to entities by the integer dimension, there is no such ambiguity.

acfisher: I see what you mean here.  How would you handle the covered by and touched by situations?
ParentsCoveredByChildren
ParentsTouchedByChildren
??

acfisher: And as far as the entities go what about boundary entities.

pazner: > I see what you mean here. How would you handle the covered by and touched by situations?
> ParentsCoveredByChildren
> ParentsTouchedByChildren
> ??

Right, those are somewhat different queries, it's something like `ParentsOfAny`, which would take a list of entities, and return a list of entities that contain *any* of the given ones, and `ParentsOfAll`, which would take a list of entities, and return a list of entities that contain *all* of the given ones.

pazner: In terms of boundary entities, in MFEM only codimension-1 entities can be marked as boundary elements, which might be a bit of a restriction. In that sense, in a 3D mesh, MFEM doesn't really have the concept of a boundary edge or vertex.

acfisher: The boundary entities can get a flag I guess {dim, i, bdrflag}.  Then we could try breaking the query into 3 pieces the kind of entity you want, the kind of entity you have, and the relationship you want to expose.

acfisher:
One thing that @samuelpmishLLNL brought up is the possibility of GPU execution of these.  While I don't want to handle that in this PR I'll want to make sure we can batch these calls in a reasonable fashion later

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,112 @@
...
+   }
+
+   struct EntityIDs
+   {
+      int dim;
+      Array<int> ids;
+      bool boundary;
+   }
+
+   //Dimension independent interface
```
acfisher, line 25: Hopefully converging on the general interface.  The question remains, do we want the call signatures of the named methods in addition to these.

acfisher: Any more comments on this interface?  If not, I would like to start writing toward it.

jandrej:
I like the entity approach (obviously). The function names don't make sense though.

`ChildrenInEntity`, there is no children in an entity. There can be children of an entity. 

`ParentsTouchedByEntities` looks like a squeezed in function to replicate a specialized query and not a generic topological operation. It is also ambiguous. Following situation in a two quadrilateral mesh where I query v2 with this. Does it return e1, e2, q1 and q2? Because topologically all of these are parents of v2.
```
           v2            
+-----------+-----------+
|   e1      |    e2     |
|           |           |
|   q1      |    q2     |
+-----------+-----------+
```

Imho we could improve the interface robustness a lot by introducing specific types for the specialized functions. See [here](https://godbolt.org/z/doaoG9TGv) for an example (courtesy of @samuelpmishLLNL)
```
enum class EntityType { Vertex, Edge, Face, Element };

template <EntityType T>
struct Index {
    int index;
};

using VertexID = Index<EntityType::Vertex>;
using EdgeID = Index<EntityType::Edge>;
using FaceID = Index<EntityType::Face>;
using ElementID = Index<EntityType::Element>;

bool IsFaceInElem(ElementID elem, FaceID face);

int main()
{
    ElementID eid = ElementID{1};
    FaceID fid = FaceID{5};

    IsFaceInElem(eid, fid); // fine
    IsFaceInElem(fid, eid); // doesn't compile. would compile and produce wrong behavior if just using "int"
}
```

The naming of functions is backwards in terms of logic. `bool IsFaceInElem(int elem, int face);` feels off. 

It itches me to say that the specialized interface is not my cup of tea. We have all the machinery available to make the interface intuitive to use. The entity based queries should be thought through in terms of naming.

pazner:
> `ChildrenInEntity`, there is no children in an entity. There can be children of an entity.

I agree, `ChildrenOfEntity` is better. Or just `Children`…

> `ParentsTouchedByEntities` looks like a squeezed in function to replicate a specialized query and not a generic topological operation. It is also ambiguous. Following situation in a two quadrilateral mesh where I query v2 with this. Does it return e1, e2, q1 and q2? Because topologically all of these are parents of v2.

My understanding (correct me if I'm wrong), is that it will return `e1, e2` if you request entities of dimension 1, and `q1, q2` if you request entities of dimension 2.

> ```c++
> enum class EntityType { Vertex, Edge, Face, Element };
> ```

I think (and my motivation for putting forward the more dimension-generic interface) we need to be careful about special-casing things like vertices, edges, etc. In a 2D mesh, are elements and faces the same thing? Why not just parameterize on the integer dimension of the entity? In which case the indices that are proposed are again just pairs of two integers (in your version, one of them is a compile-time constant, which could be useful to consider).

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,95 @@
...
+   Table *FaceToEdgeTable();
+   Table *FaceToVertTable();
+   Table *EdgeToVertTable();
+   Table *EdgeToFaceTable();
+   Table *EdgeToEdgeTable();
+   Table *EdgeToVertTable();
+   Table *VertToVertTable();
+   Table *VertToFaceTable();
+   Table *VertToEdgeTable();
+   Table *VertToVertTable();
```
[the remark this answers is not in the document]
acfisher: Agreed, A few of them are currently exposed in the Mesh class and I plan to keep those exposed and deprecate them.  I can make them private here and friend the mesh class.

acfisher:
> I like the entity approach (obviously). The function names don't make sense though.
> 
> `ChildrenInEntity`, there is no children in an entity. There can be children of an entity.

Agreed, I think "Of" makes more sense here.

> `ParentsTouchedByEntities` looks like a squeezed in function to replicate a specialized query and not a generic topological operation. It is also ambiguous. Following situation in a two quadrilateral mesh where I query v2 with this. Does it return e1, e2, q1 and q2? Because topologically all of these are parents of v2.
Yeah, naming is hard.  What I am driving at is given a set of child entities return the set of entities of a particular dimension that have have parent relationships to those children.
I thought of ParentsOfAnyEntity but that sounds like everybody
ParentsOfAnyEntityInSet?
ParentsOfEntitySet?

> Imho we could improve the interface robustness a lot by introducing specific types for the specialized functions. See [here](https://godbolt.org/z/doaoG9TGv) for an example (courtesy of @samuelpmishLLNL)
I agree that having specific types for the different kinds of entities would make things more robust.  However, that will require a significant rework of the Mesh, NCMesh, ParMesh, and ParNCMesh classes include significant changes to interfaces that users are relying on.  I agree that we should look into reworking these interfaces, but that is a very large exercise that we have to get a lot of people on board with, and thus beyond the scope of this PR.


> The naming of functions is backwards in terms of logic. `bool IsFaceInElem(int elem, int face);` feels off.
> 
> It itches me to say that the specialized interface is not my cup of tea. We have all the machinery available to make the interface intuitive to use. The entity based queries should be thought through in terms of naming.
You all are convincing me to drop the specialized interface.  I thought some of the users would want to have it, but so far I haven't run into one that really wants it.  If we run into some later I'll go ahead and add them.

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,34 @@
...
+   }
+
+   //Dimension independent interface
+   bool        IsChild(EntityID parent, EntityID child);
+   Array<int> &Neighbors(EntityID entity, int shared_dim);
+   Array<int> &ChildrenOfEntity(EntityID parent, int child_dim);
+   Array<int> &ChildrenOfEntities(EntityIDs parents, int child_dim);
+   Array<int> &ParentsOfEntity(EntityID child, int parent_dim);
+   Array<int> &ParentsOfAnyEntities(EntityIDs children, int parent_dim);
+   Array<int> &ParentsCoveredByEntities(EntityIDs children, int parent_dim);
```
termi-official, line 29: Sorry, I am a bit late to the party. If I understand correctly, then this is going to replace the internal topology interface in the mesh?

Reading trough the existing discussion the naming makes sense to me. However, I think the names `parent` and `children` might be confusing when keeping in mind that we also have NC meshes. Not super sure if this really could be an issue.

I also want to quickly comment on neighborhoods, since I had multiple discussion on this topic in my group, "neighbors" can be ambiguous in non-simplical elements. Let us take for example vertex neighborhoods in quadrilaterals on vertex 2 within the following image
```
1--2--3
|  |  |
3--4--5
```
If we define neighborhoods over the edges, then the neighborhood would be {1,3,4}. If we define the neighborhood over elements, then we obtain {1,2,3,4,5,6}. Again, not super sure how problematic this is in practice, but I just wanted to expose this possible issue, since I faced it in my group.

pazner: Good points. I think your point about the neighbors is addressed in my comment here: https://github.com/mfem/mfem/pull/2951#discussion_r840772947.

The point is that the neighborhood function would take as a parameter the "shared entity dimension", i.e. you could ask for all "edge neighbor vertices of vertex 2" or all "element neighbor vertices of vertex 2" as different queries.

@acfisher, I think the `Neighbors` function should have an additional `neighbor_dim` argument.

termi-official: Ah, yes, you are right. I read it and somehow managed to forget it when I arrived at writing the comment, sorry.

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,34 @@
...
+   struct EntityIDs
+   {
+      int dim;
+      Array<int> ids;
+      bool boundary;
+   }
+
+   //Dimension independent interface
+   bool        IsChild(EntityID parent, EntityID child);
+   Array<int> &Neighbors(EntityID entity, int shared_dim);
```
rcarson3, line 24: So, I unfortunately missed the discussion where this was initially discussed. Therefore, I'm trying to glean what this class is meant to provide based on the class names and comments from here. However, it seems like this class is meant to provide a nice way to query what elements / faces / edges / vertices neighbors are. So, it seems like this class could be very useful in building up graph analysis on mesh related data. Therefore, I have a few questions related to this class.

1. Does this class easily allow us to query how a boundary element / quantities are mapped to the normal mesh quantities. I currently have no idea how to do this in a sane manner using the existing mesh member functions, and I'd really like there to be one...
2. Would these methods return local element / face / edge / vertex numbering / neighbors? I'm only asking as I'd also be interested in the global info as well for neighbors that cross processor boundaries. Since, this info would make it simpler for me to build up a global graph to perform analysis such as connected component analysis on the global system. I guess if I just had the global IDs for the elements / vertices and had an easy to query which processors those were shared on then I could still do the global graph analysis that I might want to do.

I'm really interested in the answers to question 2 given part of my PhD thesis made us of these sorts of analysis to pry out non-intuitive insights into complicated problems.

mesh/mesh_connections.hpp
acfisher: Yeah, I realize now that I had been assuming entity_dim+1 for vertices and entity_dim-1 for other entities.  I like the idea of just naming the dimension you neighboring across.

mesh/mesh_connections.hpp
acfisher: Yes, the intent is to be able to query on boundary entities as well.  I'll put some real documentation up to make this clear now that we converging on the interface.

This object is entirely local however as it interfaces with Mesh and not ParMesh.  I have some ideas for similar features for ParMesh, but that will have to wait for another day.

acfisher: @rcarson3 an incoming update with a starting point for the documentation should make this more clear.  Stay tuned today.

acfisher:
@rcarson3 Have a look at the doxygen I added.  Please let me know if anything is unclear of if there is an unaddressed need for mesh connections.

acfisher:
All, there is one thing I want to bring up for discussion.  I currently have plural and singular calls like ChildrenOfEntity, and ChildrenOfEntities.  It would be possible to fold all of these into into the plural version and removing the EntityID struct, but it comes with the overhead of always dealing with arrays even when you are using the singular versions.  My sense that the singular versions will be called over and over and that overhead is unacceptable, but I'd like to hear other thoughts on it if possible.

jandrej:
I'd vote for plural only. Also, `MeshEntityIDs` should really be a namespace object s.t. `Mesh::EntityID` is valid.

I'm not convinced on the need of `EntityIDs`. `std::vector<EntityID>` (or `mfem::Array` if it really has to) seems more natural to me.

IMHO an `EntityID` is unique. That means if all vertices are Entity ID's from 1-10, then there can't be an entity that describes a segment with ID 1-10. @pazner is there a restriction I'm missing?

pazner:
> I'm not convinced on the need of `EntityIDs`. `std::vector<EntityID>` (or `mfem::Array` if it really has to) seems more natural to me.
> 
> IMHO an `EntityID` is unique. That means if all vertices are Entity ID's from 1-10, then there can't be an entity that describes a segment with ID 1-10. @pazner is there a restriction I'm missing?

I think the intent is that all entities in `EntityIDs` share the same dimension, but maybe an array or vector of IDs would also be fine…

data/ref-cube.mesh
```diff
@@ -11,6 +11,21 @@ MFEM mesh v1.0
...
+#   |   4---|---5
+#   |  /    |  /
+#   | /     | /
+#   |/      |/
+#   0-------1
+#
+#   y
+#   | z
+#   |/
+#   *--x
```
tzanio, line 28: I have always thought of the axes as

```suggestion
#   z
#   | y
#   |/
#   *--x
```

and there may be places in the code that already use that convention. (This is also the "default" in GLVis)

acfisher: I actually did the axes differently in a couple different places because I was having trouble with the ascii art 😅.  Let me see if I can turn them and still get the art right.  Any good ASCII artists out there?

tzanio: Can't say I am any good, but I will be certainly happy to try to help 😁 

@jakubcerveny has some similar diagrams here:

https://github.com/mfem/mfem/blob/master/mesh/ncmesh.cpp#L1321-L1332

```c++
// Tetrahedron vertex numbering:
//
//    3
//     +                         Faces: 0 back (1, 2, 3)
//     |\\_                             1 left (0, 3, 2)
//     ||  \_                           2 front (0, 1, 3)
//     | \   \_                         3 bottom (0, 1, 2)
//     |  +__  \_
//     | /2  \__ \_       Z  Y
//     |/       \__\      | /
//     +------------+     *--X
//    0              1
```

https://github.com/mfem/mfem/blob/master/mesh/ncmesh.cpp#L931-L943

```c++
// Vertex numbering is assumed to be as follows:
//
//       7             6
//        +-----------+                Faces: 0 bottom
//       /|          /|                       1 front
//    4 / |       5 / |                       2 right
//     +-----------+  |                       3 back
//     |  |        |  |                       4 left
//     |  +--------|--+                       5 top
//     | / 3       | / 2       Z Y
//     |/          |/          |/
//     +-----------+           *--X
//    0             1
```

https://github.com/mfem/mfem/blob/master/mesh/ncmesh.cpp#L1196-L1208

```c++
// Wedge vertex numbering:
//
//          5
//         _+_
//       _/ | \_                    Faces: 0 bottom
//    3 /   |   \ 4                        1 top
//     +---------+                         2 front
//     |    |    |                         3 right (1 2 5 4)
//     |   _+_   |                         4 left (2 0 3 5)
//     | _/ 2 \_ |           Z  Y
//     |/       \|           | /
//     +---------+           *--X
//    0           1
```

Here is an example from hypre: https://github.com/hypre-space/hypre/blob/master/src/examples/ex15.c#L495-L509

```c++
int ordering[48] =
{
    0,  0, -1, -1,    /* x-edge [0]-[1]                  */
    1, +1,  0, -1,    /* y-edge [1]-[2]                  */
    0,  0, +1, -1,    /* x-edge [3]-[2]     [7]------[6] */
    1, -1,  0, -1,    /* y-edge [0]-[3]     /|       /|  */
    0,  0, -1, +1,    /* x-edge [4]-[5]    / |      / |  */
    1, +1,  0, +1,    /* y-edge [5]-[6]  [4]------[5] |  */
    0,  0, +1, +1,    /* x-edge [7]-[6]   | [3]----|-[2] */
    1, -1,  0, +1,    /* y-edge [4]-[7]   | /      | /   */
    2, -1, -1,  0,    /* z-edge [0]-[4]   |/       |/    */
    2, +1, -1,  0,    /* z-edge [1]-[5]  [0]------[1]    */
    2, +1, +1,  0,    /* z-edge [2]-[6]                  */
    2, -1, +1,  0     /* z-edge [3]-[7]                  */
};
```

acfisher:
> I'd vote for plural only. Also, `MeshEntityIDs` should really be a namespace object s.t. `Mesh::EntityID` is valid.
> 
> I'm not convinced on the need of `EntityIDs`. `std::vector<EntityID>` (or `mfem::Array` if it really has to) seems more natural to me.
> 
> IMHO an `EntityID` is unique. That means if all vertices are Entity ID's from 1-10, then there can't be an entity that describes a segment with ID 1-10. @pazner is there a restriction I'm missing?
This is not how the index numbers work for the existing entity IDs.  There are seperate arrays and tables of the mesh entities.  This is why we need the dim/boundary tags to disambiguate them.

acfisher:
> > I'm not convinced on the need of `EntityIDs`. `std::vector<EntityID>` (or `mfem::Array` if it really has to) seems more natural to me.
> > IMHO an `EntityID` is unique. That means if all vertices are Entity ID's from 1-10, then there can't be an entity that describes a segment with ID 1-10. @pazner is there a restriction I'm missing?
> 
> I think the intent is that all entities in `EntityIDs` share the same dimension, but maybe an array or vector of IDs would also be fine…

That is indeed the intent of the EntityIDs struct the dim and boundary tags apply to all of the IDs on the list.  This avoids using 2x-3x the storage/bandwidth when processing lists of Entity IDs.  Maybe I should call it EntityIDList?

I also agree that the EntityID struct definitions belong in Mesh.  I'll put them there once we are settled on the interface.

data/ref-cube.mesh
acfisher: Thanks for the examples. I think I managed to cobble something together.  If only ascii has some 30 and 60 degree angles as well...

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,256 @@
...
+      3---4---5      +-2-+-3-+      +---+---+      +---+---+
+      |   |   |      9  10  11      | 2 | 3 |      6   |   7
+      7---8---9      +-4-+-5-+      +---+---+      +-2-+-3-+ 
+
+                            parent          children
+                          dim,bdry,id     dim,bdry,ids
+      ChildrenOfEntities({1,false,{2,3}}, {0,false,{}}) ->  {0,false,{3,4,5}}
+      ChildrenOfEntities({2,false,{0,2}}, {1,true ,{}}) ->  {1,true ,{4,6}}
+      ChildrenOfEntities({1,true ,{4,7}}, {0,false,{}}) ->  {0,false,{0,3,5,9}}**/
+   void ChildrenOfEntities(const EntityIDs &parents, EntityIDs &children);
```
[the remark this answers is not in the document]
YohannDudouit: What would be the semantic of `parents.boundary`? Shouldn't we have a `const Array<EntityID> &parents` instead? Same for children, wouldn't it make more sense to have `Array<EntityID> &children` or do we want to always filter them by taking `children.boundary` as input?

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,256 @@
...
+      |   |   |      6   7   8      | 0 | 1 |      4   |   5
+      3---4---5      +-2-+-3-+      +---+---+      +---+---+
+      |   |   |      9  10  11      | 2 | 3 |      6   |   7
+      7---8---9      +-4-+-5-+      +---+---+      +-2-+-3-+ 
+
+                        child        parents
+                      dim,bdry,id  dim,bdry,ids
+      ParentsOfEntity({1,false,7}, {2,false,{}}) ->  {2,false,{0,1}}
+      ParentsOfEntity({1,true ,5}, {2,false,{}}) ->  {2,false,{1}}**/
+   void ParentsOfEntity(const EntityID &child, EntityIDs &parents);
```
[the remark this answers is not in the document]
YohannDudouit: Does `parents.boundary` has a semantic as input?

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,256 @@
...
+      3---4---5      +-2-+-3-+      +---+---+      +---+---+
+      |   |   |      9  10  11      | 2 | 3 |      6   |   7
+      7---8---9      +-4-+-5-+      +---+---+      +-2-+-3-+ 
+
+                                  children              parents
+                                dim,bdry,ids          dim,bdry,ids
+      ParentsCoveredByEntities({1,false,{0,1,2,6,7}}, {2,false,{}}) ->  {2,false,{0}}, element 1 is not covered
+      ParentsCoveredByEntities({0,false,{0,1,2,3,4}}, {2,false,{}}) ->  {2,false,{0}}
+      ParentsCoveredByEntities({0,false,{0,1,2,3,7}}, {1,true ,{}}) ->  {1,true ,{0,1,4,6}}**/   
+   void ParentsCoveredByEntities(const EntityIDs &children, EntityIDs &parents);
```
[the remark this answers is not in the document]
YohannDudouit: I'm not sure I understand the example:
```
ParentsCoveredByEntities({1,false,{0,1,2,6,7}}, {2,false,{}}) -> {2,false,{0}}, element 1 is not covered
```
Why is `0` part of the result? Since `parents.boundary == false`, I would expect the result to be `{}`, since `0` is a boundary element.

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,256 @@
...
+    These kinds are represented by the @a dim and @a boundary members.  For
+    instance a mesh edge on the boundary will have dim=1 and boundary=true.  
+    This struct is used to describe the ID numbers recieved and returned by the 
+    \ref MeshConnections class.**/
+struct MeshEntityID
+{
+   int dim;
+   bool boundary;
+   int id;
+}
```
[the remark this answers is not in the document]
acfisher: Agreed, I think index is a better name and is actually what is used in Mesh.hpp.  I disagree on the boundary comment though.  There are boundary entities floating around in the mesh and they have their own 0..n index space.  In order to uniquely define an index we need to know if we are talking about the boundary version hence the flag.  To be fair there is only 1 boundary index space currently (the mesh_dim-1 case), so there are unused combos in that interface.

mesh/mesh_connections.hpp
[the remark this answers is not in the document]
acfisher: Yes

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,256 @@
...
+      7---8---9      +-4-+-5-+      +---+---+      +-2-+-3-+ 
+
+                            entity                         neighbors
+                         dim,bdry,ids                    dim,bdry,ids
+      NeighborsOfEntity({0,false,1},1,{0,false,{}}) ->  {0,false,{0,2,4}}
+      NeighborsOfEntity({0,false,1},2,{0,false,{}}) ->  {0,false,{0,1,3,4}}
+      NeighborsOfEntity({1,true ,5},0,{1,false,{}}) ->  {0,false,{1,11}},   edge numbering
+      NeighborsOfEntity({1,true ,5},0,{1,true ,{}}) ->  {0,true,{1,7}},     boundary element numbering
+      NeighborsOfEntity({0,false,1},2,{1,false,{}}) ->  error, entity.dim != neighbors.dim**/  
+   void NeighborsOfEntity(const EntityID &entity, int shared_dim, EntityIDs &neighbors);
```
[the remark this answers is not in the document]
acfisher: Because of the lazy table generation, these can indeed change the object.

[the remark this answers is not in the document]
termi-official: Just want to leave a design consideration here: I think you could also mark the caches (i.e. tables) mutable, so the method can be marked const, but still mutate the underlying data structure as a side effect. However, I am not sure what would be the better choice.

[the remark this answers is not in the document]
acfisher: Yeah, I am aware of that pattern and personally loath it.  If I am going to go through the trouble of arranging const correctness, I want the const keyword to 'really' mean it.  If it is only going to kinda mean it I would rather throw const correctness in the dustbin.  I'm open to being convinced that the pattern has value if folks have a good argument for it.

acfisher:
Alright, I think the interface has settled to the point that I can start writing tests to it.  Hopefully any further changes will be minor.

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,256 @@
...
+   /// 2D array of tables describing connections between boundary elements and other entities
+   Array<Array<Table*>> BT;
+
+   public:
+   /** @brief Sets the Mesh object connections are defined on. */
+   /** @note When this method is called, no connection tables
+       are generated so they will all be null.  They will be generated
+       lazily as access is needed.**/
+   MeshConnections(Mesh *m);
+
```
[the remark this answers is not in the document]
acfisher: I'm not sure what you mean about the boundary attribute being  not straightforward to provide.  To index into an arbitrary entity you need the dimension of the entity, whether it is boundary entity, and the index number.  So you can index it with an EntityIndex struct which has all 3 of those.  Anything less than than and you don't always have enough info to refer to the correct entity.  I think I am missing what you are saying, can you elaborate?

mesh/mesh_connections.hpp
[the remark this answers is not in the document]
pazner: I think the motivation for having the internal tables be mutable and the query functions const is something like the following:

* If the tables are not mutable, then `NeighborsOfEntity` cannot be const
* If `NeighborsOfEntity` is not const, then it cannot be called on a non-const `MeshConnections` object
* Presumably the `MeshConnections` object is obtained through an an accessor in the `Mesh` object. A non-const `MeshConnections` cannot be returned from a const Mesh.
* Consequently, none of these topological queries can be made starting from a `const` Mesh. This seems like a limitation, there are many use-cases where one wants to query the mesh topology but not modify (e.g. refine or transform) the mesh.

I think it is better in this case to consider the distinction between "logical" and "physical" state when it comes to const-correctness. Lazily constructing the tables does not modify any externally observable state of the object. There is no change to the mesh topology, geometry, or other "intrinsic" properties. It can be considered simply an implementation detail — we cache (memoize) the connectivities as an optimization.

From the point of view of user code, having a `const Mesh` should mean that the mesh itself can be queried, but not modified, and a `Mesh` object should only be passed to a function if there is some reason to really modify the mesh (do some adaptivity, refinement, etc.). If these internal state changes require using a `Mesh` instead of `const Mesh`, then basically everyone needs to use non-const `Mesh` all the time, and the distinction becomes meaningless.

There are some good explanations in the official C++ FAQ from the standards committee:
* [Should the constness of my public member functions be based on what the method does to the object’s logical state, or physical state?](https://isocpp.org/wiki/faq/const-correctness#logical-vs-physical-const)
* [How can it help me design better classes if I distinguish logical state from physical state?](https://isocpp.org/wiki/faq/const-correctness#logical-vs-physical-state)
* [What do I do if I want a const member function to make an “invisible” change to a data member?](https://isocpp.org/wiki/faq/const-correctness#mutable-data-members)

mesh/mesh_connections.hpp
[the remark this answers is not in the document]
YohannDudouit: I think I am getting confused in the different types of indices that we have. In `Mesh::FaceInformation` the `index` is independent of boundary or not, it is the "global face index". If the `MeshConnections` faces use different indices from the ones in `Mesh::FaceInformation`, that could make things quite confusing for users.

The problem I think is that there is different kinds of indices associated to the same mesh entity, e.g boundary faces have either "boundary face index" and "global face index". Other functions have index type requirements implicitly suggested by the function name, e.g. `Mesh::GetBdrAttribute(int i)` expects a "boundary face index" whereas `Mesh::GetAttribute(int i)` expects a global index. According to the methods used, users may either need the "boundary" index or the "global" index of boundary faces, e.g treating boundary conditions, or treating DG methods.

What I would suggest to consider in a future PR is to type the "boundary face indices":
```c++
struct BoundaryFaceIndex
{
   int index.

   explicit BoundaryFaceIndex(int index): index(index) { }
   
   int operator int()
   {
      return index;
   }
};
```
Then we can apply the change everywhere, for instance the interface of
```c++
Mesh::GetBdrAttribute(int i)
```
to
```c++
Mesh::GetBdrAttribute(const BoundaryFaceIndex &i)
```
and deprecate the previous version.
I would also expose methods that maps both ways "boundary face index" and "global face index", for instance
```c++
std::optional<BoundaryFaceIndex> Mesh::GetBoundaryFaceIndex(int global_face_index) const
```
and
```c++
int  Mesh::GetGlobalFaceIndex(BoundaryFaceIndex bdr_face_index) const
```
.

Therefore, I am wondering if we wouldn't want to always return the "global face index" (even for boundary faces) in all the `MeshConnections` methods, to always return indices of the same "kind", and let users be responsible for calling `std::optional<BoundaryFaceIndex> Mesh::GetBoundaryFaceIndex(int global_face_index)`.

mesh/mesh_connections.hpp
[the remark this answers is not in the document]
acfisher: Fair enough, I could see someone wanting to make a const Mesh, and that's good enough for me.

general/array.hpp
```diff
@@ -92,6 +93,10 @@ class Array
    template <typename CT, int N>
    explicit inline Array(const CT (&values)[N]);
 
+   /// Initilizer list contructor for syntax like Array<int> a = {1,2,3,4,5};
+   inline Array(std::initializer_list<T> list)
```
acfisher, line 97: I've never used one of these before so someone should make sure I didn't munge something that my quick test didn't pick up.

mesh/mesh_connections.hpp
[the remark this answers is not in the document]
YohannDudouit: I think we should make sure we agree on what entities are part of the boundary (I think this was my mistake in some of my previous comments).
- Is the boundary only described as set of the boundary faces?
- Are the elements containing these faces boundary elements?
- Should there be boundary edges for edges included in boundary faces? Same for vertices?
- Does it make sense to have isolated boundary edges in a 3D mesh?

mesh/mesh_connections.hpp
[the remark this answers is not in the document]
acfisher: Yeah I haven't changed anything about how anything is numbered in the Mesh class here.  I just rounded up all the things that were using the tables and put them in one place with a consistent interface.  After we get this in there I think we should take a long hard look at all of the indices we have floating around in there and unifying them.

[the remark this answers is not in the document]
YohannDudouit: My point was that even if the tables are working with both "global face indices" and "boundary face indices", I think we should unify the interface and only use "global face indices", and treat the "boundary face index" as an implementation detail. I think it is too bug-prone to mix "global face indices" for non-boundary faces with "boundary face indices" for boundary faces. But it also depends on users' usage, I don't know if it is more natural to manipulate "boundary face indices" or "global face indices" for boundary faces, my intuition is that it depends on the situation, therefore a unified interface only on "global face indices" makes more sense to me (with the proposed utility functions to go between "boundary face indices" and "global face indices). Once we only use "global face indices" then the `boundary` attribute becomes obsolete in many cases I think.

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,276 @@
...
+class Mesh;
+
+/** @brief Class that exposes connections between Mesh entities such as
+    vertices, edges, faces, and volumes.  This is done without assuming 
+    the elements have a particular dimension or type and will support 
+    dimensions > 3 in the future.  EntityIndex numbers are described in the 
+    \ref EntityIndex and \ref EntityIndices sturcts.  The connections are defined 
+    on the Mesh object given at the time of construction, and Table objects are built 
+    lazily as they are required so the first query of a paticular kind may be a lot slower
+    than the following queries.  It is worth noting that the our naming convention of child
```
YohannDudouit, line 31: We need to choose between "the" and "our":
```suggestion
    than the following queries.  It is worth noting that our naming convention of child
```

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,276 @@
...
+      bool boundary;
+      int index;
+   };
+
+   /** @brief Struct representing plural index numbers of geometrical mesh entities of the 
+       same kind (dimension and boundary).  In MFEM meshes there is currently support for 5 
+       (4 in 2D) entity kinds that all have their own index numberings:  vertices, edges, faces, 
+       elements, and boundary elements.  These types are represented by the @a dim and @a boundary 
+       members.  For instance faces in a 3D mesh interior will have dim=2 and boundary=false.
+       This struct is used to describe the index numbers recieved and returned by the
```
YohannDudouit, line 70:
```suggestion
       This struct is used to describe the index numbers received and returned by the 
```

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,276 @@
...
+   void NeighborsOfEntities(const EntityIndices &entities, int shared_dim, EntityIndices &neighbors);   
+
+   private:
+   /// Raw table access
+   Table* GetTable(int row_dim, bool row_bndry, int col_dim, bool col_bndry);
+};
+
+}
+
+#endif
```
YohannDudouit, line 276:
```suggestion
#endif

```

mesh/mesh.hpp
```diff
@@ -57,11 +58,16 @@ class Mesh
...
    friend class NURBSExtension;
+   friend class MeshConnections;
 
 #ifdef MFEM_USE_ADIOS2
    friend class adios2stream;
 #endif
 
+public:
+
+   MeshConnections connect;
```
YohannDudouit, line 69: I was thinking that this object would be created by the users when needed:
```c++
const Mesh mesh(...);
const MeshConnections mesh_connections(mesh);
```
But composition can work too...

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,276 @@
...
+{
+
+class Table;
+class Mesh;
+
+/** @brief Class that exposes connections between Mesh entities such as
+    vertices, edges, faces, and volumes.  This is done without assuming 
+    the elements have a particular dimension or type and will support 
+    dimensions > 3 in the future.  EntityIndex numbers are described in the 
+    \ref EntityIndex and \ref EntityIndices sturcts.  The connections are defined
```
YohannDudouit, line 28:
```suggestion
    \ref EntityIndex and \ref EntityIndices structs.  The connections are defined 
```

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,276 @@
...
+/** @brief Class that exposes connections between Mesh entities such as
+    vertices, edges, faces, and volumes.  This is done without assuming 
+    the elements have a particular dimension or type and will support 
+    dimensions > 3 in the future.  EntityIndex numbers are described in the 
+    \ref EntityIndex and \ref EntityIndices sturcts.  The connections are defined 
+    on the Mesh object given at the time of construction, and Table objects are built 
+    lazily as they are required so the first query of a paticular kind may be a lot slower
+    than the following queries.  It is worth noting that the our naming convention of child
+    and parent connections actually describes ancestor/decendent relationship eg. a node can 
+    e a "child" of edges, faces, and volumes.**/
```
YohannDudouit, line 33:
```suggestion
    be a "child" of edges, faces, and volumes.**/
```

mesh/mesh_connections.hpp
```diff
@@ -0,0 +1,276 @@
...
+       different kinds of entities (0D, 1D, 2D, 3D, 2D Boundary) respectively.  For example 
+       the Table mapping Vertices to Boundary Elements/Faces on a 3D mesh would be at T[0][3]*/
+   std::vector<std::vector<Table*>> T;
+
+
+
+   public:
+   /** @brief Struct representing the index number of a geometrical mesh entity.  In 
+       MFEM meshes there is currently support for 5 entity kinds that all have their
+       own index numberings:  vertices, edges, faces, elements, and boundary elements.
```
YohannDudouit, line 53: Should we replace "boundary elements" with "boundary faces"?

mesh/mesh_connections.hpp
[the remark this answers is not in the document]
YohannDudouit: I am having issues finding this comment (I remember it though), did you change it or did it get lost?

mesh/mesh.hpp
acfisher: Yeah, I want to put it here and deprecate the existing interfaces to the tables in Mesh.

mesh/mesh_connections.hpp
acfisher: I'm trying to avoid using the term faces to represent both the 2D things in a 3D mesh and 1D things in a 2D mesh.  Although I am using elements in 2 different ways here now.  I'll see if I an reword it in a way to avoid that.

mesh/mesh_connections.hpp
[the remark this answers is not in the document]
acfisher: We need to be able to index into the array of boundary elements.  We could special case them, but I don't think that makes the interface nicer.  We still want to be able to do queries between boundary elements and non boundary elements, so you will still need the boundary flag in both the in and out parameters of the so it would just move from the EntityIndex structs out to the method parameters.  So really it's just a matter of which numbering those queries return, the numbers that the user is more likely to need are the boundary element numbers, but I do agree that it hangs together nicely if there is only one index space for each dimensional object.  Either way I think we need mehtods to go back and forth between boundary element and standard face numberings.

mesh/mesh_connections.hpp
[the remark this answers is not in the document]
acfisher: Have a look at the comment above EntityIndex in the mesh_connection.hpp.

mesh/mesh_connections.hpp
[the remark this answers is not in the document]
YohannDudouit: > We need to be able to index into the array of boundary elements.

This is the purpose of:
```c++
std::optional<BoundaryFaceIndex> Mesh::GetBoundaryFaceIndex(int global_face_index) const

int  Mesh::GetGlobalFaceIndex(BoundaryFaceIndex bdr_face_index) const
```

>We could special case them, but I don't think that makes the interface nicer.

I think it does, because instead of mixing "boundary face indices" and "global face indices" in both the inputs and outputs, you only have "global face indices" without the need for `boundary` in `EntityID`, I think it makes the interface significantly simpler.

>We still want to be able to do queries between boundary elements and non boundary elements, so you will still need the boundary flag in both the in and out parameters of the so it would just move from the EntityIndex structs out to the method parameters.

I think there is a limitation with this, for instance when we want "either boundary or non-boundary". Also, having `boundary` as input parameter makes the interface more explicit IMHO, c.f. my remarks on the ambiguous semantic of `boundary` in some of the methods. Unless `boundary == false` means "either boundary or non-boundary".

>So really it's just a matter of which numbering those queries return, the numbers that the user is more likely to need are the boundary element numbers,

This is not so obvious to me, for instance, I had never used the "boundary face index" before, the boundaries in the DG are all handled through the "global face index". Also, users don't get access to the same functionality of MFEM with a "boundary face index" and "global face index". And finally, it is not only about the fact that the returned indices are of different kinds, but the input indices too. For instance, users that would want to iterate over the faces and look at the connectivity to apply a function `foo` would have to write code like this:
```c++
for ( int f = 0; f < fespace.GetNF(); f++ )
{
   auto face = mesh.GetFaceInformation(f);
   EntityID face_id;
   face_id.dim = mesh.Dimension() - 1;
   if ( face_id.boundary = face.IsBoundary() )
   {
      // Get the boundary face index somehow
      boundary_index = ??;
      face_id.index = boundary_index; // Warning: Using `f` would be a bug.
   }
   else
   {
      face_id.index = f;
   }
   MeshConnections mesh_conn(mesh);
   // Some function on the face
   foo( mesh_conn, face_id );
}
```
compared to
```c++
for ( int f = 0; f < fespace.GetNF(); f++ )
{
   EntityID face_id = { mesh.Dimension() - 1, f }; // boundary is unnecessary since it can be accessed through `mesh.GetFaceInformation(face_id.index).IsBoundary()`
   MeshConnections mesh_conn(mesh);
   // Some function on the face
   foo( mesh_conn, face_id );
}
```
Similar post-processing would be necessary for the output, for instance:
```c++
const Element* elem;
if ( entity_id.boundary )
{
   elem = mesh.GetBdrElement( entity_id.index );
}
else
{
   elem = mesh.GetElement( entity_id.index );
}
bar( elem );
```
compared to
```c++
bar( mesh.GetElement( entity_id.index ) );
```

>but I do agree that it hangs together nicely if there is only one index space for each dimensional object. Either way I think we need mehtods to go back and forth between boundary element and standard face numberings.

My understanding is that the non-boundary table contains the boundary elements, therefore I think the boundary table could just be used for implementing optimizations, or be exclusive to accesses with "boundary face indices".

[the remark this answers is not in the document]
acfisher: The code example you wrote up has some significant misunderstandings of how the numberings work in MFEM meshes and in my interface.  The boundary faces have both numberings floating around and given my current interface you can get at both numberings using the boundary flag.  So currently in the case where you want to run through all of the faces given the global face numbering we have:
```
for ( int f = 0; f < fespace.GetNF(); f++ )
{
   MeshConnections mesh_conn(mesh);
   EntityID face_id = { mesh.Dimension() - 1, false, f }; 
   // Some function on the face
   foo( mesh_conn, face_id );
}
```

If you want to run through all of the boundary faces in the boundary face numbering you would write:
```
for ( int f = 0; f < nbBoundaryFaces; f++ )
{
   MeshConnections mesh_conn(mesh);
   EntityID boundary_face_id = { mesh.Dimension() - 1, true, f }; 
   // Some function on the face
   foo( mesh_conn, boundary_face_id );
}
```

If you want to run through all of the boundary faces in the global face numbering you can write:
```
for ( int f = 0; f < nbBoundaryFaces; f++ )
{
   MeshConnections mesh_conn(mesh);
   EntityID face_id = { mesh.Dimension() - 1, false, mesh_conn.BoundaryElemToFaceIndex(f)}; 
   // Some function on the face
   foo(mesh_conn, face_id );
}
```
Or you can run through all of the faces and pick out the boundary faces in the global numbering like so:
```
for ( int f = 0; f < fespace.GetNF(); f++ )
{
   auto face = mesh.GetFaceInformation(f);
   if (face.IsBoundary() )
   {
      MeshConnections mesh_conn(mesh);
      EntityID face_id = {mesh.Dimension() - 1, false, f}
     
      // Some function on the face
      foo( mesh_conn, face_id );
  }
}
```
Perhaps the source if the confusion is the name of the flag boundary where boundary really is referring to the elements defined on those faces?  dim = mesh_dim-1, boundary = false indeed refers to all of the faces on the mesh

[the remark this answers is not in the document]
YohannDudouit: You're right, I indeed didn't understand a few things. Would it be correct to say that what the `boundary` flag says is which "kind" of index the `index` is, i.e "global index" or "boundary index"? In that case wouldn't it be more explicit/self-documented to use an `enum` type? For instance:
```c++
/** "Boundary" indices are indexing the subset of boundary elements in a mesh, and are
      typically used with functions having "Bdr" in their name, e.g. Mesh::GetBdrElement(int i).
      "Global" indices are indexing the integrity of elements of a given dimension in a mesh,
      and are typically used with functions not containing "Bdr" in their name, e.g.
      Mesh::GetElement(int i). */
enum class IndexType : bool { Boundary, Global };

struct MeshEntityID
{
   int dim;
   IndexType index_type;
   int index;
}
```

tzanio:
It depends on what applications need, but one way to extend this to AMR and parallel meshes is to use the conforming interpolations from the low-order `H1`, `H(curl)`, `H(div)` and `L2` spaces, call them `Pvert`, `Pedge`, `Pface` and `Pelem` to define `entity_entity` tables that account for the AMR/parallelization (and so are not identity).

For example in the AMR case `face_face = bool(Pface) x bool(Pface^T)` will mark all faces that share a master face together. This can then be used with `element_face` computed on the "cut" mesh to get an AMR version `element_face x face_face`.

In terms of the API, this `face_face` operation could be just another call.

acfisher:
I haven't forgotten this WIP.  I just pushed up some more touches on the MeshConnections interface, and also my first swipe at the changes to the Mesh interface.  One thing I am doing there is separating out the calls for the getting the entity id numbers and the the orientation information.  I think the orientation information could use some rework as well, but for now I am adding more documentation on them and saving that for another day.  Please have a look and let me know what you think of th proposed interface changes.

PR #2979, opened 2022-04-26
[WIP] Add essential boundary support in ex16 [ex16-add-dirichlet-bc]

giraldeau:
Setting essential boundary values in time-dependent laplace problem is
not totally trivial. We need a definitive example to show how to apply
the boundary condition in this situation. This patch adds dirichlet
value support to example 16.

 * Add --boundary-type option to select either natural (0) or dirichlet
   (1) boundary type.
 * Add --dirichet option to define the constant coefficient boundary
   value
 * Use K->FullMult() to actually apply diffusion involving the essential
   boundary DoF
 * For implicit method, build elimination matrix that combine mass and
   stiffness matrix and apply it to the right-hand side.

FIXME: The application of the essential boundary condition to the rhs in
the implicit method need to be verified, see comment in the source file.

Signed-off-by: Francis Giraldeau <francis.giraldeau@nrc-cnrc.gc.ca>

Please notice that this is not the final patch, just a first step to get feedback. Thanks!


examples/ex16.cpp
```diff
@@ -331,8 +353,17 @@ void ConductionOperator::Mult(const Vector &u, Vector &du_dt) const
...
+   // Essential boundary condition is constant, therefore its time derivative
+   // is set to zero. The system to solve involves only the mass matrix M^{-1},
+   // in consequence we use use this bilinear form to eliminate the boundary
+   // values. The full bilinear form K must be used to diffuse values on the
+   // boundary. If we use Kmat.Mult() only, the interior DoF are disconnected
+   // from the boundary.
+   K->FullMult(u, z);
    z.Neg(); // z = -z
+   du_dt.SetSubVector(ess_tdof_list, 0.0);
+   M->EliminateVDofsInRHS(ess_tdof_list, du_dt, z);
```
jandrej, line 366: This should eliminate true dofs

examples/ex16.cpp
```diff
@@ -331,8 +353,17 @@ void ConductionOperator::Mult(const Vector &u, Vector &du_dt) const
...
    M_solver.Mult(z, du_dt);
```
jandrej, line 367: It makes more sense to me to eliminate boundary conditions in `M` directly, then set `du_dt(tdofs)` to zero here.

examples/ex16.cpp
```diff
@@ -342,15 +373,34 @@ void ConductionOperator::ImplicitSolve(const double dt,
...
+   // system is T = M + dt K, and in consequence I think we have to multiply
+   // K->SpElim() by dt for consistency with the system, which is not possible
+   // with this method. In practice, both ways to apply essential bc seems to
+   // produce the same result, so I'm confused.
+   //
+   // K->EliminateVDofsInRHS(ess_tdof_list, du_dt, z);
+   // M->EliminateVDofsInRHS(ess_tdof_list, du_dt, z);
+
+   Te->AddMult(du_dt, z, -1.);
+   T->PartMult(ess_tdof_list, du_dt, z);
```
jandrej, line 403: This doesn't make sense

jandrej (changes requested):

examples/ex16.cpp
giraldeau: I see, du_dt is a true dofs vector, meaning that this would not work on non-conforming meshes.

giraldeau:
Thanks for the comments. Setting the boundary dofs in the rhs is somewhat hacky, but is working with non-conforming meshes, as this example shows: 

```
./ex16 -o 1 -r 0 -m ../mfem/data/amr-quad.mesh -visit --kappa 0.05 --boundary-type 1 --dirichlet -1
```

movie

v-dobrev:
Hi @giraldeau, is this PR ready for review from your perspective? If so, can you please remove the draft status and add the `ready-for-review` label?

Thank you for your contribution!

giraldeau:
Hello! I think we still have to settle on the proper way to address the Dirichlet BC in this case. Because we want to support both implicit and explicit integration scheme, and also cache the matrices when possible, I was not sure how to best handle the situation. 

I would set the boundary value on the r.h.s, because it is more general, but @jandrej suggested instead to rather overwriting the solution after the solve. This yield the same result when du_dt is zero, but not if the time derivative is different than zero.
 
I can rework the patch according to what is best. Thanks!

PR #2981, opened 2022-04-27
feature: adding demo container and automated builds for mfem

vsoch:
As discussed in https://github.com/LLNL/radiuss/issues/21#issuecomment-1110266585, this is a simple container build to provide mfem in spack, and then show example of running examples and miniapps. I've added a new README-DOCKER.md with full instructions for this! This is paired with an automated build so we always have an updated container base. This container setup can be extended to include other kinds of builds, or other use cases. In terms of the more advanced use cases, if we have a base container that matches in [rse-ops](https://rse-ops.github.io/docker-images/) we could easily just extend the base container to the current matrix and build fairly easily (but I haven't tested yet - it's much better to start out simpler and grow based on request!)

I haven't tested this CI workflow yet, so expect that I'll tweak it until it's working.

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

| [#2981](https://github.com/mfem/mfem/pull/2981) | @vsoch | @tzanio | @jandrej + @v-dobrev + @acfisher + @tzanio | 4/27/22 | 5/9/22 | 5/13/22 | |

**PR Checklist**
    
- [x] Code builds.
- [x] Update `CHANGELOG`?
- [x] Update CI configurations in `.github` if necessary


tzanio:
Thanks so much for doing this @vsoch!

Please mark with the `ready-for-review` label when you'd like us to start the review (see the full [CONTRIBUTING](https://github.com/mfem/mfem/blob/master/CONTRIBUTING.md) guidelines)

vsoch:
okay ready for review (and labeled). Note that the container build appears to have successfully built, and I'm not sure why it's lingering in the yellow "in progress" state - hopefully not a GitHub bug! But it's no worries because we have some items for discussion:
 
### Container build time

The container build is not quick, and I went off to have dinner when it was building locally and didn't realize it took so long! I don't think it should be triggered on PR, but rather just on a nightly basis. So my suggestion would be to remove the PR trigger (and we can confirm the build was successful here) and then have it build and deploy just on a scheduled nightly task and releases.  The drawback is that you don't have a test for the container build on every PR, but I think I have an approach (discussed next) that can account for that.

### A dependency base container?

We could actually have our cake and it it too by having a base container (nightly built) done that does this long build, and then a quicker testing container that runs on PRs that adds the new code and builds on top of it, so it should just need the time at the end to build mfem (but not the dependencies). If you are interested in this approach I can follow up with a second PR - I'll reorganize to have a docker/ directory with the Dockerfiles there, and a second Dockerfile.test that uses the base container.

### A Production Container?

Since we keep all the source code around, this container is bigger rather than smaller. I can also suggest we follow up with a multistage build that removes everything but the spack view (which some app developer would want to develop) to create a much smaller base. If you like this idea I can follow up with another PR for it, with the testing idea if you like that too :) This means in summary we will have:

- a base container (longer build time) that builds nightly and on releases, ghcr.io/mfem/mfem-ubuntu-base
- a testing container that only is built during a test to ensure the container build does not break (no URI needed)
- a production container that uses a multistage build to produce a smaller version of the first ghcr.io/mfem/mfem-ubuntu

### Changelog Update?

I didn't update the CHANGELOG because the change isn't related to anything in mfem proper, but let me know if this is appropriate to add and you would like a note added.

And just a comment, your CONTRIBUTING guide is impressive! I've never seen one so detailed, it almost made me nervous to mark this ready and request review. :laughing: I apologize in advance if I missed a detail!

Let me know your thoughts and I'll update the PR!

tzanio:

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# docker build -t ghcr.io/mfem/mfem-ubuntu .
+
+RUN apt-get update && \
+    apt-get install -y unzip gfortran && \ 
+    spack compiler find && \
+    spack install sw4lite && \
+    apt-get install -y libcurl4-openssl-dev libssl-dev && \
+    spack install liblas lapackpp && \
+    git clone --depth 1 https://github.com/geodynamics/sw4 /code
```
jandrej, line 11: This looks like an artifact. sw4 is not a dependency or the like.

jandrej (changes requested):

Dockerfile
vsoch: Yep an oversight (and I will remove) - I started from another Dockerfile I was working on.

Let me know about the things I proposed above and when I am allowed to push changes.

vsoch: Gotcha! I'll push changes (without force) as they are requested.

vsoch: okay that makes it much faster (53 minutes!) I'm still confused why the action doesn't end at the end - it's either cleaning something up or a legit GitHub bug. I can open a ticket if it doesn't seem to ever end, I've never seen this before.

jandrej: ♾️🐳

vsoch: @jandrej your emoji game is on point :point_right: :pencil2:

vsoch:
Why are the tests so slow? Just lack of GitHub runners? Are there runners somewhere else? :thinking:

tzanio:
> Why are the tests so slow? Just lack of GitHub runners? Are there runners somewhere else? 🤔

Are you referring to the tests like `branch-history` that haven't run above?

Those don't run for PRs from forks

vsoch:
oh gotcha! Yes I was looking at these 4 where it says "expected waiting for status to be reported" and wondering what's up. 
image

If you are doing these in GH-actions, you can do something like:

```yaml
if (github.repository_owner == 'mfem')
```
to only run the check given it's not a fork, that way it won't output the confusing checks above!

tzanio:
> oh gotcha! Yes I was looking at these 4 where it says "expected waiting for status to be reported" and wondering what's up. image
> 
> If you are doing these in GH-actions, you can do something like:
> 
> ```yaml
> if (github.repository_owner == 'mfem')
> ```
> 
> to only run the check given it's not a fork, that way it won't output the confusing checks above!

Hi @vsoch,

Here is why this is happening. I agree it is not optimal... maybe you can help us fix it 🤔 

Most of our GitHub CI tests are run on changes to PRs. They take a while and perform relatively extensive "smoke tests": see [builds-and-tests](https://github.com/mfem/mfem/blob/master/.github/workflows/builds-and-tests.yml#L24-L2) (which runs 10 serial/parallel, debug/optimized, mac/linux builds) and [build-analysis](https://github.com/mfem/mfem/blob/master/.github/workflows/mfem-analysis.yml#L17-L22) (which checks for compliance with our `.gitignore` policy). In terms of GH-actions both of these are setup as follows:

```
on:
  push:
    branches:
      - master
      - next
  pull_request:
```

However, there are quick tests that we will like to run as soon as possible for things like code styling or accidentally committing big files. We'd like this to be run these on each push, to each development branch in the repo (with or without PR). This is done in the [repo-check](https://github.com/mfem/mfem/blob/master/.github/workflows/repo-check.yml#L17-L18) script, which is setup in GH-actions as follows:

```
on:
  push:
```

This all works fine for branches inside the main mfem repo (the workflow we encourage), but what happens for PRs from forked branches is that `repo-check` is run in the fork: https://github.com/researchapps/mfem/runs/6199099644, while
`builds-and-tests` and `build-analysis` are run in the main repo.

Does this make sense?

Any suggestions?

vsoch:
hey @tzanio ! So I took a look, and it looks like the workflow runs because the trigger is a general "push" which of course will hit the forked repo before here. For the contributor, this means the only way it would run here is if they disable actions on their repository, which might be reasonable for long term contributors, but for a first-time / one-time contributor, it's going to be confusing (as it was for me!).

I looked into creative options - like can we have a step in the workflow that detects it's being run on the fork, and then sends a comment here to disable? Or better, to be able to disable the action from running on the fork? These API endpoints don't exist yet I don't think, and also that seems like an overly complex solution for a simple problem.

So I looked at your workflows, and I think this is my suggested solution: https://github.com/mfem/mfem/pull/2987/files#diff-b34025e1f15d85c58cfbcffcdf592e3f07c69ba75065de8c2ee1b6059cd6aec1. The underlying motivation for this check is to make sure that the repository adheres to good practices (docs, code style) and I would argue this should not run on every push, but rather be scoped to pushes to the master / next branches, and for pull requests. This means that it will reliably be run here (in the PR) and not on the user's fork, where nobody is likely to see it.

tzanio:
Thanks @vsoch, I left a few comments in https://github.com/mfem/mfem/pull/2987.

Sorry for this distraction, I will go now to review the main PR 😄

.github/workflows/build-container.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+  # Always have a base image ready to go - this is a nightly build
+  schedule:
+    - cron: 0 3 * * *
+
+  # Allow manual trigger of a build
+  workflow_dispatch:
+    inputs:
+      whatever:
+        description: 'This variable does not matter, its a GHA bug.'
```
tzanio, line 13:
```suggestion
        description: 'This variable does not matter, it's a GH-actions bug.'     
```

.github/workflows/build-container.yaml
```diff
@@ -0,0 +1,86 @@
...
+      whatever:
+        description: 'This variable does not matter, its a GHA bug.'     
+        
+  # On push to main we build and deploy images
+  push:
+    branches:
+      - master
+
+  # Do build to test works on PR
+  pull_request: []
```
tzanio, line 21: Remember to remove before merging

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM ghcr.io/rse-ops/cuda-ubuntu-20.04:cuda-11.0.3
```
tzanio, line 1: To keep the top directory (relatively) clean, is it possible to move this file to `config/` or maybe `config/docker/`?

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    spack env create -d . && \
+    echo "  concretization: together" >> spack.yaml && \
+    spack env activate . && \
+    spack develop --path /code mfem@master+examples+miniapps && \
+    spack add mfem@master+examples+miniapps && \
+    spack install
+
+# ensure mfem always on various paths
+RUN cd /opt/mfem-env && \
+    spack env activate --sh -d . >> /etc/profile.d/z10_spack_environment.sh
```
tzanio, line 26: This (`z10_spack_environment.sh`) is a bit cryptic, can you elaborate?

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY . /code
+
+# This is for a spack environment/view to install from there
+WORKDIR /opt/mfem-env
+RUN . /opt/spack/share/spack/setup-env.sh && \
+    spack env create -d . && \
+    echo "  concretization: together" >> spack.yaml && \
+    spack env activate . && \
+    spack develop --path /code mfem@master+examples+miniapps && \
+    spack add mfem@master+examples+miniapps && \
```
tzanio, line 21: 😎

README-DOCKER.md
```diff
@@ -0,0 +1,123 @@
+# Mfem Docker
```
tzanio, line 1: `MFEM` or `mfem`

README-DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+
+We provide a [Dockerfile](Dockerfile) to build an ubuntu base image. You can use
+this image for a demo of using mfem! 🎉️
+
+Updated containers are built and deployed on merges to the main branch and releases.
+
+### Usage
+
+Here is how to build the containers. Note that we build so it belongs to the same
```
tzanio, line 10: `container` or `they belong`

README-DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+PWD=/opt/mfem-env
+MANPATH=/opt/mfem-env/.spack-env/view/share/man:/opt/mfem-env/.spack-env/view/man:
+CMAKE_PREFIX_PATH=/opt/mfem-env/.spack-env/view
+SPACK_ENV=/opt/mfem-env
+ACLOCAL_PATH=/opt/mfem-env/.spack-env/view/share/aclocal
+LD_LIBRARY_PATH=/opt/mfem-env/.spack-env/view/lib:/opt/mfem-env/.spack-env/view/lib64
+PATH=/opt/mfem-env/.spack-env/view/bin:/opt/view/bin:/opt/spack/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+```
+
+#### Example Apps
```
tzanio, line 43:
```suggestion
#### Examples and Miniapps
```

tzanio (commented):
Initial review pass

Dockerfile
vsoch: Yes! So what that does it write a file into /etc/profile.d, which will be sourced when you shell into the container. This ensures that when you shell into the container, all the mfem deps / binaries / libs are on the path and ready to go, and the user doesn't need to confusingly figure it out. If you don't add this, you have to instruct the reader to cd to the environment, run the activation command, etc.

tzanio: Is the name standard or hard-coded somewhere? Sorry, I haven't seen that before.

vsoch: The files are sourced based on their ABC name, so a z10 is very likely to be sourced last. The rest of the name doesn't really matter except to identify what the file is doing.

vsoch: It's something I learned developing for Singularity - if you looked in the ~/.singularity/env folder you'd find a bunch of named files with that convention. https://github.com/sylabs/singularity/tree/vault/release-2.4/libexec/bootstrap-scripts/environment. We would add (to that base) files to source for apps and docker, and the naming would determine the order.

tzanio: Got it.

I'm learning several new things today 👍

README-DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+To shell into a container (here is an example with ubuntu):
+
+```bash
+$ docker run -it ghcr.io/mfem/mfem-ubuntu bash
+```
+
+Off the bat, you can see mfem libraries are in your path so you can jump into development:
+
+```bash
+env | grep MFEM
```
tzanio, line 30: This was `env | grep mfem` for me

README-DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+
+And yes, this is the working directory when you shell into the container! 
+You can find the examples here:
+
+
+```bash
+cd share/mfem/examples
+```
+```bash
+# ./ex0
```
tzanio, line 60: why not `$ ./ex0`?

README-DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+autodiff        gslib             navier   performance  tools
+```
+
+And an example in "toys"
+
+```bash
+cd toys
+```
+```bash
+# ./automata
```
tzanio, line 97: why not `$ ./automata` ?

README-DOCKER.md
```diff
@@ -0,0 +1,123 @@
+# mfem Docker
```
tzanio, line 1: I'm torn -- should we keep this here, or move to `config/docker` and mention it instead in `README` and/or `INSTALL`?

🤔

README-DOCKER.md
vsoch: Inside the container the hash bang (or whatever it's called, $PS1 is that symbol). Maybe it's been on twitter too much :) If you prefer the $ just for the docs I can change that.

README-DOCKER.md
vsoch: The main issue to not have it in config/docker is mostly just the person that browses to the repo seeing there is a Dockerfile! I put it in the root because of that. The alternative that I think is a good design is to have a link in the README, but y'all don't have a standard markdown README so we can't do that. So if we are able to make the existence clear somewhere, I do like having it be like:

```
config/docker/README.md
```
because it will render nicely in GitHub! But if there is no way to make reference in the root I think we should keep it there.

README-DOCKER.md
tzanio: Lets change it if you don't mind

README-DOCKER.md
vsoch: Fixed.

README-DOCKER.md
tzanio: Lets do that... it is time to change the `README` to `README.md`

😱

README-DOCKER.md
vsoch: All these should now be $ !

README-DOCKER.md
vsoch: make it so!  
image

And it is!
https://github.com/mfem/mfem/pull/2981/commits/8be4ab8d24ffe4e541adfa89a6fe00e760addd10

I preserved the nice emoji art in a comment block, here is how it looks rendered:

https://github.com/researchapps/mfem/tree/add/docker-build-for-demo

tzanio (approved):
This is really, really, nice. Thank you for taking the time to help us @vsoch! 🙏 

A few small comments and suggestions:

- Can you remove the whitespace in the new files? I tried to do that but for some reason I have difficulty pushing to your fork.

- Totally agree with your suggestion to "remove the PR trigger ... and then have it build and deploy just on a scheduled nightly task and releases."   

- Yes, we can follow up with the "dependency base" and "production" containers 👍  

- I do thing this should be mentioned in the `CHANGELOG`, because many folks will likely be interested in it. Doesn't have to be long, e.g. "Added a Dockerfile to build a simple MFEM container, see config/docker/README.md`

A few maybe not so small/future suggestions:

- Now that we have this container, we will of course want more 😄 ... sorry!

- When I tried running `mpirun -np 4 ex1p` in `/opt/mfem-env/.spack-env/view/share/mfem/examples` I got scary message, is there a better way to run parallel programs?

- Even when I set  `--allow-run-as-root`, I was still limited to 2 cores. I assume that's a container limitation. Any way to go around it?

- It will be awesome, if we can have glvis also be build in this container, and ideally be able to use it to visualize the mfem examples. Maybe it is as simple as adding `spack install glvis`?

vsoch:
> This is really, really, nice. Thank you for taking the time to help us @vsoch!

Totally! It wasn't a ton of work (and was fun) and I'm really happy to be able to help.

>    Can you remove the whitespace in the new files? I tried to do that but for some reason I have difficulty pushing to your fork.
>    I do thing this should be mentioned in the CHANGELOG, because many folks will likely be interested in it. Doesn't have to be long, e.g. "Added a Dockerfile to build a simple MFEM container, see config/docker/README.md`

I have these changes locally and will push after a few more local tests!

> When I tried running mpirun -np 4 ex1p in /opt/mfem-env/.spack-env/view/share/mfem/examples I got scary message, is there a better way to run parallel programs?
>     Even when I set --allow-run-as-root, I was still limited to 2 cores. I assume that's a container limitation. Any way to go around it?

So generally I wouldn't recommend this, and especially not using --privileged or running as root. There are ways to sort of hack it, but my suggestion would be instead to pull the image down via Singularity and do the same interaction, but with Singularity and not docker. When this is pushed to a registry we can give that a shot!

>   It will be awesome, if we can have glvis also be build in this container, and ideally be able to use it to visualize the mfem examples. Maybe it is as simple as adding spack install glvis?

Testing this now!

vsoch:
Apologies for the delay! I left the test build running, and came back to find my computer frozen after dinner. I did a re-build, and I see that glvis requires llvm which is always a PITA to build. So for this recent push https://github.com/mfem/mfem/pull/2981/commits/9cb757d7c065e425dc8480d2f5f0552b70f57e7d I've:

- updated the container name to have the "-base" suffix so it is clearly a simple base
- updated the CHANGELOG
- removed trailing whitespace

For this:

> It will be awesome, if we can have glvis also be build in this container, and ideally be able to use it to visualize the mfem examples. Maybe it is as simple as adding spack install glvis

my recommendation is that we do not put this in the base "simple" container, but rather use a prebuilt base with llvm as a multistage build (e.g., https://rse-ops.github.io/docker-images/r/ghcr.io-rse-ops-clang-ubuntu-22.04/) to make the image as small as possible and be able to run the demo. So once the containers are merged/deployed here, I will work on the following:

- a production container that is the base here, but a multistage build to make smaller, this will  be `ghcr.io/mfem/mfem-ubuntu`
- a glvis container based off of https://rse-ops.github.io/docker-images/r/ghcr.io-rse-ops-clang-ubuntu-22.04/ that will install mfem + glvis, and then multistage to make smaller
- a testing container that uses the base container to quickly run tests.

And for the above, I'll update the README in the docker folder with generic notes, and I'd also like to have an example / test of using the glvis container (when it exists) for examples with mpirun as you noted. I'll get started on all of the above as soon as this is merged (and I am awake, as I am going to sleep now). Cheers!

tzanio:
Thanks a lot @vsoch! This all sounds good. 

I am not sure why `glvis` pulls `llvm`, it shouldn't really it just needs some OpenGL libraries: https://github.com/GLVis/glvis/blob/master/INSTALL#L16-L39. 

Maybe @v-dobrev or @kanye-quest can help with the Spack build?

vsoch:
Sure, although my suggestion is the same that it doesn’t belong in the base container.

tzanio:
@jandrej and @acfisher -- can you please review when you get a chance?

jandrej (approved):

config/docker/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+```bash
+cd toys
+```
+```bash
+$ ./automata
+Options used:
+   --num-steps 16
+   --rule 90
+   --visualization
```
acfisher, line 107: Is the glvis server up and running automatically in the docker?  If not we should probably have the glvis launch in the example .

vsoch: It is not, so we likely should remove this example (let me know if this is your preference). I basically shelled inside and was trying random stuff and this seemed to run so I put it here :))

vsoch: Is there a better miniapp we can demo on the command line?

vsoch: Or can we just remove the `--visualization` flag and call it a day?

tzanio: Lets just use `./automata -no-vis` for now.

It is not the full story without the visualization, but including glvis seems more complicated.

vsoch: okay updated https://github.com/mfem/mfem/pull/2981/commits/87a937ab6142c60239421147b3f3992aa8455df5

acfisher (approved):

tzanio:

vsoch:
:crossed_fingers:

config/docker/README.md
```diff
@@ -0,0 +1,129 @@
+# mfem Docker
+
+We provide a [Dockerfile](Dockerfile) to build an ubuntu base image. You can use
+this image for a demo of using mfem! 🎉️
+
+Updated containers are built and deployed on merges to the main branch and releases.
```
tzanio, line 6: Mention here also how to trigger a manual build in CI?

README.md
```diff
@@ -7,13 +8,15 @@
...
 MFEM is a modular parallel C++ library for finite element methods. Its goal is
 to enable high-performance scalable finite element discretization research and
 application development on a wide variety of platforms, ranging from laptops to
 supercomputers.
 
 We welcome contributions and feedback from the community. Please see the file
-CONTRIBUTING.md for additional details about our development process.
+[CONTRIBUTING.md](CONTRIBUTING.md) for additional details about our development process.
 
 * For building instructions, see the file INSTALL, or type "make help".
```
tzanio, line 21: Link INSTALL, LICENSE and NOTICE, e.g.

```suggestion
* For building instructions, see the file [INSTALL](INSTALL), or type "make help".
```

README.md
```diff
@@ -22,6 +25,8 @@ CONTRIBUTING.md for additional details about our development process.
 * The best starting point for new users interested in MFEM's features is to
   review the examples and miniapps at https://mfem.org/examples.
 
+* Instructions for learning with docker are [here](config/docker)
```
tzanio, line 28:
```suggestion
* Instructions for learning with Docker are in [config/docker](config/docker).
```

tzanio:
Can you give merge `master` here and resolve the conflict in `CHANGELOG`.

Alternatively, can you give me permission to push to the branch?

vsoch:
okay updated! @tzanio I would have given you permission too but I don't see that little checkbox anywhere - I'm not totally keen on how that works because sometimes I see it and other times I do not.

tzanio:

.github/workflows/build-container.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+  # Always have a base image ready to go - this is a nightly build
+  schedule:
+    - cron: 0 3 * * *
+
+  # Allow manual trigger of a build
+  workflow_dispatch:
+    inputs:
+      whatever:
+        description: "This variable does not matter, it's a GH-actions bug."
```
tzanio, line 13: Looking at https://github.com/spack/spack/blob/999eee64b8d36b303a8b13d9d50d422120da355b/.github/workflows/build-containers.yml#L3-L20, could this just be

```suggestion
  workflow_dispatch:
```

vsoch: It should work like that, but when I've tried that it doesn't work to trigger it, I think it's a bug. That's why I added the variable.

tzanio: I tested it here: https://github.com/tzanio/gha-test/actions

vsoch: Then go for it - it could be my browser or something.

tzanio:
Thank you @vsoch 🙏 🚀

vsoch:
@tzanio do you plan to make the package public? https://github.com/mfem/mfem/pkgs/container/mfem-ubuntu-base. And if not already, it should be associated with the repo here so it shows up in the sidebar (and can be seen to be associated!)

v-dobrev:
Hi @vsoch, I made the package visible a enabled the option for packages to show up in the sidebar.

Does everything look okay now?

v-dobrev:
And by "visible" I meant public.

vsoch:
yes, looks perfect, thank you! I'll start on the other containers we discussed.

tzanio:
Looks great, thanks @vsoch and @v-dobrev !

Issue #2982, opened 2022-04-27
Transfer of overlapping fields in parallel with GSLIB crashes


termi-official:
The `FindPointsGSLIB` class seems to cause crashes (at least sometimes) when the interpolation coordinates can not be found. Since I could not find a way to reproduce this reliably within the currently provided miniapps I derived a MWE from the `pfindpts` and `field-interp` miniapps in the gslib folder.

```cpp
#include "mfem.hpp"
#include <iostream>

using namespace mfem;

void transfer_field_distributed(ParGridFunction* source_gf, ParGridFunction* target_gf, double default_value)
{
    *target_gf = default_value;

    const int vdim = source_gf->VectorDim();

    auto source_fes  = source_gf->ParFESpace();
    auto source_mesh = source_fes->GetParMesh();
    const int sdim   = source_mesh->SpaceDimension();

    auto target_fes  = target_gf->ParFESpace();
    auto target_mesh = target_fes->GetParMesh();

    const int target_NE                = target_fes->GetNE();
    const int target_nodes_per_element = target_fes->GetFE(0)->GetNodes().GetNPoints();

    // list of points where the grid function will be evaluated.
    Vector xyz(target_nodes_per_element*target_NE*sdim);
    for (int i = 0; i < target_NE; i++)
    {
        const FiniteElement *fe = target_fes->GetFE(i);
        const IntegrationRule ir = fe->GetNodes();
        ElementTransformation *et = target_fes->GetElementTransformation(i);

        // Obtain a "dimension x nodes per element" matrix with the coordinates in the columns.
        DenseMatrix pos;
        et->Transform(ir, pos);
        for(int d = 0;d < sdim; d++)
        {
            // xyz stores the nodal coordinates by dimension (i.e. XXXXXX...YYYYYY....ZZZZ....). We abuse 
            // this to construct a  memory reference into which we write when reading the matrix row.
            Vector row(xyz.GetData() + i*target_nodes_per_element + d*target_NE*target_nodes_per_element, target_nodes_per_element);
            pos.GetRow(d, row);
        }
    }

    const int nodes_cnt = target_nodes_per_element*target_NE;

    // Carrying out interpolation.
    Vector interp_vals(nodes_cnt*vdim);
    FindPointsGSLIB finder(MPI_COMM_WORLD);
    finder.SetDefaultInterpolationValue(NAN);
    finder.Interpolate(*source_mesh, xyz, *source_gf, interp_vals);
    finder.FreeData();

[...]

PR #2998, opened 2022-05-06
[WIP] GSLIB bug fix in InterpolateGeneral

kmittal2:
fix gslib bug for InterpolateGeneral when ncomp>1 and some points were not found

### DIFF
--- a/fem/gslib.cpp
+++ b/fem/gslib.cpp
@@ -837,7 +837,7 @@ void FindPointsGSLIB::InterpolateGeneral(const GridFunction &field_in,
 
             sarray_transfer(struct send_pt, sendpt, proc, 1, cr);
             sdpt = (struct send_pt *)sendpt->ptr;
-            for (int index = 0; index < nptorig; index++)
+            for (int index = 0; index < sendpt->n; index++)
             {
                int idx = sdpt->index + j*nptorig;
                field_out(idx) = sdpt->ival;


tzanio:

cjvogl:
@kmittal2, is there a particular case that I should test with this change (i.e. see that the original code fails while the new code succeeds)?  Otherwise I suppose the unit tests will do, although it seems like they didn't catch this bug in the first place?

termi-official:
I could find some time to update my framework and test this. The fix works like a charm, thanks! 

Do you think it make sense to add some unit tests on [FindPointsGSLIB](https://docs.mfem.org/html/classmfem_1_1FindPointsGSLIB.html)?

kmittal2:
> I could find some time to update my framework and test this. The fix works like a charm, thanks!
> 
> Do you think it make sense to add some unit tests on [FindPointsGSLIB](https://docs.mfem.org/html/classmfem_1_1FindPointsGSLIB.html)?

Hi @termi-official, yes I was thinking about that. I will try to add some unit tests (probably through a separate PR).

cjvogl:
Putting to-do list here for myself for documentation purposes:
- [ ] code compiles (using cmake)
- [ ] unit tests pass
- [ ] parallel unit tests pass
- [ ] code in #2982 builds and executes without error

cjvogl (approved):
Looks good to me!

atallah727 (approved):

tzanio:

PR #3006, opened 2022-05-13
[WIP] testing a alternate to repo-check-dev

adrienbernede:
This is simply a test to see if we could have all green checks, duplicated repo-check, but the tests really running once.
This is ugly, this is not desirable, but I thought it was still worth a test.

Relates to #3001


tzanio:
Thanks for trying @adrienbernede. It is unfortunate that you can't tell GutHub not to show the skipped jobs...

PR #3009, opened 2022-05-16
Simplified BC elimination in batched LOR

pazner:
I realized there was a simpler algorithm for eliminating the columns in the parallel off-diagonal blocks when doing the BC elimination in the batched LOR assembly (it should also be more efficient, but I haven't benchmarked).

May also be of interest to @YohannDudouit.


pazner:
Also, this should be more easily amenable to doing everything on device (including communication with device-aware MPI — the old method had to rely on thrust for scan operations).

@liruipeng, does the hypre code look ok here for GPU-aware MPI?

pazner:
Addressed in #3056

PR #3039, opened 2022-05-24
Compute l2 errors

atallah727:
modified the ComputeL2Error functions by adding an element array as an input so that the L2 error is only computed on those select elements

| [#3039](https://github.com/mfem/mfem/pull/3039) | @atallah727 | @tzanio | @brendankeith + @vladotomov + @rw-anderson | 5/31/22 | 8/17/22 | 8/23/22 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>

### DIFF
--- a/fem/gridfunc.hpp
+++ b/fem/gridfunc.hpp
@@ -485,7 +485,7 @@ public:
                                           const IntegrationRule *irs[] = NULL) const;
 
    /// Returns ||u_ex - u_h||_L2 for H1 or L2 elements
-   /* The elems input variable expects a list of markers:
+   /* The @a elems input variable expects a list of markers:
       an elem marker equal to 1 will compute the L2 error on that element
       an elem marker equal to 0 will not compute the L2 error on that element */
    virtual double ComputeL2Error(Coefficient &exsol,
@@ -577,7 +577,7 @@ public:
                                  const IntegrationRule *irs[] = NULL) const
    { return ComputeLpError(1.0, exsol, NULL, NULL, irs); }
 
-   /* The elems input variable expects a list of markers:
+   /* The @a elems input variable expects a list of markers:
     an elem marker equal to 1 will compute the L2 error on that element
     an elem marker equal to 0 will not compute the L2 error on that element */
    virtual double ComputeLpError(const double p, Coefficient &exsol,
--- a/fem/gridfunc.hpp
+++ b/fem/gridfunc.hpp
@@ -484,9 +484,14 @@ public:
    virtual double ComputeElementGradError(int ielem, VectorCoefficient *exgrad,
                                           const IntegrationRule *irs[] = NULL) const;
 
+  /// Returns ||u_ex - u_h||_L2 for H1 or L2 elements
+  /* The elems input variable expects a list of markers:
+     an elem marker equal to 1 will compute the L2 error on that element
+     an elem marker equal to 0 will not compute the L2 error on that element */
    virtual double ComputeL2Error(Coefficient &exsol,
                                  const IntegrationRule *irs[] = NULL,
-                                 Array<int> *elems = NULL) const;
+                                 Array<int> *elems = NULL) const
+  { return ComputeLpError(2.0, exsol, NULL, irs, elems); }
 
    virtual double ComputeL2Error(VectorCoefficient &exsol,
                                  const IntegrationRule *irs[] = NULL,
@@ -574,7 +579,8 @@ public:
 
    virtual double ComputeLpError(const double p, Coefficient &exsol,
                                  Coefficient *weight = NULL,
-                                 const IntegrationRule *irs[] = NULL) const;
+                                 const IntegrationRule *irs[] = NULL,
+				 Array<int> *elems = NULL) const;
 
    /** Compute the Lp error in each element of the mesh and store the results in
        the Vector @a error. The result should be of length number of elements,
--- a/fem/gridfunc.hpp
+++ b/fem/gridfunc.hpp
@@ -478,26 +478,26 @@ public:
 
    virtual double ComputeL2Error(Coefficient *exsol[],
                                  const IntegrationRule *irs[] = NULL,
-				 Array<int> *elems = NULL) const;
+                                 Array<int> *elems = NULL) const;
 
    /// Returns ||grad u_ex - grad u_h||_L2 in element ielem for H1 or L2 elements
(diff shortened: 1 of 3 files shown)


fem/gridfunc.hpp
```diff
@@ -476,21 +476,22 @@ class GridFunction : public Vector
...
-
-   virtual double ComputeL2Error(VectorCoefficient &exsol,
                                  const IntegrationRule *irs[] = NULL,
                                  Array<int> *elems = NULL) const;
 
    /// Returns ||grad u_ex - grad u_h||_L2 in element ielem for H1 or L2 elements
    virtual double ComputeElementGradError(int ielem, VectorCoefficient *exgrad,
                                           const IntegrationRule *irs[] = NULL) const;
 
+   virtual double ComputeL2Error(Coefficient &exsol,
```
vladotomov, line 487: Please add doxygen about `elems` to specify that it expects a marker instead of a list.

fem/gridfunc.hpp
```diff
@@ -476,21 +476,22 @@ class GridFunction : public Vector
                                              Array<int> &bdr_attr);
 
 
-   virtual double ComputeL2Error(Coefficient &exsol,
-                                 const IntegrationRule *irs[] = NULL) const
-   { return ComputeLpError(2.0, exsol, NULL, irs); }
```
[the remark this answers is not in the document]
atallah727: Yes, I can fix it. One thing though is that there is another ComputeL2Error function that takes "Coefficient *exsol[]" as input. In principle, we can also remove this function and use directly the ComputeLp function, but we need to decide on the type of input. The question is whether to use "Coefficient *exsol[]" or "Coefficient &exsol" as input. The ComputeLp takes the latter.

fem/gridfunc.hpp
atallah727: Sure thing. Based on your previous comment, I will add a comment in the ComputeLp function. To avoid confusion though, I will wait until we address the input issue I mentioned in the previous comment.

fem/gridfunc.hpp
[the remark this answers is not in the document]
vladotomov: I think the `*exsol[]` version of `ComputeL2Error` should stay as it is -- it has a slightly different implementation, as it assumes that the FE space has several components, and the exact solution is given as an array of Coefficient for every component.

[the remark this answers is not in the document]
atallah727: Ok, good to know. I will modify the code accordingly.

fem/gridfunc.hpp
atallah727: Sure thing.

rw-anderson: > Please add doxygen about `elems` to specify that it expects a marker instead of a list.

Is there a reason it is done this way and not using a list? I'm not sure what the use case for this is, but it would seem more efficient to only iterate over the elements in a list.

rw-anderson: Are you talking about something that is in mfem or exterior to it?

rw-anderson: A library user will have different ways of generating inputs to the interfaces. Maybe they will have to scan the whole list of elements to generate the active set before every call (as in your case it sounds like), or maybe they will already have the set and use it many times. Scanning everything and skipping in the latter case seems inefficient, especially if the set is small compared to the total size.

atallah727: So, you are suggesting I pass it only a list of elements that I want to compute the L2 error on, right? But to generate the list I would already need to use the markers to isolate those elements and pass it for L2 error computation. The element markers are computed separately. Once complete, the user passes it to the L2 error function where we compute the error on some of the elements depending on their marker.

vladotomov (approved):

fem/gridfunc.hpp
rw-anderson: Let me put it a different way. What this interface needs is a representation of a subset of elements. I think a dense representation of the subset is preferable over a sparse representation, all other things being equal, because it consumes less memory, and uses fewer iterations to traverse.

It may be true in your case that you already have a sparse marker array, but there's no reason to believe this will be true in general for a user of the library. (I think it would be more common for a subset of elements to be described by a dense list of element numbers.)

For an example elsewhere in the library of the representation of a subset of elements in an API, the AMR interface takes a user-supplied subset of elements to refine. The interface receives a (dense) list of element numbers, not a (sparse) marker array, because it is more compact and does not waste cycles iterating over a possibly large number of no-ops.

For a little more motivation, suppose we had a multi-material setting, and want to compute errors in the elements of material N or on the boundaries of material N. If you only have 2 materials, then a marker strategy is probably fine. But suppose you had 50 materials, and wanted to compute the errors for boundaries of material N. Now you'd need to have 50 sparse marker arrays, most of which would be empty, and you'd be iterating over all of the elements 50 times each, mostly doing nothing. With the dense representation, you'd only visit each element once.

It is not a big deal, we could always add the dense list version later, it just seems like the natural choice for a general library API for describing a subset of elements, and is more consistent with other interface(s) where subsets of elements to process are supplied.

atallah727: Sorry for my late reply. I think we would still have one array of size the number of elements with the markers of interest. Meaning you would just mark the elements of interest according to some algorithm. So, I don't see why we would need N markers each of size of the number of elements. At least, what I had in mind was not to create N arrays for general problems.

rw-anderson: I don't want to hold the PR up so maybe it is best to proceed with the marker array, and we could add the list version later if needed.

But to clarify what I mean, when you say "But to generate the list I would already need to use the markers to isolate those elements and pass it for L2 error computation." This may be true for the way your code is organized. But I think it would be more common for the list to already exist. For example, a common data structure for material data is to have a ragged array enumerating elements of a given material. In that case, computing errors for material N would require filling out a full sized marked array, which adds no extra information and adds no-op iteration inside the library.

But as long as it's documenting as a marker array, probably best to proceed with this implementation for now.

rw-anderson (approved):
I've already submitted my comments, so approving this now.

kmittal2 (approved):

tzanio:

PR #3072, opened 2022-06-14
Upgrades to direct solvers including `Operator::Mult` with multiple RHS

### LINKED ISSUE #3021: `Operator::Mult` or `Solver::Mult` with multiple RHS
I'm wondering if there is any interest from others in an interface for solving linear systems with multiple right-hand sides? In particular, this seems like it would be particularly useful for the interfaces to direct solvers like MUMPS or SuperLU_DIST, etc. 

One way to do so would be to have an overloaded virtual `Mult` method in `Operator` which has a default behavior of calling `Mult` on each column of the input matrix, and alternatively can be overridden by derived `Operator` or `Solver` classes if desired. There are probably several options for how the input/output RHS vectors are provided, they could each be a `DenseMatrix`, an `Array` of `Vector` objects, or a `BlockVector` with equally sized blocks, to name a few.

This would be useful for me as I am often looping over columns to solve systems with multiple RHS and I have noticed that the forward/backward solve of MUMPS is much faster when directly calling with `id->nrhs > 1`, for example. Curious if this is of interest to other developers/users before I go ahead an open a PR.

sebastiangrimberg:
Upgrades for external sparse direct solver interfaces:
- HYPRE_BIGINT and HYPRE_MIXEDINT support for MUMPS, SuperLU_Dist, and STRUMPACK interfaces. Particularly useful when using as a coarse solve for a multigrid method where the fine level matrices have > 2B nonzeros or unknowns.
- `Operator::Mult(const Array<Vector *> &X, Array<Vector *> &Y)` overrides for MUMPS, SuperLU_Dist, and STRUMPACK solver classes (depends on #3073).
- MUMPS: Block low-rank compression in MUMPS >= 5.1.0, fix bug for CMake builds where `MFEM_MUMPS_VERSION` was not being set, fix bug for matrix symmetry `enum` (had SPD and symmetric indefinite types swapped), add option to specify reordering method, add option to reuse symbolic factorization for repeated calls to `SetOperator`, automatically retry numeric factorization when errors are encountered due to the memory relaxation parameter.
- SuperLU_Dist: Enable 3D grid partitioning for SuperLU_Dist >= 7.2.0, reuse symbolic factorization/row permutation for repeated `SetOperator` calls, fix `gridexit` call.
- STRUMPACK: Enable compression settings for recent versions of STRUMPACK, mixed precision solver from STRUMPACK (requires fixes from STRUMPACK master branch above current release 6.3.1), enable symbolic factorization reuse, fixed memory leak in `STRUMPACKRowLocMatrix` (`strumpack::CSRMatrixMPI` copies all inputs).

Additional changes:
- ~~`Multigrid`: Eliminated redundant residual calculation for pre-smooth at all levels, also upgraded to use multiple RHS (take advantage at coarse solve).~~ Moved to #3075

Resolves #3021

TODO:
- [x] Need to add tests for multiple RHS interface

| [#3072](https://github.com/mfem/mfem/pull/3072) | @sebastiangrimberg | @mlstowell | @psocratis + @jandrej | 6/16/22 | ⌛due 6/30/22 | ⌛due 7/7/22 | |


mlstowell:

jandrej (changes requested):
Please separate this PR in multiple smaller ones.

* Don't change non-related comments or files in your PR
* Separate MUMPS and SuperLU changes in separate PRs

This PR has a lot of overlap with the Multigrid PR. Due to the amount of changes I'm inclined to require the review process on the Multigrid PR first and then accept this one in separated logical chunks.

Don't take this comment as a push-back. I'm just trying to make the process more likely to go through quicker for the other reviewers.

Thank you for your contribution and I hope you see the requests above as reasonable.

sebastiangrimberg:
Closing this PR: Will split into multiple and mark as WIP until #3073 or #3072 is merged.

PR #3073, opened 2022-06-15
Initial support for `Operator::Mult` with multiple RHS

sebastiangrimberg:
Interface upgrades for #3072, which is based on this branch.

Includes:
- `Operator::Mult(const Array<Vector *> &X, Array<Vector *> &Y)` overload for application to multiple RHS. Base class implementation just loops over columns of the input and calls `Mult` on each Vector.
- `Operator::AddMult` and `Operator::AddMultTranspose` methods to the base `Operator` class for derived classes which already implement this.

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#3073](https://github.com/mfem/mfem/pull/3073) | @sebastiangrimberg | @mlstowell | @psocratis + @jandrej | 6/16/22 | ⌛due 6/30/22 | ⌛due 7/7/22 | |

PR #3075, opened 2022-06-16
`Multigrid::Mult` with multiple RHS

### LINKED ISSUE #3021: `Operator::Mult` or `Solver::Mult` with multiple RHS

sebastiangrimberg:
Includes:

- `Operator::Mult(const Array<const Vector *> &X, Array<Vector *> &Y)` overload for application to multiple RHS. Base class implementation just loops over columns of the input and calls Mult on each Vector.
- `Operator::AddMult` and `Operator::AddMultTranspose` methods to the base `Operator` class for derived classes which already implement this.
- Upgraded `Multigrid` class to support multiple RHS vectors per cycle, useful when the coarse solve has an optimized routine for solving with multiple RHS.
- Eliminated redundant initial residual calculation for pre-smooth at all levels.

Resolves #3021

| [#3075](https://github.com/mfem/mfem/pull/3075) | @sebastiangrimberg | @mlstowell | @psocratis + @jandrej | 6/16/22 | 6/30/22 | 1/10/23 | |


mlstowell:

linalg/densemat.cpp
```diff
@@ -221,8 +221,13 @@ void DenseMatrix::MultTranspose(const Vector &x, Vector &y) const
    MultTranspose((const double *)x, (double *)y);
 }
 
-void DenseMatrix::AddMult(const Vector &x, Vector &y) const
+void DenseMatrix::AddMult(const Vector &x, Vector &y, const double a) const
 {
+   if (a != 1.0)
```
jandrej, line 226: Is this check stable? Can't we hit weird roundoff?

jandrej (commented):
This PR seems well executed. It's clear what the changes involve and they present a minimal API difference. I'd like an additional reviewer on this PR since it touches the main components of `Operator` and `Multigrid`.

ping @drzisga

sebastiangrimberg:
> This PR seems well executed. It's clear what the changes involve and they present a minimal API difference. I'd like an additional reviewer on this PR since it touches the main components of `Operator` and `Multigrid`.
> 
> ping @drzisga

I tried to factor out the interface changes to `Operator` into #3073, so those changes are showing up in this PR only because it is based on that branch. So this should probably only be reviewed after those changes are merged and this branch is rebased. Alternatively if these changes are a small enough addition I would be fine closing #3073 and doing the full review here.

linalg/densemat.cpp
sebastiangrimberg: Also part of #3073, sorry for the confusion. I'm not an expert here, but I was under the impression that this is safe because floating point types can always exactly represent the value 1. There may be a better way to do it Maybe there is a better way to do it.

linalg/operator.hpp
```diff
@@ -91,13 +91,34 @@ class Operator
...
+
+   /// y += A^t(x) (default) or y += a * A^t(x)
+   virtual void AddMultTranspose(const Vector &x, Vector &y,
+                                 const double a = 1.0) const;
+
+   /// Operator application on a matrix: `Y=A(X)`.
+   virtual void Mult(const Array<Vector *> &X, Array<Vector *> &Y) const;
+
+   /// Action of the transpose operator on a matrix: `Y=A^t(X)`.
+   virtual void MultTranspose(const Array<Vector *> &X, Array<Vector *> &Y) const;
```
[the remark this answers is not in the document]
sebastiangrimberg: These are just copy paste issues from the other docstrings of `Mult` (see line 88 and 91). If we want to keep them, I'll modify `AddMult` and `AddMultTranspose` to be consistent. Otherwise, I can remove them all.

Note that this change corresponds to #3073 on which this PR is based (see my comment below).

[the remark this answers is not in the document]
sebastiangrimberg: Fixed in [e5de8d6](https://github.com/mfem/mfem/pull/3075/commits/e5de8d634b1f91d37516bfaad0c16a09c272840b).

linalg/operator.cpp
```diff
@@ -48,6 +48,67 @@ void Operator::InitTVectors(const Operator *Po, const Operator *Ri,
...
+   }
+}
+
+void Operator::MultTranspose(const Array<Vector *> &X, Array<Vector *> &Y) const
+{
+   MFEM_ASSERT(X.Size() == Y.Size(),
+               "Number of columns mismatch in Operator::MultTranspose!");
+   for (int i = 0; i < X.Size(); i++)
+   {
+      MFEM_ASSERT(X[i] && Y[i], "Missing Vector in Operator::Mult!");
```
drzisga, line 83: Should be `Operator::MultTranspose` in the assertion.

linalg/operator.cpp
```diff
@@ -48,6 +48,67 @@ void Operator::InitTVectors(const Operator *Po, const Operator *Ri,
...
+}
+
+void Operator::AddMultTranspose(const Array<Vector *> &X, Array<Vector *> &Y,
+                                const double a) const
+{
+   MFEM_ASSERT(X.Size() == Y.Size(),
+               "Number of columns mismatch in Operator::MultTranspose!");
+   for (int i = 0; i < X.Size(); i++)
+   {
+      MFEM_ASSERT(X[i] && Y[i], "Missing Vector in Operator::Mult!");
```
drzisga, line 107: Here and above the name in the assertions should match the method names.

linalg/densemat.cpp
drzisga: This is stable for the default hard-coded argument `a = 1.0`. Even for round-offs, the method yields the correct results just with some performance overhead.

linalg/densemat.hpp
```diff
@@ -150,11 +150,15 @@ class DenseMatrix : public Matrix
...
-   void AddMult(const Vector &x, Vector &y) const;
+   virtual void AddMult(const Vector &x, Vector &y, const double a = 1.0) const;
 
    /// y += A^t x
-   void AddMultTranspose(const Vector &x, Vector &y) const;
+   virtual void AddMultTranspose(const Vector &x, Vector &y,
+                                 const double a = 1.0) const;
 
    /// y += a * A.x
    void AddMult_a(double a, const Vector &x, Vector &y) const;
```
drzisga, line 164: What's now the difference between `AddMult_a` and `AddMult`? From a user-perspective this looks like the same method just with differently ordered arguments.

linalg/operator.cpp
sebastiangrimberg: Fixed in [7c70dfb](https://github.com/mfem/mfem/pull/3075/commits/7c70dfb06931a82b6eb3d1d42e755d6461896f70).

linalg/operator.cpp
sebastiangrimberg: Fixed in [7c70dfb](https://github.com/mfem/mfem/pull/3075/commits/7c70dfb06931a82b6eb3d1d42e755d6461896f70).

linalg/densemat.hpp
sebastiangrimberg: That's correct. I figured keeping both allows the `a = 1.0` case to still be optimized compared to the general case at the expense of a single `if` check. It also keeps consistency with the overloads for `DenseMatrix` arguments, and means no changes to other code calling `AddMult_a` or `AddMult`.

If you think they should be unified that is fine with me and I will change it.

sebastiangrimberg:
Apologies for the rebase. I ran into a few small bugs for the `Multigrid` changes in testing so I fixed them and squashed the history.

sebastiangrimberg:
Hi all, just a follow up for review to try to get this approved and merged. Thank you!

psocratis:
> Hi all, just a follow up for review to try to get this approved and merged. Thank you!

Apologies for the delay, I got back to this today and tbh I need some more time to review it. I'll try to finish it soon

fem/multigrid.cpp
```diff
@@ -60,187 +45,232 @@ Multigrid::~Multigrid()
...
-void Multigrid::Mult(const Vector& x, Vector& y) const
+void MultigridBase::Mult(const Vector& x, Vector& y) const
 {
-   MFEM_ASSERT(NumLevels() > 0, "");
-   *X.Last() = x;
-   *Y.Last() = 0.0;
-   Cycle(GetFinestLevelIndex());
-   y = *Y.Last();
+   Array<Vector*> X_(1), Y_(1);
+   X_[0] = const_cast<Vector*>(&x);
```
drzisga, line 109: I don't like the const_cast but there is probably no other way to do it. Or is it maybe possible to use an Array of const Vector* for X_ in the method for multiple rhs?

sebastiangrimberg: Fixed in [b15b188](https://github.com/mfem/mfem/pull/3075/commits/b15b188c448d4a0086d5357fa1c376d2681dff05). I wasn't sure if this would work since I know compilers can sometimes not like `const` objects in STL vectors, but it seems to work fine here. Thanks!

psocratis (approved):
This looks good. My first impression was "why do we need all this new code for implementing multiple RHS for MG?". But eventually, I realized that this PR has a bit more. I think all changes look good but my suggestion is to change the title 😄 .  My apologies for the late review. One thing I haven't tested is the device runs. Did anyone test those?

sebastiangrimberg:
Thanks @psocratis! Indeed, originally this PR included #3081, #3082, and #3083 which is really where the multiple RHS functionality is useful. Useful for MG also if the coarse solve is some sparse direct solver, so that's the motivation for this starting PR.

tzanio:
@dylan-copeland, do you have an opinion for what should be the correct interface for a collection of vectors from your ROM work?

dylan-copeland:
> @dylan-copeland, do you have an opinion for what should be the correct interface for a collection of vectors from your ROM work?

@tzanio I have not reviewed this PR, but it appears to be using `Array<Vector*>` for a collection of vectors, which on first thought seems good for the context of multigrid with multiple RHS. I like the flexibility, since the vectors can come from anywhere and be wrapped if necessary without significant overhead. The context of ROM is different in that ROM involves dense linear algebra on a small collection of vectors, so it is most natural to store the vectors as columns of a dense matrix. That approach would not be beneficial in multigrid, and it would probably require copying data from various vectors into a matrix. Thanks for asking if there is any connection with ROM, but I don't see one, and I think this PR has a good interface.

tzanio:
thanks Dylan!

sebastiangrimberg:
@tzanio @dylan-copeland I originally thought it would be more natural to make the interface use `DenseMatrix` instead of `Array<Vector *>`, where each RHS could be individually accessed by columns of the matrix. Multigrid aside, this is how must sparse direct solver libraries are built or, for example, PETSc with its `KSPMatSolve`.

This implementation here was just inspired from the discussion in #3021 and I would be happy to change it if there is an opinion one way or another.

sebastiangrimberg:
Hi @drzisga @jandrej, any chance you could give this a final review when you have a moment? Sorry for the ping but this would be nice to move on from or I can make any desired changes.

drzisga (approved):

sebastiangrimberg:
Ping @mlstowell? I'm worried this may have gotten lost.

jandrej (approved):

mlstowell:

v-dobrev:
When building this PR with CUDA support enabled, the `nvcc` compiler produces a ton of warnings of this type:
```
../../linalg/operator.hpp(313): warning #611-D: overloaded virtual function "mfem::Operator::Mult" is only partially overridden in class "mfem::TimeDependentOperator"
```
These can be suppressed by adding `using Operator::Mult;` in sub-classes. However, another solution may be simpler in terms of changes and add some clarity -- the solution is to NOT overload the method name `Mult` and instead use a different name for the new `Mult` version, e.g. instead of
```c++
   virtual void Mult(const Array<const Vector *> &X, Array<Vector *> &Y) const;
```
use
```c++
   virtual void ArrayMult(const Array<const Vector *> &X, Array<Vector *> &Y) const;
```
The same issue affects other methods that this PR turns into overloaded methods, like `Operator::MultTranspose`, `Operator::AddMult`, etc.

We'd like to resolve these warnings -- which solutions is best is up for discussion. What do you think?

sebastiangrimberg:
I think using `ArrayMult` and avoiding the overload is a good solution here. I will make this change, thanks @v-dobrev.

mlstowell:

sebastiangrimberg:
Hi @v-dobrev @mlstowell , do you see any remaining build issues in `next`? Alternatively, is there a place I can view the build output for `next` to check and resolve issues myself if that would be helpful?

v-dobrev:
> Hi @v-dobrev @mlstowell , do you see any remaining build issues in `next`? Alternatively, is there a place I can view the build output for `next` to check and resolve issues myself if that would be helpful?

There are no issues from this PR in the nightly tests. I think it is ready to be merged. I'll merge the `master` branch here to prepare it for merging. I'll let @mlstowell merge the PR since he's the editor.

PR #3082, opened 2022-06-18
Update MUMPS solver interface

sebastiangrimberg:
Updates and upgrades for MUMPS sparse direct solver interface:

- HYPRE_BIGINT and HYPRE_MIXEDINT support.
- `Operator::Mult(const Array<Vector *> &X, Array<Vector *> &Y)` overrides for MUMPS solver classes (depends on https://github.com/mfem/mfem/pull/3075).
- Block low-rank compression in MUMPS >= 5.1..
- Fix bug for CMake builds where `MFEM_MUMPS_VERSION` was not being set (resolve differences between `config.hpp.in` for GNU make and CMake builds).
- Fix bug for matrix symmetry enum (had SPD and symmetric indefinite types swapped).
- Add option to specify reordering method.
- Add option to reuse symbolic factorization for repeated calls to `SetOperator`.
- Automatically retry numeric factorization when errors are encountered due to the memory relaxation parameter.


tzanio:

tzanio:
@jandrej and @psocratis -- can you please review when you get a chance?

tests/unit/linalg/test_direct_solvers.cpp
```diff
@@ -217,17 +217,37 @@ TEST_CASE("direct-parallel", "[Parallel], [CUDA]")
...
          mumps.SetPrintLevel(0);
          mumps.SetOperator(*A.As<HypreParMatrix>());
          mumps.Mult(B,X);
+
          Vector Y(X.Size());
          A->Mult(X,Y);
          Y-=B;
          REQUIRE(Y.Norml2() < 1.e-12);
 
+         mumps.Mult(BB,XX);
```
psocratis, line 242: There is a compile error: `error: no viable conversion from 'Array<mfem::Vector *>' to 'const mfem::Vector'`

psocratis (changes requested):
Appart from the unit test compilation error this looks good

tests/unit/linalg/test_direct_solvers.cpp
sebastiangrimberg: Whoops, missed this when rebasing. Fixed in [5fe0f7f](https://github.com/mfem/mfem/pull/3082/commits/5fe0f7fbed029ca8ac03a1db7096665a16f8b406).

psocratis:
> Thanks @psocratis, should be all good now with the test fixed as well.

Looks good!

psocratis (approved):

tzanio:
@jandrej -- can you take a look when you get a chance

tzanio:
🔔  @jandrej

tzanio:
Two questions:

1. Should this be mentioned in `CHANGELOG`
2. Do we need to update the [minimum required version](https://github.com/mfem/mfem/blob/sjg/mumps-solver-dev/INSTALL#L652) for MUMPS in `INSTALL`?

tzanio:

tzanio:

adam-sim-dev:
Hi @sebastiangrimberg , MUMPS depends on parmetis only for both parallel and sequential analysis. So parmetis may be not required.

sebastiangrimberg:
> Hi @sebastiangrimberg , MUMPS depends on parmetis only for both parallel and sequential analysis. So parmetis may be not required.

Hi @adam-sim-dev, this is correct and in general MUMPS has a lot of different configuration options which MFEM needs to know about in order to ensure correct linkage (OpenMP, Scotch/PT-Scotch, METIS/ParMETIS, etc.). Note that the values in `config/defaults.cmake` are typically meant to be overridden at configure time by the user by specifying cache values such as `MUMPS_REQUIRED_PACKAGES` (in the case of a CMake build) based on the exact configuration of MUMPS by the user. So in this case having `ParMETIS` in `defaults.cmake` is just an example for how most people might configure MUMPS. Alternatively, users can use a `user.cmake` file to customize the configuration for third-party dependencies.

PR #3119, opened 2022-07-28
Cache vcpkg only on Windows

### LINKED ISSUE #3063: Caching vcpkg in MacOS Action
In GitHub Actions, we are caching vcpkg (windows package manager) on MacOS.
This probably has no impact, but is confusing and seems unnecessary.

It seems we should we add `if: matrix.os == 'windows-2022'` in the job there:
https://github.com/mfem/mfem/blob/d3e94af7f1a286d2c551f75862327c4ab40b45a1/.github/workflows/builds-and-tests.yml#L191

What do you think @kanye-quest,

adrienbernede:
Resolves #3063

Addresses https://github.com/mfem/mfem/pull/3106#issuecomment-1183713276

### DIFF
--- a/.github/workflows/builds-and-tests.yml
+++ b/.github/workflows/builds-and-tests.yml
@@ -189,6 +189,7 @@ jobs:
         dir: ${{ env.METIS_TOP_DIR }}
 
     - name: cache vcpkg (Windows)
+      if: matrix.os == 'windows-2022'
       id: vcpkg-cache
       uses: actions/cache@v3
       with:


adrienbernede:
LGTM

adrienbernede:
@tzanio @pazner this PR is a small improvement over Will’s recent changes.

The initial motivation was to add an unique identifier to the cache key. By using the hash of the vcpkg config files, I am ensuring that the cache gets rebuilt when those change. It’s important to avoid the classic "the CI is acting as if the dependencies hadn’t been rebuilt".

Also, compared to current production state, it seems I "fix" the cache use for vcpkg:
Looking at https://github.com/mfem/mfem/actions/runs/3151491029/jobs/5125503930 it seems the cache may be empty, because metis is rebuilt anyways. That’s a 30 secs improvements.

That’s about it, the rest being incidental changes.

PR #3143, opened 2022-08-12
LinearFormExtension fixes

pazner:
- Adds common abstract base class for ElementRestriction operators
    - Adds `AddMultTranspose` to interface
- Use `DETERMINANTS` instead of `JACOBIANS` in DomainLFIntegrator
- Properly accumulate integrators in LinearFormExtension
- Add test for accumulation of integrators in LinearFormExtension

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#3143](https://github.com/mfem/mfem/pull/3143) | @pazner | @tzanio | @psocratis + @camierjs + @YohannDudouit | 8/16/22 | 9/26/22 | 9/29/22 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
      - [ ] Add the example code to the `ALL_EXE_SRCS` variable.
[...]

PR #3144, opened 2022-08-12
Support boundary linear forms on device

pazner:
Supports assembly of boundary linear forms on device. `BoundaryLFIntegrator` and `BoundaryNormalLFIntegrator` are currently support. This should be enough for the linear forms needed by the Navier miniapp. #3143 has some fixes that should be incorporated here too.

### DIFF
--- a/fem/lininteg_boundary.cpp
+++ b/fem/lininteg_boundary.cpp
@@ -1,0 +1,219 @@
+// Copyright (c) 2010-2022, Lawrence Livermore National Security, LLC. Produced
+// at the Lawrence Livermore National Laboratory. All Rights reserved. See files
+// LICENSE and NOTICE for details. LLNL-CODE-806117.
+//
+// This file is part of the MFEM library. For more information and source code
+// availability visit https://mfem.org.
+//
+// MFEM is free software; you can redistribute it and/or modify it under the
+// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// CONTRIBUTING.md for details.
+
+#include "fem.hpp"
+#include "../fem/kernels.hpp"
+#include "../general/forall.hpp"
+
+namespace mfem
+{
+
+template<int T_D1D = 0, int T_Q1D = 0> static
+void BLFEvalAssemble2D(const int vdim, const int nbe, const int d, const int q,
+                       const int *markers, const double *b,
+                       const double *detj, const double *weights,
+                       const Vector &coeff, double *y)
+{
+   const auto F = coeff.Read();
+   const auto M = Reshape(markers, nbe);
+   const auto B = Reshape(b, q, d);
+   const auto detJ = Reshape(detj, q, nbe);
+   const auto W = Reshape(weights, q);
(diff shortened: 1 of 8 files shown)


pazner:
@YohannDudouit, I also want to implement `VectorFEBoundaryFluxLFIntegrator`, but this requires getting the boundary RT DOFs, what do you think about having an `RTFaceRestriction` operator or something similar?

tzanio:

mesh/pmesh.cpp
```diff
@@ -3085,8 +3085,7 @@ int ParMesh::GetSharedFace(int sface) const
 
 int ParMesh::GetNFbyType(FaceType type) const
 {
-   MFEM_VERIFY(have_face_nbr_data,
-               "ExchangeFaceNbrData() should be called before using GetNFbyType");
+   const_cast<ParMesh*>(this)->ExchangeFaceNbrData();
```
pazner, line 3088: @YohannDudouit, this `const_cast` is kind of bad, but I'm not sure if there's a better solution… what do you think?

YohannDudouit: This probably was the reason why I only used `MFEM_VERIFY` before, though `ExchangeFaceNbrData` does not really change the `Mesh` but makes it valid, so I think the `const_cast` is okay.

jandrej (approved):

fem/lininteg_boundary.cpp
```diff
@@ -0,0 +1,302 @@
...
+      if (d==3 && q==3) { ker=BLFEvalAssemble3D<3,3>; }
+      if (d==4 && q==4) { ker=BLFEvalAssemble3D<4,4>; }
+      if (d==5 && q==5) { ker=BLFEvalAssemble3D<5,5>; }
+      if (d==2 && q==3) { ker=BLFEvalAssemble3D<2,3>; }
+      if (d==3 && q==4) { ker=BLFEvalAssemble3D<3,4>; }
+      if (d==4 && q==5) { ker=BLFEvalAssemble3D<4,5>; }
+      if (d==5 && q==6) { ker=BLFEvalAssemble3D<5,6>; }
+   }
+
+   MFEM_VERIFY(ker, "No kernel ndof " << d << " nqpt " << q);
```
camierjs, line 198: Would the `BLFEvalAssemble` kernels work as fallbacks, and if they do, would it possible to add them ?

camierjs: We should add them, I can verify the domain ones.

pazner: Sounds good, could you push that here and then I'll add the boundary?

camierjs: It turns out that that the `ker` is already initialized as `BLFEvalAssemble2D<>` or `BLFEvalAssemble3D<>`.

camierjs (approved):

tzanio:

tzanio:

tzanio:

PR #3195, opened 2022-09-07
Suggested changes in PR 3096

tzanio:
@tuckerbabcock, here is my suggested changes for https://github.com/mfem/mfem/pull/3096.

Feel free to take as much or as little as you prefer from them 😁 

I commented on the hypre PR you opened, but I don't feel it is strictly necessary here, see https://github.com/mfem/mfem/pull/3096/files#r964305927

I have not tested this extensively, so please double-check before merging ⚠️ 

cc:  @pazner, @psocratis

### DIFF
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -5050,6 +5050,7 @@ void HypreAMS::MakeSolver(int sdim, int cycle_type)
 #endif
 
    space_dim = sdim;
+   ams_cycle_type = cycle_type;
    HYPRE_AMSCreate(&ams);
 
    HYPRE_AMSSetDimension(ams, sdim); // 2D H(div) and 3D H(curl) problems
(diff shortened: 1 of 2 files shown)


tuckerbabcock:
I likely won't be able to fully test this until next week, but I wonder if after [your hypre PR](https://github.com/hypre-space/hypre/pull/730) gets merged, we might want to use an implementation similar to the one I had commented out in my original fix PR, since that would preserve all of the preconditioner options.

I'm imagining using an `#if MFEM_HYPRE_VERSION >= #####` block to determine which implementation to use (either your proposed option here, or the option preserving one that calls the hypre macros). Or, if supporting both would bloat the code too much perhaps just live with the memory leak below a certain hypre version and only fix it for hypre versions that let us access the AMS macros. Does an approach like that make sense to you?

tuckerbabcock:
Actually, I went ahead and implemented what I was suggesting, and tested it using your hypre branch. Both approaches seem to work, but the version that uses the hypre headers will preserve the preconditioner options. Please take a look and let me know what you think, but note that to test the option-preserving code path you'll have to bump the hypre version on your hypre branch.

tuckerbabcock (approved):

PR #3235, opened 2022-09-28
[SQUASH] Assembly levels for LinearForm

tzanio:
✅ Addresses the following item in https://github.com/mfem/mfem/pull/3199:

```
Decide on default for LinearForm::Assemble(bool) and necessary calls to Mesh::DeleteGeometricFactors
```

**Main changes**
1. Device assembly of linear forms now has to be explicitly enabled by calling `SetAssemblyLevel(AssemblyLevel::FULL)`

2. By default, the legacy linear form assembly is used (`AssemblyLevel::LEGACY`).

3. The boolean parameter of the `Assemble` method of `LinearForm` has been removed.

4. Added a new method, `Mesh::Update`, which should be called after the mesh node coordinates have changed.

| [#3235](https://github.com/mfem/mfem/pull/3235) | @tzanio | @mlstowell | @pazner + @camierjs + @mlstowell | 9/28/22 | 10/18/22 | ⌛due 9/30/22 | |


mlstowell:

mlstowell:
@tzanio, @camierjs, and @pazner notice that I've set a very compressed timeline for this PR. The goal is to have this ready for the mfem-4.5 release which is in just a few days. Of course, the most important thing is to release a version of MFEM that we have confidence in so please make a note of any concerns you might have. If this needs to be postponed until after mfem-4.5 then so be it. I'll need to rely on your expertise to make that determination. Thanks!

fem/assembly.hpp
```diff
@@ -0,0 +1,49 @@
...
+   /// form, i.e. a global sparse matrix in MFEM, Hypre or PETSC format.
+   /// In the case of a NonlinearForm LEGACY corresponds to an operator that
+   /// is fully evaluated on the fly.
+   /// In the case of a LinearForm LEGACY corresponds to a fully assembled
+   /// linear form, i.e. a global vector in MFEM, Hypre or PETSC format
+   /// This assembly level is ALWAYS performed on the host.
+   LEGACY = 0,
+   /// @deprecated Use LEGACY instead.
+   LEGACYFULL = 0,
+   /// Fully assembled form, i.e. a global sparse matrix or a gloable vector in
```
mlstowell, line 33: Typo: should be "global"

fem/linearform.cpp
```diff
@@ -155,18 +157,32 @@ bool LinearForm::SupportsDevice()
...
+void LinearForm::SetAssemblyLevel(AssemblyLevel assembly_level)
+{
+   if (ext)
+   {
+      MFEM_ABORT("the assembly level has already been set!");
+   }
+   assembly = assembly_level;
+   switch (assembly)
+   {
+      case AssemblyLevel::LEGACY:
```
mlstowell, line 169: I'm wondering if it would be possible to allow the user to switch back to `AssemblyLevel::LEGACY`. Would we just need to delete `ext` if it exists? If it is that simple then I think we should do it.

fem/assembly.hpp
tzanio: fixed

fem/linearform.cpp
tzanio: This mirrors the behavior of BilinearForm in https://github.com/mfem/mfem/blob/master/fem/bilinearform.cpp#L115-L141.

Currently we do not allow changing the assembly level in neither, probably because that is potentially a very expensive operation. 

We can consider doing that, but maybe that should be a separate PR?

What do you think @camierjs, @pazner, @v-dobrev?

mesh/mesh.cpp
```diff
@@ -7913,6 +7913,9 @@ void Mesh::SetNodes(const Vector &node_coord)
    {
       SetVertices(node_coord);
    }
+
+   // Invalidate the old geometric factors
+   DeleteGeometricFactors();
```
camierjs, line 7918: Couldn't we call `Update` here and for the other methods updating the nodes?

tzanio: I was debating on this, but you are probably right. I can switch them to `Update`

tzanio: Done in https://github.com/mfem/mfem/pull/3235/commits/9f3974d7c3d04306a679ced981f27ebd8cc30c9d

fem/linearform.cpp
```diff
@@ -155,18 +157,32 @@ bool LinearForm::SupportsDevice()
...
+         break;
+      case AssemblyLevel::FULL:
+         if (SupportsDevice()) { ext = new LinearFormExtension(this); }
+         break;
+      default:
+         mfem_error("Linearform does not support the given assembly level");
```
pazner, line 175:
```suggestion
         MFEM_ABORT("Linearform does not support the given assembly level");
```

fem/linearform.hpp
```diff
@@ -82,7 +86,10 @@ class LinearForm : public Vector
    /// Creates linear form associated with FE space @a *f.
    /** The pointer @a f is not owned by the newly constructed object. */
    LinearForm(FiniteElementSpace *f) : Vector(f->GetVSize())
-   { fes = f; ext = nullptr; extern_lfs = 0; UseDevice(true); }
+   {
+      fes = f; ext = nullptr; extern_lfs = 0; UseDevice(true);
```
pazner, line 90: I suggest moving each statement here to its own line (and perhaps moving the function definition to the cpp file).

fem/linearform.hpp
```diff
@@ -97,15 +104,22 @@ class LinearForm : public Vector
    /** The associated FiniteElementSpace can be set later using one of the
        methods: Update(FiniteElementSpace *) or
        Update(FiniteElementSpace *, Vector &, int). */
-   LinearForm() { fes = NULL; ext = nullptr; extern_lfs = 0; UseDevice(true); }
+   LinearForm()
+   {
+      fes = NULL; ext = nullptr; extern_lfs = 0; UseDevice(true);
```
pazner, line 109: We could have the other constructors call this default constructor to set these default values and reduce duplication.

fem/linearform.hpp
```diff
@@ -185,10 +199,22 @@ class LinearForm : public Vector
        corresponding pointer (to Array<int>) will be NULL. */
    Array<Array<int>*> *GetFLFI_Marker() { return &boundary_face_integs_marker; }
 
+   /// Set the desired assembly level.
+   /** Valid choices are:
+
+       - AssemblyLevel::LEGACY (default)
+       - AssemblyLevel::FULL
+
+       This method must be called before assembly. */
```
pazner, line 208: > This method must be called before assembly.

Not strictly true — if it's not called, it will just default to `LEGACY`

mesh/mesh.hpp
```diff
@@ -990,6 +990,11 @@ class Mesh
        for example, after the mesh nodes are modified externally. */
    void DeleteGeometricFactors();
 
+   /// This function should be called after the mesh node coordinates have
+   /// changed, e.g. after the mesh has moved. It updates internal quantities
+   /// derived from the node coordinates, suchs as the GeometricFactors.
```
pazner, line 995:
```suggestion
   /// @brief This function should be called after the mesh node coordinates
   /// have changed, e.g. after the mesh has moved.
   /** It updates internal quantities derived from the node coordinates, such as
       as the GeometricFactors. */
```

mesh/mesh.hpp
```diff
@@ -990,6 +990,11 @@ class Mesh
...
+   void Update() { DeleteGeometricFactors(); }
```
pazner, line 996: Maybe `Update` is too general/generic? It doesn't need to be called after e.g. uniform or local refinements, because those refinements will internally call `ResetLazyData`.

Perhaps `NodesHaveMoved` or `MeshHasMoved` would be more self-explanatory?

pazner (commented):
The implementation looks good to me (modulo some specific comments).

I think we should in the future (MFEM 5.0?) consider switching the default to `AssemblyLevel::FULL`. Besides from the caching of geometric factors, there really should be no difference compared with `LEGACY` (besides potentially _much_ better performance).

One feature that we lose with this design is that there is no way for users to "automatically" get GPU acceleration as we implement new integrators.

With the previous design, if we implement a new integrator, `SupportsDevice` would return true, and GPU acceleration would turn on automatically, and users would benefit from increased performance without any work on their part.

In the proposed design, users have to opt-in with `SetAssemblyLevel`. However, if an integrator is not implemented, `Assemble` will fail. That means users cannot enable `FULL` unless all the integrators are supported. If they want to benefit from future implementations, they will somehow have to keep track of which new integrators we add, and when we add support, they will have to go in and change their code.

One possible approach to fix this is to make `AssemblyLevel::FULL` just fall back on `LEGACY` if `SupportsDevice` returns false, rather than failing with an error.

fem/linearform.hpp
tzanio: Addressed in https://github.com/mfem/mfem/pull/3235/commits/c60263c1fd9ecaa9a437346bbb69e172939d7fa1

fem/linearform.hpp
tzanio: Addressed in https://github.com/mfem/mfem/pull/3235/commits/c60263c1fd9ecaa9a437346bbb69e172939d7fa1

fem/linearform.cpp
tzanio: Addressed in https://github.com/mfem/mfem/pull/3235/commits/c60263c1fd9ecaa9a437346bbb69e172939d7fa1

mesh/mesh.hpp
tzanio: I thought about it and couldn't come up with a better name... I am very open to suggestions 😄 

I didn't like nodes/moved because this really should be called any time the mesh (i.e. node coordinates and maybe mesh topology) has changed. But maybe I am overthinking it...

What do other people think?

mesh/mesh.hpp
```diff
@@ -990,6 +990,12 @@ class Mesh
        for example, after the mesh nodes are modified externally. */
    void DeleteGeometricFactors();
 
+   /// @brief This function should be called after the mesh node coordinates
+   /// have changed, e.g. after the mesh has moved.
+   /** It updates internal quantities derived from the node coordinates, such as
```
pazner, line 995:
```suggestion
   /** It updates internal quantities derived from the node coordinates, such
```

Sorry, I introduced a typo 🙈

tzanio:
> One possible approach to fix this is to make AssemblyLevel::FULL just fall back on LEGACY if SupportsDevice returns false, rather than failing with an error.

Isn't that happening already -- if SupportsDevice is false, no extension will be created and Assemble will just use the old method?

tzanio:
> I think we should in the future (MFEM 5.0?) consider switching the default to AssemblyLevel::FULL. Besides from the caching of geometric factors, there really should be no difference compared with LEGACY (besides potentially much better performance).

I absolutely agree. Part of 5.0 should be a detailed document advising users of breaking changes and adjustments they need to make to their code.

mesh/mesh.hpp
mlstowell: Personally I like `Update` because this is already used for the AMR related changes and I won't need to remember a new function name. It is general but, if needed in the future, we could add more specific `Update*` methods to provide users with more fine-grained control.

tzanio:
> One feature that we lose with this design is that there is no way for users to "automatically" get GPU acceleration as we implement new integrators.

I don't understand this comment. This change just mimics what already is happening in BilinearForm, and if the user wants GPU acceleration, they should explicitly enable it in both.

tzanio:
One thing we should consider is adding `LinearForm::SetAssemblyLevel(AssemblyLevel::FULL)` in some of the examples and miniapps?

For example the ones already setting `BilinearForm::SetAssemblyLevel`: 

- `ex1`, `ex1p`, `ex3`, `ex3p`, `ex4`, `ex4p`, `ex5`, `ex5p`, `ex6`, `ex6p`, `ex9`, `ex9p`, `ex22`, `ex22p`, `ex24`, `ex24p`, `ex25`, `ex25p`, `ex26`, `ex26p`
- `mesh-optimizer`, `pmesh-optimizer.cpp`, `minimal-surface`, `pminimal-surface`, `navier_solver`, `lor_solvers`, `plor_solvers`
- ...

@camierjs, what do you think?

camierjs:
> One thing we should consider is adding `LinearForm::SetAssemblyLevel(AssemblyLevel::FULL)` in some of the examples and miniapps?

If the `LinearForm`s involved effectively use the `Device`, then it makes sense to enable the `AssemblyLevel::FULL`.

tzanio:
> > One thing we should consider is adding `LinearForm::SetAssemblyLevel(AssemblyLevel::FULL)` in some of the examples and miniapps?
> 
> If the `LinearForm`s involved effectively use the `Device`, then it makes sense to enable the `AssemblyLevel::FULL`.

Ok, can you take a look and make the change where appropriate?

pazner:
> Isn't that happening already -- if SupportsDevice is false, no extension will be created and Assemble will just use the old method?

Ah, you're right, I missed the check for `SupportsDevice` in `SetAssemblyLevel`. Sorry about that. Looks good to me.

> I don't understand this comment. This change just mimics what already is happening in BilinearForm, and if the user wants GPU acceleration, they should explicitly enable it in both.

Yes, it mimics what is already happening in BilinearForm, but wouldn't it be nice if there was no need to call `SetAssemblyLevel` at all? For the most part, the behavior should be identical, just faster (even on CPU, because of sum factorization). If we change the default in MFEM 5.0 to `FULL` that would be fine.

>  One thing we should consider is adding LinearForm::SetAssemblyLevel(AssemblyLevel::FULL) in some of the examples and miniapps?

We could add this to **all** the examples and miniapps, that way they'll automatically get the "non-legacy" integrators for free as we add them.

mesh/mesh.hpp
pazner: Isn't it confusing though that you **should** call `FiniteElementSpace::Update` after mesh refinement, but **should not** call `Mesh::Update` after refinement?

tzanio:
> We could add this to all the examples and miniapps, that way they'll automatically get the "non-legacy" integrators for free as we add them.

If we do that, we should maybe consider also adding `SetAssemblyLevel(AssemblyLevel::FULL)` for the `BilinearForm`s, with a similar fallback to `LEGACY`.

pazner:
> If we do that, we should maybe consider also adding `SetAssemblyLevel(AssemblyLevel::FULL)` for the `BilinearForm`s, with a similar fallback to `LEGACY`.

It will be a little involved because I don't think there is `SupportsDevice` (or e.g. `SupportsFullAssembly`) for `BilinearForm`

psocratis:
May I chime in here? 😄 . In my opinion, the linear and bilinear form assembly are two very different concepts. One involves the actual assembly level (matrix-free, partial, full) the other one is about the algorithm of the assembly procedure., i.e., it's always full assembly but with traditional or sum factorization. So, to that end, I don't think there should be an assembly level enum in linear forms, but rather a single false/true flag to enable sum factorization or not. So basically as it was before. I believe assembly level in linear forms can introduce more confusion to the users. I also think that we don't have to have the two cases (linear and bilinear forms) linked, meaning we don't need to enable or disable both at the same time. Enabling sum factorization for linear forms changes nothing to the linear system, but enabling matrix-free/ PA assembly levels actually changes the form of the linear systems., i.e, the need to use other solvers/preconditioners, etc. So I agree with eventually finding a way to have the sum factorization as default for the linear forms, but still, it should be independent of the bilinear form assembly level.

tzanio:
> May I chime in here? 😄 . In my opinion, the linear and bilinear form assembly are two very different concepts. One involves the actual assembly level (matrix-free, partial, full) the other one is about the algorithm of the assembly procedure., i.e., it's always full assembly but with traditional or sum factorization. So, to that end, I don't think there should be an assembly level enum in linear forms, but rather a single false/true flag to enable sum factorization or not. So basically as it was before. I believe assembly level in linear forms can introduce more confusion to the users. I also think that we don't have to have the two cases (linear and bilinear forms) linked, meaning we don't need to enable or disable both at the same time. Enabling sum factorization for linear forms changes nothing to the linear system, but enabling matrix-free/ PA assembly levels actually changes the form of the linear systems., i.e, the need to use other solvers/preconditioners, etc. So I agree with eventually finding a way to have the sum factorization as default for the linear forms, but still, it should be independent of the bilinear form assembly level.

Fair points @psocratis 😄 

I can see this perspective, and if folks prefer a different resolution for this I will support it.

One issue is that there is a bit of unfortunate entangling of the notions of assembly and device execution. What I was thinking about was that if one wants the LinearForm to use the device, it is likely they want the same for the BilinearForm (even if both are assembled fully).

Also, I am not sure if it is advisable, but it is possible to consider `ELEMENT` and `PARTIAL` assembly for a `LinearForm`. In general one can think of a linear form as the corresponding `BilinearForm` with test function 1, e.g. $\int f \phi$ is $P^T G^T B^T d$, where `d` is a $Q$-vector with entries $d_k = \omega_k f(q_k) |J(q_k)|$.

mesh/mesh.hpp
mlstowell: I think this is where finer control would be useful. Would there be any harm in calling `Mesh::Update` after refinement (aside from potential performance penalties)? If not then the naive user could call `Update` on virtually everything after any type of mesh change; topological or geometrical. The more advanced user could update only what is necessary after a particular type of change. As long as we have sufficient documentation so that the more advanced user can make intelligent choices.

psocratis:
> What I was thinking about was that if one wants the LinearForm to use the device, it is likely they want the same for the BilinearForm (even if both are assembled fully).

This is indeed very likely. 

> Also, I am not sure if it is advisable, but it is possible to consider `ELEMENT` and `PARTIAL` assembly for a `LinearForm`. In general one can think of a linear form as the corresponding `BilinearForm` with test function 1, e.g. ∫fϕ is PTGTBTd, where `d` is a Q-vector with entries dk=ωkf(qk)|J(qk)|.

Fair point, though this is just a conceptual thing.

All I am saying is that I would prefer that enabling/disabling sum factorization should have a more descriptive flag for the users, rather than relying on the comments we put in the assembly levels enum. I will however go with the popular vote 😄

tzanio:
@psocratis, @pazner, and @camierjs -- I made the changes we discussed today, please take a look and let me know if you have any comments.

@mlstowell, reminder to squash-merge this when approved.

fem/plinearform.cpp
```diff
@@ -43,18 +43,9 @@ void ParLinearForm::MakeRef(ParFiniteElementSpace *pf, Vector &v, int v_offset)
...
 
-void ParLinearForm::Assemble(bool use_device)
+void ParLinearForm::Assemble()
 {
-   bool all_supports_device = use_device;
-
-   if (use_device)
-   {
-      bool supports_device = SupportsDevice();
-      MPI_Allreduce(&supports_device, &all_supports_device, 1,
```
camierjs, line 53: This logic was needed when `LinearForm::SupportsDevice()` did not return the same value in parallel (`face_info` related?). If we remove it here, we might need to move it inside `SupportsDevice`.

tzanio: This sounds like a bug to me, but if there is indeed a reason for `SupportsDevice()` to return different values on different MPI ranks, than I suggest we make it virtual and use this in the implementation of `ParLinearForm::SupportsDevice()`.

fem/tmop_tools.cpp
```diff
@@ -266,7 +266,7 @@ void ParAdvectorCGOper::Mult(const Vector &ind, Vector &di_dt) const
 
    if (al == AssemblyLevel::PARTIAL)
    {
-      K.ParFESpace()->GetParMesh()->DeleteGeometricFactors();
+      K.ParFESpace()->GetParMesh()->NodesUpdated();
    }
```
pazner, line 270: Ditto

> Is the `if` statement necessary? Seems simpler/clearer to just always call `NodesUpdated`, if the form is not using PA, then the geometric factors won't be requested, and no harm will be done...

fem/linearform.hpp
```diff
@@ -185,10 +209,15 @@ class LinearForm : public Vector
        corresponding pointer (to Array<int>) will be NULL. */
    Array<Array<int>*> *GetFLFI_Marker() { return &boundary_face_integs_marker; }
 
+   /// Which assembly algorithm to use: the new device-compatible fast assembly
+   /// (true), or the legacy CPU-only algorithm (false). If not set, the default
+   /// value is false.  If used, this method must be called before assembly. */
```
pazner, line 214:
```suggestion
   /// @brief Which assembly algorithm to use: the new device-compatible fast
   /// assembly (true), or the legacy CPU-only algorithm (false).
   /** If not set, the default value is false.  If used, this method must be
       called before assembly. */
```

fem/linearform.hpp
```diff
@@ -185,10 +209,15 @@ class LinearForm : public Vector
...
+   void UseFastAssembly(bool use_fa);
+
    /// Assembles the linear form i.e. sums over all domain/bdr integrators.
-   /// When @a use_device is set to true and the linearform assembly is
+   /// When UseFastAssembly(true) has been called and the linearform assembly is
    /// compatible with device execution, it will be executed on the device.
```
pazner, line 219:
```c++
   /// Assembles the linear form i.e. sums over all domain/bdr integrators.
   /** When @ref UseFastAssembly "UseFastAssembly(true)" has been called and the
       linearform assembly is compatible with device execution, it will be
       executed on the device. */
```

fem/plinearform.hpp
```diff
@@ -114,9 +114,9 @@ class ParLinearForm : public LinearForm
    void MakeRef(ParFiniteElementSpace *pf, Vector &v, int v_offset);
 
    /// Assembles the ParLinearForm i.e. sums over all domain/bdr integrators.
-   /// When @a use_device is set to true and all the ParLinearForm assembly is
+   /// When UseFastAssembly(true) has been called and the linearform assembly is
    /// compatible with device execution, it will be executed on the device.
```
pazner, line 118:
```c++
   /// Assembles the ParLinearForm i.e. sums over all domain/bdr integrators.
   /** When @ref LinearForm::UseFastAssembly "UseFastAssembly(true)" has been
       called and the linear form assembly is compatible with device execution,
       the assembly will be executed on the device. */
```

fem/linearform.hpp
```diff
@@ -32,6 +32,10 @@ class LinearForm : public Vector
    /** @brief Extension for supporting different assembly levels. */
    LinearFormExtension *ext;
 
+   /// Should we use the device-compatible fast assembly algorithm (false by
+   /// default)
```
pazner, line 36:
```suggestion
   /// @brief Should we use the device-compatible fast assembly algorithm (false
   /// by default)
```

fem/linearform.hpp
```diff
@@ -78,11 +82,23 @@ class LinearForm : public Vector
    /// Copy construction is not supported; body is undefined.
    LinearForm(const LinearForm &);
 
+   /// Defaults values to be used in constructors
+   void Init()
+   {
+      ext = nullptr;
+      extern_lfs = 0;
+      fast_assembly = false;
```
pazner, line 90: Maybe just provide these as the default values for the data members and then we can delete `LinearForm::Init` entirely?

pazner (commented):
I think this change is looking good! Please forgive the Doxygen nitpicks 😇 

I also had a few other comments/suggestions

psocratis (approved):
This looks good to me!

fem/plinearform.hpp
tzanio: Fixed in https://github.com/mfem/mfem/pull/3235/commits/343e7ca7c8c5f4b432f97f8632702f8ce5ef2c6c

fem/linearform.hpp
tzanio: Fixed in https://github.com/mfem/mfem/pull/3235/commits/343e7ca7c8c5f4b432f97f8632702f8ce5ef2c6c

fem/linearform.hpp
tzanio: Fixed in https://github.com/mfem/mfem/pull/3235/commits/343e7ca7c8c5f4b432f97f8632702f8ce5ef2c6c

fem/tmop_tools.cpp
```diff
@@ -193,7 +193,7 @@ void SerialAdvectorCGOper::Mult(const Vector &ind, Vector &di_dt) const
 
    if (al == AssemblyLevel::PARTIAL)
    {
-      K.FESpace()->GetMesh()->DeleteGeometricFactors();
+      K.FESpace()->GetMesh()->NodesUpdated();
    }
```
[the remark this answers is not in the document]
tzanio: I like this, changed in https://github.com/mfem/mfem/pull/3235/commits/8a01807a924921ede00b0566367d6bee979c3526

fem/tmop_tools.cpp
tzanio: Changed in https://github.com/mfem/mfem/pull/3235/commits/8a01807a924921ede00b0566367d6bee979c3526

fem/tmop.cpp
```diff
@@ -2352,9 +2352,9 @@ void AdaptivityEvaluator::SetParMetaInfo(const ParMesh &m,
 void AdaptivityEvaluator::ClearGeometricFactors()
 {
 #ifdef MFEM_USE_MPI
-   if (pmesh) { pmesh->DeleteGeometricFactors(); }
+   if (pmesh) { pmesh->NodesUpdated(); }
 #else
-   if (mesh) { mesh->DeleteGeometricFactors(); }
+   if (mesh) { mesh->NodesUpdated(); }
 #endif
 }
```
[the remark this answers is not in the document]
tzanio: @vladotomov, can we remove or rename this method?

fem/linearform.cpp
```diff
@@ -155,17 +156,22 @@ bool LinearForm::SupportsDevice()
...
 }
 
-void LinearForm::Assemble(bool use_device)
+void LinearForm::UseFastAssembly(bool use_fa)
 {
-   Array<int> vdofs;
-   ElementTransformation *eltrans;
-   DofTransformation *doftrans;
-   Vector elemvect;
+   fast_assembly = use_fa;
```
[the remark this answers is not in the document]
tzanio: Good idea, done in https://github.com/mfem/mfem/pull/3235/commits/1324cc88309a42f3ea43937511e57cd6157efa1f

tests/unit/fem/test_linearform_ext.cpp
```diff
@@ -233,11 +233,13 @@ TEST_CASE("Linear Form Extension", "[LinearFormExtension], [CUDA]")
...
       LinearForm d1(&L2fes);
       d1.AddDomainIntegrator(new DomainLFIntegrator(norm2_grad_x));
+      d1.UseFastAssembly(true);
       d1.Assemble();
 
       LinearForm d2(&L2fes);
       d2.AddDomainIntegrator(new DomainLFIntegrator(norm2_grad_x));
-      d2.Assemble(false);
+      d2.UseFastAssembly(false);
+      d2.Assemble();
```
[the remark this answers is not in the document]
tzanio: I see what you mean, but I think the code looks cleaner the way it is... how will you rewrite it?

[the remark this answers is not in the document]
pazner: Something like this, but not critical...

```c++
d.UseFastAssembly(true);
d.Assemble();
Vector d_fast = d;

d.UseFastAssembly(false);
d.Assemble();
d -= d_fast;
```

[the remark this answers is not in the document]
tzanio: I have a slight preference for the current version, but either one is fine for me.

@psocratis, @camierjs and @mlstowell -- what do you think?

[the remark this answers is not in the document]
psocratis: The current version looks slightly cleaner to me but no biggie!

fem/plinearform.cpp
tzanio: @camierjs and @pazner, am I correct that `LinearForm::SupportsDevice()` can return different values on different MPI ranks because of the following code?

https://github.com/mfem/mfem/pull/3144/files#diff-c6385ef7dfee72c05f1246af1a530036f80e2cefb776734e0c03328530c56572R124-R141

If so, should I add `ParLinearForm::SupportsDevice()` as suggested above?

fem/tmop.cpp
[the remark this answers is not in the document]
vladotomov: Actually this method was written by @tomstitt and is used in BLAST.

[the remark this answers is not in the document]
tzanio: In that case maybe it is best to restore the old implementation?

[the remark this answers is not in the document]
tomstitt: Thanks @vladotomov ! Yes, we use this in blast

fem/plinearform.cpp
tzanio: See https://github.com/mfem/mfem/pull/3235/commits/6eb6556a048d768ffe6141bc58e16056177e2ae2

fem/tmop.cpp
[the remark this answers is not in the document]
tzanio: See https://github.com/mfem/mfem/pull/3235/commits/6eb6556a048d768ffe6141bc58e16056177e2ae2

tzanio:
I believe I have addressed all outstanding comments on this PR, please let me know if I missed something, or if you have any additional suggestions. Thanks!

pazner (approved):
Looks great, thanks for the effort on this PR @tzanio!

camierjs (approved):

tzanio:

PR #3319, opened 2022-11-10
Reproduce bug for nc meshes and face quadrature space

artv3:
`mpirun -n 2 ./ex9p -pa ` will give the following error in debug mode

```
ref - mesh.GetNumFacesWithGhost() = 120
ref - mesh.GetNumFacesWithGhost() = 105


Assertion failed: (i>=0 && i<size) is false:
 --> Access element 79 of array, size = 60
 ... in function: const T &mfem::Array<mfem::Element *>::operator[](int) const [T = mfem::Element *]
 ... in file: mesh/../general/array.hpp:746
```

### DIFF
--- a/examples/ex9p.cpp
+++ b/examples/ex9p.cpp
@@ -1,4 +1,4 @@
-//                       MFEttM Example 9 - Parallel Version
+//                       MFEM Example 9 - Parallel Version
 //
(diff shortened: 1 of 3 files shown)


pazner:
Fixed in #3320

PR #3326, opened 2022-11-15
Ns flat plate

Jeff-Hadley:
New miniapp example - navier_flat_plate.cpp 

Solves incompressible flow over a flat plate. 

This is the skeleton file to properly build up the example. Values of flow variables still need to be updated before running the example. Also need the mesh file.


tzanio:
@Jeff-Hadley, did you mean to close this? (just checking)

tzanio:
no worries

PR #3327, opened 2022-11-16
Add const Vector& overload for Distance

pazner:
Adds overload to `Distance` to avoid calling `HostRead` at the call site.

Target is PR #3194.


mlstowell:
Thanks Will!

mlstowell (approved):

tzanio (approved):

Issue #3311, opened 2022-11-07
Development of New Example


como2099:
@Jeff-Hadley @Mateus-A-R-Braga and I will be developing a new example. It will solve the incompressible Navier-Stokes equations for a two-dimensional stagnation flat plate geometry, with a comparison study to the incompressible Blasius flat plate solution (time permitting). We will build off of existing examples 9, 14, 17, 18 and the Navier miniapp. This is a project for a class at CU Boulder this fall semester, so we plan to have this "issue" resolved by 11/28/2022 via a pull request that will introduce our new example (in order to fit into the timeline necessitated by the class).

@jedbrown

tzanio:
Welcome @como2099, @Jeff-Hardley and @Mateus-A-R-Braga!

Please let us know if you have any questions.

PR #3345, opened 2022-11-28
[WIP] New Navier Miniapp Example: 2d Flat Plate Boundary Layer

Jeff-Hadley:
This is a contribution based on the original issue ticket number #3311 as part of a class project at CU Boulder. 

We have begun construction of a flat plate boundary layer example using the existing navier_solver.cpp source code. 

The new example's problem description is detailed in the file `navier_flat_plate.cpp`, and the 3 meshes accompanying the example are `flat_plate_coarse.msh`, `flat_plate_coarser.msh`, and `flat_plate_coarsest.msh` (included in the pull request). The example can be built in it's default state and runs with the `flat_plate_coarse.msh` file (see image below). The boundary layer develops, but there is an unexplained acceleration of the flow in the center of the domain.

Additionally, during our testing of the example we ran into stability issues. We ran the example for all three meshes and order 1 through 3 shape functions (9 trials total), finding that some cases would converge to a somewhat correct solution while some would diverge and blow up, with seemingly no correlation between convergence and mesh/shape function selection. We were unable to achieve convergence running any cases of order 3 or higher. We also tried changing the problem by prescribing the top boundary of the domain to be a slip-wall condition (as opposed to the stress free condition originally prescribed), and ran into essentially the same problems with even more non-physical artifacts.

Seeing as there is no correlation between mesh refinement or shape function order and convergence, we believe the current state of `navier_solver.cpp` may not be applicable to the flat plate case we are trying to develop. We have completed this project with regards to our course requirements; however, we are open to continuing this discussion in search of an explanation for the irregular results we have obtained. 

@Mateus-A-R-Braga, @como2099, @jedbrown 

image (8)


tzanio:
Please let us know if you need any help with resolving the CI failure

jandrej:
Let's try to figure out the issues with convergence here. First of all, and don't take this the wrong way, the mesh is horrible. It doesn't represent the transition between farfield and plate on an element edge and also is way too coarse on the plate boundary as well as too resolved in the farfield.

After mesh resolution you have to look at aspect ratios and structure of the elements. Preferably you create a plate-aligned mesh that is block structured. Take a look at proper grids for this problem [here](https://turbmodels.larc.nasa.gov/flatplate_grids.html).

You didn't mention how you came up with the the parameters you chose. Is this due to non-dimensionalization? Your grid coordinates are not representing the benchmark problem, hence the question.

After these issues are resolved and you still don't get adequate convergence you need to go into adaptive time stepping based on CFL numbers.

Let me know if this is a problem you really need to see working with Navier and I'll take the time and work with you on getting benchmark results.

PR #3359, opened 2022-12-11
Adding support for empty partitions in AMS and ADS

mlstowell:
Setup methods for HypreAMS and HyperADS require the basis function order. This cannot be obtained from the `FiniteElement` when a partition has no elements. As a result empty partitions use an order of 1 regardless of the true order. This then leads to the setup methods hanging in collective MPI calls when the true order is greater than 1.

This PR obtains the order from the `FiniteElementCollection` instead.


mlstowell:
@tzanio, I haven't worked much with trace spaces. I'd appreciate it if you could take a look at these changes to see if I've determined the order correctly in this context.

PR #3466, opened 2023-02-15
Small fix in Adams Moulton and Adams Bashforth integrators [ab-am-fix]

IdoAkkerman:
During the first few steps, when the history is not build yet, the integrator runs RK. In those steps the time was incremented twice.

This PR fixes this, and also improves the unit test in order to capture these errors.

| [#3466](https://github.com/mfem/mfem/pull/3466) | @IdoAkkerman | @tzanio | @atallah727 + @cjvogl | 2/23/23 | 3/25/23 | 3/30/23 | |

<details>
    
- [x] Code builds.
- [x] Update `CHANGELOG`:
    - [x] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [x] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [x] Update `INSTALL`:
    - [x] `.github`
    - [x] `.appveyor.yml`
- [x] Update `.gitignore`:
- [x] New examples:
- [x] New miniapps:
- [x] New capability:
</details>


IdoAkkerman:
This fixes the issue raised in #2343

bensworth:
This looks good; I think we should still add a check for nonuniform time stepping. In principle one can derive a general polynomial interpolation for multistep methods through non uniform points, but I don't think you can just apply this method naively. I think this is particularly relevant given some of the one-line adaptivity hidden in some of the examples, where one might not know the effect on integrators. What I have done in implementations is effectively restart if dt changes. So rather than just initialize using RK on the first step, have an initialization flag, and if dt changes you set that flag to true and start over with an RK one-step method.

IdoAkkerman:
That makes sense. I will add that.
To make sure that dt is not continuously changed and you are effectively running RH instead of AB or AM, I will add mfem::out statement to make the user aware of what is happening.

bensworth:
Nice, this looks good. Only problem at this point is the tests are failing. Once the tests pass should be good to go

linalg/ode.cpp
```diff
@@ -446,6 +470,7 @@ AdamsMoultonSolver::AdamsMoultonSolver(int s_, const double *a_)
    smax = std::min(s_+1,5);
    a = a_;
    k = new Vector[5];
```
bensworth, line 472: Why make a pointer of size 5 rather than smax?

linalg/ode.cpp
```diff
@@ -407,6 +417,20 @@ void AdamsBashforthSolver::Init(TimeDependentOperator &f_)
```
bensworth, line 417: The % sign in `idx[i] = (smax-i)%smax;` is unnecessary right, this is just ordering the idx backwards, idx[0]=smax,...,idx[smax-1]=1?

bensworth:
Above comments apply to AM and AB due to code similarity.

Few other questions/comments on orders to include:
- The first order methods are just forward and backward Euler. @pazner do you want these included, or should we remove to avoid multiple equivalent definitions?
- The 5th order Adams Moulton method is not viable under the current MFEM methods, because there is no start up method of order 5. As a result, the method will not give 5th order accuracy; if your start up is 4th order, in general it doesn't matter if you integrate exactly thereafter, you will still have 4th order in time. As a result, I don't think this method should be included, or a 5th order DIRK/ESDIRK method should be added as well. The stability region of high order AM methods is horrendous anyways, so I think it is okay to not include.
- The 5th order AB method needs an appropriate order start up method. Fortunately, MFEM has an explicit RK6 now that you can use for this, but this should be added to code

linalg/ode.cpp
```diff
@@ -11,6 +11,9 @@
 
 #include "operator.hpp"
 #include "ode.hpp"
+#ifdef MFEM_USE_MPI
+#include "../general/communication.hpp"
+#endif
```
pazner, line 16: I don't think the `#ifdef` guard is needed.

linalg/ode.cpp
```diff
@@ -407,6 +417,20 @@ void AdamsBashforthSolver::Init(TimeDependentOperator &f_)
...
+      dt_ = dt;
+
+      if (print)
+      {
+         mfem::out << "WARNING:" << std::endl;
+         mfem::out << " - Time step changed" << std::endl;
+         mfem::out << " - Purging Adams-Bashforth history" << std::endl;
+         mfem::out << " - Will run Runge-Kutta to rebuild history" << std::endl;
+      }
+   }
```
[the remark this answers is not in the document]
pazner: Good point. It would be good to omit this output on the first step.

linalg/ode.cpp
IdoAkkerman: You are right. This should be smax. Note:
1)  This is not part of the PR.
2)  This issue is fixed in #3480

linalg/ode.cpp
IdoAkkerman: For i=0 this is not the case, though.
We could also consider `idx[i] = smax-i -1`.
Perhaps that is nicer then the current.
This could perhaps better be done in #3480

linalg/ode.cpp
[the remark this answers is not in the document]
IdoAkkerman: Agree. Will work on that.

linalg/ode.cpp
IdoAkkerman: I was not sure. I was a bit struggling with the making sure only root was outputting. I will remove the guard and check if things still work.

IdoAkkerman: Done

linalg/ode.cpp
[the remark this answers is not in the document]
IdoAkkerman: I initialize` dt_=-1` and the warning is not printed when ` dt_ < 0`.

IdoAkkerman:
Requests related to changes in this PR have been accommodated.

Other comments regarding the AB and AM classes have been handled in #3480

bensworth (approved):
Looks great, thanks @IdoAkkerman

tzanio:

linalg/ode.cpp
```diff
@@ -362,6 +364,12 @@ AdamsBashforthSolver::AdamsBashforthSolver(int s_, const double *a_)
    {
       RKsolver = new RK4Solver();
    }
+
+#ifdef MFEM_USE_MPI
+   print = mfem::Mpi::IsInitialized() ? mfem::Mpi::Root() : true;
```
cjvogl, line 369: What does `mfem::MPI::Root()` return if `mfem::Mpi::IsInitialized == false`? Asking to gain an understanding for this conditional.

linalg/ode.hpp
```diff
@@ -232,6 +232,8 @@ class AdamsBashforthSolver : public ODESolver
    Vector *k;
    Array<int> idx;
    ODESolver *RKsolver;
+   double dt_;
+   bool print;
```
cjvogl, line 236: Instead of adding a member variable, can all conditionals that depend on `print` use `mfem::Mpi::Root()` instead?  

From a design point of view, I think it's better to leave the concept of which rank is root encapsulated within the `Mpi` singleton. While I don't think this is really a use case quite yet, such an approach would avoid an issue if the communicator is modified in such a way that root is changed between the construction of the `AdamsBashforthSolver` and the calls to `Step`

linalg/ode.hpp
```diff
@@ -314,6 +316,8 @@ class AdamsMoultonSolver : public ODESolver
...
```
cjvogl, line 320: See comment for addition of `print` member variable in `AdamsBashforthSolver`....

linalg/ode.cpp
pazner: I believe it will result in a runtime MPI error. `Mpi::Root()` has to make a call to the MPI API, which will error if `MPI_Init` has not been called yet.

pazner: Also, I don't think these need to be qualified with `mfem::`. It should be fine to use `Mpi::IsInitialized` and `Mpi::Root` since this is within the `mfem` namespace.

cjvogl: Ah I see that now in the `Mpi` singleton. In my opinion, the correct behavior if this line is reached and `MPI_Init` has not yet been called is an error is generated. If there's agreement on that, I recommend the conditional be removed and any error handling for the case when the user has forgotten to call `Mpi::Init` be left for implementation in the `Mpi` singleton.

pazner: It's perfectly valid to compile MFEM with MPI, but not call `MPI_Init` (as long as none of the parallel functionality is used). This is what happens in all of the serial examples when MFEM is configured to use MPI.

cjvogl: That makes sense.  What do the other MFEM classes that are called from both serial and parallel examples do when they want to print to the terminal?

pazner: I think most instances have similar logic. (Yes, there is probably a good opportunity for refactoring, or adding something like `ParPrint` that handles this automatically, but I would recommend doing that separately from this PR).

linalg/ode.hpp
cjvogl: Noting the [conversation above](https://github.com/mfem/mfem/pull/3466#discussion_r1115963315) regarding calling `Mpi::Root()` from serial examples, I would suggest here moving the conditional logic into a private member function instead of store the value in a member variable:
```c++
inline bool print()
{
#ifdef MFEM_USE_MPI
   return Mpi::IsInitialized() ? Mpi::Root() : true;
#else
  return true;
#endif
}

bensworth:
How come function vs variable?

cjvogl:
> How come function vs variable?

My concern is if which rank is root changes between the time the variable is set and the variable is read, then the value of the variable is now inconsistent which which rank is root.  More generally, I'd prefer that the concept of what is root should remain encapsulated within the `Mpi` singleton for extensibility (e.g., different communicators considered on heterogeneous computing environments that ASCR seems to be very interested in lately).

linalg/ode.hpp
IdoAkkerman: @cjvogl I agree! I was just a bit rusty and did not get it to work.. so went for something simple.
Your code snippet is very much appreciated.

linalg/ode.hpp
IdoAkkerman: I have made the change but put it in PR #3480.

cjvogl: Please excuse my confusion: what is the motivation for putting the change in PR #3480 instead of this one?

linalg/ode.hpp
cjvogl: Thanks for making the change @IdoAkkerman!

IdoAkkerman:
@cjvogl I made the root print change in the upstream PR.

cjvogl:
> @cjvogl I made the root print change in the upstream PR.

I suppose my confusion is what the change wasn't made to this PR?

linalg/ode.hpp
cjvogl: Noting here that this conversation has been continued in the comments.

cjvogl (changes requested):
With the due date approaching tomorrow, I'm just formally noted there are two open topics.

linalg/ode.hpp
IdoAkkerman: This PR was meant to fix the double dt issue.
The other PR is a refactorisation and small feature addition. I expected this to be small PR. The dt check was already edit per review request. Perhaps remove it here and really make it a dt fix PR?

IdoAkkerman: Done.

IdoAkkerman:
@cjvogl I had to change the includes back to the ones with `..`. Otherwise the ubuntu-latest-make-dbg-seq-int32  test (and at least one other test) fails.

cjvogl:
> @cjvogl I had to change the includes back to the ones with `..`. Otherwise the ubuntu-latest-make-dbg-seq-int32 test (and at least one other test) fails.

Arg, ok so I presume the Makefile build system doesn't properly set the `-I` flags as the CMake build does.  Sorry to suggest something that breaks the Makefile build!

cjvogl (approved):
Looks good to me!
- All concerns addressed
- All unit tests passed locally (i.e., `make test`)

tzanio:

PR #3524, opened 2023-03-06
Complete domain integrator marker support

sebastiangrimberg:
Adds support for domain markers for `MixedBilinearForm`, `DiscreteLinearOperator`, and `NonlinearForm` classes.

| [#3524](https://github.com/mfem/mfem/pull/3524) | @sebastiangrimberg | @mlstowell | @jandrej + @dylan-copeland | 3/7/23 | 11/6/23 | ⌛due 3/28/23 | |


mlstowell:

fem/bilinearform.hpp
```diff
@@ -1059,6 +1070,9 @@ class DiscreteLinearOperator : public MixedBilinearForm
    /// Adds a domain interpolator. Assumes ownership of @a di.
    void AddDomainInterpolator(DiscreteInterpolator *di)
    { AddDomainIntegrator(di); }
+   void AddDomainInterpolator(DiscreteInterpolator *di,
+                              Array<int> &elem_marker)
+   { AddDomainIntegrator(di, elem_marker); }
```
dylan-copeland, line 1075: `DiscreteLinearOperator::Assemble` does not check the markers. Should that check be added?

sebastiangrimberg: Good catch, fixed in [052b5e0](https://github.com/mfem/mfem/pull/3524/commits/052b5e0cf20d35216e7dfb87d538bb4aceca074e). Thanks!

dylan-copeland (approved):

fem/bilinearform.cpp
```diff
@@ -1895,41 +1923,56 @@ void DiscreteLinearOperator::Assemble(int skip_zeros)
...
+               elmat += elemmat;
+            }
          }
          if (ran_dof_trans || dom_dof_trans)
          {
-            TransformPrimal(ran_dof_trans, dom_dof_trans, totelmat);
+            TransformPrimal(ran_dof_trans, dom_dof_trans, elemmat);
          }
-         mat->SetSubMatrix(ran_vdofs, dom_vdofs, totelmat, skip_zeros);
+         mat->SetSubMatrix(test_vdofs, trial_vdofs, elemmat, skip_zeros);
```
jandrej, line 1975: Inconsistent naming of Range and Domain space by using Test and Trial here.

sebastiangrimberg: Yes, but it is now consistent with Trial and Test FE spaces above (like for `MixedBilinearForm`):

```
trial_fes->GetFaceVDofs(i, trial_vdofs);
test_fes->GetFaceVDofs(i, test_vdofs);
```

Would you still like me to revert this?

jandrej: If it was in `Trial and Test FE spaces above (like for MixedBilinearForm` already you can leave it.

sebastiangrimberg: Indeed the previous looked like 

```
trial_fes->GetFaceVDofs(i, dom_vdofs);
test_fes->GetFaceVDofs(i, ran_vdofs);
```

and I just updated to match `MixedBilinearForm` and use the `trial_vdofs` and `test_vdofs` members inherited from `MixedBilinearForm`.

sebastiangrimberg:
@jandrej, sorry to bother but is there anything holding up your approval of this minor PR? Thanks a lot, just hoping to get it merged when ready.

sebastiangrimberg:
Hi @mlstowell, I came across an issue today testing this branch on ex31p, using the sample run: `mpirun -np 4 ex31p -m ../data/hexagon.mesh -o 2`. I had previously tried to replace the element assembly in `DiscreteLinearOperator::Assemble` from:

```
{
   ...

   domain_integs[0]->AssembleElementMatrix2(*dom_fe, *ran_fe, *T, totelmat);
   for (int j = 1; j < domain_integs.Size(); j++)
   {
      domain_integs[j]->AssembleElementMatrix2(*dom_fe, *ran_fe, *T, elmat);
      totelmat += elmat;
   }
   if (ran_dof_trans || dom_dof_trans)
   {
      TransformPrimal(ran_dof_trans, dom_dof_trans, totelmat);
   }
   mat->SetSubMatrix(ran_vdofs, dom_vdofs, totelmat, skip_zeros);
}
```

to:

```
{
   ...

   elmat.SetSize(test_vdofs.Size(), trial_vdofs.Size());
   elmat = 0.0;
   for (int k = 0; k < domain_integs.Size(); k++)
   {
      if (domain_integs_marker[k] == NULL ||
           (*(domain_integs_marker[k]))[elem_attr-1] == 1)
      {
         domain_integs[k]->AssembleElementMatrix2(*trial_fes->GetFE(i),
                                                  *test_fes->GetFE(i),
                                                  *eltrans, elemmat);
         elmat += elemmat;
      }
   }
   if (ran_dof_trans || dom_dof_trans)
   {
      TransformPrimal(ran_dof_trans, dom_dof_trans, elmat);
   }
   mat->SetSubMatrix(test_vdofs, trial_vdofs, elmat, skip_zeros);
}
```

But, this actually doesn't work for a `ND_R2D_FECollection` with Hypre's AMS solver. In this case, for the Nedelec interpolation matrix mapping `H1^d` to `ND`, the constructed H1^d space by `HypreAMS` has `vdim = 2`, and thus `trial_vdofs` are for `vdim = 2`, but the matrix returned by `AssembleElementMatrix2` for the `IdentityInterpolator` assumes `vdim = 3` because of the `ND_R2D_FiniteElement`. Thus, `SetSubMatrix` never used to throw an error even though `elmat`'s number of columns was actually larger than `trial_vdofs` dimension. The error was only thrown after trying to do `elmat += elemmat` after my change.

Maybe this is something I can fix in this PR as this is where the bug showed up. One approach would be to use a `dynamic_cast` in `HypreAMS::MakeGradientAndInterpolation` for any of `ND_R2D_FECollection` or `ND_R2D_Trace_FECollection`, or `ND_R1D_FECollection`, but I'm not familiar enough with these to understand what the right behavior would be. Is the idea that in this case we should always be configuring the AMS solver with `dim = 3`?

Thanks for your help. I can also just leave the code as is so that ex31p throws an error instead of silently having the unexpected behavior it did before and this issue could be fixed in a follow-up PR.

sebastiangrimberg:
Ping @mlstowell, sorry to bother you again. I'd like to move forward with this branch is possible but the only thing blocking is this uncovered issue with the restricted ND FE collections. It would be nice to resolve here if possible.

mlstowell:
Hi, @sebastiangrimberg, you're right to press me on this. I've let this slide for far too long. I remember having a similar issue with AMS when developing the R2D classes. I'm not seeing a problem with the command line you posted above. It's possible that I've implemented a work-around in my local version of HYPRE. I'll track this down and get back to you.

What version of HYPRE are you using?

sebastiangrimberg:
I was using the most recent `master` of `HYPRE` for this. Note that the head commit on this branch will not reveal the issue even though it is still there, but running ex31p with the previous commit ec729707 should throw the size mismatch error when assembling the `DiscreteLinearOperator`.

mlstowell:
Right, I realized that after my comment. I'll make that change and have a closer look. Thanks!

mlstowell:
Okay, the problem arises because of an incorrect assumption in the `HypreAMS` solver which can be found [here](https://github.com/mfem/mfem/blob/ef6074c42089a7a0eceaa8d12f69ba8462eaa26e/linalg/hypre.cpp#L5451). This line assumes that the number of components in the vector field matches the space dimension.

I think the best fix is to make the vector dimension available through the `FiniteElementCollection` but we don't currently have this in the interface and it may require some negotiation to reach a suitable function for providing this information.

A much quicker fix, which you could include in this PR as a temporary measure, would be something like this:
1) use dynamic casts to check if the `edge_fec` (in `HypreAMS::MakeGradientAndInterpolation`) is an `ND_R2D_FECollection` or `ND_R1D_FECollection`. If it is then set a new variable called `vdim` to 3. Otherwise this new variable can equal `sdim`.
2) use the new `vdim` rather than `sdim` on line 5451.
This should resolve the issue. In the future we could simply set `vdim` by querying `edge_fec` using whatever interface we settle on.

sebastiangrimberg:
@mlstowell, I have just pushed a branch `sjg/fecoll-vdim-fix` which takes the first approach you mention: See https://github.com/mfem/mfem/compare/sjg/bilinearform-integs-marker...sjg/fecoll-vdim-fix.

Depending on how controversial this is, I can rebase it onto `master` and open a new PR for review, or if the changes are OK I can just merge into this PR.

Adding `GetVDim` and using this in `HypreAMS` does indeed resolve the issue with `ex31p` as we expected. Proof that the code in `master` was buggy is shown by the results of `mpirun -np 4 ./ex31p -o 2`: Running with the tip of this branch (`sjg/bilinearform-integs-marker`) without the fix converges in 90 iterations, while with the fix from `sjg/fecoll-vdim-fix` only 27 iterations are required.

mlstowell:
I like what you've done and it's tempting to just merge it into this PR. However, AMS is Tzanio's creation and I'd like to give him a chance to comment on those changes. Please re-base those changes and create a separate PR so we can request a review from @tzanio.

sebastiangrimberg:
> I like what you've done and it's tempting to just merge it into this PR. However, AMS is Tzanio's creation and I'd like to give him a chance to comment on those changes. Please re-base those changes and create a separate PR so we can request a review from @tzanio.

See #3774. Thanks @mlstowell!

jandrej (approved):

tzanio:
@sebastiangrimberg -- what's the status of this PR given that #3774 is now in `next`?

sebastiangrimberg:
Once #3774 is merged here, we can revert commit ef6074c42089a7a0eceaa8d12f69ba8462eaa26e and this will be ready.

sebastiangrimberg:
Now that https://github.com/mfem/mfem/pull/3774 is merged into `master`, this has been finalized and is ready for `next`.

mlstowell:

PR #3533, opened 2023-03-13
Switching to single point Dirichlet BC in IrrotationalProjector

mlstowell:
Only a single point is required to make the diffusion operator non-singular and the results are usually much better.

This improvement was noticed in connection with discussion #3523.


tzanio:
@mlstowell, do you want to revive this?

PR #3541, opened 2023-03-15
Fix bugs in L2ZZErrorEstimator

hughcars:
I believe the L2ZZErrorEstimator has two bugs in it:
* It does not make use of the `DofTransformation` pointers to perform `InvTransformPrimal` and `TransformPrimal`, thus is incorrect for ND tets for p>1
* It uses `AddElementVector` to store the flux value into the resulting field. In the case of a zero initialized fully discontinuous field (i.e. L2^d as used for a Poisson flux) this has no effect. But for an ND space this results in double counting. 

The first of these I believe is a plain bug, the second of these I believe to be a bug but could be a misunderstanding on my part. I have also propagated fixing the first issue to what I believe are other instances in `gridfunc.cpp` and `pgridfunc.cpp`. I am not totally sure of these fixes so would appreciate feedback on whether I have misunderstood their application.

This was diagnosed by comparing a flux grid function computed identically to that in the `L2ZZErrorEstimator` function call, to one computed using the CurlInterpolator. The code to construct the interpolated version is
```
using namespace mfem;
mfem::Vector field_vec; // This is populated with the solution vector in ND.

ParGridFunction(&fes); // fes is the FiniteElementSpace associated with field_fes
ParDiscreteLinearOperator curl(&fes, &flux_fes); // ND and RT respectively
curl.AddDomainInterpolator(new CurlInterpolator);
curl.Assemble();
curl.Finalize()
auto *curlop = std::unique_ptr<HypreParMatrix>(curl.ParallelAssemble());

Vector flux(flux_fes.GetTrueVSize());
curlop->Mult(field, flux);

ParGridFunction flux_func(&flux_fes);
flux_func.SetFromTrueDOFs(flux);
```

| [#3541](https://github.com/mfem/mfem/pull/3541) | @hughcars | @tzanio | @mlstowell + @psocratis + @rw-anderson | 3/25/23 | 4/29/23 | 5/4/23 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:

tzanio:
@psocratis and @rw-anderson -- can you please review when you get a chance?

fem/pgridfunc.cpp
```diff
@@ -325,8 +343,8 @@ void ParGridFunction::GetVectorValue(int i, const IntegrationPoint &ip,
    if (nbr_el_no >= 0)
    {
       Array<int> dofs;
-      DofTransformation * doftrans = pfes->GetFaceNbrElementVDofs(nbr_el_no,
-                                                                  dofs);
+      const auto * const doftrans = pfes->GetFaceNbrElementVDofs(nbr_el_no,
+                                                                 dofs);
```
mlstowell, line 347: What is the advantage of using `auto` here?

mlstowell:
Hi, @hughcars,

I think your usage of `DofTransformation` is correct here and I thank you for catching these issues. I am curious why you choose to use `const auto * const` rather than `DofTransformation *`. If they are equivalent, I would prefer the second option because it is more specific and more amenable to the use of tools like `grep`.

Regarding the double counting i.e. `AddElementVector` vs. `SetSubVector`, I'm sure you are correct that this double counting is a problem. However, I'm not sure of the appropriate fix. Switching to `SetSubVector` will remove the double counting but it introduces a dependence on the order of element traversal. I think what we want is to build a right-hand-side vector `b` by summing contributions from neighboring elements. I'm not sure why the code was not written this way initially (@rw-anderson or @psocratis may have keener insight on this matter.) 

My inclination is to rewrite this portion of `L2ZZErrorEstimator` but your fix is an improvement over the double counting error that currently exists.

Best wishes,
Mark

mlstowell:
Regarding the use of `auto`, I don't think the quoted C++ guidelines apply in this case. There is no repetition of the typename and the function call doesn't give any clues to the reader indicating what type should be expected. This would require the reader to lookup the function in the header files to figure out what object is being used. Unless the other reviewers disagree I would prefer using `DofTransformation *` (adding the `const` keywords is fine).

I see your point about the flux computation. Without coefficients and assuming the user chooses the correct space for the flux everything should be the same regardless of traversal order.

hughcars:
> Regarding the use of `auto`, I don't think the quoted C++ guidelines apply in this case. There is no repetition of the typename and the function call doesn't give any clues to the reader indicating what type should be expected. This would require the reader to lookup the function in the header files to figure out what object is being used. Unless the other reviewers disagree I would prefer using `DofTransformation *` (adding the `const` keywords is fine).
> 
> I see your point about the flux computation. Without coefficients and assuming the user chooses the correct space for the flux everything should be the same regardless of traversal order.

Fixed in https://github.com/mfem/mfem/pull/3541/commits/1c25aeb978448c354da5140aad68c52c4ed4c364

mlstowell (approved):

psocratis (approved):

tzanio:

hughcars:
> Can we mention in `CHANGELOG` so users are not surprised to see changes?

Yes, I can write an entry, does this work?
```
- Bugfix for ZZ error estimation with Nedelec spaces: interface unknowns are no longer double counted, and element transformations for p >= 2 Nedelec elements are now accounted for correctly.
```

PR #3546, opened 2023-03-17
Issue: MAGMA variable batch LU solver example

artv3:
This is PR holds an example of trying to exercise MAGMA's variable batch LU factorization and solver routines. 

At the moment I have the factorization working; however since there is no direct companion solver routine I am trying to use the `magmablas_dtrsm_vbatched` routines to form a solver. Unfortunately I am not sure how to call the function properly. I'm hoping if someone from the MAGMA team has some time the could take a look at main.cpp.


examples/main.cpp
```diff
@@ -0,0 +1,222 @@
...
+   //Current error:
+   // hip error code: 'hipErrorInvalidValue':1 at /long_pathname_so_that_rpms_can_package_the_debug_info/data/driver/rocBLAS/library/src/rocblas_auxiliary.cpp:633
+   //
+
+
+   //L(Ux) = b //lower solve
+   magmablas_dtrsm_vbatched(MagmaLeft, MagmaLower, MagmaNoTrans,
+                            MagmaNonUnit,
+                            num_of_rows.ReadWrite(),
+                            rhs_num_of_cols.ReadWrite(),
```
nbeams, line 193: I believe that error is coming from `magma_getvector`, which is called in `magmablas_dtrsm_vbatched`: 

```
    // compute the max. dimensions
    magma_imax_size_2(m, n, batchCount, queue);
    magma_int_t max_m, max_n;
    magma_getvector(1, sizeof(magma_int_t), &m[batchCount], 1, &max_m, 1, queue);
    magma_getvector(1, sizeof(magma_int_t), &n[batchCount], 1, &max_n, 1, queue);
    ```
    so both the `m` and `n` arrays need to be of size `NE + 1`.  The `magma_imax_size_2` function will put the max value in the "extra" spot at the end.  That's why `trsm_vbatched` is trying to copy the last elements of the arrays, assuming that the arrays are of length `batchCount + 1`.
    
  So, both `num_of_rows` and `rhs_num_of_cols` need to be 1 element larger, but you don't have to initialize the last element (MAGMA will do that).

File: examples/main.cpp [hunk 2]
```diff
@@ -0,0 +1,222 @@
...
+   //
+   //Question: How do we perform the batch LU/Solve ?
+   //
+
+
+   //rhs only has 1 column
+   Array<int> rhs_num_of_cols(NE);
+   rhs_num_of_cols = 1;
+
+   //Question: Do we need to perform the pivoting ourselfs?
```
nbeams, line 166: Yes, I believe so.  The `dgetrs_batched` routine uses `magma_zlaswp_rowserial_batched` with the `dipiv_array`.  But, it looks like the `vbatched` version is slightly different ("left" and "right" versions).

artv3:
Thanks @nbeams! and to confirm, we have to apply the pivoting ourselves to the RHS prior to making the solve call?

nbeams:
Correct.

PR #3618, opened 2023-04-21
Unified `BilinearFormIntegrator::GetRule` Interface

sebastiangrimberg:
NOTE: This branch is based on https://github.com/mfem/mfem/pull/3526 (in review)

The current interface in `BilinearFormIntegrator`, `NonlinearFormIntegrator`, and `LinearFormIntegrator` classes for selecting quadrature rules leads to code duplication and makes extensibility difficult. This PR introduces a top level `GetRule` member for integrator classes to override. This allows for several lifestyle improvements:

- This unifies the choice of integration rule and removes any code duplication between `AssemblyLevel::Legacy` element matrix assembly and newer variants (`Partial`, `None`, `Full`, ...). Currently there are integrators which instead of implementing their own static `GetRule` are using a different one which already exists ([here](https://github.com/mfem/mfem/blob/master/fem/bilininteg_hcurl.cpp#L960) and [here](https://github.com/mfem/mfem/blob/master/fem/bilininteg_hdiv.cpp#L1496) for example), leading to different quadrature rules between assembly levels which is not ideal.
- It removes the need for customizations for each integrator to be duplicated in the [libCEED interface](https://github.com/mfem/mfem/blob/master/fem/ceed/interface/util.cpp#L194).
- Users and application codes, if wishing to use an existing integrator but with a custom integration order, currently only have the option of using `SetIntRule` which does not work for mixed meshes. The new structure allows easy construction of a derived integrator class which only overrides the `GetRule` method and does not have to fully reimplement `AssembleElementMatrix` just to use a custom integration rule.

I realize that this does touch a lot of code which has not been touched for a long time, and so there may be some reluctancy to adopting this PR. Hence I'd love to hear what people think about reviewing and merging this. I have checked and there are only two spots where these modifications have actually modified the existing default quadrature order in order to reduce some code duplication where I thought it was OK:

- `BoundaryMassIntegrator` gets an added `Trans.GetOrderW()` in order to match `MassIntegrator`
- `GroupConvectionIntegrator` gets an added `Trans.GetOrder()` in order to match `ConvectionIntegrator`.

Also I apologize for a lot of whitespace changes, I can revert any of these as desired if it is decided to go forward with this PR.


sebastiangrimberg:
Closing now that the dependent PRs have been closed and moved to downstream repositories.

PR #3654, opened 2023-05-07
Add missing header to fix build for GCC 13

topazus:
Ref: https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes


pazner:
Dup of #3630?

topazus:
> Dup of #3630?

okay, got it.

Issue #3691, opened 2023-05-24
Bug in `ParNCMesh` with p > 2 H1 space on conformal prism mesh


hughcars:
I have a mesh `p1_prism.msh`, available on branch `cavity-prism-par-ncmesh-bug`, which when run with `ex1p` runs fine, with and without a uniform refinement:
```
mpirun -np 3 ex1p -m ../../data/p1_prism.msh -o 2
```

The initial `p1_prism` mesh as shown by glvis is
<img width="535" alt="Screenshot 2023-05-24 at 10 08 13 AM" src="https://github.com/mfem/mfem/assets/114775781/40a18eea-a8f5-4583-938e-10656b2fed00">
and the uniformly refined mesh is 
<img width="535" alt="Screenshot 2023-05-24 at 10 08 50 AM" src="https://github.com/mfem/mfem/assets/114775781/f067eab9-e385-4e02-9144-a1465854f28c">

Passing the same mesh to `ex6p` however causes a range of issues. If a uniform refinement is done on the serial mesh before `EnsureNCMesh` is called, calling `mpirun -np 3 ex6p -m ../../data/p1_prism.msh -o 2 -h1` (using `-h1` avoids triggering a related bug for RT spaces, see https://github.com/mfem/mfem/issues/3671) gives these error messages:
```
Verification failed: (key >= 0) is false:
 --> R1 index 35 not found.
 ... in function: const NCMesh::MeshId &mfem::NCMesh::NCList::LookUp(int, int *) const
 ... in file: [redact]/mfem/mesh/ncmesh.cpp:3474

Verification failed: (key >= 0) is false:
 --> R0 index 13 not found.
 ... in function: const NCMesh::MeshId &mfem::NCMesh::NCList::LookUp(int, int *) const
 ... in file: [redact]/mfem/mesh/ncmesh.cpp:3474
```
In the initial phase of `BuildParallelConformingInterpolation` where local true dofs are being sent to other ranks [here](https://github.com/mfem/mfem/blob/master/fem/pfespace.cpp#L2509). The `R0` and `R1` in the above error messages are the ranks of the failing processors.

[...]

PR #3709, opened 2023-06-05
Bug fixes for NC prism meshes

hughcars:
This PR contains fixes to resolve the issues identified in https://github.com/mfem/mfem/issues/3691 and https://github.com/mfem/mfem/issues/3671.

These were tested using 
```
#!/bin/bash
for np in {2,3,6}
do
  for o in {1,2,3}
  do
    mpirun -np $np ex6p -m ../../data/p1_prism.msh -o $o
    mpirun -np $np ex6p -m ../../data/p2_prism.msh -o $o
  done
done
```
which previously errored for all. 

The test prism meshes are included. The most pernicious issue was the use of `std::lower_bound` in the case of a non bijective map from face entity to dofs. Addressing this appears to have resolved the other issues as well (at least the previously failing tests now pass). Using `std::upper_bound` and stepping backwards ensures that only entities that increment the dof count are found, rather than the first entity at a given total dof count. 

Some style/readability changes were made particularly within `pfespace.cpp`, these were made to help debugging, to reduce the number of nested closures and refactor some home rolled algorithms to use those provided by the STL. Also fixed some warnings of unused variables/leaks in some tests.

| [#3709](https://github.com/mfem/mfem/pull/3709) | @hughcars | @mlstowell | @dylan-copeland + @jakubcerveny + @mlstowell | 6/5/23 | 7/14/23 | ⌛due 6/26/23 | |


tests/unit/linalg/test_operator.cpp
```diff
@@ -31,23 +31,19 @@ TEST_CASE("Operator", "[Operator]")
    {
       // When cg is in a product (on the right), we require cg->iterative_mode to be false.
 
-      ProductOperator *product = NULL;
-      RAPOperator *rap = NULL;
-      TripleProductOperator *triple = NULL;
```
hughcars, line 36: These variables were all lighting up for unused variables (in error) but I opted to fix it anyway.

mlstowell:

fem/pfespace.cpp
```diff
@@ -1842,9 +1842,13 @@ int ParFiniteElementSpace::PackDof(int entity, int index, int edof) const
...
-   const int* pos = std::lower_bound(array, end, value);
-   MFEM_VERIFY(pos != end, "value not found");
-   return pos - array;
+   const int* pos = std::upper_bound(array, end, value);
+   MFEM_VERIFY(pos != array, "value not found");
+   if (pos == end)
+   {
+      MFEM_VERIFY(*(array+size - 1) == value, "Last entry must be exact")
+   }
+   return pos - array - 1;
```
hughcars, line 1851: The impact of this can be best illustrated by considering a sequence of P2 triangles and quads. p2 triangles have no face dofs, so {Q, T, T, Q, Q} would gives {0, 1, 1, 1, 2} as the dof numbering. Searching with lower bound for 1, would give the first triangle (T), whereas the true target is the second quad (Q). This was the scenario that allowed me to catch this bug, but this fix appears to have resolved a lot of other failing cases I was working on. My hypothesis for this was that this error was causing the NC mesh to get into a mess such that the indexing no longer worked for any other dofs either.

fem/pfespace.cpp
```diff
@@ -2161,11 +2162,21 @@ void NeighborRowMessage::Decode(int rank)
...
+#ifdef MFEM_DEBUG_PMATRIX
+         mfem::out << "Rank " << pncmesh->MyRank << " receiving from " << rank
+                   << ": ent " << ent << ", index " << id.index
+                   << ", edof " << edof << " (id " << id.element << "/"
+                   << int(id.local) << ")" << std::endl;
+#endif
+
+         // If edof arrived with a negative index, flip it, and the scaling.
+         double s = (edof < 0) ? -1.0 : 1.0;
+         edof = (edof < 0) ? -1 - edof : edof;
```
hughcars, line 2174: This change was caught in noticing some RT dofs, when sent over the wires were arriving with negative indices once the mesh had been adapted for a while.

fem/pfespace.cpp
```diff
@@ -2440,15 +2447,50 @@ ::BuildParallelConformingInterpolation(HypreParMatrix **P_, SparseMatrix **R_,
...
+            msg << x << ' ';
+         }
+         msg << '}';
+      }
+
+      int entity, index, edof;
+      UnpackDof(dof, entity, index, edof);
+      msg << " entity " << entity << " index " << index << " edof " << edof;
+      return msg.str();
+   };
```
hughcars, line 2491: This lambda was extremely helpful during debugging, when trying to keep track of an individual failing dof. I leave it in case anyone runs into a similar morass in future, as it should print all the helpful information they might want.

tests/unit/mesh/test_ncmesh.cpp
```diff
@@ -510,16 +512,83 @@ TEST_CASE("FaceEdgeConstraint",  "[Parallel], [NCMesh]")
...
+      {
+         CheckL2Projection(pmesh, smesh, p, exact_soln);
+      }
    }
+
+
 } // test case
 
 
+
```
dylan-copeland, line 591: There are a few empty lines that could be removed.

dylan-copeland (approved):

tests/unit/mesh/test_ncmesh.cpp
hughcars: Addressed in https://github.com/mfem/mfem/pull/3709/commits/0ca88cb295f116fa9e3181737c3302c27096ee12

hughcars:
@jakubcerveny @mlstowell If there's no objections, it would be nice to merge this as it will make further NC development simpler.

mlstowell:
Hi, @hughcars,

I wanted to make what I thought would be a small request but I may have opened a whole can of worms.

The two meshes being proposed for addition may not be generally useful (although I do like `p2_prism.mesh`) so I was thinking of just building them in the unit test. For example we could add the following:
```
Mesh CylinderMesh(bool quadratic)
{
   double c[3];

   Mesh mesh(3, 15, 8);

   for (int i=0; i<3; i++)
   {
      c[0] = 0.0;  c[1] = 0.0;  c[2] = 2.74 * i;
      mesh.AddVertex(c);

      for (int j=0; j<4; j++)
      {
         c[0] = 2.74 * ((j + 1) % 2) * (1 - j);
         c[1] = 2.74 * (j % 2) * (2 - j);
         c[2] = 2.74 * i;
         mesh.AddVertex(c);
      }
   }

   for (int i=0; i<2; i++)
   {
      for (int j=0; j<4; j++)
      {
         /*
         mesh.AddWedge(5*i, 5*i+j+1, 5*i+(j+1)%4+1,
                       5*(i+1), 5*(i+1)+j+1, 5*(i+1)+(j+1)%4+1);
         */
         /*
         mesh.AddWedge(5*i+(j+1)%4+1, 5*i, 5*i+j+1,
                       5*(i+1)+(j+1)%4+1, 5*(i+1), 5*(i+1)+j+1);
         */
         mesh.AddWedge(5*i+j+1, 5*i+(j+1)%4+1, 5*i,
                       5*(i+1)+j+1, 5*(i+1)+(j+1)%4+1, 5*(i+1));
      }
   }

   mesh.FinalizeTopology();

   if (quadratic)
   {
      mesh.SetCurvature(2);

      auto quad_cyl = [](const Vector& x, Vector& d)
      {
         d.SetSize(3);
         d = x;
         double ax = fabs(x[0]);
         double ay = fabs(x[1]);
         double r = ax + ay;
         if (r < 1e-6) { return; }

         double sx = copysign(1.0, x[0]);
         double sy = copysign(1.0, x[1]);

         double t = ((2.0 - (1.0 + sx) * sy) * ax +
                     (2.0 - sy) * ay) * 0.5 * M_PI / r;
         d[0] = r * cos(t);
         d[1] = r * sin(t);

         return;
      };

      mesh.Transform(quad_cyl);
   }

   return mesh;
}
``` 
This code should reproduce meshes equivalent to `p1_prism.msh` or `p2_prism.msh`. The nodes and elements are ordered differently but that shouldn't matter.

Notice that I include three ways to add the wedge elements. Two of these lead to failing tests with differences in the L2 errors on the order of 1e-9. Hence the can of worms. Does this small difference expose an error in our H1 wedge basis functions or is the unit test overly optimistic? I'm honestly not sure.

hughcars:
> Notice that I include three ways to add the wedge elements. Two of these lead to failing tests with differences in the L2 errors on the order of 1e-9. Hence the can of worms. Does this small difference expose an error in our H1 wedge basis functions or is the unit test overly optimistic? I'm honestly not sure.

Ah, that is potentially troublesome. Good to catch whatever it is though. I assume that tightening up the linear solver tolerances did not improve things? An alternative might be hitting it with a direct, as that would at least convince that it is something within the basis or the mesh. If it's the wedge basis functions, I'm not sure I would know where to start in debugging those.

mlstowell:
> Ah, that is potentially troublesome. Good to catch whatever it is though. I assume that tightening up the linear solver tolerances did not improve things? An alternative might be hitting it with a direct, as that would at least convince that it is something within the basis or the mesh. If it's the wedge basis functions, I'm not sure I would know where to start in debugging those.

I did not try to tighten the tolerances because they are already at 1e-16. I did try adding comparable tests on a tetrahedral mesh and saw comparable errors. I'm tempted to try the same tests on a hexahedral mesh for peace of mind.

Incidentally, I just committed the modified unit test. Feel free to modify it in any way you see fit.

hughcars:
Thanks for the updated test: some tests good, more tests better. I was taking another look at the cavity resonator I was using to generate these meshes to think about it some more, and I actually now have even more dispiriting news:

Screenshot 2023-07-13 at 5 02 24 PMScreenshot 2023-07-13 at 5 02 04 PM

This is with one rank, cubic ND elements, and some NC tetrahedral adaptation. I can also reproduce it doing a 6 rank adapt. The first image is a crinkle cut, and the second has the NC mesh visible. Clearly there's something up, the nice thing is being a single rank it at least rules out my modification for ParNCMesh for p>=2 ND elements in https://github.com/mfem/mfem/pull/3611.

mlstowell (approved):
The unit test has been expanded to contain a Hex mesh.

mlstowell:
I don't expect @dylan-copeland to have any concerns with the more recent changes since they only add more mesh options to the unit test. Assuming the checks all pass I'll try to merge this into `next` this afternoon. Thanks @hughcars for your continuing improvements and your patience.

hughcars:
Thank you @mlstowell for the help with the PR, and for all the heavy lifting on ND elements in the first place.

mlstowell:

Issue #3727, opened 2023-06-13
`Submesh::From::Boundary` for nonconforming mesh


sebastiangrimberg:
Right now, both the `SubMesh` and `ParSubMesh` constructors don't allow construction from an NC parent mesh (checked in the constructor). Upon first glance, I think that if the use case is `From::Boundary` and the boundary is an external one (all faces have elements only on one side) that this should work just fine. I just want to check my understanding here, and also if there is an easy way to check this for the user in the constructor that the boundary attributes are true external boundaries.

Pinging @jandrej @mlstowell as the experts, thanks!

jandrej:
I think what you are stating here is correct. Iff the `SubMesh` is extracted from a boundary, the conforming and non-conforming cases are equivalent. 

I did not involve a constructor from `NCMesh` since this class is vastly different from the regular `Mesh` and unfortunately requires a separate implementation of `SubMesh`.

sebastiangrimberg:
OK, makes sense. I am just testing out something like the following and it seems to work fine:

```
diff --git a/mesh/submesh/psubmesh.cpp b/mesh/submesh/psubmesh.cpp
index b316d8b41..2a0e264a6 100644
--- a/mesh/submesh/psubmesh.cpp
+++ b/mesh/submesh/psubmesh.cpp
@@ -38,10 +38,8 @@ ParSubMesh ParSubMesh::CreateFromBoundary(const ParMesh &parent,
 ParSubMesh::ParSubMesh(const ParMesh &parent, SubMesh::From from,
                        Array<int> &attributes) : parent_(parent), from_(from), attributes_(attributes)
 {
-   if (Nonconforming())
-   {
-      MFEM_ABORT("SubMesh does not support non-conforming meshes");
-   }
+   MFEM_VERIFY(from == SubMesh::From::Boundary || !Nonconforming(),
+               "ParSubMesh does not support non-conforming meshes with From::Domain");

    MyComm = parent.GetComm();
    NRanks = parent.GetNRanks();
diff --git a/mesh/submesh/submesh.cpp b/mesh/submesh/submesh.cpp
index ac2058c57..2620bfb0f 100644
--- a/mesh/submesh/submesh.cpp
+++ b/mesh/submesh/submesh.cpp
@@ -31,10 +31,8 @@ SubMesh SubMesh::CreateFromBoundary(const Mesh &parent,
 SubMesh::SubMesh(const Mesh &parent, From from,
[...]

PR #3766, opened 2023-06-30
Allow `Submesh::From::Boundary` for nonconforming mesh

sebastiangrimberg:
If a mesh is nonconforming, we should still be able to construct a `SubMesh` using `SubMesh::From::Boundary` when the boundary is a true external boundary and thus all the faces in the submesh are actually conforming. This PR aims to enable this. There is a check during `SubMesh` construction that the criteria for correctness are actually met. In addition, it was tested in `tests/unit/mesh/test_submesh.cpp` by manually setting the new parameter `nonconforming` to `true`. This should work both serially and in parallel, but the parallel unit test doesn't have any true exterior boundaries so we only test on the serial case.

Resolves https://github.com/mfem/mfem/issues/3727.

### DIFF
--- a/mesh/submesh/psubmesh.cpp
+++ b/mesh/submesh/psubmesh.cpp
@@ -348,7 +348,7 @@ void ParSubMesh::FindSharedEdgesRanks(Array<int> &rhe)
    sedge_comm.Bcast<int>(rhe, 0);
 }
 
-void ParSubMesh::FindSharedFacesRanks(Array<int>& rht, Array<int> &rhq)
+void ParSubMesh::FindSharedFacesRanks(Array<int> &rht, Array<int> &rhq)
 {
    GroupCommunicator squad_comm(parent_.gtopo);
    parent_.GetSharedQuadCommunicator(squad_comm);
@@ -522,7 +522,7 @@ void ParSubMesh::AppendSharedEdgesGroups(ListOfIntegerSets &groups,
 }
 
 void ParSubMesh::AppendSharedFacesGroups(ListOfIntegerSets &groups,
-                                         Array<int>& rht, Array<int> &rhq)
+                                         Array<int> &rht, Array<int> &rhq)
 {
    IntegerSet quad_group;
 
@@ -607,8 +607,8 @@ void ParSubMesh::AppendSharedFacesGroups(ListOfIntegerSets &groups,
    }
 }
 
-void ParSubMesh::BuildVertexGroup(int ngroups, const Array<int>& rhvtx,
-                                  int& nsverts)
+void ParSubMesh::BuildVertexGroup(int ngroups, const Array<int> &rhvtx,
+                                  int &nsverts)
 {
    group_svert.MakeI(ngroups);
    for (int i = 0; i < rhvtx.Size(); i++)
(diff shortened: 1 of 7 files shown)


tzanio:

jandrej (approved):
Looks good to me.

mlstowell:
Unfortunately, I think the premise that the external boundary faces form a conforming mesh is incorrect. It is certainly true that there exist special cases where the external boundary will be conforming and we could support these. 

I think the source of confusion is the `NCFace` argument of `Mesh::GetFaceInfos()`. If this returns `-1` for every boundary face, indicating that these faces are conforming, that does not mean that the `SubMesh` created using this boundary will be a conforming mesh. What it means is that the boundary faces conform to the faces of the volume elements adjacent to this boundary.

Consider the mesh uploaded by @jandrej in issue #3727. Clearly, the boundary of this 3D non-conforming mesh will be a 2D non-conforming mesh. If we use the code in this PR to build a `SubMesh` from the boundary of the pictured mesh I believe that `SubMesh` will contain cuts along each of the nonconforming edges. Such a mesh would appear correct but probably not work as expected although projections and transfers between parent and sub-mesh would probably work fine. Adding a calculation of the sub-meshes Euler number in the unit test could confirm this suspicion.

sebastiangrimberg:
Thanks for catching my mistake/oversight @mlstowell. It won't be easy, but I still can imagine this PR being extended to support a `SubMesh` on the boundary of a nonconforming `Mesh` or `ParMesh`. As you said, the transfer operations should work fine since the `SubMesh` is single valued. The ugly work will go into constructing the `SubMesh` with an associated `NCMesh` (or `ParNCMesh`) data structure. However, since all this information should be available from the parent mesh I think it should in principle still be possible to make work. Correct me if I'm wrong!

I'll mark this as WIP for now.

hughcars:
I believe the trickiest part about constructing such a mesh, will be reverse engineering the root mesh structure. I believe it's a fairly low level assumption of the NCMesh structure, that the root mesh (initial elements) form a conformal mesh. In principle it can be done though, I believe by recursively coarsening any nc faces found and counting the number of times any given element needs to be refined to reach the initial state. This bares a lot of similarity to loading an NCMesh from file, it's quite plausible we could reuse that functionality for the construction of the relevant NCMesh member variable. At least for topologically simple boundary meshes.

jandrej:
I'd prefer if the code involves more `NCMesh` involvement that we attempt a dedicated `NCMesh` version of `SubMesh`.

mlstowell:
Yes, @sebastiangrimberg, I do think we have access to the information we need to make this work. I think @hughcars has the correct plan in mind. And, of course, I agree with @jandrej that a specialized `SubMesh` designed to support non-conforming meshes would be wise.

I'd just like to add one more complication if I could. I'm sure we all realize that we will need to support adapting either the parent or sub-mesh after the `SubMesh` object has been created. I think we should keep this in mind as we design these changes to the `SubMesh` class. We don't need to solve this problem immediately but it may be useful to envision how this type of mesh synchronization will work.

For example, It would be simple enough to use the existing `GridFunction` transfer scheme to exchange the elements tagged for refinement in either direction but I suspect we would need more information to update the transfer maps.

PR #3557, opened 2023-03-24
Parallel H1 HO <-> LOR Transfer Operator

ebchin:
Updates `L2ProjectionGridTransfer::L2ProjectionH1Space` to work on `ParFiniteElementSpace`s.  New miniapp (`miniapps/tools/lor-transferp.cpp`) demonstrates this capability.

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#3557](https://github.com/mfem/mfem/pull/3557) | @ebchin | @tzanio | @pazner + @tzanio | 3/24/23 | 8/12/23 | 8/22/23 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
      - [ ] Add the example code to the `ALL_EXE_SRCS` variable.
      - [ ] Make sure `THIS_TEST_OPTIONS` is set correctly for the new example.
[...]

PR #3815, opened 2023-08-07
Proposed changes to #3557

pazner:
Some proposed changes to PR #3557:

* Remove the class `ParL2ProjectionH1Space`. Instead the serial and parallel versions are in a single class, `L2ProjectionH1Space`. Only the constructor is different.
* Make sure to do RAP in serial when the prolongation matrices are not identity (e.g. AMR meshes)
* Simplify mass computation in the miniapp code
* Remove the optimization where the DOF and true-DOF vectors would alias each other in certain cases. This way is clearer and safer, but a bit less efficient. With a parallel build this optimization isn't going to matter.

### DIFF
--- a/fem/transfer.cpp
+++ b/fem/transfer.cpp
@@ -670,7 +670,7 @@ void L2ProjectionGridTransfer::L2ProjectionH1Space::Mult(
       Y.SetSubVector(vdofs_list, Y_dim);
    }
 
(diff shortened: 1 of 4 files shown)


ebchin (approved):
Thanks for the suggestions and improvements @pazner.  All of them look good to me!

PR #4028, opened 2023-12-13
Use Array<int> instead of Memory<int> in Table

pazner:
This PR replaces `Memory<int>` with `Array<int>` in the `Table` class. The advantage is that this delegates the all the ownership semantics to the `Array` class, so that `Table` doesn't need to deal with resource management. In particular, this means that it can have default destructor, move, and copy constructors that all do the right thing ([rule of zero](https://en.cppreference.com/w/cpp/language/rule_of_three#Rule_of_zero)).

I think this generally simplifies the code and makes things easier to understand. If we build more complicated classes out of building blocks like `Array<T>` that have proper copy and move semantics, then we should be able to inherit these behaviors for free.

If people like this approach, we could do similar things in other classes (`DenseMatrix`, for example).


pazner:
Marking this as ready for review… open to feedback and suggestions about this idea.

In the branch [densemat-array](https://github.com/mfem/mfem/tree/densemat-array) I also made the changes to `DenseMatrix` and `DenseTensor`. We can pull those in here if we want to review them together.

tzanio:

sebastiangrimberg (approved):
This is a great change in my opinion and makes the `Table` class much easier to read and use. Everything looks good to me.

I would strongly be in favor of making similar changes to `DenseMatrix` as proposed, and might also add that `Vector` could also be a high-value target as well unless I am missing something. These can certainly be left to follow-up PRs, but maybe adding an issue so that we can keep track of that proposal would make sense (unless you are close to opening a PR already, @pazner).

pazner:
> I would strongly be in favor of making similar changes to `DenseMatrix` as proposed, and might also add that `Vector` could also be a high-value target as well unless I am missing something. These can certainly be left to follow-up PRs, but maybe adding an issue so that we can keep track of that proposal would make sense (unless you are close to opening a PR already, @pazner).

Thanks, @sebastiangrimberg! I am totally fine with making those changes to this PR as well. I agree `Vector` is also a good candidate. (Really in some sense `Vector` should just inherit from `Array<double>`. A large part of the interface is duplicated between these two classes). The slight complication is that derived classes (`HypreVector`, `GridFunction`, etc.) may need some small changes, but I don't think this is a dealbreaker.

sebastiangrimberg:
> Thanks, @sebastiangrimberg! I am totally fine with making those changes to this PR as well. I agree `Vector` is also a good candidate. (Really in some sense `Vector` should just inherit from `Array<double>`. A large part of the interface is duplicated between these two classes). The slight complication is that derived classes (`HypreVector`, `GridFunction`, etc.) may need some small changes, but I don't think this is a dealbreaker.

I'm definitely supportive of making those changes in this PR to accelerate the process. Happy to include in this review.

pazner:
I merged the `DenseMatrix` and `DenseTensor` branch into this PR. `Vector` might make sense to do as a separate PR. Making sure everything works properly with e.g. `PetscParVector` will be more involved.

sebastiangrimberg:
Changes to `DenseMatrix` and `DenseTensor` also look good to me, and I'm OK with waiting on `Vector`. I agree that there will probably be room for simplifications with the derived vector classes and so attacking that in a separate PR makes sense.

pazner:
Just updated this and resolved conflicts with the single-precision support. Thoughts on this @artv3, @v-dobrev?

tzanio:
@pazner, can you please resolve conflicts?

pazner:
> @pazner, can you please resolve conflicts?

Done. Any chance to get a review on this?

v-dobrev:
I'll try to take a closer look soon.

tomstitt:
We're hitting a case where we segfault while creating a `DenseTensor` from another `DenseTensor` that was `NewDataAndSize`'d, something like:

```
DenseTensor t;
t.NewDataAndSize(mem, i, j, k);
DenseTensor t2(t);
```

`t.Mk` is created with `(NULL, i, j)` during `NewDataAndSize` and when we create `t2` we call `Array`'s copy constructor, which segfaults when we try to copy from `t.Mk`. This might be a general concern if `DenseTesnor::Mk` is ever setup? I don't think we actually want to ever copy the data from `Mk`

pazner:
Thanks for checking @tomstitt! Can you give it another try with the latest commit?

tomstitt:
> Thanks for checking @tomstitt! Can you give it another try with the latest commit?

Yes! Works, thanks!

linalg/densemat.cpp
helloworld922, line 1: I think `DenseMatrix`'s copy assign operator can be removed to get the default?

general/table.hpp
```diff
@@ -28,166 +28,161 @@ struct Connection
...
+
+   int *GetI() { return I.GetData(); }
+   int *GetJ() { return J.GetData(); }
+   const int *GetI() const { return I.GetData(); }
+   const int *GetJ() const { return J.GetData(); }
+
+   Memory<int> &GetIMemory() { return I.GetMemory(); }
+   Memory<int> &GetJMemory() { return J.GetMemory(); }
+   const Memory<int> &GetIMemory() const { return I.GetMemory(); }
+   const Memory<int> &GetJMemory() const { return J.GetMemory(); }
```
helloworld922, line 134: Do we also want a way to get the underlying Array?

linalg/densemat.hpp
```diff
@@ -1238,21 +1192,21 @@ class DenseTensor
...
-   real_t *Data() { return tdata; }
+   real_t *Data() { return tdata.GetData(); }
 
-   const real_t *Data() const { return tdata; }
+   const real_t *Data() const { return tdata.GetData(); }
 
-   Memory<real_t> &GetMemory() { return tdata; }
-   const Memory<real_t> &GetMemory() const { return tdata; }
+   Memory<real_t> &GetMemory() { return tdata.GetMemory(); }
+   const Memory<real_t> &GetMemory() const { return tdata.GetMemory(); }
```
helloworld922, line 1230: Similar, does it make sense to be able to get the Array?

general/array.hpp
```diff
@@ -113,9 +113,12 @@ class Array
    /// Destructor
    inline ~Array() { data.Delete(); }
 
-   /// Assignment operator: deep copy from 'src'.
+   /// Copy assignment operator: deep copy from 'src'.
    Array<T> &operator=(const Array<T> &src) { src.Copy(*this); return *this; }
```
helloworld922, line 114:
```suggestion
   /// Copy assignment operator: deep copy from 'src'.
   Array<T> &operator=(const Array<T> &src) { if(&src != this) {src.Copy(*this);} return *this; }
```

general/array.hpp
```diff
@@ -113,9 +113,12 @@ class Array
...
 
+   /// Move assignment operator
+   Array<T> &operator=(Array<T> &&src) { Swap(src, *this); return *this; }
```
helloworld922:
```suggestion
   /// Move assignment operator
   Array<T> &operator=(Array<T> &&src) { if(&src != this) { Swap(src, *this); } return *this; }
```

general/array.hpp
helloworld922, line 1: `Array2D` should also be able to have its copy operators removed to gain move semantics.

helloworld922 (changes requested):

general/array.hpp
pazner: Is this check needed? I think self `Copy` should be harmless (it looks like `Memory<T>::CopyFrom` and `MemoryManager::Copy_` guard against self-copy).

general/array.hpp
pazner: Done in dd68bbe9fd2ff660e399552937d452db0171dd25.

general/array.hpp
pazner: I made `Array`'s move constructor and assignment operator explicitly defaulted in d1efdf93be5c913ca7b5ebefa5a053cf8a1a1bf2.

Does that look OK to you? (`Memory<T>`'s move assignment guards against self-assignment).

general/table.hpp
pazner: We can definitely it add if it's needed — not sure if there's a compelling use-case.

v-dobrev:
@pazner, there are some CI failures after I merged `master` here.

v-dobrev:
The issue in `tmop_pa_tests_debug` is the following: on this line https://github.com/mfem/mfem/blob/bdb627a5ae0d2f355568c8443ccfe569ae0cbfd5/fem/tmop.cpp#L2835
the L.H.S. `DenseMatrix` has memory type `HOST_DEBUG` while the R.H.S. has memory type `HOST`. This is true in this branch and in `master`. The problem in this branch is that the `Array` copy assignment wants to preserve the memory type of the R.H.S., so it resets the L.H.S. not to be an alias and copies the data in a different place, not the intended alias memory.

We can change the logic in the `Array` copy assignment (in several ways) to fix this use case but I'm not sure that's a change we really want to make for all use cases of `Array` copy.

v-dobrev:
The logic used in the copy-assignment operators of `Array` and `Vector` is different. Maybe we should adopt the behavior from `Vector` in `Array` -- we probably don't need to duplicate exactly the memory type from the source to the destination. It could be sufficient to just focus on the "use device" flags as we do in class `Vector`.

What do you think?

general/array.hpp
```diff
@@ -103,14 +98,17 @@ class Array
...
 
    /// Destructor
    inline ~Array() { data.Delete(); }
 
-   /// Assignment operator: deep copy from 'src'.
+   /// Copy assignment operator: deep copy from 'src'.
    Array<T> &operator=(const Array<T> &src) { src.Copy(*this); return *this; }
 
+   /// Move assignment operator
+   Array<T> &operator=(Array<T> &&src) = default;
```
v-dobrev: This is incorrect -- the current content of the `Array` is leaked.

general/array.hpp
```diff
@@ -103,14 +98,17 @@ class Array
    explicit inline Array(std::initializer_list<CT> values);
 
    /// Move constructor ("steals" data from 'src')
-   inline Array(Array<T> &&src) : Array() { Swap(src, *this); }
+   Array(Array<T> &&src) = default;
```
v-dobrev: Does anyone know if the default implementation resets `src.size` to `0`? The requirement is that `src` is in a valid state after the assignment, however, empty `data` with non-zero `size` is not exactly valid. Invoking the destructor is ok but some other operations can lead to errors.

general/array.hpp
pazner: Good point - thanks. Does the following look OK to you?

```cpp
   /// Move constructor ("steals" data from 'src')
   Array(Array<T> &&src)
   {
      *this = std::move(src);
   }

   /// Move assignment operator
   Array<T> &operator=(Array<T> &&src)
   {
      std::swap(data, src.data);
      std::swap(size, src.size);
      src.DeleteAll(); // <-- optional, might be more "intuitive" behavior
      return *this;
   }
```

Also, should we clear the `src` array? (I lean towards yes, but it could also be correct to just swap).

general/array.hpp
pazner: No, I don't think `src.size` will be reset to zero - I think it will be unchanged. The implementation proposed here is probably better: https://github.com/mfem/mfem/pull/4028#discussion_r2264006726

pazner:
> The logic used in the copy-assignment operators of `Array` and `Vector` is different. Maybe we should adopt the behavior from `Vector` in `Array` -- we probably don't need to duplicate exactly the memory type from the source to the destination. It could be sufficient to just focus on the "use device" flags as we do in class `Vector`.
> 
> What do you think?

> The logic used in the copy-assignment operators of `Array` and `Vector` is different. Maybe we should adopt the behavior from `Vector` in `Array` -- we probably don't need to duplicate exactly the memory type from the source to the destination. It could be sufficient to just focus on the "use device" flags as we do in class `Vector`.
> 
> What do you think?

I think assignment for `Vector` and `Array` should have the semantics. Probably `Vector`'s version is less surprising?

(This has always been a bit of a weird thing with `operator=` and aliases. If `alias` is e.g. a subvector of `base`, then `alias = vec;` _may_ change `base` or it may not)

general/array.hpp
v-dobrev: For the move assignment, I think we should keep the `src.DeleteAll()` and we just need to add a check for self-assignment:
```c++
template <class T>
inline Array<T> &Array<T>::operator=(Array<T> &&src)
{
   if (this == &src) { return *this; }
   std::swap(data, src.data);
   std::swap(size, src.size);
   src.DeleteAll();
   return *this;
}
```

For the move construction: if we use the move assignment, we first need to make sure `*this` is initialized. Instead, a simpler implementation is this:
```c++
template <typename T>
inline Array<T>::Array(Array<T> &&src)
   : data(std::move(src.data)), size(src.size)
{
   src.size = 0;
}
```

general/array.hpp
pazner: Resolved in 5a7edc9548de0d9aaa82be63e92533cb19394534.

general/array.hpp
pazner: Done in 5a7edc9548de0d9aaa82be63e92533cb19394534.

pazner:
For the TMOP failures, I have some potential fixes (click to expand):

<details>
<summary>Add HostRead/HostReadWrite in DenseMatrix</summary>

```diff
diff --git a/linalg/densemat.cpp b/linalg/densemat.cpp
index 0b6996a504..d780170f55 100644
--- a/linalg/densemat.cpp
+++ b/linalg/densemat.cpp
@@ -72,6 +72,9 @@ DenseMatrix::DenseMatrix(const DenseMatrix &mat, char ch)
    {
       data.SetSize(capacity);
 
+      HostWrite();
+      mat.HostRead();
+
       for (int i = 0; i < height; i++)
       {
          for (int j = 0; j < width; j++)
@@ -145,11 +148,14 @@ void DenseMatrix::AbsMult(const Vector &x, Vector &y) const
    kernels::AbsMult(height, width, HostRead(), x.HostRead(), y.HostWrite());
 }
 
-real_t DenseMatrix::operator *(const DenseMatrix &m) const
+real_t DenseMatrix::operator*(const DenseMatrix &m) const
 {
    MFEM_ASSERT(Height() == m.Height() && Width() == m.Width(),
                "incompatible dimensions");
 
+   HostRead();
+   m.HostRead();
+
    const int hw = height * width;
    real_t a = 0.0;
    for (int i = 0; i < hw; i++)
@@ -206,8 +212,8 @@ void DenseMatrix::AddMult(const Vector &x, Vector &y, const real_t a) const
    MFEM_ASSERT(height == y.Size() && width == x.Size(),
                "incompatible dimensions");
 
-   const real_t *xp = x.GetData(), *d_col = data;
-   real_t *yp = y.GetData();
+   const real_t *xp = x.HostRead(), *d_col = HostRead();
+   real_t *yp = y.HostReadWrite();
    for (int col = 0; col < width; col++)
    {
       real_t x_col = xp[col];
@@ -230,7 +236,9 @@ void DenseMatrix::AddMultTranspose(const Vector &x, Vector &y,
    MFEM_ASSERT(height == x.Size() && width == y.Size(),
                "incompatible dimensions");
 
-   const real_t *d_col = data;
+   x.HostRead();
+   y.HostReadWrite();
+   const real_t *d_col = HostRead();
    for (int col = 0; col < width; col++)
    {
       real_t y_col = 0.0;
@@ -248,11 +256,8 @@ void DenseMatrix::AddMult_a(real_t a, const Vector &x, Vector &y) const
    MFEM_ASSERT(height == y.Size() && width == x.Size(),
                "incompatible dimensions");
 
-   HostRead();
-   x.HostRead();
-   y.HostReadWrite();
-   const real_t *xp = x.GetData(), *d_col = data;
-   real_t *yp = y.GetData();
+   const real_t *xp = x.HostRead(), *d_col = HostRead();
+   real_t *yp = y.HostReadWrite();
    for (int col = 0; col < width; col++)
    {
       const real_t x_col = a*xp[col];
@@ -270,7 +275,9 @@ void DenseMatrix::AddMultTranspose_a(real_t a, const Vector &x,
    MFEM_ASSERT(height == x.Size() && width == y.Size(),
                "incompatible dimensions");
 
-   const real_t *d_col = data;
+   x.HostRead();
+   y.HostReadWrite();
+   const real_t *d_col = HostRead();
    for (int col = 0; col < width; col++)
    {
       real_t y_col = 0.0;
@@ -287,6 +294,8 @@ real_t DenseMatrix::InnerProduct(const real_t *x, const real_t *y) const
 {
    real_t prod = 0.0;
 
+   HostRead();
+
    for (int i = 0; i < height; i++)
    {
       real_t Axi = 0.0;
@@ -506,6 +515,8 @@ real_t DenseMatrix::Det() const
                << "  Height() = " << Height()
                << ", Width() = " << Width());
 
+   HostRead();
+
    switch (Height())
    {
       case 1:

```

</details>

<details>
<summary>Explicitly turn off the device flag in TMOP</summary>

```diff
diff --git a/fem/tmop.cpp b/fem/tmop.cpp
index ac7b7c86e6..2afa570622 100644
--- a/fem/tmop.cpp
+++ b/fem/tmop.cpp
@@ -2301,6 +2301,7 @@ void TargetConstructor::ComputeElementTargets(int e_id, const FiniteElement &fe,
                               nodes->FESpace()->GetFE(e_id) : NULL;
    const DenseMatrix &Wideal =
       Geometries.GetGeomToPerfGeomJac(fe.GetGeomType());
+   Wideal.GetMemory().UseDevice(false);
    MFEM_ASSERT(Wideal.Height() == Jtr.SizeI(), "");
    MFEM_ASSERT(Wideal.Width() == Jtr.SizeJ(), "");
 
@@ -2798,6 +2799,7 @@ void DiscreteAdaptTC::ComputeElementTargets(int e_id, const FiniteElement &fe,
       {
          const DenseMatrix &Wideal =
             Geometries.GetGeomToPerfGeomJac(fe.GetGeomType());
+         Wideal.GetMemory().UseDevice(false);
          const int ndofs = tspec_fesv->GetFE(e_id)->GetDof(),
                    ntspec_dofs = ndofs*ncomp;
 
@@ -3014,6 +3016,7 @@ void DiscreteAdaptTC::ComputeElementTargetsGradient(const IntegrationRule &ir,
       {
          const DenseMatrix &Wideal =
             Geometries.GetGeomToPerfGeomJac(fe->GetGeomType());
+         Wideal.GetMemory().UseDevice(false);
          const int dim = Wideal.Height(),
                    ndofs = fe->GetDof(),
                    ntspec_dofs = ndofs*ncomp;

```

</details>


<details>
<summary>Change DenseMatrix copy assignment to always copy to host</summary>

(Should probably also change copy constructor)

```diff
diff --git a/linalg/densemat.cpp b/linalg/densemat.cpp
index 0b6996a504..8067d96be4 100644
--- a/linalg/densemat.cpp
+++ b/linalg/densemat.cpp
@@ -631,6 +631,16 @@ DenseMatrix &DenseMatrix::operator=(const real_t *d)
    return *this;
 }
 
+DenseMatrix &DenseMatrix::operator=(const DenseMatrix &m)
+{
+   height = m.Height();
+   width = m.Width();
+   data.SetSize(m.TotalSize());
+   HostWrite();
+   data.GetMemory().CopyFrom(m.GetMemory(), m.TotalSize());
+   return *this;
+}
+
 DenseMatrix &DenseMatrix::operator+=(const real_t *m)
 {
    kernels::Add(Height(), Width(), m, (real_t*)data);
diff --git a/linalg/densemat.hpp b/linalg/densemat.hpp
index 347382f195..747f0fbca7 100644
--- a/linalg/densemat.hpp
+++ b/linalg/densemat.hpp
@@ -62,7 +62,7 @@ public:
    DenseMatrix(DenseMatrix &&) = default;
 
    /// Copy assignment (deep copy).
-   DenseMatrix &operator=(const DenseMatrix &) = default;
+   DenseMatrix &operator=(const DenseMatrix &);
 
    /// Move assignment.
    DenseMatrix &operator=(DenseMatrix &&) = default;
```

</details>

<details>
<summary>In TMOP, only call Read on a copy of a global matrix, rather than a reference</summary>

```diff
diff --git a/fem/tmop/tmop_pa_da3.cpp b/fem/tmop/tmop_pa_da3.cpp
index 65cbf56364..c9770b9bda 100644
--- a/fem/tmop/tmop_pa_da3.cpp
+++ b/fem/tmop/tmop_pa_da3.cpp
@@ -255,7 +255,7 @@ void DiscreteAdaptTC::ComputeAllElementTargets(const FiniteElementSpace &pa_fes,
                "mixed meshes are not supported");
    MFEM_VERIFY(!fes->IsVariableOrder(), "variable orders are not supported");
    const FiniteElement &fe = *fes->GetTypicalFE();
-   const DenseMatrix &W = Geometries.GetGeomToPerfGeomJac(fe.GetGeomType());
+   const DenseMatrix W = Geometries.GetGeomToPerfGeomJac(fe.GetGeomType());
    const DofToQuad::Mode mode = DofToQuad::TENSOR;
    const DofToQuad &maps = fe.GetDofToQuad(ir, mode);
    const Array<real_t> &B = maps.B;
diff --git a/fem/tmop/tmop_pa_tc2.cpp b/fem/tmop/tmop_pa_tc2.cpp
index ae012f102c..d7752f4c57 100644
--- a/fem/tmop/tmop_pa_tc2.cpp
+++ b/fem/tmop/tmop_pa_tc2.cpp
@@ -124,7 +124,7 @@ TargetConstructor::ComputeAllElementTargets<2>(const FiniteElementSpace &fes,
    MFEM_VERIFY(!fes.IsVariableOrder(), "variable orders are not supported");
    const FiniteElement &fe = *fes.GetTypicalFE();
    MFEM_VERIFY(fe.GetGeomType() == Geometry::SQUARE, "");
-   const DenseMatrix &W = Geometries.GetGeomToPerfGeomJac(Geometry::SQUARE);
+   const DenseMatrix W = Geometries.GetGeomToPerfGeomJac(Geometry::SQUARE);
    const DofToQuad::Mode mode = DofToQuad::TENSOR;
    const DofToQuad &maps = fe.GetDofToQuad(ir, mode);
    const int D1D = maps.ndof;
diff --git a/fem/tmop/tmop_pa_tc3.cpp b/fem/tmop/tmop_pa_tc3.cpp
index 52f7676bfc..cc8c34ea3c 100644
--- a/fem/tmop/tmop_pa_tc3.cpp
+++ b/fem/tmop/tmop_pa_tc3.cpp
@@ -128,7 +128,7 @@ TargetConstructor::ComputeAllElementTargets<3>(const FiniteElementSpace &fes,
    MFEM_VERIFY(!fes.IsVariableOrder(), "variable orders are not supported");
    const FiniteElement &fe = *fes.GetTypicalFE();
    MFEM_VERIFY(fe.GetGeomType() == Geometry::CUBE, "");
-   const DenseMatrix &W = Geometries.GetGeomToPerfGeomJac(Geometry::CUBE);
+   const DenseMatrix W = Geometries.GetGeomToPerfGeomJac(Geometry::CUBE);
    const DofToQuad::Mode mode = DofToQuad::TENSOR;
    const DofToQuad &maps = fe.GetDofToQuad(ir, mode);
    const Array<real_t> &B = maps.B;
```

</details>

general/array.hpp
```diff
@@ -103,13 +98,35 @@ class Array
...
+   /// Copy assignment operator: deep copy from 'src'.
+   Array<T> &operator=(const Array<T> &src)
+   {
+      SetSize(src.Size());
+      data.UseDevice(UseDevice() || src.UseDevice());
+      // keep 'data' where it is, unless 'use_dev' is true
+      if (UseDevice()) { Write(); }
+      data.CopyFrom(src.data, src.Size());
+      return *this;
+   }
```
v-dobrev: I think it's better to modify the `Copy` method -- all versions of copy should behave the same.

pazner: Done in 7fea47019297325dfadf21b6b5aee4305dd726d8.

v-dobrev:
@pazner, @helloworld922, as we discussed, I made PR #4978 (toward this PR) with the additions and changes I'm proposing to include in this PR.

pazner:
Thanks @v-dobrev, I'll take a look at #4978.

I added some more options to my previous comment. (cc @helloworld922).

v-dobrev:
> I added some more options to my previous comment. (cc @helloworld922).

Yet another option is to keep the `DenseMatrix` copy-assigment (and copy-ctor) to be `= default` and add a new method that does a host copy, e.g. `DenseMatrix::HostCopyFrom(const DenseMatrix &)`. This can then be used instead of copy-assignment in https://github.com/mfem/mfem/blob/bdb627a5ae0d2f355568c8443ccfe569ae0cbfd5/fem/tmop.cpp#L2835

This option, however, will be a bit trickier for users if they run into similar issue (`DenseMatrix` copy that happens on device) -- they'll have to figure out that the copy-assignment is done on device and then figure out that it can be done on host with `HostCopyFrom`.

On the other hand, there may be contexts where one wants to perform the copy on device, so I'm not sure that there's a universally good solution.

pazner:
Of the proposed fixes above, this is the one that makes the most sense to me. It changes the TMOP code to avoid setting the device flag on global data. It's similar to this other case, where calling `Read` on global data structures can cause errors in other parts of the code. Maybe it's just better to avoid calling `Read` on global objects when possible.

https://github.com/mfem/mfem/blob/f0cb31088ce75a56d7fd868de16c5a584f83e7db/fem/hybridization_ext.cpp#L52-L57

```diff
diff --git a/fem/tmop/tmop_pa_da3.cpp b/fem/tmop/tmop_pa_da3.cpp
index 65cbf56364..c9770b9bda 100644
--- a/fem/tmop/tmop_pa_da3.cpp
+++ b/fem/tmop/tmop_pa_da3.cpp
@@ -255,7 +255,7 @@ void DiscreteAdaptTC::ComputeAllElementTargets(const FiniteElementSpace &pa_fes,
                "mixed meshes are not supported");
    MFEM_VERIFY(!fes->IsVariableOrder(), "variable orders are not supported");
    const FiniteElement &fe = *fes->GetTypicalFE();
-   const DenseMatrix &W = Geometries.GetGeomToPerfGeomJac(fe.GetGeomType());
+   const DenseMatrix W = Geometries.GetGeomToPerfGeomJac(fe.GetGeomType());
    const DofToQuad::Mode mode = DofToQuad::TENSOR;
    const DofToQuad &maps = fe.GetDofToQuad(ir, mode);
    const Array<real_t> &B = maps.B;
diff --git a/fem/tmop/tmop_pa_tc2.cpp b/fem/tmop/tmop_pa_tc2.cpp
index ae012f102c..d7752f4c57 100644
--- a/fem/tmop/tmop_pa_tc2.cpp
+++ b/fem/tmop/tmop_pa_tc2.cpp
@@ -124,7 +124,7 @@ TargetConstructor::ComputeAllElementTargets<2>(const FiniteElementSpace &fes,
    MFEM_VERIFY(!fes.IsVariableOrder(), "variable orders are not supported");
    const FiniteElement &fe = *fes.GetTypicalFE();
    MFEM_VERIFY(fe.GetGeomType() == Geometry::SQUARE, "");
-   const DenseMatrix &W = Geometries.GetGeomToPerfGeomJac(Geometry::SQUARE);
+   const DenseMatrix W = Geometries.GetGeomToPerfGeomJac(Geometry::SQUARE);
    const DofToQuad::Mode mode = DofToQuad::TENSOR;
    const DofToQuad &maps = fe.GetDofToQuad(ir, mode);
    const int D1D = maps.ndof;
diff --git a/fem/tmop/tmop_pa_tc3.cpp b/fem/tmop/tmop_pa_tc3.cpp
index 52f7676bfc..cc8c34ea3c 100644
--- a/fem/tmop/tmop_pa_tc3.cpp
+++ b/fem/tmop/tmop_pa_tc3.cpp
@@ -128,7 +128,7 @@ TargetConstructor::ComputeAllElementTargets<3>(const FiniteElementSpace &fes,
    MFEM_VERIFY(!fes.IsVariableOrder(), "variable orders are not supported");
    const FiniteElement &fe = *fes.GetTypicalFE();
    MFEM_VERIFY(fe.GetGeomType() == Geometry::CUBE, "");
-   const DenseMatrix &W = Geometries.GetGeomToPerfGeomJac(Geometry::CUBE);
+   const DenseMatrix W = Geometries.GetGeomToPerfGeomJac(Geometry::CUBE);
    const DofToQuad::Mode mode = DofToQuad::TENSOR;
    const DofToQuad &maps = fe.GetDofToQuad(ir, mode);
    const Array<real_t> &B = maps.B;
```

helloworld922:
Does that make a deepcopy of the densematrix each time? That seems like a relatively expensive thing to do?

pazner:
> Does that make a deepcopy of the densematrix each time? That seems like a relatively expensive thing to do?

Yes, but the matrices are at most 3x3, so the cost should be negligible.

helloworld922:
It's not so much the size, but I think it's always a malloc/free for the copy constructor? Plus another device malloc/free when read is called.

helloworld922:
I think adding `HostRead` to `DenseMatrix` makes the most sense, it makes writing incorrect code harder, and the cost when no transfer is required is minimal; when I benchmarked it, it was about 3 clock cycles, which should be on the order of 1ns for all relatively modern platforms.

Changing just the copy-assign to always copy to host also seems bad since it guarantees you can never do device to device dense matrix copy.

pazner:
> I think adding `HostRead` to `DenseMatrix` makes the most sense, it makes writing incorrect code harder, and the cost when no transfer is required is minimal; when I benchmarked it, it was about 3 clock cycles, which should be on the order of 1ns for all relatively modern platforms.
> 
> Changing just the copy-assign to always copy to host also seems bad since it guarantees you can never do device to device dense matrix copy.

I'm good with this. I'll make the change here and if needed we can discuss further.

helloworld922:
Another option if we want to keep changes to `DenseMatrix` minimal is to have `TargetConstructor` store an internal copy of `W`, either as static members or object members. This still leaves potentially surprising behavior to users if they try to use a `DenseMatrix` on host and device, but also avoids introducing any unexpected new surprising behavior (are there any?) for anyone expecting this to be a host-only class.

Something like this. I'm not entirely sure why the tmop debug test still fails, there might be somewhere else that a DenseMatrix/other buffer is not located somewhere expected?

```diff
diff --git a/fem/tmop.hpp b/fem/tmop.hpp
index ae55788090..c9c8613153 100644
--- a/fem/tmop.hpp
+++ b/fem/tmop.hpp
@@ -1520,6 +1520,13 @@ protected:
    const TargetType target_type;
    bool uses_phys_coords; // see UsesPhysicalCoordinates()
 
+  
+   /// for storing a device-usable dense matrix
+   mutable DenseMatrix W;
+   mutable real_t detW;
+   mutable Geometry::Type current_W_type = Geometry::INVALID;
+
+
 #ifdef MFEM_USE_MPI
    MPI_Comm comm;
 #endif
diff --git a/fem/tmop/tmop_pa_da3.cpp b/fem/tmop/tmop_pa_da3.cpp
index 65cbf56364..24608d54d8 100644
--- a/fem/tmop/tmop_pa_da3.cpp
+++ b/fem/tmop/tmop_pa_da3.cpp
@@ -255,7 +255,13 @@ void DiscreteAdaptTC::ComputeAllElementTargets(const FiniteElementSpace &pa_fes,
                "mixed meshes are not supported");
    MFEM_VERIFY(!fes->IsVariableOrder(), "variable orders are not supported");
    const FiniteElement &fe = *fes->GetTypicalFE();
-   const DenseMatrix &W = Geometries.GetGeomToPerfGeomJac(fe.GetGeomType());
+   if (current_W_type != fe.GetGeomType())
+   {
+      current_W_type = fe.GetGeomType();
+      W = Geometries.GetGeomToPerfGeomJac(current_W_type);
+      // is this ever used?
+      detW = Geometries.GetGeomToPerfGeomJac(current_W_type).Det();
+   }
    const DofToQuad::Mode mode = DofToQuad::TENSOR;
    const DofToQuad &maps = fe.GetDofToQuad(ir, mode);
    const Array<real_t> &B = maps.B;
diff --git a/fem/tmop/tmop_pa_tc2.cpp b/fem/tmop/tmop_pa_tc2.cpp
index ae012f102c..23356c3be6 100644
--- a/fem/tmop/tmop_pa_tc2.cpp
+++ b/fem/tmop/tmop_pa_tc2.cpp
@@ -55,6 +55,7 @@ MFEM_REGISTER_TMOP_KERNELS(bool, TC_IDEAL_SHAPE_GIVEN_SIZE_2D_KERNEL,
                            const Array<real_t> &b_,
                            const Array<real_t> &g_,
                            const DenseMatrix &w_,
+                           const real_t detW,
                            const Vector &x_,
                            DenseTensor &j_,
                            const int d1d,
@@ -63,7 +64,6 @@ MFEM_REGISTER_TMOP_KERNELS(bool, TC_IDEAL_SHAPE_GIVEN_SIZE_2D_KERNEL,
    constexpr int DIM = 2;
    constexpr int NBZ = 1;
 
-   const real_t detW = w_.Det();
    const int D1D = T_D1D ? T_D1D : d1d;
    const int Q1D = T_Q1D ? T_Q1D : q1d;
 
@@ -124,7 +124,13 @@ TargetConstructor::ComputeAllElementTargets<2>(const FiniteElementSpace &fes,
    MFEM_VERIFY(!fes.IsVariableOrder(), "variable orders are not supported");
    const FiniteElement &fe = *fes.GetTypicalFE();
    MFEM_VERIFY(fe.GetGeomType() == Geometry::SQUARE, "");
-   const DenseMatrix &W = Geometries.GetGeomToPerfGeomJac(Geometry::SQUARE);
+   if (current_W_type != Geometry::SQUARE)
+   {
+      current_W_type = Geometry::SQUARE;
+      W = Geometries.GetGeomToPerfGeomJac(current_W_type);
+      auto detW = Geometries.GetGeomToPerfGeomJac(current_W_type).Det();
+   }
+   //
    const DofToQuad::Mode mode = DofToQuad::TENSOR;
    const DofToQuad &maps = fe.GetDofToQuad(ir, mode);
    const int D1D = maps.ndof;
@@ -152,7 +158,7 @@ TargetConstructor::ComputeAllElementTargets<2>(const FiniteElementSpace &fes,
          R->Mult(*nodes, X);
          MFEM_ASSERT(nodes->FESpace()->GetVDim() == 2, "");
          MFEM_LAUNCH_TMOP_KERNEL(TC_IDEAL_SHAPE_GIVEN_SIZE_2D_KERNEL,
-                                 id,NE,B,G,W,X,Jtr);
+                                 id,NE,B,G,W,detW,X,Jtr);
       }
       case GIVEN_SHAPE_AND_SIZE: return false;
       default: return false;
diff --git a/fem/tmop/tmop_pa_tc3.cpp b/fem/tmop/tmop_pa_tc3.cpp
index 52f7676bfc..55b34dc695 100644
--- a/fem/tmop/tmop_pa_tc3.cpp
+++ b/fem/tmop/tmop_pa_tc3.cpp
@@ -56,6 +56,7 @@ MFEM_REGISTER_TMOP_KERNELS(bool, TC_IDEAL_SHAPE_GIVEN_SIZE_3D_KERNEL,
                            const Array<real_t> &b_,
                            const Array<real_t> &g_,
                            const DenseMatrix &w_,
+                           const real_t detW,
                            const Vector &x_,
                            DenseTensor &j_,
                            const int d1d,
@@ -63,7 +64,6 @@ MFEM_REGISTER_TMOP_KERNELS(bool, TC_IDEAL_SHAPE_GIVEN_SIZE_3D_KERNEL,
 {
    constexpr int DIM = 3;
 
-   const real_t detW = w_.Det();
    const int D1D = T_D1D ? T_D1D : d1d;
    const int Q1D = T_Q1D ? T_Q1D : q1d;
 
@@ -128,7 +128,12 @@ TargetConstructor::ComputeAllElementTargets<3>(const FiniteElementSpace &fes,
    MFEM_VERIFY(!fes.IsVariableOrder(), "variable orders are not supported");
    const FiniteElement &fe = *fes.GetTypicalFE();
    MFEM_VERIFY(fe.GetGeomType() == Geometry::CUBE, "");
-   const DenseMatrix &W = Geometries.GetGeomToPerfGeomJac(Geometry::CUBE);
+   if (current_W_type != Geometry::CUBE)
+   {
+      current_W_type = Geometry::CUBE;
+      W = Geometries.GetGeomToPerfGeomJac(current_W_type);
+      detW = Geometries.GetGeomToPerfGeomJac(current_W_type).Det();
+   }
    const DofToQuad::Mode mode = DofToQuad::TENSOR;
    const DofToQuad &maps = fe.GetDofToQuad(ir, mode);
    const Array<real_t> &B = maps.B;
@@ -155,7 +160,7 @@ TargetConstructor::ComputeAllElementTargets<3>(const FiniteElementSpace &fes,
          R->Mult(*nodes, X);
          MFEM_ASSERT(nodes->FESpace()->GetVDim() == 3, "");
          MFEM_LAUNCH_TMOP_KERNEL(TC_IDEAL_SHAPE_GIVEN_SIZE_3D_KERNEL,
-                                 id,NE,B,G,W,X,Jtr);
+                                 id,NE,B,G,W,detW,X,Jtr);
       }
       case GIVEN_SHAPE_AND_SIZE: return false;
       default: return false;
```

pazner:
> Another option if we want to keep changes to `DenseMatrix` minimal is to have `TargetConstructor` store an internal copy of `W`, either as static members or object members. This still leaves potentially surprising behavior to users if they try to use a `DenseMatrix` on host and device, but also avoids introducing any unexpected new surprising behavior (are there any?) for anyone expecting this to be a host-only class.

OK, this is what I implemented in 5c04ba55f6b7323c80d2c9ef4d1fa61335edb788. (Rather than cache the determinant, only the matrix is cached, and `HostRead` is called before `Det` - this shouldn't cause superfluous copies, since the device `Read` will leave the host data valid). It looks like all the tests are passing now.

v-dobrev:
My concern is the following: in this PR we are changing some of the behavior of class `DenseMatrix` ans we are changing MFEM to work with this new behavior. What happens in user codes that get broken by the new `DenseMatrix` behavior? Everyone that runs into this issue will have to spend time figuring out what happened and figuring out a solution. I'd like to avoid this type of issues whenever possible, and in this case it seems possible.

The main change of behavior is that the copy-ctor and the copy-assignment now can copy the data on device. If the original behavior is recovered, the changes in TMOP will not be necessary. Hopefully, this will also apply to other existing user code.

pazner:
> My concern is the following: in this PR we are changing some of the behavior of class `DenseMatrix` ans we are changing MFEM to work with this new behavior. What happens in user codes that get broken by the new `DenseMatrix` behavior? Everyone that runs into this issue will have to spend time figuring out what happened and figuring out a solution. I'd like to avoid this type of issues whenever possible, and in this case it seems possible.
> 
> The main change of behavior is that the copy-ctor and the copy-assignment now can copy the data on device. If the original behavior is recovered, the changes in TMOP will not be necessary. Hopefully, this will also apply to other existing user code.

I think this may be a case of [Hyrum's Law](https://www.hyrumslaw.com). Is this behavior of DenseMatrix documented or communicated somewhere?

If an object has device support (providing `Read`, `Write`, `ReadWrite` operations, backed by `Memory<T>`, like `DenseMatrix`), then it makes sense to support copying on device. This is how basically all other MFEM objects work (`Array`, `Vector`, `SparseMatrix`, `Table`, etc.).

This particular failure is a bit complicated, because it has to do both with DenseMatrix copy operations and how slicing a DenseTensor works (which I think does not register alias memory), but one primary cause is that calling `Read`, which is a const member function, actually changes the state and behavior of the object by setting the device flag. At least to me, this was surprising when I first learned about it.

---

All that being said, if we are OK precluding device-to-device copies of DenseMatrix, then we can definitely change the copy operations. I think it would be a bit unfortunate, since there are legitimate use-cases, but it's not the end of the world.

v-dobrev:
@helloworld922, what are your thoughts on https://github.com/mfem/mfem/pull/4028#issuecomment-3232239220?

helloworld922:
I think that since there are device methods and uses for DenseMatrix already, it makes sense to allow device copy. Maybe `operator=` should try to preserve the locations where the LHS is valid. Since copy is a write operation on the destination the user at least in principle should be calling `Write` or `HostWrite` to specify where in the LHS is being written to, and by default it should copy to host since I think Memory defaults to being valid on the Host. It shouldn't matter where the data on the RHS is located.

v-dobrev (approved):

tzanio:

v-dobrev:

tzanio:
Should this change be mentioned somehow in `CHANGELOG`?

general/table.hpp
```diff
@@ -28,166 +28,161 @@ struct Connection
...
 
 public:
    /// Creates an empty table
    Table() { size = -1; }
 
-   /// Copy constructor
-   Table(const Table &);
-
-   /** Merge constructors
-       This is used to combine two or three tables into one table.*/
```
v-dobrev, line 62: Why was this comment deleted?

pazner: It got moved to the three-table constructor rather than adding a new comment by mistake (previously it was one comment for two constructors). The missing one should be added back.

v-dobrev: Do you want to add it back in #5057?

pazner: Done in 9cc9765e158b6f85d1cac517da953c0b439fd9a4.

tzanio: thanks @pazner !

PR #3806, opened 2023-08-02
Miniapp for block preconditioned elasticity with partial and full assembly on GPUs [full-assembly-elast-dev]

victor-decaria-nnl:
This PR aims to provide an example of how to extend some functionality of end-to-end GPU acceleration in the style of [this paper](https://arxiv.org/abs/2210.12253) to vector valued PDEs.

The miniapp is installed under `miniapps/solvers/lor_elast`. The linear elastic beam problem is used for the miniapp. Rather than assembling the whole vector valued low-order refined matrix, only the diagonal component blocks are assembled. This reduces storage and allows reuse of the existing full assembly code for scalar systems. The approach is extensible to other PDEs.

Summary of contribution:
- Miniapp which demonstrates a block diagonal preconditioning strategy for elasticity where the LOR preconditioner is constructed on GPUs. The conditioning is h-independent.

To make the miniapp possible, the following changes to mfem were made:
- For `ElasticityIntegrator`, added overrides of `AssemblePA`, `AssembleEA`, `AssembleDiagonalPA`, `AddMultPA` and `AddMultTransposePA` and corresponding GPU kernels.
- `BilinearFormIntegrator* ElasticityIntegrator::ComponentIntegrator(const int I, const int J)` - Returns a new `ElasticityIntegrator` which only acts in the `(I,J)`th vector component. The component integrators created from this function shares resources (quadrature data) with the original integrator. This is a possible blueprint for an interface for other vector integrators if the solver in the miniapp is generalized.
- `const FiniteElementSpace* GetFESpace() const` - Integrator holds a pointer to the fespace after partial assembly and this returns that. It is a scalar fespace if integrator was constructed with `ComponentIntegrator`. This could also generalize to other integrators. Is there already a good way to get a scalar component of a vector fespace that could be used here?
- Optionally disable use of tensor product basis in `BilinearFormExtension`. This seemed necessary since the new kernels use `DofToQuad::Mode::FULL` exclusively.

The miniapp also includes an option to test the individual components of partial assembly for elasticity, and an operator `BlockFESpaceOperator` to facilitate this. `BlockFESpaceOperator` is meant to be general for other block systems with different spaces, such as Stokes, and can be used with `FormLinearSystem`. To make managing the lifetime of offsets easier, a change is suggested to allow `BlockOperator` to own its offsets rather than storing references to them.

[...]

PR #4043, opened 2023-12-20
Proposed changes for PR #3806

pazner:
Breaks out the individual component integrators in `ElasticityIntegrator` to a separate class `ElasticityComponentIntegrator`. Simplifies some of the surrounding logic and miniapp example code.

If this looks OK to you @victor-decaria-nnl, please feel free to merge this PR into your branch.

### DIFF
--- a/fem/bilininteg.cpp
+++ b/fem/bilininteg.cpp
@@ -3016,6 +3016,12 @@ void VectorDiffusionIntegrator::AssembleElementVector(
    }
 }
 
+ElasticityComponentIntegrator::ElasticityComponentIntegrator(
+   ElasticityIntegrator &parent_, int i_, int j_)
+   : parent(parent_),
+     i_block(i_),
+     j_block(j_)
+{ }
 
 void ElasticityIntegrator::AssembleElementMatrix(
    const FiniteElement &el, ElementTransformation &Trans, DenseMatrix &elmat)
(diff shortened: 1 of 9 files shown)


victor-decaria-nnl:
Thanks for the suggestion @pazner. I see this nicely simplifies the lexicographic map creation.

I'm not sure that the component integrator interface became simpler for the user:
- The user now has to ensure the parent outlives the component integrators.
- Without a function like `GetFESpace`, the user is responsible for constructing the appropriate scalar version of the vector fespace that matches the partially constructed data.

But I am ok if this is how you would like the interface, it's still easy to use. Maybe we could revisit this in the future if we do this again with another integrator? I was thinking of moving the interface to `BilinearForm`, but it's perhaps hard to justify with this being the only component integrator.

pazner:
Thanks for taking a look, @victor-decaria-nnl. You make good and valid points. Let me make some general points about the design, and then respond to the specific items you brought up.

I think the most important thing is that each integrator class should perform a well-defined mathematical operation. So, `ElasticityIntegrator` should represent the same integrand, independent of context. This necessitates the component integrators to be separate classes. The only reason they are connected by the `parent` in this implementation is to share coefficient data for efficiency reasons (i.e. it's just an optimization).

To give a couple examples why having separate classes is important: in the previous implementation, `ElasticityIntegrator::AssembleEA` only works for the component integrator, whereas `ElasticityIntegrator::AssembleDiagonalPA` only works for the monolithic integrator, yet this are both overridden in the same class.

A second point is regarding the scalar spaces. The way integrators are defined in MFEM, they are supposed to be independent of the spaces (the mathematical object that they model is an integrand, only when they are part of a bilinear form defined on a space does that information become relevant). Now, in PA this becomes a little bit mixed up, because some PA data (like the space, maps, and geom objects) are cached within the integrator. This should just be an implementation detail, and shouldn't be user-facing. To illustrate, in the previous implementation, the following code appears to be well-formed (and semantically makes perfect sense), yet it segfaults:

```cpp
ConstantCoefficient lambda(1.0), mu(1.0);
ElasticityIntegrator elast(lambda, mu);
auto component_integ = elast.ComponentIntegrator(0,0);
```

The reason being that integrators are created without knowledge of the space, yet getting the component integrator needs to create the scalar space. On the other hand, the equivalent code with the design that I'm proposing works as expected:

```cpp
ConstantCoefficient lambda(1.0), mu(1.0);
ElasticityIntegrator elast(lambda, mu);
ElasticityComponentIntegrator comp(elast, 0, 0);
```

The structure of a typical MFEM program is that the creation of the spaces is the user's responsibility, and happens at a level "above" the integrator, and this design works the same way.

Now, we may want to introduce a convenience constructor for finite element spaces that lets us "clone" a space with a different `vdim`. (In fact, I think we maybe should think about a design that allows having the same type of space with different vdims without as much duplication (almost all of the internal tables will be the same, just the vdim integer will be different)).

> I'm not sure that the component integrator interface became simpler for the user:
> 
> * The user now has to ensure the parent outlives the component integrators.

This can be alleviated with `shared_ptr` like you used in your PR. We haven't adopted `shared_ptr` in the library, but I would be OK with this. It can sometimes be hard to trace ownership with `shared_ptr`, but in this case it should be clear enough.

> * Without a function like `GetFESpace`, the user is responsible for constructing the appropriate scalar version of the vector fespace that matches the partially constructed data.

Yes, true. I think this could be made easier by providing a mechanism to easily create scalar/vector versions of a given space with different vdim.

PR #4044, opened 2023-12-22
Named attribute sets

mlstowell:
Adding support for named attribute sets. This is mainly a change to the MFEM mesh format which adds two new sections `attribute_sets` and `bdr_attribute_sets` after the lists of elements and boundary elements. I had considered adding the new sections at the end of the mesh file but this was complicated by the reading of nodal GridFunctions which are read after returning from `ReadMFEMMesh`.

The new attribute sets are quite simple. There are a handful of new `Mesh` methods for defining, deleting, augmenting, and accessing the new sets which are stored internally as `Array<int>` objects contained in STL map containers keyed by STL strings. Two separate containers have been added; one for element attributes and the other for boundary element attributes. Likewise the accessor methods are provided in two flavors. The sets do not need to be disjoint or to span the full range of attribute IDs. The same names can be used for boundary and element sets without creating name clashes.

This PR includes a simple Gmsh script for producing a small mesh with multiple named physical regions and boundaries. The Gmsh reader has been updated to process these named entities whenever present and populate the corresponding attribute sets. There is also a modified version of `ex1p` which reads this Gmsh mesh and uses the new set methods to create new sets from the sets in the file before using these new sets to define sources, material regions, and boundary conditions.

| [#4044](https://github.com/mfem/mfem/pull/4044) | @mlstowell | @tzanio | @pazner + @dylan-copeland + @tzanio | 1/16/24 | 4/2/24 | 4/3/24 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


mlstowell:
@rieben1, @brunner6, and @wnissen, I think this PR is almost ready for review. I plan to clean up the new example code and add a serial version as well. There will also be a companion PR documenting the changes to the mesh format on the mfem.org web pages.

I'd be very interested to hear any feedback you might have. The new features added in this PR are quite simple but I think they cover the basics. If there's anything else you think we should add at this point just let me know.

wnissen:
> @rieben1, @brunner6, and @wnissen, I think this PR is almost ready for review. I plan to clean up the new example code and add a serial version as well. There will also be a companion PR documenting the changes to the mesh format on the mfem.org web pages.
> 
> I'd be very interested to hear any feedback you might have. The new features added in this PR are quite simple but I think they cover the basics. If there's anything else you think we should add at this point just let me know.

Thanks, I am just about to walk out the door for the year but will take a look when I get back on the 8th. Glad to see string attributes!

brunner6:
@mlstowell : Cool!  I see the functions are well documented, and there's an example, but it looks like it's geared toward exhaustive testing rather than an introduction on how to use this.  Is there a place to describe the simplest way to use this?  Questions I want to ask are:

1. How can I get attribute to apply a boundary condition with it?
2. Can I loop over elements in a set and do things like compute a material model?
3. How does this work with refinement, either uniform for the high-order to low-order, or full-up AMR?  (I'm OK if this is coming soon.  But I'd like a path from what's here to full AMR support.)

mlstowell:
Hi, @brunner6,
Your third question is perhaps the easiest to address. Attribute numbers are propagated to any new elements that are created during any type of mesh refinement. The rule is simply that any new child elements are assigned the same attribute number as the parent element from which they were created. This PR does not alter this behavior in any way so the new named sets will work on refined meshes without any modification.

The other two questions and the overall query about an introduction involve so many special cases they are difficult to answer in detail. We don't currently have any introductory documentation but we do intend to add something to our web pages. I think the best answer I can give at this point is that if you have a code which currently applies boundary conditions or builds material models based on attribute numbers then this PR merely adds the option to refer to these attribute numbers by name rather than number but the rest of your code will continue to work as it currently does. The only possible change would be if your code is applied to individual attribute numbers it could now be applied to sets of attribute numbers referred to by name.

This PR does not add any significant functionality to MFEM. It literally just adds the ability to refer to sets of attribute numbers by name as a convenience. More significant functionality could be added in the future but much of this would only consist of wrapper functions which take a name as an argument, find the corresponding set of attribute numbers, and then use the existing attribute number based functions.

You mention looping over elements in a set. The way MFEM currently does this is to loop over all elements and check to see if each is in the desired set. This works well if the set is expected to be large and the overhead of ignoring elements outside the desired set is relatively small. If, on the other hand, the desired set is small it would probably be more efficient to somehow keep track of this small set of elements and loop over them when needed. I've already done some work to support this but it has not been a priority. If you have a need for this type of functionality we should discuss your requirements and perhaps adjust our priorities.

Best wishes,
Mark

mesh/mesh.cpp
```diff
@@ -1581,6 +1581,120 @@ void Mesh::SetAttributes()
    }
 }
 
+void Mesh::GetAttributeSetNames(std::set<std::string> &names) const
+{
+   names.clear();
+
+   std::map<std::string,Array<int> >::const_iterator it;
```
paul-hilscher, line 1588: minor:

Following [cpp core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-for-range) I am wondering whether a range-for loop should be preferred here:

```
{
   names.clear();

   for(auto const &attr_set: attr_sets)
   {
      names.insert(attr_set.first);
   }
}   
```
? Same in some other places (GetBdrAttributeSetNames, ...)

brunner6:
@mlstowell : Thanks.  I had a long conversation with @rieben1 and @tzanio yesterday about this, and I think I'm conflating at least two feature requests into one, like what you suggested.  This is a great addition, and we can have other discussions about more features when we really need them.  I think what I'm really hoping is that I never need to know about the actual ID numbers as I go from mesh generation to the code.  Maybe this means more work for @wnissen to auto generate attributs in Pmesh from whatever sets I apply to the mesh.

brunner6:
I'm not authorized to approve the MR, but I support this change.

mesh/mesh.hpp
```diff
@@ -1014,6 +1022,138 @@ class Mesh
...
+       reference.
+    */
+   Array<int> & GetAttributeSet(const std::string & set_name);
+
+   /// @brief Access a named boundary attribute set
+   /**
+       @param[in] set_name The name of the set being accessed
+
+       @note If the named set does not exist a new empty set will be created
+       and returned.
```
paul-hilscher, line 1148: minor/optional:

just wondering if there is a use case for this ? Alternatively a more defensive approach
```
Array<int> & Mesh::GetBdrAttributeSet(const std::string & set_name)
{
   return bdr_attr_sets.at(set_name);
}
```
could be used to avoid this side-effect (in which case the code will throw an exception).

mesh/mesh_readers.cpp
```diff
@@ -2735,6 +2868,26 @@ void Mesh::ReadGmshMesh(std::istream &input, int &curved, int &read_gf)
       }
    } // we reach the end of the file
 
+   // Process set names
```
paul-hilscher, line 2871: nitpick/optional:

maybe:
```
   if(!phys_names_by_dim.empty()) 
   {
      // Process boundary attribute set names
      for(auto const &attr: phys_names_by_dim[Dim-1])
      {
         AddToBdrAttributeSet(attr.second, attr.first);
      }

      // Process element attribute set names
      for(auto const &attr: phys_names_by_dim[Dim])
      {
         AddToAttributeSet(attr.second, attr.first);
      }
   }
```

mesh/mesh.cpp
```diff
@@ -1581,6 +1581,116 @@ void Mesh::SetAttributes()
...
+void Mesh::AddToBdrAttributeSet(const std::string &set_name,
+                                const Array<int> &attr)
+{
+   bdr_attr_sets[set_name].Append(attr);
+   bdr_attr_sets[set_name].Sort();
+   bdr_attr_sets[set_name].Unique();
+}
+
+void Mesh::RemoveFromAttributeSet(const std::string &set_name, int attr)
+{
```
paul-hilscher, line 1657: I wonder if this could be replaced with:
```
{
    Array<int> &attrs = attr_sets[set_name];

    attrs.DeleteFirst(attr);
}
```

1) This would require that 
```
void Mesh::SetAttributeSet(const std::string &set_name, const Array<int> &attr)
void Mesh::SetBdrAttributeSet(const std::string &set_name,  const Array<int> &attr)
```
do actually call `Sort()` and `Unique()`. 

2) Ideally `Array<T>` should have a member function named `DeleteAll(const T &el)` which does "the right thing" based on whether an internal flag indicates if it  the array is sorted and/or unique. This would remove later the duplication for `Mesh::RemoveFromBdrAttributeSet`.

mesh/mesh.cpp
```diff
@@ -1581,6 +1581,116 @@ void Mesh::SetAttributes()
...
+   for (auto const &bdr_attr_set : bdr_attr_sets)
+   {
+      names.insert(bdr_attr_set.first);
+   }
+}
+
+
+void Mesh::SetAttributeSet(const std::string &set_name, const Array<int> &attr)
+{
+   attr_sets[set_name] = attr;
```
paul-hilscher, line 1607: What is the motivation that `attr` is not sorted and unified here ?

mesh/mesh.hpp
mlstowell: Probably not. I was thinking of justifying it as a way to create a new set when you don't yet know which attributes will belong to it. This is a flimsy justification.

To be honest I was avoiding the need to figure out how to handle the returned reference. I didn't want to halt execution in this situation. I considered issuing a warning message. But in this case what Array<int> do I return? I don't have an empty array in the Mesh class that will persist beyond the call to this method. I could add one as member data but that seems odd. I could allocate an array object just to return a reference to it but then the caller will need to delete it at some point. Do you (@paul-hilscher) happen to know how this situation is typically handled?

mesh/mesh_readers.cpp
mlstowell: I missed this one. I'm glad you found it.

mesh/mesh.cpp
mlstowell: No motivation. Just an oversight. Good catch!

mesh/mesh.cpp
mlstowell: Good idea. This has been done. I also added a warning message if the named sets don't exist (in just these two "RemoveFrom*" methods).

mesh/mesh.hpp
pazner: I also think automatically creating a new empty attribute set is not the right behavior here. There are a couple of options to handle this:

* `MFEM_ABORT`. This will halt execution, but it probably indicates a bug that needs to be fixed anyway?
* Change the return type to `Array<int>*`, allowing for `nullptr` as a return value.
* When we eventually (hopefully soon!) move to C++17, we can start using `std::optional` (in this case it's a little more complicated since we would need to use `std::reference_wrapper` as well)

pazner (commented):
This is going to be a useful feature, especially when used together with Gmsh's physical surfaces/volumes/etc. Thanks for the implementation @mlstowell.

I left a couple of minor comments.

Perhaps a bigger suggestion: what do you think about extracting this functionality to a separate class (and separate source file), e.g. an `AttributeSets` class? The `Mesh` class is already large and unwieldy, and the source files are too long (the `[p]mesh.?pp` files total over 23,000 lines). To me this seems like pretty discrete functionality that can be cleanly separate from the rest of the mesh class.

mesh/mesh.cpp
```diff
@@ -4104,6 +4249,18 @@ Mesh::Mesh(const Mesh &mesh, bool copy_nodes)
...
+      // Copy attribute and bdr_attribute names
+      for (auto const &attr_set : mesh.attr_sets)
+      {
+	 attr_set.second.Copy(attr_sets[attr_set.first]);
+      }
+      for (auto const &bdr_attr_set : mesh.bdr_attr_sets)
+      {
+	 bdr_attr_set.second.Copy(bdr_attr_sets[bdr_attr_set.first]);
+      }
+   }
```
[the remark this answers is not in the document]
mlstowell: I'm not sure and I did not try that. I noticed that all of the other Array objects are copied using Array::Copy so I wanted to use the same mechanism here.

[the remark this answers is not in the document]
pazner: I think it should work, and since this pattern appears in a number of places, it should simplify things quite a bit.

Note that the copy assignment operator for `Array<T>` is defined using `Array<T>::Copy`, so it should be essentially identical
```cpp
Array<T> &operator=(const Array<T> &src) { src.Copy(*this); return *this; }
```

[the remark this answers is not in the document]
v-dobrev: The use of `Copy` here probably dates back to when `Array` did not have copy-assignment.

mesh/mesh.hpp
wnissen: In some of the simulation codes I'm familiar with, an attribute is checked for existence to determine if a particular kind of setup or physics is enabled, like slide lines. Having it return an empty set breaks the semantics, though it's not the end of the word.

data/compass.mesh
```diff
@@ -0,0 +1,95 @@
...
+"Eastern Boundary" 2 1 8
+"NNE" 1 2
+"NNW" 1 3
+"Northern Boundary" 2 2 3
+"SSE" 1 7
+"SSW" 1 6
+"Southern Boundary" 2 6 7
+"WNW" 1 4
+"WSW" 1 5
+"Western Boundary" 2 4 5
```
wnissen, line 77: This is a totally reasonable way to do attribute sets, and easy to implement on the meshing side.

pazner:
Hi @mlstowell, hope you don't mind me chiming in… I think you can simplify/reduce a lot of boilerplate in the `ArraysByName` class, see [this branch](https://github.com/mfem/mfem/compare/named-attr-sets...named-attr-sets-simplify) for a suggestion.

mlstowell:
Thanks, @pazner, of course I don't mind. That is tempting. If I can retain the functionality I need it would certainly be good to eliminate the iterator code. I'll give it a try.

pazner:
I would also suggest removing all the copy constructors and copy assignment operators (the implicitly-declared ones will do the job just fine). `ArraysByName<T>::Copy` also becomes unnecessary. Personally I also don't see the need for the overload of `GetNames` that copies into an existing set. Is the purpose optimization? I suspect there is no meaningful improvement in efficiency compared with the version that returns the requested set.

mlstowell:
I have no particular attachment to the `ArraysByName<T>::Copy` method (or the copy constructor) but I feel that it is important to provide an interface which is analogous to the `Array` class (and other MFEM classes). I think, therefore, that the decision to remove such methods should extend to the library as a whole but I don't want to make that decision in this PR.

The `GetNames(std::set<std::string> &names)` method is useful because sets can be created or deleted so updating an existing set of names may be preferable to copying the set. I also don't think this is much of a performance optimization (these are likely to be small containers after all). Partly I prefer the slightly shorter line of code because it reflects how I expect such things to work:
```
set<string> names = abn.GetNames();
...
abn.GetNames(names);
/// vs.
names = abn.GetNames();
```
I should also admit to an old bias against any unnecessary copying (which probably goes hand-in-hand with my expectations). Again I doubt the performance difference would be noticeable in most applications but why copy a set of items twice when once is sufficient?

pazner:
> I have no particular attachment to the `ArraysByName<T>::Copy` method (or the copy constructor) but I feel that it is important to provide an interface which is analogous to the `Array` class (and other MFEM classes). I think, therefore, that the decision to remove such methods should extend to the library as a whole but I don't want to make that decision in this PR.

If the copy constructor/assignment operator is removed, it will be implicitly declared/defined, so the interface remains consistent with other copyable objects (both in MFEM, and externally). If anything, I would argue that `Array<T>::Copy` is the outlier (as far as I can tell, only `Table` and `GroupTopology` have such `Copy` member functions).

The other important thing about not providing user-defined copy constructors and assignment operators is that those will cause there to be no implicitly declared move constructors or move assignment operators, which prevents some optimizations (see [rule of three/five/zero](https://en.cppreference.com/w/cpp/language/rule_of_three)). The [recommendation](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero) in the C++ core guidelines is to try to avoid defining these special member functions.

> The `GetNames(std::set<std::string> &names)` method is useful because sets can be created or deleted so updating an existing set of names may be preferable to copying the set. I also don't think this is much of a performance optimization (these are likely to be small containers after all). Partly I prefer the slightly shorter line of code because it reflects how I expect such things to work:
> 
> ```
> set<string> names = abn.GetNames();
> ...
> abn.GetNames(names);
> /// vs.
> names = abn.GetNames();
> ```
> 
> I should also admit to an old bias against any unnecessary copying (which probably goes hand-in-hand with my expectations). Again I doubt the performance difference would be noticeable in most applications but why copy a set of items twice when once is sufficient?

The second one is only three characters longer :)
And if you really want to save typing, `names=abn.GetNames();` is only one character longer.

In this case, it will be a move rather than a copy, so I don't think there should be any unnecessary copying.

My suggestion is motivated mostly by two factors: value semantics are easier to reason about and less likely to result in bugs (and with move semantics, are not meaningfully less performant), and reducing API surface area reduces maintenance, documentation, and development burden. But of course some of these are subjective/judgement calls, so I understand that others may see things differently

tzanio:

mlstowell:
> The second one is only three characters longer :)
> 
> And if you really want to save typing, names=abn.GetNames(); is only one character longer.
> 
> In this case, it will be a move rather than a copy, so I don't think there should be any unnecessary copying.
> 
> My suggestion is motivated mostly by two factors: value semantics are easier to reason about and less likely to result in bugs (and with move semantics, are not meaningfully less performant), and reducing API surface area reduces maintenance, documentation, and development burden. But of course some of these are subjective/judgement calls, so I understand that others may see things differently

Okay, you've convinced me. I'll remove the `GetNames<set<string>&>` methods.

general/arrays_by_name.cpp
```diff
@@ -0,0 +1,66 @@
...
+// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// CONTRIBUTING.md for details.
+
+#include "arrays_by_name.hpp"
+
+namespace mfem
+{
+
+template <class T>
+void ArraysByName<T>::Print(std::ostream &os, int width) const
```
pazner, line 18: Is there a reason these are in the cpp file and only instantiated for `ArraysByName<T>` (rather than having everything in the hpp file)?

general/arrays_by_name.hpp
```diff
@@ -0,0 +1,256 @@
...
+   /// @brief Copy constructor: deep copy from @a src
+   ArraysByName(const ArraysByName &src) = default;
+
+   /// @brief Move constructor: deep move from @a src
+   ArraysByName(ArraysByName &&src) noexcept = default;
+
+   /// @brief Copy constructor (deep copy) from 'src', an ArraysByName
+   /// container of convertible type.
+   template <typename CT>
+   inline ArraysByName(const ArraysByName<CT> &src);
```
pazner, line 52: I know this is analogous to a similar constructor in `Array<T>`, but I really don't think this is needed ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)). If down the line we have a need for something like this, we can always add it. (The downside of adding something now rather than waiting until later is that we have to commit to backwards compatibility, maintaining this code in the future, etc.).

general/arrays_by_name.hpp
```diff
@@ -0,0 +1,256 @@
...
+
+   /// @brief Return the number of named arrays in the container
+   inline int Size() const { return data.size(); }
```
pazner, line 55:
```suggestion
   int Size() const { return data.size(); }
```
Don't need `inline` if the member function is defined in the class body (it is automatic).

general/arrays_by_name.hpp
```diff
@@ -0,0 +1,256 @@
...
+
+public:
+
+   /// @brief Default constructor
+   ArraysByName() = default;
+
+   /// @brief Copy constructor: deep copy from @a src
+   ArraysByName(const ArraysByName &src) = default;
+
+   /// @brief Move constructor: deep move from @a src
```
pazner, line 46:
```suggestion
   /// @brief Move constructor
```

Not sure "deep move" is so helpful, since move is kind of like a shallow copy. Probably just "Move constructor" is enough.

general/arrays_by_name.hpp
```diff
@@ -0,0 +1,256 @@
...
+   inline void SortAll();
+
+   /// @brief Remove duplicates from each sorted named array
+   ///
+   /// @note Identical entries may exist in multiple arrays but will only occur
+   /// at most once in each array.
+   inline void UniqueAll();
+
+   /// STL-like begin.  Returns pointer to the first element of the array.
+   inline iterator begin() { return data.begin(); }
```
pazner, line 132:
```suggestion
   iterator begin() { return data.begin(); }
```
And similarly for the `begin` and `end` below.

general/arrays_by_name.hpp
```diff
@@ -0,0 +1,256 @@
...
+        it1 != LHS.end() && it2 != RHS.end(); it1++, it2++)
+   {
+      if (it1->first != it2->first) { return false; }
+      if (it1->second != it2->second) { return false; }
+   }
+   return true;
+}
+
+template <typename T> template <typename CT>
+inline ArraysByName<T>::ArraysByName(const ArraysByName<CT> &src)
```
pazner, line 158:
```suggestion
ArraysByName<T>::ArraysByName(const ArraysByName<CT> &src)
```
Don't need `inline` on the function definitions if there is already `inline` on the declarations.

mesh/attribute_sets.hpp
```diff
@@ -0,0 +1,194 @@
...
+class AttributeSets
+{
+public:
+   /// Named sets of element attributes
+   ArraysByName<int> attr_sets;
+
+   /// Named sets of boundary attributes
+   ArraysByName<int> bdr_attr_sets;
+
+   AttributeSets() {}
```
pazner, line 35:
```suggestion
   AttributeSets() = default;
```
And we can do the same for the copy and move constructors as in `ArraysByName<T>`. (Or, equivalently, just remove all the  constructor declarations — they will be declared and defined implicitly).

mesh/attribute_sets.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+   /// @brief Create a copy of the internal data to the provided @a copy.
+   void Copy(AttributeSets &copy) const;
```
pazner, line 38: Can remove this in favor of copy/move constructors and `operator=` (as in `ArraysByName<T>`).

mesh/mesh.hpp
```diff
@@ -274,6 +275,9 @@ class Mesh
    /// A list of all unique boundary attributes used by the Mesh.
    Array<int> bdr_attributes;
 
+   /// Named sets of element and boundary element attributes
+   AttributeSets attribute_sets;
```
pazner, line 279: What do people think: should we have public member data or accessor member function? I'm fine either way

mesh/mesh.hpp
```diff
@@ -1010,6 +1014,44 @@ class Mesh
...
+   /// attributes
+   ///
+   /// @param max_attr Number of entries to create in the @a marker array
+   /// @param attrs    An array of attribute numbers which should be activated
+   /// @param marker   Output array indicating active/inactive attributes
+   ///
+   /// The marker array will be of size max_attr and it will contain only zeroes
+   /// and ones. Ones indicate which attribute numbers are present in the attrs
+   /// array. In the special case when attrs has a single entry equal to -1 the
+   /// marker array will contain all ones.
```
pazner, line 1027: I'm not sure this special case is desirable.

mesh/mesh.hpp
```diff
@@ -1010,6 +1014,44 @@ class Mesh
...
+   void AttrToMarker(int max_attr, const Array<int> &attrs,
+                     Array<int> &marker) const;
```
pazner, line 1029: Does this need to be a (non-static) `Mesh` member function? To me it seems like it could either be a free function or a static member function

examples/ex39.cpp
pazner, line 1: It might be good to base this example on something like ex0 rather than ex1. This example has a lot of "incidental complexity" — it supports static condensation, device execution, isoparametric FEs, algebraic CEED preconditioning, direct and iterative solvers, etc.

I think basing it on a more "pared down" example will help illustrate more clearly using the named attribute sets, rather than all the other stuff. (Same goes for ex39p).

general/arrays_by_name.hpp
```diff
@@ -0,0 +1,256 @@
...
+   /// @brief Copy assignment operator: deep copy from 'src'.
+   ArraysByName<T> &operator=(const ArraysByName<T> &src) = default;
+
+   /// @brief Move assignment operator: deep move from 'src'.
+   ArraysByName<T> &operator=(ArraysByName<T> &&src) noexcept = default;
+
+   /// @brief Assignment operator (deep copy) from @a src, an ArraysByName
+   /// container of convertible type.
+   template <typename CT>
+   inline ArraysByName &operator=(const ArraysByName<CT> &src);
```
pazner, line 108: Same as for the convertible type constructor

general/arrays_by_name.hpp
```diff
@@ -0,0 +1,256 @@
...
+   /// message and halt execution. This is intended to call attention to
+   /// possible typos or other errors. To handle such errors more gracefully
+   /// consider first calling EntryExists.
+   inline void DeleteArray(const std::string &name);
+
+   /// @brief Copy assignment operator: deep copy from 'src'.
+   ArraysByName<T> &operator=(const ArraysByName<T> &src) = default;
+
+   /// @brief Move assignment operator: deep move from 'src'.
+   ArraysByName<T> &operator=(ArraysByName<T> &&src) noexcept = default;
```
pazner, line 103:
```suggestion
   /// @brief Move assignment operator
   ArraysByName<T> &operator=(ArraysByName<T> &&src) noexcept = default;
```

pazner (commented):
This is a very useful feature, and it is well-implemented (and well-tested). Thank you @mlstowell! (Also, the compass mesh is super cool and ex39 is a nice illustration of the features).

I have a couple of mostly minor comments. I think the example code can be simplified/pared down to better focus on the new features (I can also help with this if you would like).

examples/ex39.cpp
mlstowell: A good point. I'll look into trimming it

general/arrays_by_name.cpp
mlstowell: The main reason is that I struggled with supporting different printed  formats which would add complexity to the Print method and I didn't want these longer methods in the header (I also wanted the files to be recompiled while I was testing formats without manually deleting select object files). Now that the formats are more or less settled I can move them into the header. I can also add more instantiation types.

mesh/mesh.hpp
mlstowell: Right, this should be static.

general/arrays_by_name.cpp
pazner: No need to add more (or any, really) instantiations — if it's in the header, the templates will be instantiated automatically when they're used. (And without `extern` declarations, I don't think it will speed up compile times).

mesh/mesh.hpp
mlstowell: It's something I use frequently. It is a convenient shortcut when specifying a long list of attributes would be a hassle. Why would it be undesirable?

pazner: To me, it's non-obvious behavior. If I saw passing -1 as an attribute, I would first think that it's a bug or mistake in the logic, and only until reading the (last sentence of the) documentation would it make sense.

Isn't it equivalent to this:

```
mesh.AttrToMarker(mesh.attributes, markers);
```

and

```
mesh.BdrAttrToMarker(mesh.bdr_attributes, bdr_markers);
```

which to me is much more self-documenting (i.e., intent is clearly indicated in the code)?

mesh/mesh.hpp
pazner: Note that `FiniteElementSpace` also has static member functions `MarkerToList` and `ListToMarker`, which do essentially the same thing.

Maybe these should be factored out to a general-purpose utility function, since the functionality is not specific to `Mesh` (nor to `FiniteElementSpace`).

mesh/mesh.hpp
mlstowell: I look at this from the point of view of an application developer. This is a shortcut I like to use so it appears in many applications that I have written. This means that in each application I need to write some small piece of code to offer such an option to the user and then parse that option and translate it into a line like those that you have suggested. It would be easier for me if the library recognized this special case. Then the application can simply parse the user selected set of attributes from the command line and pass them directly to the library. If the user specifies '-1' the library knows what to do with this and no code needs to be written in any application to process this special case.

That being said you also make a good point from the library developer point of view. If you want me to remove this special case I will.

mlstowell: Another relevant point is that this behavior has existed, for years, in the standalone function `AttrToMarker` located in the `miniapps/common/mesh_extras.?pp` files. This function. has just never been needed in the example codes until now.

pazner: You're right, I wasn't looking at it from the perspective of passing user input to these functions, and I can see how that would be convenient. But my vote would still be to remove this special case — a helper function to translate a special input (e.g. "-1") to mean all attributes belongs better in application code rather than in the core library (different applications could have different conventions for how BCs are specified)

mesh/mesh.hpp
pazner: `FiniteElementSpace::MarkerToList` and `FiniteElementSpace::ListToMarker` operate on the host. Are you worried that in the future this might change?

mesh/mesh.cpp
```diff
@@ -1624,6 +1624,37 @@ void Mesh::SetAttributes()
...
+         MFEM_VERIFY(attr > 0, "Attribute number less than one!");
+         marker[attr-1] = 1;
+      }
+   }
+}
+
+void Mesh::AttrToMarker(const Array<int> &attrs, Array<int> &marker) const
+{
+   AttrToMarker(attributes.Max(), attrs, marker);
+}
```
pazner, line 1651: This function should probably still work if both `Mesh::attributes` and `attrs` are empty (in which case `marker` will just be an empty array). But currently it will fail on the call to `attributes.Max()` (and same for `BdrAttrToMarker`).

mesh/mesh.hpp
mlstowell: Nothing here would prevent an application from using a different convention.

mesh/mesh.hpp
mlstowell: I'm not worried about anything I just see these as very different cases and I like having separate implementations in these two classes because DoFs are naturally associated with the `FiniteElementSpace` and domain/boundary regions with the `Mesh`.

mesh/mesh.cpp
mlstowell: Well isn't that annoying...

I don't believe `Mesh::attributes` can be empty in a valid mesh. On the other hand `attrs` certainly could be and we should handle that case properly. Good catch!

pazner: Ah, you're right about `attributes`, but `Mesh::bdr_attributes` can be empty (e.g. periodic meshes)

mesh/mesh.hpp
pazner: I see what you mean. Yeah, seems not worth it to unify them (especially since the attribute indices have to be shifted by -1. But besides for this difference, the implementations are actually the same).

mesh/mesh.cpp
mlstowell: Good point. I'll resolve that.

mesh/mesh.hpp
dylan-copeland: What is the reason for having any public access to `attribute_sets`? Isn't it used only internally by `Mesh`? Why not make it `protected`?

tests/unit/general/test_arrays_by_name.cpp
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) 2010-2023, Lawrence Livermore National Security, LLC. Produced
```
dylan-copeland, line 1: 2024

tests/unit/general/test_arrays_by_name.cpp
```diff
@@ -0,0 +1,129 @@
...
+// MFEM is free software; you can redistribute it and/or modify it under the
+// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// CONTRIBUTING.md for details.
+
+#include "mfem.hpp"
+#include "unit_tests.hpp"
+
+using namespace mfem;
+
+void FillArraysByName(ArraysByName<int> &abn)
```
dylan-copeland, line 17: Perhaps include duplicates in the arrays, and verify that duplicates are removed?

mesh/mesh.hpp
dylan-copeland: If we had another function, e.g. `MarkAllAttr(int max_attr, Array<int> &marker)` that sets `marker` to all 1's of size `max_attr`, it would be clearer and avoid the `-1` encoding. I suppose this would be less convenient for the application, since it has to check for its special case and call a different function, e.g.

`if (attrs[0] == -1) MarkAllAttr(max_attr, marker); else AttrToMarker(max_attr, attrs, marker);`

Is that inconvenient enough to justify complicating the interface to this function? It seems like there would be some logic to set `attrs[0] = -1` in the application. If it is already setting it to `-1` in some special case, calling `MarkAllAttr` in that case wouldn't be a significant inconvenience, right? Or am I underestimating the inconvenience?

dylan-copeland (changes requested):

examples/ex39.cpp
```diff
@@ -0,0 +1,287 @@
...
+   ConstantCoefficient defaultCoef(1.0e-6);
+   ConstantCoefficient baseCoef(1.0);
+   ConstantCoefficient roseCoef(2.0);
+
+   Array<int> base_marker;
+   Array<int> rose_marker;
+   mesh.AttrToMarker(attr_sets.GetAttributeSet("Base"), base_marker);
+   mesh.AttrToMarker(attr_sets.GetAttributeSet("Rose Even"), rose_marker);
+
+   a.AddDomainIntegrator(new DiffusionIntegrator(defaultCoef));
```
dylan-copeland, line 246: Comment that `defaultCoef` is the coefficient in the complement of "Base" and "Rose Even", which is "Rose Odd"?

mesh/attribute_sets.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+   /// @brief Return true if any named sets are currently been defined
+   bool SetsExist() const;
+
+   /// @brief Return all attribute set names as an STL set
+   std::set<std::string> GetAttributeSetNames() const;
+
+   /// @brief Return all boundary attribute set names as an STL set
+   std::set<std::string> GetBdrAttributeSetNames() const;
```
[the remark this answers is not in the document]
dylan-copeland: Good point about adding more sets in the future, which may be the most compelling reason to make the change.

mesh/mesh.hpp
vladotomov: Since `attributes` and `bdr_attributes` are public (right above), this can probably follow the pattern.

general/arrays_by_name.hpp
```diff
@@ -0,0 +1,273 @@
...
+
+#include <iostream>
+#include <map>
+#include <set>
+#include <string>
+
+namespace mfem
+{
+
+template <class T>
```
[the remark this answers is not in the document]
mlstowell: Thanks for catching this.

mesh/mesh.hpp
mlstowell: Yes, @vladotomov is correct. I was just following this pattern (and hoping no one would notice). I'm also fine either way.

examples/ex39.cpp
mlstowell: Actually the default coef is applied everywhere but it's so small it doesn't significantly alter diffusion in the other regions. Yes, this is sloppy. In any case more explanation is an excellent idea.

tests/unit/general/test_arrays_by_name.cpp
mlstowell: ...and the new tests fail...

mlstowell: I guess I'm still learning how to use c++-11

mesh/mesh.hpp
mlstowell: The main convenience is seen on the command line. If a user wants to specify all attributes they don't need to explicitly list them on the command line. Once we're inside an application I'm not overly concerned with the logic. This is just a convenient way to avoid duplicating the same logic in different applications. @dylan-copeland, if you also think this is confusing, I can remove it and just add this function to the two examples that are added in this PR.

dylan-copeland: If I understood correctly, the advantage is an application can have a command-line argument with an array of attributes, e.g. `-att 1 2 3 4 5`, or `-att -1` for all attributes, e.g. `miniapps/electromagnetics/maxwell.cpp`. Without supporting the `-1` case, the application would need an argument like `-allatt` and then have logic to call `MarkAllAttr` in the case that `-allatt` was used. That's not a serious inconvenience, but that logic would need to be repeated in every such application. I don't think it matters too much, and I'm fine with keeping it as is, if there is a reasonable justification like this.

pazner: I think best way to handle `-1` in this case would be a one-liner in application code after reading the command-line arguments:
```cpp
if (attrs.Size() == 1 && attrs[0] == -1) { attrs = mesh.attributes; }
```
No need to introduce a special case to the library API in my opinion.

examples/ex39.cpp
```diff
@@ -0,0 +1,287 @@
...
+   args.AddOption(&visualization, "-vis", "--visualization", "-no-vis",
+                  "--no-visualization",
+                  "Enable or disable GLVis visualization.");
+   args.Parse();
+   if (!args.Good())
+   {
+      args.PrintUsage(cout);
+      return 1;
+   }
+   args.PrintOptions(cout);
```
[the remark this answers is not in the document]
dylan-copeland: @pazner why hasn't this simplification been made in existing examples like `ex1p`?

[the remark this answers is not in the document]
pazner: I think it's just because of code churn — making the change would introduce new diffs that don't change functionality in any way. But this is a new example, so might as well use the newer way.

examples/ex39p.cpp
```diff
@@ -0,0 +1,330 @@
...
+      sol_name << "sol." << setfill('0') << setw(6) << myid;
+
+      ofstream mesh_ofs(mesh_name.str().c_str());
+      mesh_ofs.precision(8);
+      pmesh.Print(mesh_ofs);
+
+      ofstream sol_ofs(sol_name.str().c_str());
+      sol_ofs.precision(8);
+      x.Save(sol_ofs);
+   }
```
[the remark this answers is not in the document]
mlstowell: I think that small inconsistency is to avoid the odd name "mesh.mesh" in the serial case.

mesh/mesh.hpp
mlstowell: In the weekly PR meeting we decided to remove these new helper functions from the mesh class

general/arrays_by_name.hpp
```diff
@@ -0,0 +1,273 @@
...
+   /// @brief Load the contents of the container from an input stream
+   ///
+   /// @note This method will not first empty the container. First call
+   /// DeleteAll if this behavior is needed.
+   void Load(std::istream &in);
+
+   /// @brief Sort each named array in the container
+   inline void SortAll();
+
+   /// @brief Remove duplicates from each sorted named array
```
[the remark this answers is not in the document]
mlstowell: This was badly worded. I've tried to clarify it a little

mesh/mesh.hpp
```diff
@@ -1010,6 +1014,44 @@ class Mesh
    /// store these in the Mesh::attributes and Mesh::bdr_attributes arrays.
    virtual void SetAttributes();
 
+   /// @brief Prepares a marker array corresponding to an array of element
+   /// attributes
+   ///
+   /// @param max_attr Number of entries to create in the @a marker array
```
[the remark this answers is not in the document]
mlstowell: These methods have all been deleted

mesh/attribute_sets.hpp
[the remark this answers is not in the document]
mlstowell: Done

mlstowell:
@dylan-copeland, I saw a couple of your comments that I missed the first time. Could you please take another look and let me know if I missed anything else?

dylan-copeland (approved):

tests/unit/general/test_array.cpp
```diff
@@ -25,3 +25,46 @@ TEST_CASE("Array init-list construction", "[Array]")
       REQUIRE(a[i] == b[i]);
    }
 }
+
+TEST_CASE("Array entry sorting", "[Array]")
+{
+   int ContigData[6] = {6, 5, 4, 3, 2, 1};
+   Array<int> a(ContigData, 6);
+   Array<int> b({1.0, 2.0, 3.0, 3.0, 2.0, 1.0});
```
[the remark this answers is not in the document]
mlstowell: A fair question. This was copied from the test above and then modified. I cannot say why there are doubles in that test but they are not needed in this one. I can change this.

tests/unit/general/test_arrays_by_name.cpp
```diff
@@ -0,0 +1,202 @@
...
+{
+   ArraysByName<int> abn;
+
+   FillArraysByName(abn);
+
+   // Test standard Iterator
+   int i = 1;
+   for (auto a : abn)
+   {
+      REQUIRE(a.second.Size() == i);
```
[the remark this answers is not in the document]
mlstowell: Right. I prefer this method because potential failures will produce an error that shows which entry lead to the failure.

general/arrays_by_name.hpp
```diff
@@ -0,0 +1,291 @@
...
+   same underlying generic type T, which must be a trivial type, see
+   `std::is_trivial`.
+
+   In order to provide some level of protection against typos this class will
+   not create new named arrays when access to unrecognized names is requested.
+   New named arrays must be explicitly created using `CreateArray()`. To
+   facilitate this behavior and avoid such errors the method `EntryExists()` is
+   provided.
+
+   This container does not store pointers to pre-existing arrays it will copy
```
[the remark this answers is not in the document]
mlstowell: I have never been good at recognizing run-on sentences. If you could suggest a re-write I'll gladly use it.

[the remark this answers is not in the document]
mlstowell: Actually, I think I see a way to fix it. Feel free to edit it if you prefer.

tzanio:
Now that PR https://github.com/mfem/mfem/pull/3922 has been merged in `master`, please update this PR by:

1. merging `master` in here and resolving conflicts
2. replacing use of `double` with `real_t`
3. guarding double/single-only sections with `MFEM_USE_DOUBLE` and `MFEM_USE_SINGLE`
4. making sure tests run cleanly with default and `MFEM_PRECISION=single` configurations

See https://github.com/orgs/mfem/discussions/4207 for more details and/or ask @dylan-copeland for help.

tzanio (approved):
Thanks @mlstowell !

mlstowell:
Just conformed that all tests pass when running `make test` with `MFEM_USE_SINGLE`.

examples/ex39.cpp
```diff
@@ -0,0 +1,288 @@
+//                                MFEM Example 39
+//
+// Compile with: make ex39
+//
+// Sample runs:  ex39
+//               ex39 -ess "Southern Boundary"
+//               ex39 -src Base
+//
+// Device sample runs:
+//               ex39 -fa -d cuda
```
v-dobrev, line 10: Uncovered by the nightly tests: this sample run is not valid -- the option `-fa` is not supported at the moment.

examples/ex39p.cpp
```diff
@@ -0,0 +1,317 @@
+//                       MFEM Example 39 - Parallel Version
+//
+// Compile with: make ex39p
+//
+// Sample runs:  mpirun -np 4 ex39p
+//               mpirun -np 4 ex39p -ess "Southern Boundary"
+//               mpirun -np 4 ex39p -src Base
+//
+// Device sample runs:
+//               mpirun -np 4 ex39p -fa -d cuda
```
v-dobrev, line 10: Uncovered by the nightly tests: this sample run is not valid -- the option `-fa` is not supported at the moment.

tzanio:

tzanio:
Thanks so much for adding this @mlstowell, I am sure it will be really useful for our users!

PR #4065, opened 2024-01-11
Arena-type allocator for workspace/temporary vectors

pazner:
We commonly need "workspace" or "scratch" or other temporary vectors to perform some calculations. For example, `Operator::AddMult` is implemented as follows:
```cpp
void Operator::AddMult(const Vector &x, Vector &y, const double a) const
{
   mfem::Vector z(y.Size());
   Mult(x, z);
   y.Add(a, z);
}
```
The vector `z` is a short-lived temporary vector. Especially on GPU, the cost of performing these allocations (especially in a loop, e.g. in a CG iteration or multigrid) can be non-negligible.

A solution often used currently is to move `z` to be a mutable data member of the class. This way, the allocation is performed once only. However, the storage for `z` is not freed as a long as the object is alive, and it cannot be used for other purposes.

This PR introduces a mechanism designed at helping solve this kind of issue. In this PR, `Operator::AddMult` is implemented as follows:
```cpp
void Operator::AddMult(const Vector &x, Vector &y, const double a) const
{
   WorkspaceVector z = Workspace::NewVector(y.Size());
   Mult(x, z);
   y.Add(a, z);
}
```
Temporary vectors of this kind are created using `Workspace::NewVector`. The `Workspace` is essentially a "bump allocator" specifically designed for short-lived workspace vectors. When `z` goes out of scope, the backing vector (called a "chunk") remains allocated (i.e. there is no call to `delete` or `cudaFree`), however the `Workspace` is now free to reuse the memory for other workspace vectors. There is a mechanism to consolidate the allocated chunks to reduce fragmentation.

On GPU, creating a workspace vector with `Workspace::NewVector` (once the memory pool has been initialized) is orders of magnitude faster than allocating a new vector (incurring the overhead of `cudaMalloc`).

There are a couple of advantages of this approach:

* Obtain roughly the same performance as the "mutable vector" approach, but without adding member data, and keeping the code clearer (the scope/lifetime of the temporaries becomes obvious, and classes don't need to manage a bunch of mutable "temp" vectors)
* The memory used for workspace vectors can be reused between completely unrelated classes (e.g. when calling `AddMult` on two completely distinct `Operator`s, the same buffer is used for the temporary `z` in both instances)

The main downside that I am aware of is that the workspace retains the storage needed, and this may increase memory usage in some situations.

| [#4065](https://github.com/mfem/mfem/pull/4065) | @pazner | @tzanio | @sebastiangrimberg + @v-dobrev + @tzanio + @artv3 | 1/27/24 | ⌛due 2/10/24 | ⌛due 2/17/24 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


sebastiangrimberg:
This is really awesome! Great that it supports vectors of different sizes. I could see this being a big win when, for example, you have a lot of parallel or constrained operators and each one no longer needs to store it's own internal temporary l-vectors for the local multiplications. It should work with nesting as well, as far as I can tell, like for `P1^T P2^T A P2 P1`.

pazner:
Yes, it will definitely work with nesting. The first time the operator is applied, you won't see any benefit (since the allocator has to set up the buffers initially), but for subsequent operator applications, the allocations should be essentially free. The memory usage will be the maximum sum of the sizes of temporary vectors that are alive simultaneously.

sebastiangrimberg:
I am happy to be a reviewer on this once that starts. Looking forward to trying it out.

tzanio:

pazner:
@v-dobrev suggested an alternative approach, where the workspace/"arena allocator" is available as a `MemoryType`. This would have some advantages (for example, other data structures, such as arrays or matrices, could be allocated in the workspace).

I made a prototype of this idea in [this branch](https://github.com/mfem/mfem/compare/workspace-memspace). I'm not super confident that the implementation is fully correct (I don't know well all the internals of the memory manager).

sebastiangrimberg:
> @v-dobrev suggested an alternative approach, where the workspace/"arena allocator" is available as a `MemoryType`. This would have some advantages (for example, other data structures, such as arrays or matrices, could be allocated in the workspace).
> 
> I made a prototype of this idea in [this branch](https://github.com/mfem/mfem/compare/workspace-memspace). I'm not super confident that the implementation is fully correct (I don't know well all the internals of the memory manager).

I just took a brief look as I was working on testing out this PR today. I'm not sure I agree with the design decision to make it a `MemoryType`. In the original implementation, the `Workspace` sits above the abstraction level of the memory manager and is  compatible with any memory manager backend (like unified memory, Umpire, or just the defaults), without the user doing anything special in their code when they interact with the `Workspace`. I like this encapsulation. On the other hand, as a type of `MemoryType`, we lose this ability I'm pretty sure. Even if we don't lose it, it makes the code more difficult to understand as now the workspace allocation for the bump allocator is all tied together with the memory manager backend code.

general/workspace.hpp
```diff
@@ -0,0 +1,135 @@
...
+/// created with Workspace::NewVector.
+///
+/// A WorkspaceVector is created using the Workspace bump allocator. The
+/// allocator can quickly and cheaply create new vectors, so that these vectors
+/// can be created and destroyed in loops without incurring the memory
+/// allocation and deallocation overhead.
+///
+/// WorkspaceVector%s should be used only for short-lived temporary storage;
+/// they are not intended to be stored in other classes.
+class WorkspaceVector : public Vector
```
sebastiangrimberg, line 34: Just from a style point of view, I think newlines between the various declarations in the class are easier to read and also more consistent with the rest of the library. I don't think the space savings adds much here, and it's nicer with whitespace.

sebastiangrimberg:
Also, I since this is meant to be used in application codes and not just an internal header (I think?), a `
#include "general/workspace.hpp"` line should be added to the top-level `mfem.hpp`.

general/workspace.hpp
```diff
@@ -0,0 +1,135 @@
...
+{
+   // using internal::WorkspaceChunk;
+   friend class internal::WorkspaceChunk;
+   /// The WorkspaceChunk containing the data for this vector.
+   class internal::WorkspaceChunk &chunk;
```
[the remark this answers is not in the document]
pazner: Good point, done in 5426236427dafa521fa4e016652ff284e3f44a60.

general/workspace.hpp
pazner: Done in 399370db07a7ccd9ef53900307067ed56e4fa396.

general/workspace.hpp
```diff
@@ -0,0 +1,176 @@
...
+   /// Private constructor, create with Workspace::NewVector() instead.
+   WorkspaceVector(internal::WorkspaceChunk &chunk_, int offset_, int n);
+
+public:
+   /// @brief Move constructor. The moved-from WorkspaceVector has @a
+   /// size_in_chunk set to zero.
+   WorkspaceVector(WorkspaceVector &&other);
+
+   /// No copy constructor.
+   WorkspaceVector(WorkspaceVector &other) = delete;
```
hughcars, line 61: Why no `const` here and on the copy assignment?

pazner: Here I don't think it matters (since it's deleted), but for the copy assignment it should be `const`.

sebastiangrimberg: Fixed in [3c8c3ee](https://github.com/mfem/mfem/pull/4065/commits/3c8c3ee7f5e2b2b4ebcc0f6e6fd5c3952e10118b)

hughcars: > Here I don't think it matters (since it's deleted), but for the copy assignment it should be `const`.

But the `delete` would be removing the non-`const`, so i think as it was, there was still a `const` copy constructor.

sebastiangrimberg:
I have a strange observation when using `mfem::Workspace` on GPU, something related to the `MemoryManager` I believe. I am seeing the following error:

```
CUDA error: (cudaMemcpy(dst, src, bytes, cudaMemcpyDeviceToHost)) failed with error:
 --> misaligned address
 ... in function: void* mfem::CuMemcpyDtoH(void*, const void*, size_t)
 ... in file: /data/home/sjg/software/palace/build-cuda-debug-x86_64/extern/mfem/general/cuda.cpp:162
```

These errors don't seem to be consistently reproducible in terms of where they show up, and changing different parts of the code  which are fully unrelated can change if it occurs or not. I've seen similar behaviors related to aliasing underlying `Vector` objects using `MakeRef`, and wonder if this is related. There isn't anything that sticks out as part of this PR which would create problems when running on GPUs, unless there are underlying issues with multiple levels of pointer aliasing for `Memory` objects.

@pazner, did you test this on GPU at all? Any ideas?

sebastiangrimberg:
Updating here after debugging this odd `misaligned address` error. Walking through with `cuda-gdb`, the error is actually being encountered inside of cuSPARSE's spMV, and in particular here is the true error message thrown by cuSPARSE:

```
CUDA Exception: Warp Misaligned Address
The exception was triggered at PC 0x7f659b8a98d0

Thread 1 "palace-x86_64.b" received signal CUDA_EXCEPTION_6, Warp Misaligned Address.
[Switching focus to CUDA kernel 0, grid 554, block (80,0,0), thread (32,0,0), device 0, sm 0, warp 10, lane 0]
0x00007f659b8a9900 in void cusparse::csrmv_v3_partition_kernel<256, cusparse::VectorScalarMultiplyPolicy, int, int, double, double, double>(int const*, int, int, int, int*, cusparse::KernelCoeff<double>, int, double*)<<<(3767,1,1),(256,1,1)>>> ()
```

I'm using CUDA 12.2. When I use `mfem::Workspace` vectors, I can end up with a case where a vector starts at an alignment that seems to be rejected by cuSPARSE. Forcing the initial pointer for a vector to align to 16 bytes, instead of 8 which happens naturally, solves the errors. The issue thus arrises when putting a `WorkspaceVector` in a `WorkspaceChunk` following an odd-sized vector. I cannot find anywhere in the cuSPARSE or CUDA docs where there is a requirement for 16 byte alignment of the first entry of vector data, but it does seem to be that cuSPARSE is requiring this at least in this function. Seems very bizarre to me but maybe someone else has experience with this and it's expected behavior.

Here is the small patch required to force the non-standard alignment:

```
index ad8210a61..69622ec70 100644
--- a/general/workspace.cpp
+++ b/general/workspace.cpp
@@ -15,11 +15,11 @@ namespace mfem
 {

 WorkspaceVector::WorkspaceVector(
-   internal::WorkspaceChunk &chunk_, int offset_, int n)
+   internal::WorkspaceChunk &chunk_, int offset_, int n, int padding)
    : Vector(chunk_.GetData(), chunk_.GetOffset(), n),
      chunk(chunk_),
      offset(offset_),
-     original_size(n)
+     original_size(n + padding)
 {
    UseDevice(true);
 }
@@ -47,9 +47,12 @@ WorkspaceChunk::WorkspaceChunk(int capacity)

 WorkspaceVector WorkspaceChunk::NewVector(int n)
 {
-   MFEM_ASSERT(HasCapacityFor(n), "Requested vector is too large.");
-   WorkspaceVector vector(*this, offset, n);
-   offset += n;
+   constexpr int alignment = 16;
+   const int s = (n * sizeof(double)) % alignment;
+   const int padding = (s > 0) ? (alignment - s) / sizeof(double) : 0;
+   MFEM_ASSERT(HasCapacityFor(n + padding), "Requested vector is too large.");
+   WorkspaceVector vector(*this, offset, n, padding);
+   offset += n + padding;
    vector_count += 1;
    return vector;
 }
diff --git a/general/workspace.hpp b/general/workspace.hpp
index b5c0ebae8..e41e71b11 100644
--- a/general/workspace.hpp
+++ b/general/workspace.hpp
@@ -50,7 +50,8 @@ class WorkspaceVector : public Vector
    bool moved_from = false;

    /// Private constructor, create with Workspace::NewVector() instead.
-   WorkspaceVector(internal::WorkspaceChunk &chunk_, int offset_, int n);
+   WorkspaceVector(internal::WorkspaceChunk &chunk_, int offset_, int n,
+                   int padding = 0);

 public:
    /// @brief Move constructor. The moved-from WorkspaceVector has @a
@@ -104,6 +105,9 @@ public:
    /// Create a WorkspaceChunk with the given @a capacity.
    WorkspaceChunk(int capacity);

+   /// Return the data offset.
+   int GetOffset() const { return offset; }
+
    /// @brief Return the available capacity (i.e. the largest vector that will
    /// fit in this chunk).
    int GetAvailableCapacity() const { return data.Size() - offset; }
@@ -115,15 +119,12 @@ public:
    /// "original capacity" remains unchained.
    int GetOriginalCapacity() const { return original_capacity; }

-   /// Return the data offset.
-   int GetOffset() const { return offset; }
+   /// Returns true if this chunk can fit a new vector of size @a n.
+   bool HasCapacityFor(int n) const { return n <= GetAvailableCapacity(); }

    /// Sets whether the chunk is in the front of the list
    void SetFront(bool front_) { front = front_; }

-   /// Returns true if this chunk can fit a new vector of size @a n.
-   bool HasCapacityFor(int n) const { return n <= GetAvailableCapacity(); }
-
    /// Returns true if this chunk is empty.
    bool IsEmpty() const { return vector_count == 0; }
```

We could expose the `alignment` variable with a default of `sizeof(double)` (or `sizeof(real_t)` eventually) so users could set this. I have to imagine we would always want the default and would hope this issue with cuSPARSE is a bug or issue with my CUDA installation perhaps, but am not certain.

pazner:
> > @v-dobrev suggested an alternative approach, where the workspace/"arena allocator" is available as a `MemoryType`. This would have some advantages (for example, other data structures, such as arrays or matrices, could be allocated in the workspace).
> > I made a prototype of this idea in [this branch](https://github.com/mfem/mfem/compare/workspace-memspace). I'm not super confident that the implementation is fully correct (I don't know well all the internals of the memory manager).
> 
> I just took a brief look as I was working on testing out this PR today. I'm not sure I prefer the design where it is implemented as another `MemoryType`. In the original implementation, the `Workspace` sits above the abstraction level of the memory manager and is compatible with any memory manager backend (like unified memory, Umpire, or just the defaults), without the user doing anything special in their code when they interact with the `Workspace`. Though I think this still stands with the new implementation, I think it makes the code more difficult to understand as now the workspace allocation for the bump allocator is all tied together with the memory manager backend code when it doesn't seem like it has to be.
> 
> The enums for memory types were previously exclusive, and the new design doesn't seem to reflect that. A memory object with `HOST_WORKSPACE` or `DEVICE_WORKSPACE` type actually has underlying memory which has one of the other types associated with it. This makes me suspicious of whether this is the right place to insert this new functionality and whether or not it will play nicely with the other memory types.

I would like to return to working on this PR. These are all good points. @v-dobrev, what are your thoughts about this? (Having thought about it a bit more, I do think the ability to have non-Vector scratch memory could be very useful).

Also, regarding the alignment issues, it appears that the 16-byte alignment is indeed a requirement for cuSPARSE ([see here](https://docs.nvidia.com/cuda/cusparse/index.html?highlight=alignment#cusparseaxpby)). We can probably just always enforce 16-byte alignment for new allocations. Apparently cudaMalloc always enforces 128-byte alignment. We could do the same when allocating device memory?

pazner:
I substantially re-worked this PR after some discussion and feedback. (There has been a force-push to this branch—sorry!).

The fast arena allocator is now implemented using the memory manager. This likely incurs some amount of overhead (there are some hash map lookups, there are some virtual function calls in the memory manager, etc.). We should benchmark to see if the performance impact is significant. My guess is that the performance is still acceptable. I intend to add a mechanism to configure the memory manager to use e.g. Umpire for these allocations instead of the build-in arena allocator. The memory type is retrieved using `Device::GetDeviceTemporaryMemoryType()`, which is pretty verbose—we probably want something shorter/more convenient to type?

The deallocation method is more flexible. We don't strictly require stack (LIFO) ordering for the allocations, because the individual allocations are tracked more closely (using the hash map; this incurs some overhead, but is more flexible).

All arena allocations are aligned to 256 bytes (CUDA API calls also align to 256 bytes, so this is consistent, and should work with all the CUDA libraries).

There are still a few rough edges, but I think it's ready for another look at. @artv3, @hughcars, would you mind seeing if this implementation works with your applications (and if there are significant performance implications)?

helloworld922:
I measured the overhead of requesting a new temporary vector to previously allocated memory, and it appears to be about 1/3 microsecond per temporary vector, regardless of requested vector length N.
gpu_arena_vec

I also benchmarked `BlockOperator::Mult` setup for partial assembly in the same way as in ex5.cpp (Darcy problem) for a 2D rectangular domain with N elements along each side, with degree 1-3 elements.

gpu_darcy_block_op
gpu_darcy_block_op_zoom

Overall the two perform similarly, though there are a few discrepancies which I suspect are due to some kind of systematic issue with how I setup the benchmark.

najlkin:
A great improvement indeed 😉 , but I have a question. Is the implementation _thread safe_? A can imagine that someone will accidentally or intentionally try to allocate thread local vectors with this... 🤔

fem/linearform_ext.cpp
```diff
@@ -61,6 +61,7 @@ void LinearFormExtension::Assemble()
       }
 
       // Assemble the linear form
+      Vector b(elem_restrict_lex->Height(), Device::GetDeviceTemporaryMemoryType());
```
v-dobrev, line 64: We need to make sure `b.UseDevice(true)` is called, either directly or as part of the above constructor under some conditions.

One option for the above `Vector` constructor is to check if the `MemoryType` argument is in the currently-configured device `MemoryClass` as returned by `Device::GetDeviceMemoryClass()`, i.e. when `MemoryClassContainsType(Device::GetDeviceMemoryClass(), mt)` is true, and in that case call `UseDevice(true)`.

fem/linearform_ext.cpp
```diff
@@ -104,6 +105,8 @@ void LinearFormExtension::Assemble()
       }
 
       // Assemble the linear form
+      Vector bdr_b(bdr_restrict_lex->Height(),
+                   Device::GetDeviceTemporaryMemoryType());
```
v-dobrev, line 109: Same as above. There are probably other similar situations that need to be handled.

fem/linearform_ext.cpp
pazner: Good point. I think that's already handled here (line 972 below): 

https://github.com/mfem/mfem/blob/40f6b5658e9bae7a67b0dfb8841d5b47eba4c6db/general/mem_manager.hpp#L962-L973

Does that look OK to you?

pazner:
> A great improvement indeed 😉 , but I have a question. Is the implementation _thread safe_? A can imagine that someone will accidentally or intentionally try to allocate thread local vectors with this... 🤔

Good point, I haven't really thought about it, which means that it's very likely not thread-safe 😅 I would imagine that some kind of locks or mutexes would be needed to avoid data races.

On this topic: is the memory manager itself generally thread-safe? There are global maps that store metadata associated with allocated pointers (e.g. device pointer and alias information), I don't know if we can guarantee that there aren't data races, etc.

I don't know if this is important for the design. Maybe documentation would suffice for now?

tzanio:
Can you please merge the updated `master` and resolve conflicts?

najlkin:
@pazner , I just quickly checked and I think the memory manager is roughly thread-safe only on host, which is probably more or less ok, as no one mixes legacy OpenMP with GPUs hopefully 🤔 . However, I would consider the workspace vectors even for host, so we could get rid of the guarded ( with `MFEM_THREAD_SAFE`) local allocations of vectors, which are missing at many places anyway. This PR does not make the situation worse, but could make it better 😉 
A straightforward and performant way how to achieve thread safety could be to split the arena, so there would be multiple instances, each for one thread. There would then some accessing stub, returning the corresponding arena based on the thread id. This would work at least for OpenMP with `omp_get_thread_num()`. However, it can be deferred to a later PR and some documentation notes/warnings for now 😉 .

pazner:
Thanks for checking @najlkin.

I agree that the MFEM_THREAD_SAFE allocation of local vectors could be improved significantly here. I guess two options are (1) to have thread-local arenas, or (2) to make the arena thread safe.

I propose that we postpone this to a future PR (I'd be happy to review, but I don't have the expertise to implement). In the meantime I will write some documentation and mention this.

general/mem_manager.cpp
```diff
@@ -1301,6 +688,20 @@ bool MemoryManager::IsAlias_(const void *h_ptr)
    return maps->aliases.find(h_ptr) != maps->aliases.end();
 }
 
+bool MemoryManager::IsDanglingAlias_(const void *h_ptr)
+{
+   const auto alias_it = maps->aliases.find(h_ptr);
+   if (alias_it != maps->aliases.end())
+   {
+      const internal::Alias &alias = alias_it->second;
+      const void *base_h_ptr = static_cast<const char*>(h_ptr) + alias.offset;
```
v-dobrev, line 697: Just something I noticed here: I think the offset should subtracted from the alias address to get the base address.

general/device.hpp
```diff
@@ -273,6 +273,18 @@ class Device
    */
    static inline MemoryType GetDeviceMemoryType() { return Get().device_mem_type; }
 
+   static inline MemoryType GetDeviceTemporaryMemoryType()
```
v-dobrev, line 276: This type should be user configurable, e.g. users may want to use Umpire quick pool (setup as `MemoryType::DEVICE_UMPIRE_2`) for temporaries.

fem/linearform_ext.cpp
v-dobrev: The idea for the `USE_DEVICE` was that it is independent of the `Memory` class -- no methods in the class use the value or depend on the value, except the constructors (initialize to false) and the two methods that get and set the value.

I'm leaning toward keeping the current approach -- call `b.UseDevice(true)`. Modifying the `Vector` constructor as described above can have some undesired effects, e.g. creating a vector with the current host memory type will set the device flag when using the OpenMP backend (in some cases this is what one may want but in other cases it may not be).

pazner: How about the above, simply setting the device flag if `IsDeviceMemory(mt))` is true?

Is there a situation when a user will specifically construct a vector with a device memory type, but not intend for vector operations to be performed on device?

general/device.hpp
pazner: Yes, that's the plan. Should it just be a static function in `Device` like `Device::SetDeviceTemporaryMemoryType`?

Also, `GetDeviceTemporaryMemoryType` seems verbose. Is there a shorter alternative that you would prefer?

general/mem_manager.cpp
pazner: Thanks, fixed in d463bb916c6d6308144ab275717dc33ea6c49547.

fem/linearform_ext.cpp
v-dobrev: In this context, we want to have code that will set the device execution flag for OpenMP too, so using `IsDeviceMemory(mt))` is not sufficient.

general/device.hpp
v-dobrev: > Yes, that's the plan. Should it just be a static function in `Device` like `Device::SetDeviceTemporaryMemoryType`?

The user configuration for the memory types is now done with `Device::SetMemoryTypes`. One option is to add a version of `Device::SetMemoryTypes` with 4 args -- 2 new args for host-temp-type and device-temp-type. Another option is to add a new method `Device::SetTempMemoryTypes`. In either case, the 2-argument `Device::SetMemoryTypes` should set defaults for the temp-types that make sense. Also, we should check if any relevant changes are needed in `Device::UpdateMemoryTypeAndClass` and in `MemoryManager::Configure`.

> Also, `GetDeviceTemporaryMemoryType` seems verbose. Is there a shorter alternative that you would prefer?

The only thing that comes to mind is to shorten `Temporary` to `Temp`: `GetDeviceTempMemoryType`.

fem/linearform_ext.cpp
pazner: Ah yeah, I see. Maybe we should have a constructor for `Vector` that both sets the memory type to `Device::GetDeviceTemporaryMemoryType()` and sets `Vector::UseDevice(true)`.

Both as a convenience thing, and to avoid forgetting to call `UseDevice(true)` and then getting a performance penalty.

helloworld922:
@pazner something which has occurred to me is operations like `AddMult` sometimes are asynchronous with respect to the host, so we'll need some way to ensure kernel executions complete before the temporary memory is reused/de-allocated.

pazner:
> @pazner something which has occurred to me is operations like `AddMult` sometimes are asynchronous with respect to the host, so we'll need some way to ensure kernel executions complete before the temporary memory is reused/de-allocated.

Is this only possible on non-default CUDA streams? I don't immediately see how an issue could arise — do you have a concrete example?

helloworld922:
> > @pazner something which has occurred to me is operations like `AddMult` sometimes are asynchronous with respect to the host, so we'll need some way to ensure kernel executions complete before the temporary memory is reused/de-allocated.
> 
> Is this only possible on non-default CUDA streams? I don't immediately see how an issue could arise — do you have a concrete example?

It would require at least two streams which can execute asynchronously. Here's an example which uses the default stream and another stream which can execute asynchronous to the default stream:

```
#include "mfem.hpp"

#include "general/forall.hpp"

#include <cmath>

__global__ void zero_kernel(mfem::real_t *b, int n) {
  int i = threadIdx.x + blockIdx.x * blockDim.x;
  if (i < n) {
    for (int j = 0; j < 1000; ++j) {
      b[i] *= 0;
    }
  }
}

__global__ void copy_kernel(const mfem::real_t *a, mfem::real_t *b, int n) {
  int i = threadIdx.x + blockIdx.x * blockDim.x;
  if (i < n) {
    b[i] = a[i];
  }
}

__global__ void double_kernel(const mfem::real_t *a, mfem::real_t *b, int n) {
  int i = threadIdx.x + blockIdx.x * blockDim.x;
  if (i < n) {
    // dummy work
    for (int j = 0; j < 1000; ++j) {
      b[i] = 2 * a[i];
    }
  }
}

__global__ void add_kernel(const mfem::real_t *a, mfem::real_t *b,
                           mfem::real_t val, int n) {
  int i = threadIdx.x + blockIdx.x * blockDim.x;
  if (i < n) {
    b[i] = a[i] + val;
  }
}

int main(int argc, char **argv) {
  mfem::Device device("gpu");
  device.Print();

  mfem::Vector input(0x1000000);
  mfem::Vector output(0x1000000);
  mfem::Vector output2(0x1000000);
  input.HostWrite();
  for (int i = 0; i < input.Size(); ++i) {
    input[i] = i;
  }

#ifdef MFEM_USE_HIP
  hipStream_t stream;
  hipStreamCreateWithFlags(&stream, hipStreamNonBlocking);
#elif defined(MFEM_USE_CUDA)
  cudaStream_t stream;
  cudaStreamCreateWithFlags(&stream, cudaStreamNonBlocking);
#endif

  {
    mfem::Vector z = mfem::Vector::NewTemporary(input.Size());
    constexpr int team_size = 256;
    int nblocks = (input.Size() + team_size - 1) / team_size;
    copy_kernel<<<nblocks, team_size, 0, 0>>>(input.Read(), z.Write(),
                                                      input.Size());
    double_kernel<<<nblocks, team_size, 0, 0>>>(
        z.Read(), output.Write(), output.Size());
  }
  {
    mfem::Vector z = mfem::Vector::NewTemporary(input.Size());
    constexpr int team_size = 256;
    int nblocks = (input.Size() + team_size - 1) / team_size;
    zero_kernel<<<nblocks, team_size, 0, stream>>>(z.Write(), z.Size());
    add_kernel<<<nblocks, team_size, 0, stream>>>(z.Read(), output2.Write(), 1,
                                                   output2.Size());
  }

  MFEM_DEVICE_SYNC;
  output.HostRead();
  output2.HostRead();
  for (int i = 0; i < output.Size(); ++i) {
    if (fabs(output[i] - 2 * i) > 1e-1) {
      std::cout << "output[" << i << "] = " << output[i] << std::endl;
      break;
    }
  }

  for (int i = 0; i < output.Size(); ++i) {
    if (fabs(output2[i] - 1) > 1e-1) {
      std::cout << "output2[" << i << "] = " << output2[i] << std::endl;
      break;
    }
  }
#ifdef MFEM_USE_HIP
  hipStreamDestroy(stream);
#elif defined(MFEM_USE_CUDA)
  cudaStreamDestroy(stream);
#endif
}
```
Here is a screenshot of rocprof output from Tuolumne showing that `zero_kernel` is executing simultaneously to `double_kernel`.
Screenshot 2025-07-31 160423

If we're just limiting things to the default stream, then it's not an issue, though it might be worth adding a note somewhere if someone wants to use asynchronous queues.

general/mem_internal.hpp
```diff
@@ -0,0 +1,589 @@
...
+
+#include "mem_manager.hpp"
+#include "forall.hpp"
+#include <unordered_map>
+#include <cstdint>
+
+// Uncomment to try _WIN32 platform
+//#define _WIN32
+//#define _aligned_malloc(s,a) malloc(s)
+
```
helloworld922, line 23:
```suggestion

#ifdef MFEM_USE_UMPIRE
#include <umpire/Umpire.hpp>
#include <umpire/strategy/QuickPool.hpp>
#endif

```

general/mem_internal.hpp
```diff
@@ -0,0 +1,589 @@
...
+private:
+   umpire::strategy::AllocationStrategy *strat;
+public:
+   UmpireHostMemorySpace(const char * name)
+      : HostMemorySpace(),
+        UmpireMemorySpace(name, "HOST"),
+        strat(allocator.getAllocationStrategy()) {}
+   void Alloc(void **ptr, size_t bytes) override
+   { *ptr = allocator.allocate(bytes); }
+   void Dealloc(void *ptr) override { allocator.deallocate(ptr); }
```
helloworld922:
```suggestion
   void Dealloc(Memory &base) override { allocator.deallocate(base.h_ptr); }
```

general/mem_internal.hpp
```diff
@@ -0,0 +1,589 @@
...
+class UmpireDeviceMemorySpace : public DeviceMemorySpace,
+   public UmpireMemorySpace
+{
+public:
+   UmpireDeviceMemorySpace(const char * name)
+      : DeviceMemorySpace(),
+        UmpireMemorySpace(name, "DEVICE") {}
+   void Alloc(Memory &base) override
+   { base.d_ptr = allocator.allocate(base.bytes); }
+   void Dealloc(Memory &base) override { rm.deallocate(base.d_ptr); }
```
helloworld922:
```suggestion
   void Dealloc(Memory &base) override { allocator.deallocate(base.d_ptr); }
```

helloworld922 (changes requested):
I'm getting some unit test failures when I try running this on Tuolumne. I can work on fixing the Umpire related ones, though I do see one in test_dgmassinv.cpp when run in gpu_unit_tests with and without Umpire.

helloworld922:
Here's a patch which fixes the Umpire tests. It moves the Umpire-specific tests into their own executable to isolate them from other tests. I was able to modify the CMake get this to build correctly, however I'm not sure how to modify the makefile. @pazner or @v-dobrev could one of you help me with this?

```
diff --git a/tests/unit/CMakeLists.txt b/tests/unit/CMakeLists.txt
index 33cfdb1dc6..f05978f34d 100644
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -26,7 +26,6 @@ set(UNIT_TESTS_SRCS
   general/test_error.cpp
   general/test_mem.cpp
   general/test_text.cpp
-  general/test_umpire_mem.cpp
   general/test_zlib.cpp
   linalg/test_cg_indefinite.cpp
   linalg/test_chebyshev.cpp
@@ -211,6 +210,19 @@ if (MFEM_USE_HIP)
    endif()
 endif()
 
+if (MFEM_USE_UMPIRE)
+   if(MFEM_USE_CUDA OR MFEM_USE_HIP)
+      set(UMPIRE_UNIT_TESTS_SRCS umpire_unit_test_main.cpp general/test_umpire_mem.cpp)
+      mfem_add_executable(umpire_unit_tests ${UMPIRE_UNIT_TESTS_SRCS})
+      if(MFEM_USE_CUDA)
+         set_property(SOURCE ${UMPIRE_UNIT_TESTS_SRCS} PROPERTY LANGUAGE CUDA)
+      endif()
+      target_link_libraries(umpire_unit_tests mfem)
+      add_dependencies(${MFEM_ALL_TESTS_TARGET_NAME} umpire_unit_tests)
+      add_test(NAME umpire_unit_tests COMMAND umpire_unit_tests)
+   endif()
+endif()
+
 #-----------------------------------------------------------
 # SERIAL SEDOV + TMOP TESTS:
 #   sedov_tests_{cpu,debug,gpu,gpu_uvm}
diff --git a/tests/unit/general/test_umpire_mem.cpp b/tests/unit/general/test_umpire_mem.cpp
index 1d4fa12d3d..cfe5c31c4c 100644
--- a/tests/unit/general/test_umpire_mem.cpp
+++ b/tests/unit/general/test_umpire_mem.cpp
@@ -20,19 +20,12 @@
 #include "umpire/Umpire.hpp"
 #include <umpire/strategy/QuickPool.hpp>
 
-#ifdef MFEM_USE_CUDA
-#include <cuda.h>
-constexpr const char * device_name = "cuda";
-#elif defined(MFEM_USE_HIP)
-constexpr const char * device_name = "hip";
-#endif
-
 using namespace mfem;
 
 constexpr unsigned num_elems = 1024;
-constexpr unsigned num_bytes = num_elems * sizeof(double);
-constexpr double host_val = 1.0;
-constexpr double dev_val = -1.0;
+constexpr unsigned num_bytes = num_elems * sizeof(real_t);
+constexpr real_t host_val = 1.0;
+constexpr real_t dev_val = -1.0;
 
 static std::size_t alloc_size(const char * name)
 {
@@ -43,20 +36,7 @@ static std::size_t alloc_size(const char * name)
 
 static bool is_pinned_host(void * h_p)
 {
-   unsigned flags;
-#ifdef MFEM_USE_CUDA
-   auto err = cudaHostGetFlags(&flags, h_p);
-   cudaGetLastError(); // also resets last error
-   if (err == cudaSuccess) { return true; }
-   else if (err == cudaErrorInvalidValue) { return false; }
-#elif defined(MFEM_USE_HIP)
-   auto err = hipHostGetFlags(&flags, h_p);
-   hipGetLastError(); // also resets last error
-   if (err == hipSuccess) { return true; }
-   else if (err == hipErrorInvalidValue) { return false; }
-#endif
-   fprintf(stderr, "fatal (is_pinned_host): unknown return value: %d\n", err);
-   return false;
+   return Device::QueryMemoryType(h_p) == MemoryType::HOST_PINNED;
 }
 
 static void test_umpire_device_memory()
@@ -83,26 +63,14 @@ static void test_umpire_device_memory()
    rm.makeAllocator<umpire::strategy::QuickPool, true>(device_temp_alloc_name,
                                                        rm.getAllocator("DEVICE"), 0, 0);
 
-   // set the default host and device memory types; they will be made dual to
-   // each other
-   Device::SetMemoryTypes(MemoryType::HOST, MemoryType::DEVICE_UMPIRE);
-
-   // update some dual memory types
-   MemoryManager::SetDualMemoryType(MemoryType::DEVICE_UMPIRE_2,
-                                    MemoryType::HOST);
-   MemoryManager::SetDualMemoryType(MemoryType::HOST_PINNED,
-                                    MemoryType::DEVICE_UMPIRE);
-
    // set the Umpire allocators used with MemoryType::DEVICE_UMPIRE and
    // MemoryType::DEVICE_UMPIRE_2
    MemoryManager::SetUmpireHostAllocatorName(host_alloc_name);
    MemoryManager::SetUmpireDeviceAllocatorName(device_perm_alloc_name);
    MemoryManager::SetUmpireDevice2AllocatorName(device_temp_alloc_name);
-   Device device(device_name);
 
-   REQUIRE(device.GetHostMemoryType() == MemoryType::HOST);
-   REQUIRE(device.GetDeviceMemoryType() == MemoryType::DEVICE_UMPIRE);
-   device.Print();
+   REQUIRE(Device::GetHostMemoryType() == MemoryType::HOST);
+   REQUIRE(Device::GetDeviceMemoryType() == MemoryType::DEVICE_UMPIRE);
 
    printf("All pools should be empty at startup:");
    REQUIRE(alloc_size(host_alloc_name) == 0);
@@ -137,16 +105,19 @@ static void test_umpire_device_memory()
    // with the above constructor, host_temp is valid on device, so we cannot
    // directly access its host pointer; switch to valid on host without copying
    // data from device to host:
+   host_temp.UseDevice(false);
    host_temp.HostWrite();
    REQUIRE(!is_pinned_host(host_temp.GetData()));
+   REQUIRE(!host_temp.UseDevice());
    CHECK_PERM(num_bytes);
    CHECK_TEMP(0);
    PRINT_SIZES();
    host_temp = host_val; // done on host since UseDevice() is not set
+   REQUIRE(host_temp[0] == host_val);
 
    // allocate in temporary device memory
    printf("ReadWrite %u bytes in temporary memory: ", num_bytes);
-   double * d_host_temp = host_temp.ReadWrite();
+   real_t * d_host_temp = host_temp.ReadWrite();
    mfem::forall(num_elems, [=] MFEM_HOST_DEVICE (int i) { d_host_temp[i] = dev_val; });
    CHECK_PERM(num_bytes);
    CHECK_TEMP(num_bytes);
@@ -182,7 +153,7 @@ static void test_umpire_device_memory()
    PRINT_SIZES();
 
    printf("Write %u more bytes in temporary memory: ", num_bytes);
-   double * d_dev_temp = dev_temp.Write();
+   real_t * d_dev_temp = dev_temp.Write();
    mfem::forall(num_elems, [=] MFEM_HOST_DEVICE (int i) { d_dev_temp[i] = dev_val; });
    CHECK_PERM(num_bytes*2);
    CHECK_TEMP(num_bytes*2);
@@ -272,11 +243,14 @@ static void test_umpire_device_memory()
    printf("host=%zu\n", alloc_size(host_alloc_name));
 }
 
-TEST_CASE("UmpireMemorySpace", "[MemoryManager]")
+TEST_CASE("UmpireMemorySpace", "[MemoryManager][GPU]")
 {
    SECTION("Device")
    {
-      test_umpire_device_memory();
+      if (Device::Allows(Backend::DEVICE_MASK))
+      {
+         test_umpire_device_memory();
+      }
    }
 }
 
diff --git a/tests/unit/umpire_unit_test_main.cpp b/tests/unit/umpire_unit_test_main.cpp
new file mode 100644
index 0000000000..b1200c6b38
--- /dev/null
+++ b/tests/unit/umpire_unit_test_main.cpp
@@ -0,0 +1,36 @@
+//
+//
+
+#define CATCH_CONFIG_RUNNER
+#include "mfem.hpp"
+#include "run_unit_tests.hpp"
+
+{
+   // unit/general/test_umpire_mem.cpp has some strange requirements, isolate it
+   // from all other tests
+
+   // set the default host and device memory types; they will be made dual to
+   // each other
+   mfem::Device::SetMemoryTypes(mfem::MemoryType::HOST,
+                                mfem::MemoryType::DEVICE_UMPIRE);
+
+   // update some dual memory types
+   mfem::MemoryManager::SetDualMemoryType(mfem::MemoryType::DEVICE_UMPIRE_2,
+                                          mfem::MemoryType::HOST);
+   mfem::MemoryManager::SetDualMemoryType(mfem::MemoryType::HOST_PINNED,
+                                          mfem::MemoryType::DEVICE_UMPIRE);
+   mfem::Device device("gpu");
+
+   // Include only tests labeled with GPU. Exclude parallel tests.
+   int res = RunCatchSession(argc, argv, {"[GPU]", "~[Parallel]"});
+   return res;
+}
```

fem/qinterp/det.hpp
```diff
@@ -227,18 +219,21 @@ inline void Det3D(const int NE,
...
       const int max_qd = std::max(max_q1d, max_d1d);
       const int mem_size = max_qd * max_qd * max_qd * 9;
-      d_buff->SetSize(2*mem_size*GRID);
-      GM = d_buff->Write();
+      buffer_size = 2*mem_size*GRID;
    }
 
+   // if SMEM is true, d_buff will be empty (zero size)
+   Vector d_buff = Vector::NewTemporary(buffer_size);
+   real_t *GM = SMEM ? d_buff.Write() : nullptr;
```
helloworld922:
```suggestion
   real_t *GM = SMEM ? nullptr : d_buff.Write();
```

helloworld922: This fixes test_dgmassinv

pazner:
> Here's a patch which fixes the Umpire tests. It moves the Umpire-specific tests into their own executable to isolate them from other tests. I was able to modify the CMake get this to build correctly, however I'm not sure how to modify the makefile. @pazner or @v-dobrev could one of you help me with this?

Awesome, thanks! I added it to the makefile (hopefully correctly) in b6d94cc5ae754825365b0eabeb1b73456b721c5d

helloworld922:
Should we add a way to specify temporary host-pinned or managed buffers separately from the default host/device memory types? For example [mfem::reduce](https://github.com/mfem/mfem/blob/b6d94cc5ae754825365b0eabeb1b73456b721c5d/general/reducers.hpp#L532) uses a temporary host-pinned memory workspace.

It would also be helpful to have some equivalent members `Array::NewTemporary` and `Array::NewHostTemporary` for a similar reason.

PR #4078, opened 2024-01-22
Various minor bug/deficiency fixes

najlkin:
This PR combines various smaller bug or implementation deficiency fixes, following the effort of #1951 to align master with [najlkin/mfem](https://github.com/najlkin/mfem/). In particular, there are the following:

- A minor optimization of `SparseMatrix::EliminateCol()` to stop the linear search for the column when it is found.
- Added `const` qualifiers to `Operator`s in `BlockOperator` as its primary purpose is to apply ("multiply") the operators and has no methods to modify its operators, so write access is unnecessary.
- Added access to `FaceIsTrueInterior` in `ParMesh`, which is implemented in `Mesh` where it has no good usage and becomes useful only on distributed meshes.
- Added `const` qualifier to the quadrature function parameter of `QuadratureFunctionCoefficient`.
- Added default `NULL` integration rule to `(Vector)QuadratureLFIntegrator`, where it is not used anyway.
- Added `ClosedGL` quadrature among the closed ones in the `CheckClosed` method.
- Fixed boundary trace face integration in `MixedBilinearForm`. This, I believe, tracks back to the unification of the face transformations, where the behavior at the boundary is not fully backward compatible and `ElementNo` is not the face index, but the boundary face.
- Fixed `VectorDiffusionOperator` to use correct dimensions, which under normal circumstances does not matter as they are equal, but in some more advanced scenarios might give segfaults.
- Added `const` qualifiers to the boundary attributes markers in the projection methods of `GridFunction`.
- Added support of integral finite elements to calculation of the L2 error through `GridFunction::ComputeL2Error`.

| [#4078](https://github.com/mfem/mfem/pull/4078) | @najlkin | @tzanio | @vladotomov + @sebastiangrimberg | 1/23/24 | 2/1/24 | 2/3/24 | |

<details>
    
- [x] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>

### DIFF
--- a/fem/bilinearform.cpp
+++ b/fem/bilinearform.cpp
@@ -1585,9 +1585,10 @@ void MixedBilinearForm::Assemble(int skip_zeros)
          ftr = mesh->GetBdrFaceTransformations(i);
          if (ftr)
          {
-            trial_fes->GetFaceVDofs(ftr->ElementNo, trial_vdofs);
+            const int iface = mesh->GetBdrElementFaceIndex(i);
+            trial_fes->GetFaceVDofs(iface, trial_vdofs);
             test_fes->GetElementVDofs(ftr->Elem1No, test_vdofs);
-            trial_face_fe = trial_fes->GetFaceElement(ftr->ElementNo);
+            trial_face_fe = trial_fes->GetFaceElement(iface);
             test_fe1 = test_fes->GetFE(ftr->Elem1No);
             // The test_fe2 object is really a dummy and not used on the
             // boundaries, but we can't dereference a NULL pointer, and we don't
--- a/linalg/sparsemat.cpp
+++ b/linalg/sparsemat.cpp
@@ -1767,6 +1767,7 @@ void SparseMatrix::EliminateCol(int col, DiagonalPolicy dpolicy)
             if (aux->Column == col)
             {
                aux->Value = 0.0;
+               break;
             }
          }
       }
--- a/mesh/pmesh.hpp
+++ b/mesh/pmesh.hpp
@@ -491,6 +491,8 @@ public:
    void GenerateOffsets(int N, HYPRE_BigInt loc_sizes[],
                         Array<HYPRE_BigInt> *offsets[]) const;
 
+   using Mesh::FaceIsTrueInterior;
+
    void ExchangeFaceNbrData();
    void ExchangeFaceNbrNodes();
 
--- a/mesh/pmesh.hpp
+++ b/mesh/pmesh.hpp
@@ -491,7 +491,9 @@ public:
    void GenerateOffsets(int N, HYPRE_BigInt loc_sizes[],
                         Array<HYPRE_BigInt> *offsets[]) const;
 
-   using Mesh::FaceIsTrueInterior;
+   /** Return true if the face is interior or shared. In parallel, this
+       method only works if the face neighbor data is exchanged. */
+   inline bool FaceIsTrueInterior(int FaceNo) const { return Mesh::FaceIsTrueInterior(FaceNo); }
 
    void ExchangeFaceNbrData();
    void ExchangeFaceNbrNodes();
--- a/miniapps/solvers/div_free_solver.cpp
+++ b/miniapps/solvers/div_free_solver.cpp
@@ -587,7 +587,7 @@ void DivFreeSolver::Mult(const Vector & x, Vector & y) const
       ch.Clear();
       ch.Start();
 
-      auto M = dynamic_cast<const HypreParMatrix&>(ops_.Last()->GetBlock(0, 0));
+      auto& M = dynamic_cast<const HypreParMatrix&>(ops_.Last()->GetBlock(0, 0));
       M.Mult(-1.0, correction.GetBlock(0), 1.0, resid.GetBlock(0));
       SolvePotential(resid.GetBlock(0), correction.GetBlock(1));
       blk_y.GetBlock(1) += correction.GetBlock(1);
--- a/miniapps/solvers/div_free_solver.cpp
+++ b/miniapps/solvers/div_free_solver.cpp
@@ -314,7 +314,8 @@ void SaddleSchwarzSmoother::Mult(const Vector & x, Vector & y) const
    blk_y.GetBlock(1) -= coarse_l2_projection;
 }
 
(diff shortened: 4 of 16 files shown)


fem/bilinearform.cpp
```diff
@@ -1585,9 +1585,10 @@ void MixedBilinearForm::Assemble(int skip_zeros)
          ftr = mesh->GetBdrFaceTransformations(i);
          if (ftr)
          {
-            trial_fes->GetFaceVDofs(ftr->ElementNo, trial_vdofs);
+            const int iface = mesh->GetBdrElementFaceIndex(i);
+            trial_fes->GetFaceVDofs(iface, trial_vdofs);
```
vladotomov, line 1589: I might have a code that uses this, I'll check how it affects the results.

tzanio:

linalg/sparsemat.cpp
```diff
@@ -1767,6 +1767,7 @@ void SparseMatrix::EliminateCol(int col, DiagonalPolicy dpolicy)
             if (aux->Column == col)
             {
                aux->Value = 0.0;
+               break;
```
sebastiangrimberg, line 1770: I'm not sure this is correct behavior. The linked list implementation of a sparse matrix allows for a column nonzero to be repeated in a given row, indicating the values should be summed. Thus by breaking early you miss zero-ing all the entries for that column.

najlkin: If it that so, it is definitely wrong. I believed that the columns are unique for each row as the add methods search the row first. I also do not see any summation in `Finalize()` if I am not wrong. Can you point me to a part doing that addition?

sebastiangrimberg: Ah, you're probably right and I just assumed that was the case here. Apologies!

najlkin: No worries, but actually it is a good point. The/some methods could check internal consistency of the LIL if it does not contain repeated columns. At the moment, it goes under the radar if it happens somehow. A suggestion for a follow-up! 😉 Do you want to take it?

mesh/pmesh.hpp
```diff
@@ -491,6 +491,8 @@ class ParMesh : public Mesh
    void GenerateOffsets(int N, HYPRE_BigInt loc_sizes[],
                         Array<HYPRE_BigInt> *offsets[]) const;
 
+   using Mesh::FaceIsTrueInterior;
```
sebastiangrimberg, line 494: A better approach in my opinion would be to just move the definition of `FaceIsTrueInterior` in `mesh.hpp` into the `public methods`, rather than exposing it here. I would support that change, since `Mesh::FaceIsInterior` is already `public`.

miniapps/solvers/div_free_solver.cpp
```diff
@@ -586,7 +587,7 @@ void DivFreeSolver::Mult(const Vector & x, Vector & y) const
       ch.Clear();
       ch.Start();
 
-      auto M = dynamic_cast<HypreParMatrix&>(ops_.Last()->GetBlock(0, 0));
+      auto M = dynamic_cast<const HypreParMatrix&>(ops_.Last()->GetBlock(0, 0));
```
sebastiangrimberg, line 590: Nit: `auto &M`

mesh/pmesh.hpp
najlkin: I do not know, the thing is there is no good use of `FaceIsTrueInterior` for a non-distributed mesh, because it should coincide with the plain `FaceIsInterior`. It makes only sense for `ParMesh`, but it is not implemented there due to technical reasons, so this approach is maybe better.

miniapps/solvers/div_free_solver.cpp
najlkin: A good point, it is not related to the change, but should be fixed too 😉

mesh/pmesh.hpp
sebastiangrimberg: Because `FaceIsTrueInterior` uses only data members of the base `Mesh` class (in the face informations), its definition should be kept there and moving it would also be a breaking change. I think it is more clean to just make it `public` there and remove this added `using` statement.

najlkin: I agree with you that the implementation should stay there, just I am not sure if we should pollute `Mesh` by something useless there and potentially confusing (the TDOF terminology comes with the distributed meshes). This was probably the reason why this method has been made protected there.

sebastiangrimberg: I'm not sure I would call moving the method from protected to public polluting, and it adds zero lines of code vs. this change. I'd suggest just going ahead and doing it, or reverting this and opening a separate PR if you think it warrants further discussion.

najlkin: Ok, @vladotomov , any opinion on this?

vladotomov: In principle `using` should not be used this way, as it defeats the purpose of the `protected` declaration in the parent; this feels like a hack. But I also agree that `FaceIsTrueInterior` should not be a public method of `Mesh`.

You can create a new public `ParMesh` method that calls `Mesh::FaceIsTrueInterior`, this avoids both of the above.

najlkin: Ok, that sounds to me as a fair compromise 😉

fem/bilinearform.cpp
najlkin: Have you found time to get to it? However, this is (/should be) the right behavior, where the dofs are retrieved for the associated face. It had been mixing the face and boundary element indices since the unification of the `FaceTransformations`. Before that, the code was valid, because the actual transformation was always a face transformation.

vladotomov: I looked at it, but the code I was thinking about didn't go this path. Maybe it will show up in some of the autotests.

najlkin: That will imply to change the autotests then 😊 , but we will see. So ready to approve the PR to move it there?

sebastiangrimberg (approved):

vladotomov (approved):

tzanio:

PR #3443, opened 2023-01-31
Clarify ODE ImplicitSolve documentation

jandrej:
Closes #3395

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#3443](https://github.com/mfem/mfem/pull/3443) | @jandrej | @pazner | @cjvogl + @bensworth + @v-dobrev | 2/8/23 | ⌛due 2/22/23 | ⌛due 3/1/23 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
      - [ ] Add the example code to the `ALL_EXE_SRCS` variable.
      - [ ] Make sure `THIS_TEST_OPTIONS` is set correctly for the new example.
   - [ ] List the new example in `doc/CodeDocumentation.dox`.
   - [ ] If new examples directory (e.g.`examples/pumi`), list it in `doc/CodeDocumentation.conf.in`
[...]

PR #4115, opened 2024-02-06
Refactored TimeDependentOperator Documentation [operator-doc]

cjvogl:
This PR attempts to address issues identified in #3443 by improving the documentation for the `TimeDependentOperator`, among others that appear to stem from some documentation being written towards a DAE while other documentation was written towards a general ODE form.

A summary of the changes (and the motivation) is
1. the general description of the algebraic equations with $F$ and $G$ tended to switch between using $y$, $k$, and $f$ in a manner that seemed to cause confusion... this PR refactored the documentation to describe an operator from $(u,t)$ to $k(u,t)$ such that $F(u,k,t) = G(u,t)$, reserving $y$ for specific exampled of $F$ and $G$ (as they pertain to DAEs/ODEs) and avoiding the use of $f$ (and any confusion with it not being related to $F$).
2. for ODEs of the form $M \frac{dy}{dt} = g(y,t)$, it was not clear what $F$ and $G$ are nor which should contain the mass matrix... this PR refactored the documentation to directly describe $F$ and $G$, perhaps in a pedantic fashion, and how the mass matrix is included depending on whether the derived class uses $G(u,t) = M^{-1} g(u,t)$ or $G(u,t) = g(u,t)$ (especially important for `SUNImplicitSetup`).

| [#4115](https://github.com/mfem/mfem/pull/4115) | @cjvogl | @tzanio | @acfisher + @jandrej | 2/10/24 | 5/18/24 | 5/20/24 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


cjvogl:
@acfisher, I hope this is along the lines of what you were looking for.  I certainly welcome input from @jandrej and @bensworth as to whether this makes things more clear or not.

tzanio:

linalg/operator.hpp
```diff
@@ -309,16 +309,23 @@ class Operator
 
 
 /// Base abstract class for first order time dependent operators.
-/** Operator of the form: (x,t) -> f(x,t), where k = f(x,t) generally solves the
+/** Operator of the form: (x,t) -> k(x,t), where k generally solves the
```
jandrej, line 312: `TimeDependentOperator::Mult(x,y)` at time `t` gives you `f(x,t)` right?

cjvogl: What is your definition of `F` and `G`?  My interpretation of this class was that `Mult` gives `G(x,t)` (mainly because this is how Petsc defines `ExplicitMult`).  So if `F(x, k, t)` is defined as `k`, then `Mult` does indeed give you `k(x,t)` (or `f(x,t)` in the prior notation).

jandrej: I agree with it viewing it from the perspective of the F = G representation.

bensworth:
@cjvogl Overall I like the changes. I think my only complaint is I don't love the use of `x` as arguments for `F` and `G`, because in PDEs it is very natural to think of `x` specifically in the context of the spatial domain, not as a variable. Personally I would change that to `y` everywhere. Otherwise, thanks for clarifying documentation.

cjvogl:
> @cjvogl Overall I like the changes. I think my only complaint is I don't love the use of `x` as arguments for `F` and `G`, because in PDEs it is very natural to think of `x` specifically in the context of the spatial domain, not as a variable. Personally I would change that to `y` everywhere. Otherwise, thanks for clarifying documentation.

@bensworth, that makes sense to me.  I would like to avoid use of `y` in the arguments for `F` and `G` only because it appears as the "out" variable in `Mult`... would `u` work? i.e., `F(u,k,t) = G(u,t)`?

jandrej:
I don't have a preference

cjvogl:
@bensworth , I ended up going with `u` instead of `y` just to more clearly denote when talking about a general `F` and `G` (using `u`) and typical `F` and `G` for solving ODEs/DAEs (using `y`).  Let me know if this doesn't work for you.

bensworth:
Hey @cjvogl ya `u` is great, just anything but `x`. It could be worth adding a comment that `k` generally corresponds to `dy/dt`, but this is just a suggestion, I don't feel super strong about it

linalg/operator.hpp
cjvogl: Thinking about this more, I suppose `TimeDependentOperator::Mult` should return `k` if the underlying `Operator` is defined as a mapping to `k`.  I'm running an idea by the SUNDIALS devs to see if the `ARKStepSolver` can use `TimeDependentOperator::ExplicitMult` when `ARKStepSolver` is looking for `g(y,t)` instead of `M^{-1} g(y,t)`.  If that change can be made, then I'll refactor the new documentation to say 
1. `TimeDependentOperator::Mult` returns the `k` that solves `F(x,k,t) = G(x,t) for a given `x` at time `t`.
2. For solving ODEs of the form `M dy/dt = g(y,t)`, `F(x,k,t)` is defined as `M k` and `G` is defined as `g(y,t)`

acfisher:
@cjvogl Sorry I didn't get to this till now.  Yes this is exactly what I was looking for.  Thanks for putting it together.

acfisher (approved):

linalg/operator.hpp
```diff
@@ -309,30 +309,37 @@ class Operator
...
+    1. F(u,k,t) = k and G(u,t) = inv(M) g(u,t)
+    2. F(u,k,t) = M k and G(u,t) = g(u,t)
+    */
 class TimeDependentOperator : public Operator
 {
 public:
    enum Type
    {
-      EXPLICIT,   ///< This type assumes F(x,k,t) = k, i.e. k = f(x,t) = G(x,t).
+      EXPLICIT,   ///< This type assumes F(u,k,t) is linear in k.
```
v-dobrev, line 328: I don't like the change of meaning here -- this is not backward compatible.

If we want to have a type where `F` is linear in `k` (i.e. in `du/dt`) then it should be a new separate type.

linalg/operator.hpp
```diff
@@ -309,30 +309,37 @@ class Operator
...
+/** Operator of the form: (u,t) -> k(u,t), where k generally solves the
+    algebraic equation F(u,k,t) = G(u,t). The functions F and G represent the
+    _implicit_ and _explicit_ parts of the operator, respectively.
+
+    A common use for this class is representing a differential algebraic
+    equation of the form \f$ F(y,\frac{dy}{dt},t) = G(y,t) \f$. Consider an
+    ordinary differential equation of the form \f$ M \frac{dy}{dt} = g(y,t) \f$,
+    where F and G are defined as one of the following:
+    1. F(u,k,t) = k and G(u,t) = inv(M) g(u,t)
+    2. F(u,k,t) = M k and G(u,t) = g(u,t)
```
v-dobrev, line 321: You are changing `x` to `u` -- why use `y` here and not `u`? It is best to stay consistent.

linalg/operator.hpp
```diff
@@ -309,30 +309,37 @@ class Operator
...
       NORMAL,
-      /** Assuming additive split, f(x,t) = f1(x,t) + f2(x,t), evaluate the
-          first term, f1. */
+      /** Assuming additive split, G(u,t) = G1(u,t) + G2(u,t), evaluate with
+          F(u,k,t) = G1(u,t). */
       ADDITIVE_TERM_1,
-      /** Assuming additive split, f(x,t) = f1(x,t) + f2(x,t), evaluate the
-          second term, f2. */
+      /** Assuming additive split, G(u,t) = G1(u,t) + G2(u,t), evaluate with
+          F(u,k,t) = G2(u,t). */
```
v-dobrev, line 342: This is also changing meaning in a way that is not backward compatible.

v-dobrev:
I commented above about a few places where the meaning of constants/methods is being changed in a way that is not backward compatible. There are probably more places like this -- I did not look further. I think we should avoid changing things in a way that is not backward compatible.

linalg/operator.hpp
cjvogl: I have to admit that I didn't quite understand the use of `EXPLICIT` vs `IMPLICIT`... is there an example somewhere that differentiates between the two?

In solving $M \frac{dy}{dt} = g(y,t)$, is the related `TimeDependentOperator` considered `EXPLICIT`?

linalg/operator.hpp
cjvogl: I tried to stay consistent with using $F(u,k,t)$ and $G(u,t)$ for general algebraic equations, reserving $y$ for when the specialized case when the algebraic equations are represent either an ODE or a DAE.

linalg/operator.hpp
cjvogl: Can you elaborate?  When solving $M \frac{dy}{dt} = g_1(y,t) + g_2(y,t)$, where $F(u,k,t) = M k$ and $G(u,t) = g_1(u,t) + g_2(u,t)$, wouldn't an additive split method solve either $M \frac{dy}{t} = g_1(y,t)$ or $M \frac{dy}{t} = g_2(y,t)$?

cjvogl:
> I commented above about a few places where the meaning of constants/methods is being changed in a way that is not backward compatible.  There are probably more places like this -- I did not look further. I think we should avoid changing things in a way that is not backward compatible.

@v-dobrev , I agree that this refactor in documentation should be consistent with current derived classes / backwards compatible.  My intention was to have this all be backwards compatible, so either
1. I didn't explain things in a way that the backwards compatibility is clear
2. I don't understand how the non-compatible derived classes interpreted the structure of `TimeDependentOperator`

Do you have a feeling for which it is?  If you think it is 2, can you point out some derived classes that this documentation will not be backwards compatible with?

linalg/operator.hpp
cjvogl: For what it's worth, I had assumed the meaning of `EXPLICIT` vs `IMPLICIT` was equivalent to the PETSc `TSEquationType` documented at https://petsc.org/release/manualpages/TS/TSEquationType/

image

v-dobrev: The idea for the purpose of the `Type` is not well explained: the idea is that classes that use `TimeDependentOperator`s (like time-steppers) can query a given `TimeDependentOperator` about the form of ODE it represents. For example, if a user defines a derived class in which the form `du/dt = f(u,t)` is used, then they should setup their class to return `EXPLICIT` which will also imply that there is no difference between `Mult()` and `ExplicitMult()`, in particular. Of course, it is not wrong for the user class to return `IMPLICIT` (which is more general than `EXPLICIT`) in this case, however, the class that uses such `TimeDependentOperator` may try (if it makes decisions based on the returned type) to treat the ODE in its general (`IMPLICIT`) form and call methods designed for that case like `ImplicitMult()`, `ExplicitMult()`, or `GetImplicitGradient()`.

This is the intended use for the "type". However, I'm not really sure if any of the time steppers actually make use of it. Instead they just call the methods they want, assuming they are implemented and potentially assuming one of the "types" without actually checking. In some cases, not checking is perfectly fine, e.g. if the time stepper uses methods designed to work in all cases like `Mult()` or `ImplicitSolve()`.

v-dobrev: > I have to admit that I didn't quite understand the use of `EXPLICIT` vs `IMPLICIT`... is there an example somewhere that differentiates between the two?
> 
> In solving Mdydt=g(y,t), is the related `TimeDependentOperator` considered `EXPLICIT`?

It depends on the how the user-defined subclass of `TimeDependentOperator` treats the ODE: if `ExplicitMult()` returns `g(y,t)` then the class should return `IMPLICIT`; if `ExplicitMult()` returns `inv(M) g(y,t)` then it can (and typically should) return `EXPLICIT` but it is not wrong to return `IMPLICIT`.

Note that independent of the "type", `Mult()` should always return `inv(M) g(y,t)`, if it is implemented (some time-steppers may just not use `Mult()` so it will be fine to leave it undefined.)

cjvogl: Thank you for clarifying, I think I have a good handle on things now.  

>Note that independent of the "type", Mult() should always return inv(M) g(y,t),

Ah ok, this was where I think I was getting confused: currently `ARKStepSolver` will assume `Mult` returns $g(y,t)$ if there has been a call to `ARKStepSolver::UseMFEMMassSolver`.  With the change I proposed to have `ARKStepSolver` instead use `ExplicitMult` to get $g(y,t)$ when there has been a call to `ARKStepSolver::UseMFEMMassSolver`, this will all be consistent.

I'll push a commit with an iteration of changes.

linalg/operator.hpp
cjvogl: My understanding is that `TimeDependentOperator` can be used for things other than DAEs/ODEs... so I reserved the $u$ for full generality and $y$ when `TimeDependentOperator` represents a DAE/ODE.

linalg/operator.hpp
v-dobrev: What I'm thinking about is the general case where $F(u,k,t)$ is NOT linear in $k$. In that case assuming additive splitting for $f(u,t)$ is not the same as assuming additive splitting for $G(u,t)$. Basically, assuming the relations: $$F(u,f(u,t),t) = G(u,t), \quad F(u,f_1(u,t),t) = g_1(u,t), \quad F(u,f_2(u,t),t) = g_2(u,t)$$ then:

1. if $G(u,t) = g_1(u,t) + g_2(u,t)$ then, in general, $f(u,t) \ne f_1(u,t) + f_2(u,t)$;
2. and vice-versa: if $f(u,t) = f_1(u,t) + f_2(u,t)$, then, in general $G(u,t) \ne g_1(u,t) + g_2(u,t)$.

If this type of additive split is never used when $F(u,k,t)$ is NOT linear in $k$, then we can state that assumption (F is linear in k) for all methods and type-constants related to the additive split -- and then an additive split for $f$ is equivalent to an additive split of `G`, so the documentation changes will be backward-compatible (with the previously-not-mentioned assumption that F is linear in k 😁).

linalg/operator.hpp
v-dobrev: Okay, I see. It's hard to convey that intent in the text without stating it explicitly, at least I didn't get the idea. 😁

However, as I said above, it is a minor thing, so don't feel like you need to change `y` to `u`.

cjvogl: I'll also note I found using `u` as the independent variable that `F` depends on and `y` the solution to the DAE/ODE to be useful in explaining `ImplicitSolve`

linalg/operator.hpp
```diff
@@ -384,62 +391,83 @@ class TimeDependentOperator : public Operator
...
+       2. F(u,k,t) = M k and G(u,t) = g(u,t)
+
+       This function then computes one of the following:
+       1. @a v = inv(M) g(@a u, t)
+       2. @a v = g(@a u, t) */
+   virtual void Mult(const Vector &u, Vector &v) const override;
+
+   /** @brief Solve for unknown @a k at current time t that satisfies
+       F(@a u + @a gamma @a k, @a k, t) = G(@a u + @a gamma @a k, t).
+
```
v-dobrev, line 423: I see you removed some of the existing documentation for `ImplicitSolve()` -- I personally think it is best to keep the previous text and just add more explanations.

linalg/operator.hpp
cjvogl: I'm not clear on what the $f(u,t)$ is.  In the following DAE: $y \frac{dy}{dt} = g_1(y,t) + g_2(y,t)$, what is the $f(u,t)$?

linalg/operator.hpp
```diff
@@ -384,62 +391,83 @@ class TimeDependentOperator : public Operator
...
+       For solving an ordinary differential equation of the form
+       \f$ M \frac{dy}{dt} = g(y,t) \f$, recall F and G are defined as one of
+       the following:
+       1. F(u,k,t) = k and G(u,t) = inv(M) g(u,t)
+       2. F(u,k,t) = M k and G(u,t) = g(u,t)
+
+       This function then computes one of the following:
+       1. @a v = inv(M) g(@a u, t)
+       2. @a v = g(@a u, t) */
```
v-dobrev, line 418: As discussed here: https://github.com/mfem/mfem/pull/4115/files#r1498480826 this text is incorrect -- this function should always return the slope `k`.

cjvogl: I'm working on a revised version right now.

linalg/operator.hpp
v-dobrev: Following the definitions in the class doxygen $f(u,t)$ is the solution to $F(u,f(u,t),t) = G(u,t)$, so in the equation you asked about (I'm not sure why you call it DAE, to me this is just an ODE), $f(y,t)$ (you ask about $f(u,t)$ but I'm not sure if there is a difference) is simply $f(y,t) = (1/y)(g_1(y,t)+g_2(y,t)$, so the ODE is simply $\frac{dy}{dt} = f(y,t)$.

v-dobrev: Note that the documentation was written exclusively with ODEs in mind, so it was assumed that $F(u,f(u,t),t)=G(u,t)$ can be inverted and solved for $f(u,t)$. In the case when that is not true and some equations in $F(u,k,t)=G(u,t)$ are independent of $k$, then proper generalizations must be considered and the documentation updated carefully.

cjvogl: >I'm not sure why you call it DAE, to me this is just an ODE

Yeah good point.  I think I have what I need to take another stab at this.  Stay tuned.

linalg/operator.hpp
cjvogl: I believe that existing documentation for `ImplicitSolve` was causing a lot of confusion (see #3443 ).  I personally found
```
The input vector @a x corresponds to time index (or cycle) n, while the
currently set time, #t, and the result vector @a k correspond to time
index n+1. The time step @a dt corresponds to the time interval between
cycles n and n+1.
```
to be particularly confusing for a DIRK method, where 
1. the stage solves should evaluate the RHS of the ode at t_n + c_i \Delta t (not time index n +1) 
2. the input vector x is the solution a time index n plus a combination of other stage values
3. the dt should correspond to a_ii \Delta t... which is not always the same as c_i \Delta t 

This was the motivation for my replacing of the text above with
```
A DIRK integrator can use @a k from this function, with @a u set to
\f$ y(t) + \Delta t \sum_{j=1}^{i-1} a_{ij} k_j \f$ and @a gamma set to
\f$ a_{ii} \Delta t \f$, for \f$ k_i \f$.
```

linalg/operator.hpp
cjvogl: I just pushed a commit that effectively reverted this line (but using the new notation).

linalg/operator.hpp
cjvogl: I just pushed a commit where I opted to just stick with the original definition with the new notation.

linalg/operator.hpp
cjvogl: This should be corrected now!

linalg/operator.hpp
v-dobrev: The paragraph you found confusing can be fixed by prefixing it by "For, example, in the case of backward Euler ..." -- it is a simple case that is easy to understand and gives some idea how the method works, so I think it is worth keeping with the fix.

Also, the other deleted paragraph expressing the functionality in terms of $f(u,t)$ is also worth having.

v-dobrev: Now that the method parameter is called `gamma`, the paragraph about backward Euler should mention that in this case `gamma` is `dt` -- "the time interval between cycles n and n+1".

cjvogl: > The paragraph you found confusing can be fixed by prefixing it by "For, example, in the case of backward Euler ..." -- it is a simple case that is easy to understand and gives some idea how the method works, so I think it is worth keeping with the fix.

I will add back the case of backward Euler.

>Also, the other deleted paragraph expressing the functionality in terms of $f(u,t)$ is also worth having.

I believe this is captured, with the new notation, in line 430:

```
In either case, this function solves for @a k in M @a k = g(@ a + @a gamma @k, t).
```
What do you think?

v-dobrev: I'm looking at the latest version, I'll comment there.

linalg/operator.hpp
```diff
@@ -309,30 +309,38 @@ class Operator
...
+    _implicit_ and _explicit_ parts of the operator, respectively.
+
+    A common use for this class is representing a differential algebraic
+    equation of the form $ F(y,\frac{dy}{dt},t) = G(y,t) $. Consider an
+    ordinary differential equation of the form $ M \frac{dy}{dt} = g(y,t) $,
+    where F and G are defined as one of the following:
+    1. F(u,k,t) = k and G(u,t) = inv(M) g(u,t)
+    2. F(u,k,t) = M k and G(u,t) = g(u,t)
+    3. F(u,k,t) = M k - g(u,t) and G(u,t) = 0
+    */
```
v-dobrev, line 323:
```suggestion
    equation of the form $ F(y,\frac{dy}{dt},t) = G(y,t) $.

    For example, consider an ordinary differential equation of the form
    $ M \frac{dy}{dt} = g(y,t) $. There are various ways of expressing this ODE
    as a TimeDependentOperator depending on the choices for F and G. Here are
    some common choices:

      1. F(u,k,t) = k and G(u,t) = inv(M) g(u,t),
      2. F(u,k,t) = M k and G(u,t) = g(u,t),
      3. F(u,k,t) = M k - g(u,t) and G(u,t) = 0.

    Note that depending on the ODE solver, some of the above choices may be
    preferable to the others.
*/
```
Some of the text formatting changes are to make the comment more readable in the editor.

linalg/operator.hpp
```diff
@@ -309,30 +309,38 @@ class Operator
...
 class TimeDependentOperator : public Operator
 {
 public:
    enum Type
```
v-dobrev, line 327:
```suggestion
   /// Enum used to describe the form of the time-dependent operator.
   /** The type should be set by classes derived from TimeDependentOperator to
       describe the form, in terms of the functions F and G, used by the
       specific derived class. This information can be queried by classes or
       functions (like time stepping algorithms) to make choices about the
       algorithm to use, or to ensure that the TimeDependentOperator uses the
       form expected by the class/function.

       For example, assume that a derived class is implementing the ODE
       $M \frac{dy}{dt} = g(y,t)$ and chooses to define $F(u,k,t) = M k$ and
       $G(u,t) = g(u,t)$. Then it cannot use type EXPLICIT, unless $M = I$, or
       type HOMOGENEOUS, unless $g(u,t) = 0$. If, on the other hand, the derived
       class chooses to define $F(u,k,t) = k$ and $G(u,t) = M^{-1} g(y,t)$, then
       the natural choice is to set the type to EXPLICIT, even though setting it
       to IMPLICIT is also not wrong -- doing so will simply fail to inform
       methods that query this information that it uses a more specific
       implementation, EXPLICIT, that may allow the use of algorithms that
       support only the EXPLICIT type. */
   enum Type
```

linalg/operator.hpp
```diff
@@ -384,62 +392,87 @@ class TimeDependentOperator : public Operator
...
+
+       For solving an ordinary differential equation of the form
+       $ M \frac{dy}{dt} = g(y,t) $, recall F and G are defined as one of the
+       following:
+       1. F(u,k,t) = k and G(u,t) = inv(M) g(u,t)
+       2. F(u,k,t) = M k and G(u,t) = g(u,t)
+       3. F(u,k,t) = M k - g(u,t) and G(u,t) = 0
+       Regardless of the choice of F and G, this function computes
+       @a k = inv(M) g(@a u, t) */
+   virtual void Mult(const Vector &u, Vector &v) const override;
```
v-dobrev, line 419:
```suggestion
   /** @brief Perform the action of the operator (u,t) -> k(u,t) where t is the
       current time set by SetTime() and @a k satisfies
       F(@a u, @a k, t) = G(@a u, t).

       For solving an ordinary differential equation of the form
       $ M \frac{dy}{dt} = g(y,t) $, recall that F and G can be defined in
       various way, e.g.:

         1. F(u,k,t) = k and G(u,t) = inv(M) g(u,t)
         2. F(u,k,t) = M k and G(u,t) = g(u,t)
         3. F(u,k,t) = M k - g(u,t) and G(u,t) = 0.

       Regardless of the choice of F and G, this function should always compute
       @a k = inv(M) g(@a u, t). */
   virtual void Mult(const Vector &u, Vector &v) const override;
```

linalg/operator.hpp
```diff
@@ -384,62 +392,87 @@ class TimeDependentOperator : public Operator
...
+       $ y(t + \Delta t) = y(t) + \Delta t k_0 $ where
+       $ M k_0 = g \big( y(t) + \Delta t k_0 \big) $. A backward Euler
+       integrator can use @a k from this function, with @a u set to $ y(t) $ and
+       @a gamma set to $ \Delta t$, for $k_0$. Generalizing further, consider a
+       diagonally implicit Runge-Kutta (DIRK) method defined by
+       $ y(t + \Delta t) = y(t) + \Delta t \sum_{i=1}^s b_i k_i $ where
+       $ M k_i = g \big( y(t) + \Delta t \sum_{j=1}^i a_{ij} k_j, t + c_i \Delta t \big) $.
+       A DIRK integrator can use @a k from this function, with @a u set to
+       $ y(t) + \Delta t \sum_{j=1}^{i-1} a_{ij} k_j $ and @a gamma set to
+       $ a_{ii} \Delta t $, for $ k_i $.
```
v-dobrev, line 442:
```suggestion
   /** @brief Solve for the unknown @a k, at the current time t, the following
       equation:
       F(@a u + @a gamma @a k, @a k, t) = G(@a u + @a gamma @a k, t).

       For solving an ordinary differential equation of the form
       $ M \frac{dy}{dt} = g(y,t) $, recall that F and G can be defined in
       various ways, e.g.:

         1. F(u,k,t) = k and G(u,t) = inv(M) g(u,t)
         2. F(u,k,t) = M k and G(u,t) = g(u,t)
         3. F(u,k,t) = M k - g(u,t)

       Regardless of the choice of F and G, this function should solve for @a k
       in M @a k = g(@a u + @a gamma @a k, t).

       To see how @a k can be useful, consider the backward Euler method defined
       by $ y(t + \Delta t) = y(t) + \Delta t k_0 $ where
       $ M k_0 = g \big( y(t) + \Delta t k_0, t + \Delta t \big) $. A backward
       Euler integrator can use @a k from this function for $k_0$, with the call
       using @a u set to $ y(t) $, @a gamma set to $ \Delta t$, and time set to
       $t + \Delta t$. See class BackwardEulerSolver.

       Generalizing further, consider a diagonally implicit Runge-Kutta (DIRK)
       method defined by
       $ y(t + \Delta t) = y(t) + \Delta t \sum_{i=1}^s b_i k_i $ where
       $ M k_i = g \big( y(t) + \Delta t \sum_{j=1}^i a_{ij} k_j,
                         t + c_i \Delta t \big) $.
       A DIRK integrator can use @a k from this function, with @a u set to
       $ y(t) + \Delta t \sum_{j=1}^{i-1} a_{ij} k_j $ and @a gamma set to
       $ a_{ii} \Delta t $, for $ k_i $. For example, see class SDIRK33Solver.
```

linalg/operator.hpp
```diff
@@ -384,62 +392,87 @@ class TimeDependentOperator : public Operator
...
+       $ M \frac{dy}{dt} = g(y,t) $, recall F and G are defined as one of
+       the following:
+       1. F(u,k,t) = k and G(u,t) = inv(M) g(u,t)
+       2. F(u,k,t) = M k and G(u,t) = g(u,t)
+       3. F(u,k,t) = M k - g(u,t) and G(u,t) = 0
+       This function performs setup to solve $ A x = b $ where A is either
+       1. A(@a y,t) = I - @a gamma inv(M) J(@a y,t)
+       2. A(@a y,t) = M - @a gamma J(@a y,t)
+       3. A(@a y,t) = M - @a gamma J(@a y,t)
+       with J = dg/dy (or a reasonable approximation thereof).
```
v-dobrev, line 472: For correct Doxygen formatting and improved editor formatting, I suggest:
```suggestion
   /** @brief Setup a linear system as needed by some SUNDIALS ODE solvers.

       For solving an ordinary differential equation of the form
       $ M \frac{dy}{dt} = g(y,t) $, recall that F and G can be defined as one
       of the following:

         1. F(u,k,t) = k and G(u,t) = inv(M) g(u,t)
         2. F(u,k,t) = M k and G(u,t) = g(u,t)
         3. F(u,k,t) = M k - g(u,t) and G(u,t) = 0

       This function performs setup to solve $ A x = b $ where A is either

         1. A(@a y,t) = I - @a gamma inv(M) J(@a y,t)
         2. A(@a y,t) = M - @a gamma J(@a y,t)
         3. A(@a y,t) = M - @a gamma J(@a y,t)

       with J = dg/dy (or a reasonable approximation thereof).
```

v-dobrev:
One last suggestion: extend the comment here: https://github.com/mfem/mfem/blob/70c6f713d5a6f6790320c072b82b46b469efb4b9/linalg/operator.hpp#L349

to include a ref to Type with its documentation:
```c++
   Type type; /**< @brief Describes the form of the TimeDependentOperator, see
                   the documentation of #Type. */
```

v-dobrev (approved):

linalg/operator.hpp
cjvogl: I believe this is resolved, please feel free to re-open this thread if it is not.

linalg/operator.hpp
cjvogl: I believe this is resolved, please feel free to re-open this thread if it is not.

linalg/operator.hpp
cjvogl: I believe this is resolved, please feel free to re-open this thread if it is not.

linalg/operator.hpp
cjvogl: Change is made: `ARKStepSolver` uses `ExplicitMult` when either a `UseMFEMMassLinearSolver` or `UseSundialsMassLinearSolver` is called

cjvogl:
@acfisher, I believe all suggestions have been applied & concerns addressed.  If am I missing any, I am happy to quickly tackle them.

v-dobrev:
This may be ready to go. @cjvogl recommends to merge this before #4049.

tzanio:

tzanio:
Thanks @cjvogl !

PR #4127, opened 2024-02-09
Fix ParaView/VTU output for prism meshes

sebastiangrimberg:
Fixes a bug for prism mesh VTU output with `high_order_output = false` and `ref > 1`.

NOTE: This is still a WIP. There are still issues with high order output of prism elements. See below for a clip of a prism mesh in ParaView with `ref = 4` for a 4th order solution. The left is using `high_order_output = false`, using the `PRISM` type, where the connectivity of the nodes in each prism looks correct, as does the solution. The right is with `high_order_output = true` which uses the `LAGRANGE_PRISM` type, and is incorrect. I suppose maybe it could be a bug with ParaView itself or a strange visualization artifact, though I'm not sure. This is using ParaView 5.11.2. Tetrahedral and hexahedral meshes in both cases look fine.

Tagging @pazner @jonesholger in case either of you has suggestions here.


tzanio:
try to extend to the high-order case?

pazner:
This fix and the fix for the high-order case (which actually came from a bug in the VTK library) is now in #4797, so I think that PR can supersede this one, and we can close this PR.

tzanio:
Closing in favor of https://github.com/mfem/mfem/pull/4797

PR #4175, opened 2024-03-05
Device HO-LOR transfer

artv3:
- [x] Add 3D version
- [x] Further optimize with vendor linear algebra libraries
- [ ] Clean up / organize


miniapps/tools/lor-transfer.cpp
```diff
@@ -167,6 +172,10 @@ int main(int argc, char *argv[])
    {
       gt = new L2ProjectionGridTransfer(fespace, fespace_lor);
    }
+
+   gt->UseDevice(true);
```
artv3, line 176: Use mfem's device configuration instead of having a new method

linalg/densemat.hpp
```diff
@@ -1283,6 +1283,22 @@ void BatchLUFactor(DenseTensor &Mlu, Array<int> &P, const double TOL = 0.0);
...
+
+void BatchInverseMatrix(const Vector &LU,
+                        const int m,
+                        const int NE,
+                        const Array<int> &P,
+                        Vector &INV);
+
+void BatchInverseMatrix(const DenseTensor &LU,
+                        const Array<int> &P,
+                        DenseTensor &INV);
```
tomstitt, line 1300: It looks like there are existing `BatchLUFactor` and BatchLUSolve` methods, can we combine those and these?

linalg/densemat.cpp
```diff
@@ -4392,4 +4393,268 @@ void BatchLUSolve(const DenseTensor &Mlu, const Array<int> &P, Vector &X)
...
+            {
+               //Swap<double>(X[i+k*m], X[i+piv_k*m]);
+               mfem::kernels::internal::Swap<double>(X[i + k * m], X[i + piv_k * m]);
+            }
+         }
+      }
+   });
+}
+
+void BatchInverseMatrix(const DenseTensor &LU,
```
tomstitt, line 4578: I think the implementation of this is the same as the one above besides getting `m` and `NE` from the tensor, can you implement one in terms of the other to avoid duplicate code?

linalg/densemat.hpp
tomstitt: For generality please consider using `len`/`length` instead of `NE`

linalg/batchlinalg.cpp
```diff
@@ -0,0 +1,391 @@
...
+   {
+      BatchInverseMatrix(LUMatrixBatch_, P_, InvMatBatch);
+   }
+}
+
+void LibBatchSolver::SolveLU(const mfem::Vector &b, mfem::Vector &x) const
+{
+
+   x = b;
+#if defined(MFEM_USE_CUDA)
```
tomstitt, line 239:
```suggestion
#if defined(MFEM_USE_CUDA_OR_HIP)
```

linalg/densemat.cpp
artv3: yes, one thought is we just need to make sure wrap the memory object in the mfem object. There are just in consistencies in using vector and densetensor types to store batches of matrices.

tomstitt: you could have a single version that takes the sizes along with c-arrays and just have both call that?

linalg/densemat.cpp
```diff
@@ -2224,6 +2224,7 @@ void DenseMatrix::Print(std::ostream &os, int width_) const
    ios::fmtflags old_flags = os.flags();
    // output flags = scientific + show sign
    os << setiosflags(ios::scientific | ios::showpos);
+   //os << setiosflags(ios::scientific | ios::showpos);
```
tomstitt, line 2227:
```suggestion
```

linalg/batchlinalg.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include "../config/config.hpp"
+#include "../general/globals.hpp"
+#include "matrix.hpp"
+#include "densemat.hpp"
+
+namespace mfem
+{
+
+class LibBatchSolver
```
tomstitt, line 23: Just call it `BatchSolver`?

tomstitt: You probably want to document what this class does

linalg/batchlinalg.hpp
```diff
@@ -0,0 +1,83 @@
...
+   LibBatchSolver() = delete;
+
+   LibBatchSolver(const SolveMode mode);
+
+   LibBatchSolver(const mfem::DenseTensor &MatrixBatch, const SolveMode mode);
+
+   void AssignMatrices(const mfem::DenseTensor &MatrixBatch);
+
+   void AssignMatrices(const mfem::Vector &vMatrixBatch, const int ndofs,
+                       const int num_matrices);
```
tomstitt, line 70: Perhaps `size` instead of `ndofs`?

linalg/batchlinalg.cpp
```diff
@@ -0,0 +1,391 @@
...
+      default: mfem_error("Case not supported");
+   }
+}
+
+void LibBatchSolver::ReleaseMemory()
+{
+   MFEM_RELEASE_MEM(LUMatrixBatch_);
+   MFEM_RELEASE_MEM(InvMatrixBatch_);
+   MFEM_RELEASE_MEM(P_);
+   MFEM_RELEASE_MEM(lu_ptr_array_);
```
tomstitt, line 387: I vote to remove the macro and just call `GetMemory().DeleteDevice(false)`, we don't know how others will want to use this. Please also change the name to `ReleaseDeviceMemory`.

linalg/batchlinalg.cpp
```diff
@@ -0,0 +1,391 @@
...
+   if (lu_valid_)
+   {
+      return;
+   }
+
+#if defined(MFEM_USE_CUDA_OR_HIP)
+   if (Device::Allows(Backend::DEVICE_MASK))
+   {
+
+      mfem::Array<int> info_array(num_matrices_); // need to move to temp mem
```
tomstitt, line 163: Can you add an optional `MemoryType` to the constructor and use that throughout? You can default to `MemoryType::DEFAULT`

linalg/batchlinalg.cpp
```diff
@@ -0,0 +1,391 @@
...
+#include <hip/hip_runtime_api.h>
+#define MFEM_cu_or_hip(stub) hip##stub
+#define MFEM_Cu_or_Hip(stub) Hip##stub
+#define MFEM_CU_or_HIP(stub) HIP##stub
+#define MFEM_CUDA_or_HIP(stub) HIP##stub
+#else
+#define MFEM_cu_or_hip(stub)
+#define MFEM_Cu_or_Hip(stub)
+#define MFEM_CU_or_HIP(stub)
+#define MFEM_CUDA_or_HIP(stub)
```
tomstitt, line 43: we have these defined in a few places now, do you want to move them to a shared header? Or should we do that in a followup ?

artv3: I think now would be good to do

linalg/batchlinalg.cpp
artv3: yes

linalg/batchlinalg.cpp
```diff
@@ -0,0 +1,391 @@
...
+      auto status = MFEM_cu_or_hip(blasCreate)(&handle);
+      MFEM_VERIFY(status == MFEM_CU_or_HIP(BLAS_STATUS_SUCCESS),
+                  "Cannot initialize GPU BLAS");
+      atexit([]()
+      {
+         MFEM_cu_or_hip(blasDestroy)(handle);
+         handle = nullptr;
+      });
+   }
+   return handle;
```
tomstitt, line 67: Let's change this to let a library user set the blas handle:

```suggestion
static MFEM_cu_or_hip(blasHandle_t) device_blas_handle = nullptr;

const MFEM_cu_or_hip(blasHandle_t) & DeviceBlasHandle()
{
   if (!device_blas_handle)
   {
      auto status = MFEM_cu_or_hip(blasCreate)(&device_blas_handle);
      MFEM_VERIFY(status == MFEM_CU_or_HIP(BLAS_STATUS_SUCCESS),
                  "Cannot initialize GPU BLAS");
      atexit([]()
      {
         MFEM_cu_or_hip(blasDestroy)(device_blas_handle);
         device_blas_handle = nullptr;
      });
   }
   return device_blas_handle;
```

Then just put an extern definition in header!

linalg/batchlinalg.cpp
artv3: The name of the method? i.e. `DeleteDevice -> ReleaseDeviceMemory'?

tomstitt: > The name of the method? i.e. `DeleteDevice -> ReleaseDeviceMemory'?

Yes please !

fem/transfer.cpp
```diff
@@ -401,11 +453,471 @@ L2ProjectionGridTransfer::L2ProjectionL2Space::L2ProjectionL2Space(
...
+            }
+
+         }
+      });
+   } // end of mixed assembly mass matrix
+
+
+   //R = inv(M_L) * M_mixed
+   //Need to compute M_L
+   MassIntegrator mi;
```
chrissytseng, line 751: add qfunctionCoefficient here

fem/transfer.cpp
```diff
@@ -401,11 +453,471 @@ L2ProjectionGridTransfer::L2ProjectionL2Space::L2ProjectionL2Space(
...
+            mfem::forall(nel_ho, [=] MFEM_HOST_DEVICE (int iho)
+            {
+               for (int iref = 0; iref < nref; ++iref)
+               {
+                  const int lo_el_id = iref + nref*iho;
+                  for (int qx=0; qx<Q1D; ++qx)
+                  {
+
+                     const real_t detJ = J(qx, lo_el_id);
+                     d_D(qx, iref, iho) = W(qx) * detJ;
```
chrissytseng, line 590: add coefficient/weight in quadrature here

fem/transfer.cpp
chrissytseng, line 751: add qfunctionCoefficient here with existing MassIntegrator device

PR #4208, opened 2024-03-22
Exodus II Writer

EdwardPalmer99:
**Brief**
- This PR adds ExodusII mesh outputs.
- The writer can handle multiple element types (one type per element block).
- Added writer unit tests (some are commented-out until the reader is updated).
- The following element types are supported:

**First-Order Element Types**
- Hex8
- Tet4
- Wedge6
- Pyramid5

**Second-Order Element Types**
- Hex27
- Tet10
- Wedge18
- Pyramid14

**Caveats**
- Cannot mix first- and second-order element types
- Wedge18/Pyramid14 are not supported by the current ExodusII reader
- Limited file options (will extend in future)
- Issues with Tet node ordering if refine=1 (set to zero to avoid issues)
- No support for multiple timesteps
- Assumptions about element order are made

**Future Work**
- Once #4178 is merged, the commented-out mixed mesh and wedge/pyramid tests should be uncommented. These have been tested.
- It would make sense to move the ReadCubit code into the same file in a future PR and create a reader class similar to the ExodusIIWriter.
- Add option to write-out coordinates in blob format (xyz, xyz, xyz) rather than large file format (xxx, yyy, zzz). The reader should be updated to handle reading-in files written using the blob format.

| [#4208](https://github.com/mfem/mfem/pull/4208) | @EdwardPalmer99 | @tzanio | @mlstowell + @bslazarov + @jandrej | 4/7/24 | 6/11/24 | 6/14/24 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:
Thanks @EdwardPalmer99!

tzanio:

mlstowell:
Hi, @EdwardPalmer99,
Thanks for submitting this. It will be a good addition to MFEM. I have a few notes that could make this more MFEM-like. I'll comment in the files.
Best wishes,
Mark

mesh/mesh.hpp
```diff
@@ -298,6 +298,11 @@ class Mesh
    // (true) is set in mesh_readers.cpp.
    static bool remove_unused_vertices;
 
+#ifdef MFEM_USE_NETCDF
```
mlstowell, line 301: It would be better to group this much further down in the file near the `PrintXG`, `PrintVTK`, and similar member functions. The precise location is not important but somewhere within the Doxygen group `Print/Save/Export methods` would be appropriate.

Also, the name `PrintExodusII` would follow the present pattern.

mesh/exodus_writer.cpp
```diff
@@ -0,0 +1,1000 @@
...
+   }\
+}
+
+namespace mfem
+{
+
+#ifdef MFEM_USE_NETCDF
+
+// Variable labels
+const char * EXODUS_TITLE_LABEL = "title";
```
mlstowell, line 35: We are reluctant to place variables directly into the `mfem` namespace. Even when they have such specific names as these. I think we would most often define these as `static` variables within something like the `ExodusIIWriter` class. If you prefer they could probably be placed within a separate utility class. Other possibilities exist. If you have a strong inclination towards a particular solution we can certainly discuss options.

tests/unit/mesh/test_exodus_writer.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+TEST_CASE("ExodusII Write Hex8", "[Mesh]")
+{
+#ifdef MFEM_USE_NETCDF
+   // Load Exodus II mesh from file.
+   std::string fpath_original = "data/simple-cube-hex8.e";
+   Mesh original_mesh = Mesh::LoadFromFile(fpath_original, 0, 0, true);
+
+   // Write generated Exodus II mesh to file.
+   std::string fpath_generated = "data/simple-cube-hex8-out.e";
```
mlstowell, line 104: Two small points.

First, we don't typically write in the `data` directory. This is a minor point but it may be better to place these temporary output files in the current directory.

Second, the test should remove its temporary files to avoid clutter and false positives in future tests. For example, this is done in a similar unit test using `remove` as seen here: https://github.com/mfem/mfem/blob/41a40ebf57e982c54eaaaab514b487c4182ae195/tests/unit/fem/test_datacollection.cpp#L137

tests/unit/data/simple-cube-pyramid14.e
pazner, line 1: Perhaps these mesh files (which add up to about 0.5 MB) should be moved to the [mfem/data](https://github.com/mfem/data) repository. The unit tests can then be tagged with the tag `[MFEMData]`, so that they will only be run if the unit test executable is passed the `--data` command line argument.

mesh/exodus_writer.cpp
EdwardPalmer99: Okay, sure. How about plonking them into an "ExodusIILabels" namespace? I think this is a better solution than putting them into a utility class.

mlstowell: As long as that namespace is inside the `mfem` namespace I think that would be fine.

mesh/exodus_writer.cpp
```diff
@@ -0,0 +1,1018 @@
...
+
+protected:
+   /// @brief Closes any open file and creates a NetCDF file using selected flags.
+   void OpenExodusII(std::string fpath, int flags);
+
+   /// @brief Closes any open file.
+   void CloseExodusII();
+
+   /// @brief Generates blocks based on the elements in the mesh. We iterate
+   /// over the mehs elements and use the attributes as the element blocks. We
```
mlstowell, line 148: typo: `mehs`

mesh/exodus_writer.cpp
```diff
@@ -0,0 +1,1018 @@
...
+      else
+      {
+         auto & block_element_ids = _element_ids_for_block_id.at(block_id);
+         block_element_ids.push_back(ielement);
+
+         // Safety check: ensure that the element type matches what we have on record
+         // for the block.
+         if (element_type != _element_type_for_block_id.at(block_id))
+         {
+            MFEM_ABORT("Multiple element types are defined for block: " << block_id);
```
mlstowell, line 869: I suggest we change this so that the attempt to write an output file stops but program execution can continue. We could print a warning message stating either that mixed meshes are not yet supported or that Exodus II meshes do not support mixed element types. Whichever is more accurate.

While we're on this topic: if mixed meshes are supported but mixed blocks are not then I would like to see an outline of how we might improve such support in the future. Might it be as simple as pre-processing the mesh and dividing regions with a common attribute number into multiple blocks based on element type?

mesh/exodus_writer.cpp
```diff
@@ -0,0 +1,1018 @@
...
+   DefineDimension(label, front_element->GetNFaces(),
+                   &num_fac_per_el_id);
+
+   // 5. Define element node connectivity for block.
+   WriteNodeConnectivityForBlock(block_id);
+
+   // 6. Define the element type.
+   std::string element_type;
+
+   bool higher_order = (_mesh.GetNodes() != nullptr);
```
mlstowell, line 549: We need to be a little more discerning here. At the very least we should confirm that the `Nodes` defined in the mesh do indeed correspond to a 2nd order H1 space (because that would seem to be the assumption made later on). We can then exit gracefully with a warning message if printing this particular type of mesh is not supported.

In the future we can perhaps address issues such as `Nodes` based on L2 spaces. We could even project higher order `Nodes` to 2nd order nodes as long as we print a warning message. These improvements do not need to be handled at this time but we should be clear about the limitations of the current implementation.

mesh/exodus_writer.cpp
```diff
@@ -0,0 +1,1018 @@
...
+      }
+   }
+
+   for (int ibdr_element = 0; ibdr_element < _mesh.GetNBE(); ibdr_element++)
+   {
+      int boundary_id = _mesh.GetBdrAttribute(ibdr_element);
+      int bdr_element_face_index = _mesh.GetBdrElementFaceIndex(ibdr_element);
+
+      // Locate match.
+      auto & element_face_info = mfem_face_index_info_for_global_face_index.at(
```
mlstowell, line 979: I'm not certain but I believe this call to `at` may be failing on interior surfaces such as those found in `miniapps/multidomain/multidomain-hex.mesh`. Some would argue that "boudary" faces should never exist within the interior of a mesh but MFEM meshes do support this. Unfortunately this means that this writer must determine how best to handle such things. Does the Exodus II format support interior "boundary" faces? Does it store such things in a different manner? At the very least the current implementation of this writer should recognize these "interior" boundary faces and skip them if necessary (and issue a warning message to that effect).

mesh/exodus_writer.cpp
EdwardPalmer99: Mixed meshes are supported. I've added two tests to check mixed first- and second-order meshes (although these are commented-out because they are dependent on  #4178 being merged (for pyramid and wedges). An ExodusII limitation is that only a single element type is supported per block. You can have multiple blocks defined each with a different element type but not a single block with multiple types. It is an error and that's why I'm using an abort. The error message could be clearer -- perhaps "Detected multiple element types on an element block. Only a single element type can be defined per block".

Yes, that would work. Provided that the ExodusII file we generate has a single element type per block we should be okay. We'd just have to be careful with propagating the new attribute numbers.

mesh/exodus_writer.cpp
EdwardPalmer99: Agreed. I've added a method to check the Nodal FESpace. It calls an abort if the space is not of the correct type. A warning would definitely be better although it will require more major changes to the writer class to do this elegantly.

mesh/exodus_writer.cpp
EdwardPalmer99: Right, thanks for spotting this. I've had a quick scan through the ExodusII file format specification and it is not entirely clear. The best solution for now would be to identify the interior boundary faces and skip them. I'll look into doing this.

EdwardPalmer99: We could either call _RemoveInternalBoundaries_ to remove these or probably better is to check the boundary by calling _FaceIsInterior_. This would avoid modifying the mesh we are writing to the ExodusII file.

EdwardPalmer99: Okay, I've pushed a commit which addresses the issue with miniapps/multidomain/multidomain-hex.mesh. It calls FaceIsInterior to check whether a boundary element is internal and if it is then we print a warning and skip it.

tests/unit/data/simple-cube-pyramid14.e
pazner: I see that you moved the mesh files to the `data` subdirectory in 4f27e913ca6f0c9b59932280e7c7569128c1988e. What I was suggesting is that you delete these mesh files entirely from this PR and move them to the data repository.

That way (if this PR is squash-merged or if we edit the commit history) we can avoid having these relatively large files in the MFEM repo.

pazner: Thanks @EdwardPalmer99.

@tzanio, when merging this PR we can squash-merge or just delete these files from the history (I can do this e.g. with [this tool](https://rtyley.github.io/bfg-repo-cleaner/))

mesh/exodus_writer.cpp
jandrej: They are marked by and integer id, yes.

mesh/exodus_writer.cpp
```diff
@@ -0,0 +1,1075 @@
...
+   /// @brief Writes the number of timesteps. Currently, we do not support
+   /// multiple timesteps.
+   void WriteTimesteps();
+
+   /// @brief Writes a dummy variable. This is to circumvent a bug in LibMesh where
+   /// it will skip the x-coordinate when reading in an ExodusII file if the id of
+   /// the x-coordinates is 0. To prevent this, we define a dummy variable before
+   /// defining the coordinates. This ensures that the coordinate variable IDs have
+   /// values greater than zero.
+   void WriteDummyVariable();
```
jandrej, line 237: This can't be fixed in LibMesh?

mesh/exodus_writer.cpp
```diff
@@ -0,0 +1,1075 @@
...
+
+   int num_el_in_blk_id;
+   DefineDimension(label, block_element_ids.size(),
+                   &num_el_in_blk_id);
+
+   // 2. Define number of nodes per element.
+   label = GenerateLabel("num_nod_per_el%d", block_id);
+
+   int num_node_per_el_id;
+   if (_mesh.GetNodes())
```
jandrej, line 517: Please no underscore prefixed variables due to C/CPP standard reserved scoped.

mesh/exodus_writer.cpp
EdwardPalmer99: It's an issue for libMesh users without HDF5 caused by a rather subtle bug in an older version of their exodus reader.

[#libMesh/3283](https://github.com/libMesh/libmesh/issues/3823)

tzanio:
Note: we may want to squash-merge this PR

mesh/exodus_writer.cpp
jandrej: Thanks for the context. We can leave it in for now but should track if there's a fix in libmesh and remove it. I'd reference the github issue https://github.com/libMesh/libmesh/issues/3823 in the comment.

jandrej (approved):

tzanio:

PR #4282, opened 2024-05-03
Fix for #4280

tzanio:
Resolve https://github.com/mfem/mfem/issues/4280

### DIFF
--- a/fem/datacollection.cpp
+++ b/fem/datacollection.cpp
@@ -429,14 +429,15 @@ void VisItDataCollection::RegisterField(const std::string& name,
    }
 
    DataCollection::RegisterField(name, gf);
-   if (gf->FESpace()->FEColl()->GetContType() == FiniteElementCollection::CONTINUOUS)
+   if (gf->FESpace()->FEColl()->GetContType() ==
+       FiniteElementCollection::CONTINUOUS)
    {
       // H1-continuous space
       field_info_map[name] = VisItFieldInfo("nodes", gf->VectorDim(), LOD);
    }
    else
    {
-      // any other space with partial continuity L2, H(curl), H(div), etc.
+      // any other space with partial continuity: L2, H(curl), H(div), etc.
       field_info_map[name] = VisItFieldInfo("elements", gf->VectorDim(), LOD);
    }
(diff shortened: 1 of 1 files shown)


sethwatts:
This LGTM, @tzanio.

v-dobrev:
I don't like the idea of changing the output format to accommodate a change in the reader, especially if the change is not backward compatible.

I think the better solution is to fix this in the reader. I can do that among other fixes in the reader I'm planning to do (@cyrush is aware of my plans) -- I just have not had time to work on them recently. @cyrush, if I start this work soon, what branch should I use as the starting point? I plan to propose my changes through a fork -- as we've discussed it in the past.

tzanio:
@sethwatts, we changed this PR after discussion with @cyrush.

Any chance you can try it in your application? We expect that it should resolve your issue and still work in older versions of VisIt, but we would like to confirm before including it in `mfem-4.7` (to be released today)

sethwatts:
@tzanio, I won't be able to rebuild my application with the mfem 4.7 branch today since I have other dependencies which themselves depend on mfem, and  I won't have time today to get everything updated correctly. However, I _can_ confirm that changing the assoc field from "nodes" to "elements" in the `*.mfem_root` files, with no other changes, does indeed fix the visualization issue in visit 3.4.1. Hopefully that is sufficient to include this change to the 4.7 release.

tzanio:
Thanks @sethwatts that's exactly what we need.

One follow-up: does "elements" works also in VisIt 3.3?

sethwatts:
No, "elements" does _not_ work when rendered with visit-3.3; instead, it throws an error message and does not render anything. As I understood the discussion on visit issue [19469](https://github.com/visit-dav/visit/issues/19469), the change in expectation for "elements" is new in visit-3.4.

cyrush:
Bummer. Let’s reevaluate then.
________________________________

tzanio:
Since there is no backward-compatible solution, and we don't want to break the support for VisIt 3.3, we will address this by updates in the VisIt plugin.

@sethwatts, we recommend that you either use VisIt 3.3 for now, or add a post-processing step to change "nodes" to "elements" in your root files. I'm happy to explain the rationale for this in more detail if you are interested.

PR #4307, opened 2024-05-15
[WIP] Enhance VisIt Data Collection to write out field basis names [task/JustinPrivitera/05_14_24/visit_data_coll_write_out_basis_name]

JustinPrivitera:
Pull request for comments about the integration of task/JustinPrivitera/05_14_24/visit_data_coll_write_out_basis_name into master.
 - Add `basis` string to the `VisItFieldInfo` class.
 - Add logic to handle `basis` string where appropriate.

These changes should bring the VisIt Data Collection more on par with the Conduit Data Collection. Once this feature is added, I will make changes within the MFEM reader inside VisIt to look for the basis name and use it to make appropriate choices about how to represent MFEM data, which will address this issue: https://github.com/visit-dav/visit/issues/19469.

TODO
 - [ ] understand what is the best path for quadrature elements. What should be provided as the basis name? An empty string? They are already element associated, which leads me to think that they have never worked with VisIt. If that's true, that sounds like a separate issue. What should we do in the meantime?
 - [ ] how can I test these changes?

### DIFF
--- a/fem/datacollection.cpp
+++ b/fem/datacollection.cpp
@@ -429,7 +429,10 @@ void VisItDataCollection::RegisterField(const std::string& name,
    }
 
    DataCollection::RegisterField(name, gf);
-   field_info_map[name] = VisItFieldInfo("nodes", gf->VectorDim(), LOD);
+   field_info_map[name] = VisItFieldInfo("nodes", 
+                                         gf->FESpace()->FEColl()->Name(),
+                                         gf->VectorDim(),
+                                         LOD);
    visit_levels_of_detail = std::max(visit_levels_of_detail, LOD);
 }
 
@@ -448,7 +451,8 @@ void VisItDataCollection::RegisterQField(const std::string& name,
    }
 
    DataCollection::RegisterQField(name, qf);
-   field_info_map[name] = VisItFieldInfo("elements", 1, LOD);
+   // TODO ask mfem folks what the right path forward is for quadrature functions
+   field_info_map[name] = VisItFieldInfo("elements", "", 1, LOD);
    visit_levels_of_detail = std::max(visit_levels_of_detail, LOD);
 }
 
@@ -673,6 +677,7 @@ std::string VisItDataCollection::GetVisItRootString()
         it != field_info_map.end(); ++it)
    {
       ftags["assoc"] = picojson::value((it->second).association);
+      ftags["basis"] = picojson::value((it->second).basis);
       ftags["comps"] = picojson::value(to_string((it->second).num_components));
       ftags["lod"] = picojson::value(to_string((it->second).lod));
       field["path"] = picojson::value(path_str + it->first + file_ext_format);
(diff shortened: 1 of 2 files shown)


jandrej:
I'm not sure if QuadratureFunctions worked with VisIt before. They definitely do with ParaView, where they are just displayed as a "node" or "point" in space associated with values. They don't require interpolation to visualize as they're point-discontinuous and not element-discontinuous.

JustinPrivitera:
@jandrej I know that quadrature functions do not work with VisIt. However, both quadrature functions and grid functions use `VisItFieldInfo`. If I add basis name to that class, I need to handle the grid function case and the quadrature function case. What is the right thing to provide for quadrature functions for basis name? An empty string?

tzanio:
Maybe instead of the basis name, we should consider saving `gf->FESpace()->FEColl()->GetContType()` for a given grid function `gf`. The [possible values](https://docs.mfem.org/html/classmfem_1_1FiniteElementCollection.html#a174fa5ed3d062dd7c92e9df3df9a0721a346183a0c130fc8c0dfc9de300277698) for this are:

- `FiniteElementCollection::CONTINUOUS` - e.g. this is what you get for $H^1$
- `FiniteElementCollection::TANGENTIAL` - e.g. this is what you get for $H(curl)$
- `FiniteElementCollection::NORMAL` - e.g. this is what you get for $H(div)$
- `FiniteElementCollection::DISCONTINUOUS` - e.g. this is what you get for $L^2$

Based on these values, in the VisIt plugin then we can:

- Plot nodal field on a LOR mesh for `CONTINUOUS`
- Plot nodal field on a disjoint mesh refined with `GeometryRefiner` for all the other options
- (Optionally) Plot an element field on a LOR mesh for `DISCONTINUOUS` order 0 (i.e. p.w. const grid function)

JustinPrivitera:
@tzanio this was my thought as well. However, I discussed with @cyrush, and he felt it would be better to bring the VisIt outputs in closer parity with the Conduit outputs, which do not handle the quadrature function case and _do_ include the basis name. We can wait for him to chime in if you'd like, or organize a time to meet and discuss.

tzanio:
Regarding `QuadratureFunction`, this has been discussed previously, but there are several options:

1. plot colored bubbles at each point
2. interpret the values as dofs of L2 grid function at the points and plot that
3. plot p.w. const values on LOR mesh where each LOR element contains exactly one of the points

My personal preference is for option 3

tzanio:
Check also the code-style [CI failure](https://github.com/mfem/mfem/actions/runs/9099837569/job/25013318760?pr=4307)

JustinPrivitera:
@cyrush can we let this be automatically closed since you have solved this in VisIt? Or do we still need this?

cyrush:
We will still need some logic in mfem proper. 
The root file needs to identify the quadrature function case as a unique case from nodes or elements.

cyrush:
@JustinPrivitera  this can be closed in lieu of #5097  - we can continue discussion in the PR.

PR #4123, opened 2024-02-08
D(P)G elements support in MixedBilinearForm

najlkin:
This PR introduces support for D(P)G elements in mixed bilinear forms, i.e. discontinuous finite elements in forms with different trial and test spaces. It should be stressed that this is something different from the current trace integrators in mixed forms, which assume one of the spaces to be continuous, giving a trace face element directly. In particular, there are these changes:

- New interior and boundary face integration in `MixedBilinearForm`.
- Extension of `DGTraceIntegrator` for the new mixed DGs, including integral elements following #4088 .
- ~~New mixed elements integration in non-linear forms.~~
- Methods for assembly of the face matrices for the mixed forms (`Compute(Bdr)FaceElementMatrix()`), including also the `const` qualifier following #4082 .
- Methods for retrieval of the mixed face integrators from `MixedBilinearForm` (`Get(B)FBFI()`)

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#4123](https://github.com/mfem/mfem/pull/4123) | @najlkin | @tzanio | @sohailreddy + @psocratis + @tzanio | 5/11/24 | 10/4/24 | 10/14/24 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
[...]

PR #4082, opened 2024-01-24
Boundary/face integration in (Mixed)BilinearForm and Hybridization

najlkin:
This pull request collects enhancements related to integration on boundaries or faces (from [najlkin/mfem](https://github.com/najlkin/mfem/)). Most of them built on the existing things and only complete/improve them, but a new feature is the boundary integration for `Hybridization`. In particular, there are the following things:

- Boundary face constraint integrators in `Hybridization`, which are useful for constraint boundary conditions.
- New `VectorFEBoundaryFluxIntegrator` and `VectorFEBoundaryNormalLFIntegrator` classes for integration of RT elements on the boundaries. These are linear form variants of the corresponding bilinear integrators, which are useful for boundary conditions mainly.
- `Compute(Bdr)TraceFaceElementMatrix` methods in `MixedBilinearForm` for convenience following the `BilinearForm`.
- `Compute(Bdr)FaceElementMatrix` in `BilinearForm` which were missing.
- The `const` qualifier for all `Compute*ElementMatrix` methods of `(Mixed)BilinearForm`.

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#4082](https://github.com/mfem/mfem/pull/4082) | @najlkin | @tzanio | @cjvogl + @mlstowell | 5/11/24 | 7/16/24 | 7/30/24 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
[...]

Issue #91, opened 2016-10-11
Solving HDG matrix


rhebergens:
Hi,

We have a working version of HDG for a Poisson equation and are able to get super-convergence results as expected from literature. We are having some issues regarding solving the system in a smart way. The HDG discretization results in the following block-matrix form:

[A B][ u ] = [ f ]
[C D][ q ]    [ g ]

The nice thing about HDG is that A is a block-diagonal matrix so that it should be fairly cheap to compute its inverse. We can then eliminate u so find a system only for q.

Our problem though, is that A has contributions from elements and faces. This is why we cannot directly use something like `InverseIntegrator` as used in the DPG example 8. We were hoping you might have a suggestion how best to solve the global system. We have the following two ideas:

* re-write our code such that we have a loop over elements and somehow, given an element, find the faces of the element boundary, add these contributions to our element matrix, and then use `InverseIntegrator`. The disadvantage here, though, is that we would visit each face twice.

* Our second idea is use Pardiso to compute the inverse of A. But then we'd need to do matrix-matrix multiplication to compute the Schur complement  -CA^{-1}B + D.

Any help is appreciated. We would be happy to contribute this code as an example also once we have things running nicely...

regards,
Sander, Tamas, Abdullah

rhebergens:
If I'm not mistaken, I think the hybridization that is already in MFEM gives us more or less what we need for HDG after some modifications.
regards, Sander

tzanio:
Hi @rhebergens,

Yes, you should be able to use the general hybridization framework (see `BilinearForm::EnableHybridization` and `fem/hybridization.hpp`) by specifying an integrator for the mixed form `C` between the `u` and `q` spaces. We have one where `u` is in `RT` and `q` is discontinuous on the mesh skeleton/interface.

Sorry for the delayed reply, we've been busy with a new project.

Hope this helps,
Tzanio

tzanio:
Hi @rhebergens, were you able to use the included hybridization for your problem?

rhebergens:
Hi Tzanio,

We have things running in serial. It was actually a lot easier than expected. We first create all global blocks of our HDG matrix. Since A is a block diagonal matrix, we can then simply write a for loop over all elements, extract the blocks, compute the inverse of the element blocks, and overwrite the A matrix by its inverse that way. Using the RAP operators we were able to then construct our Schur complement, S = - B A^{-1} B^T + C. 

[...]

PR #4350, opened 2024-06-11
[WIP] Hybridization of mixed systems (HRT, HDG) [darcy-hdg-dev]

najlkin:
In this PR, the capabilities of MFEM are developed to handle mixed systems in an easy way and hybridize them effortlessly. The particular family of problems with symmetric weak mixed formulation (_not_ necessarily with symmetric coefficients though!) is considered, which appears commonly in physics, like Darcy problem, heat diffusion or Maxwell's equations. They can be written as the equation for the _flux_ (velocity, heat flux, electric field) and _potential_ (pressure, temperature, magnetic field). The potential equation then can include a convection term, giving the largely universal convection-diffusion framework. We concentrate mainly on discontinuous potentials, but the basic functionality is offered also for continuous elements, while fluxes can be either continuous (RT, ND, ...) or discontinuous (DG). The main idea is that the user can formulate the mixed system only and easily hybridize it with a minimal effort, like in the case of `BilinearForm` and `Hybridization`. The main classes offered here are:

- `(Par)DarcyForm` - The main class for formulation of the mixed system. It wraps the `BilinearForm` for the flux and optionally for the potential (definite or convection problems), and the `MixedBilinearForm` for the action of the flux (divergence or curl typically). Alternatively, `NonlinearForm` can be used for the potentials (like for non-linear diffusion-reaction/convection). It constructs the transpose operator/matrix of the flux and the underlying `BlockOperator`. The easy-to-use methods like `FormLinearSystem()` then can be called to obtain the linear system, where elimination of the essential DOFs of fluxes is performed in the whole mixed system. The most importantly, it offers the method `EnableHybridization()`, which performs the hybridization automatically, without the need to reformulate the system. The user needs to specify only the flux constraint (like in `BilinearForm::EnableHybridization()`), while the potential constraint is collected from the interior face integrators of the potential mass form, which implement the new method(s) dedicated to collective assembly of the (broken space) element matrices, constraint, action of the Lagrange multiplier and the trace stabilization together (see :point_down:  for details). This way, the identical parameters are used as without the hybridization, consistency on the algebraic level is guaranteed, and the assembly process is accelerated. The boundary terms are handled in a similar fashion, adding the natural BCs as constraint boundary terms.
- `DarcyHybridization` - The class designed for hybridization of the mixed system. Using the given flux and potential constraint integrators, it performs the hybridization of the system in a (from the outer perspective) similar manner to `Hybridization`, but on both equations at the same time, weakly enforcing continuity of the total flux (instead of the flux itself). The methodology of the HDG hybridization follows the paper: [Nguyen, N. C., Peraire, J., & Cockburn, B. (2009). An implicit high-order hybridizable discontinuous Galerkin method for linear convection–diffusion equations. Journal of Computational Physics, 228(9), 3232–3254.](https://doi.org/10.1016/j.jcp.2009.01.030) for linear convection and its extension [Nguyen, N. C., Peraire, J., & Cockburn, B. (2009). An implicit high-order hybridizable discontinuous Galerkin method for nonlinear convection–diffusion equations. Journal of Computational Physics, 228(23), 8841–8855.](https://doi.org/10.1016/j.jcp.2009.08.030) for non-linear convection. Internally, (unlike `Hybridization`) the construction is block-wise without construction of the sparse matrices of the mixed system, supporting also elimination of the essential DOFs.
- `DarcyReduction` - The abstract class for `DarcyFluxReduction` and `DarcyPotentialReduction`, which perform elimination of the flux or potential from the mixed system. This process is automatically triggered when `DarcyForm::EnableReduction()` is called prior to assembly of the form. However, be aware of that the reduction is possible only in certain cases. For elimination of the flux, the LDG formulation must be used, which offers an efficient alternative for HDG for low-order elements. For elimination of the potential, the face integrators, typically corresponding to convection, must not be present and only element contributions are allowed. Currently, only the linear regime is supported.

Two regimes of operation are supported:
 - _linear_ - The mass term of the potential is linear (like in linear convection/reaction) and a `BilinearForm` is used for the potential part. In the case of hybridization, the system is reduced to a sparse matrix and vector of the right hand side (similarly to `Hybridization`). Classes `HDGConvectionCenteredIntegrator` and `HDGConvectionUpwindedIntegrator` are offered for the centered and upwinded HDG advection schemes respectively. They are both derived from `DGTraceIntegrator`, working in the (L)DG (non-hybridized) regime as well, where the standard `AssembleFaceMatrix()` interface is used for calculation of the numerical fluxes. In the HDG regime, the new methods `AssembleHDGFaceMatrix()` of the integrators are used instead.
 - _non-linear_ - The mass term(s) are non-linear (like in non-linear convection/reaction) and a `NonlinearForm` is used for their assembly. Additionally, a `BlockNonlinearForm` can be specified for non-linear diffusion. The hybridization process is different in these cases and the `DarcyHybridization` object itself acts as the system operator. Its action (`Mult()`) on a given vector of the traces then involves the local (element-wise) solver of choice, which can be either LBFGS/LBB or the Newton solver (with GMRES for linear iterations). The former requires only the action of the numerical flux obtained through `AssembleHDGFaceVector()`, while the latter relies also on the Jacobians assembled in `AssembleHDGFaceGrad()`.
For non-linear convection, the class `HyperbolicFormIntegrator` has been extended and one of the following numerical fluxes can be used:
   - `HDGFlux` - HDG schemes I and II from the referenced paper (and Rusanov flux without hybridization)
   - `RusanovFlux` - Classical Rusanov flux (aka local Lax-Friedrichs) stabilized by the integral formula from the paper for hybridization
   - `GodunovFlux` - Classical Godunov flux, also stabilized by the integral formula from the paper for hybridization

   The framework for non-linear diffusion relies on `MixedConductionNLFIntegrator`, which is combined with a ` MixedFluxFunction`. Currently implemented, there are `LinearDiffusionFlux` and `FunctionDiffusionFlux`.

❔ This PR presents an alternative to #303 , addressing the original issue #91 .
❔ It has also a great value for simulations of transport in anisotropic media, like in tokamaks investigated within [RF-SciDAC](https://www.rfscidac4.org/).

#### Dependencies: ↪️ ####

- #4123 
- #4082 
- #4356 
- #4513 
- #4799 
- #4907 
- #4968 
- #4659 
- #5066 
- #4091 
- #4132 
- more should be spun off...

#### Integration: 🚀 ####

- #5384
- more should be generated...

#### TODO: 📑 ####

- [x] non-linear convection (HDG-I/II, Rusanov, Godunov)
- [x] gradient construction for Newton solver
- [x] elimination of flux/potential from the linear mixed system
- [x] non-linear diffusion
- [x] integrate the core classes to the folder structure
- [x] MPI parallelization
- [x] add reconstruction of continuous fluxes
- [x] reconstruction of super-converging fluxes and potentials (linear)
- [x] broken RT elements
- [x] regression tests
- [x] convergence tests
- [x] non-conforming meshes (mixed, reduction, linear hybridization)
- [x] AMR (hybridization without projection)
- [x] H1 traces
- [x] elasticity
- [x] essential Dirichlet conditions
- [x] parallel operator-based assembly
- [x] documentation
- [x] clean examples
- [x] check/add sample runs
- [x] update the links in [Gallery](https://mfem.org/gallery/) with the merge
- [ ] algebraic constraints
- [ ] GPU support?
- [ ] hybridization of Maxwell?

#### Examples: 🌟 ####

- `ex5(p).cpp` - A minimal example showing usage of `(Par)DarcyForm` instead of `BlockOperator` in example 5 with an effortless hybridization of the RT elements.
- `hdg/ex5(p).cpp` - Identical to the previous one, but adds DG elements on top of it and their reduction (LDG) or hybridization (HDG).
- `hdg/ex6(p).cpp` - Poisson problem solved by the hybridized formulation (H(B)RT/HDG) with AMR.
- `hdg/ex14(p).cpp` - Poisson problem solved by the mixed or hybridized formulation with (B)RT/DG.
- `hdg/ex17(p).cpp` - Elasticity problem with Dirichlet BCs solved by the L/HDG formulations.
- `hdg/ex21(p).cpp` - Elasticity problem with a load BC solved by HDG with AMR.
- `hdg/ex41(p).cpp` - Implicit-explicit convection-diffusion solved by LDG or H(B)RT/HDG.

#### Miniapps: 🌟 ####

- `hdg/(p)anisodiff.cpp` - Demonstrates the asymptotic or steady-state heat diffusion with an anisotrotropic heat conductivity tensor:
  - steady diffusion - (anti)symmetric diffusion of a sine profile (also in `ex5-heat.cpp`)
  - diffusion ring - strongly anisotropic diffusion along a circle
  - boundary layer - exponentially decaying boundary layer from [C. Vogl, I. Joseph & M. Holec, Mesh refinement for anisotropic diffusion in magnetized plasmas (2023). Computers and Mathematics with Applications, 145, 159-174.](https://doi.org/10.1016/j.camwa.2023.06.019)
  - Sovinec problem - a classical problem for measuring the artificial diffusion in the perpendicular direction to the dominant one of anisotropy from [C. R. Sovinec et al., Nonlinear magnetohydrodynamics simulation using high-order finite elements (2004). Journal of Computational Physics, 195, 355–386.](https://doi.org/10.1016/j.jcp.2003.10.004)
  - Umansky problem - a test of artificial diffusion widening a transition layer from [M. V. Umansky, M. S. Day & T. D. Rognlien (2005), On Numerical Solution of Strongly Anisotropic Diffusion Equation on Misaligned Grids, Numerical Heat Transfer, Part B: Fundamentals, 47(6), 533-554.](https://doi.org/10.1080/10407790590928946)
  - single and double null problems - two-wire, large aspect ratio approximation of tokamak with X point(s) adopted from  [C. Vogl, I. Joseph & M. Holec, Mesh refinement for anisotropic diffusion in magnetized plasmas (2023). Computers and Mathematics with Applications, 145, 159-174.](https://doi.org/10.1016/j.camwa.2023.06.019)
  - ...
- `hdg/mfem-logo.cpp` - [The winning visualization of MFEM Workshop 2024](https://mfem.org/gallery/#nikl). (To reproduce run `./mfem-logo -nx 200 -o 3 -a 1e3 -ks 1e-3 -c 2e4 -dg -hb`)
- `hdg/(p)convdiff.cpp` - Shows all the examples from the referenced paper:
  - steady diffusion
  - steady advection-diffusion
  - steady advection
  - non-steady advection(-diffusion)
  - Kovasznay flow
  - steady Burgers advection-diffusion
  - non-steady Burgers advection-diffusion
-  `plasma/braginskii_hdg.cpp` - anisotropic diffusion with Euler equations with anisotropy direction (magnetic field) loaded from a file (IMEX for now, but fully implicit soon 🚧 ) [WIP 👷 ]
- `plasma/scatter.cpp` - miniapp demonstrating implicit coupling with Maxwell equations  [WIP 👷 ]

_Kovasznay flow_ (40x40 Q2 centered HDG dt=0.1, k=1-2, SDIRK23)
`./ex5-nguyen -p 5 -nx 40 -ny 40 -nt 100 -k 1e-2 -sx 10 -sy 2.5 -o 2 -ode 3 -tf 10 -dg -hb`
Kovasznay-Q2-40x40-nt100-k1e-2-ode3-dg-hb-t10


lindsayad:
Does `Hybridization` require a mixed formulation? I am currently researching HDG methods for advection-dominated Navier-Stokes with velocity-pressure formulation like in https://epubs.siam.org/doi/abs/10.1137/16M1083839 as opposed to gradient-velocity-pressure formulations

najlkin:
> Does `Hybridization` require a mixed formulation? I am currently researching HDG methods for advection-dominated Navier-Stokes with velocity-pressure formulation like in https://epubs.siam.org/doi/abs/10.1137/16M1083839 as opposed to gradient-velocity-pressure formulations

The classical `Hybridization` is for a single bilinear form, the new `DarcyHybridization` is then for a mixed system. Darcy or Stokes problems are classical examples. However, a problem with the formulation in the cited paper is that there are two traces, while here we consider only a single one 🤔 .

lindsayad:
> However, a problem with the formulation in the cited paper is that there are two traces

Yea, these kinds of formulations are what @thorvath12 was referring to in https://github.com/mfem/mfem/pull/303#issuecomment-2014017463. So if I understand correctly, this PR would not be a substitute for #303 for those formulations

najlkin:
@lindsayad , it is an alternative approach, not only implementation, to solution of advection-diffusion equation (or similar) mentioned in #91 . Conceptually, the work of G. N. Wells weakly enforces continuity of all quantities without a difference, while Nguyen & Cockburn only apply this on the total flux. Such formulation then can benefit from the local conservation property and nicely work even without hybridization like RTDG or LDG (which we keep full compatibility with here, so a single line turns on the hybridization or not :wink: ). Moreover, there might be physical cases when discontinuity of the potential is desirable like Riemann problems. Anyway, there might be good reasons for the G. N. Wells approach as well, but it is closer to the technique of static condensation (with some extra stabilization), which is offered in MFEM on the algebraic level for `BilinearForm`, but not mixed systems unfortunately. This would be an interesting question if these techniques really coincide for discretely compatible trace spaces and the stabilization is not needed or not... 🤔

lindsayad:
It's very interesting work and I really do like the capability to toggle on hybridization with very little additional effort from the user. I'll need to understand better how this is done in the LDG case for which classic LDG is not itself hybridizable as I understand Cockburn's papers

najlkin:
You are right, it must be pointed out that LDG and HDG formulations are no longer equivalent here. HDG has better stabilization compared LDG, which misses one part as Cockburn mentions. This can be clearly seen when you try the centered scheme for advection-dominated problems. LDG performs poorly showing some oscillations, while HDG gives nicely smooth solution 👍 .

najlkin:
⚠️ Multiple examples were moved and renamed, constituting the new `miniapps/hdg` folder and contributing to `miniapps/plasma` folder. Namely, there are these miniapps:

- `hdg/anisodiff` - test problems of anisotropic diffusion (formerly `ex5-aniso`)
- `hdg/(p)nguyen` - (convection-)diffusion test problems (formerly `ex5-nguyen`)
- `plasma/braginskii_hdg` - modified version of Euler convection-diffusion from `ex18-hdg` with anisotropic diffusion [WIP👷]
- `plasma/scatter` - example of non-linear waves of the medium and EM waves (formerly `ex5-scatter`) ▶️  going to be rewritten

najlkin:
⚠️ I merged `master`, resolving conflicts. Everything seems to be working, but please cleanly recompile the code with new makefile defaults.

fem/darcy/darcyhybridization.hpp
sshiraiwa, line 1: Thanks. One typo here.
```
    bool own_m_nlfi_p{};
    bool own_m_nlfi_p{};
```
One of this should have been ```own_m_nlfi_u```

fem/darcy/darcyreduction.hpp
sshiraiwa, line 1:
```
   NonlinearFormIntegrator *m_nlfi_p{};
   NonlinearFormIntegrator *m_nlfi_p{};
```
The same here. One of them should be ```*m_nlfi_u{};```

fem/darcy/darcyhybridization.hpp
najlkin: Thanks, fixed just now :wink:

fem/darcy/darcyreduction.hpp
najlkin: Fixed, I wonder how many typos I made in a single commit X) Submitting from a conference does not lead to the best code really :)

sshiraiwa:
Currently, ex5-hdg.cxx uses miniapps/hdg/darcyop.hpp. This looks to me not organized well. 
I naively think that at least top level exampels should run without extra dependency other than libmfem.

najlkin:
As discussed elsewhere, it is not meant for master and was added just to try SolutionController. I will revert it.
Update: Reverted ✔️

najlkin:
❓ Please notice I moved the examples to `hdg` subfolder of `examples`.

ianabel:
Hi,

Looking to help out, add some extra generality to this branch, and help get it into MFEM. My current plan is to branch a gf-hdg-dev branch from this one (gf = GridFire, the company supporting this work), and make periodic pull requests to merge into darcy-hdg-dev those things that naturally go together. 

How does this sound @najlkin ? Also feel free to point me at things which you want doing to accelerate the ultimate merge!

Ian

najlkin:
Hi @ianabel , that sounds amazing! 👍 We would love to learn more about your application and what parts are the most interesting for you or missing from your perspective... 🤔 
Besides, let me mention that we are very close with the phase 1 of merging this work, see #5384 and mfem/web#380 👀 .

ianabel:
@najlkin -- let me summarise what I'm trying to use this for and what I'm up to right now:

So, from an abstract point of view, I want something that implements the Peraire, Nguyen, Cockburn HDG schemes for convection-diffusion systems (including elliptic and time-dependent problems). I like the algorithms and want to be able to point people at a real implementation rather than 'ask the authors for some vintage MATLAB' :) 

On top of this, I want to add the Cockburn-Solano method of solving on curved domains by extension from polygonal subdomains without needing non-simplicial elements. I also watn to investigate how hard adding HDG<->BEM coupling into MFEM will be (again, looks like a very neat self-contained set of algorithms that tie together cleanly)

Practically, this will be used for the following problems:

1) 1D-2V Fokker-Planck simulations using fully implicit time advances and fully nonlinear collisions in Rosenbluth potential form (this also requires the coupling-at-a-distance between HDG and BEM to give the correct behavior at infinity).

2) A fast adaptive Grad-Shafranov solver based on the papers by Cerfon & Sanchez-Vizuet, free boundary capability added again using the HDG-BEM coupling at a semicircular interface and again using the extension from subdomain method to pin at the circular boundary

Looking at the first-phase pull request #5384 , I see one qeury arises over test suites. My initial claude-assisted pass, pushed as [branches/gf-hdg-dev](https://github.com/mfem/mfem/tree/gf-hdg-dev) contains a bunch of tests and checks.  I would also note that the regression tests you use are implementation-specific (so fail if I use a different point release of UMFPack or similar) as they compare beyond the resolution the algorithm gives.

Feel free to pick any commits you want from the branch. Currently the only feature I've added is the ability to truly run multiple systems of equations (clearly this was on your future plan but time is short and life is busy!), and the hooks needed to provide a fully general stabilization function.

The extended internal API for the stabilization function would allow the HDG-I HDG-II, etc, methods to just be different stabilization functions, rather than looking like fundamentally different schemes.

The existing stabilization functions have not been altered so existing results are bit-for-bit identical (or should be). However, the default behavior your branch ships is tau ~ 1/h (even for purely diffusive problems), which is specifically not what NPC suggest ( kappa / l in the papers is called out specifically as not being mesh-size-dependent, l is some physical scale) and so you get lower-than-ideal convergence rates for some problems with 'td=const' which looks at first glance like a reasonable choice. 

Some hidden bugs did get fixed in the meantime (list according to claude --  I've been kicking it but there might still be a mistake in here, and I need to re-review):

│ 1   │ IsothermalFlux::ComputeAvgFlux buffer overflow — d*(2+dim) stride copied from EulerFlux (which has dim+2 equations) into a class with dim+1. Wrong answer plus malloc(): invalid size │ fem/hyperbolic.cpp               │ 5233301e64 │
│ 2   │ SumBlockNLFIntegrator::AllocBlock read uninitialized pointers after SetSize; the Array2D overload also indexed a by b's extents                                                       │ fem/nonlininteg.cpp              │ fa843767fc │
│ 3   │ LeftScaling on a 1×dim matrix in LinearDiffusionFlux/FunctionDiffusionFlux — scaled every component by ikappa(0). Found by the FD Jacobian test                                       │ fem/nonlininteg_mixed.cpp        │ 285ea01fb3 │
│ 4   │ 2 * (el_u.GetOrder(), el_p.GetOrder()) — a comma operator discarding the flux order, under-integrating whenever it exceeds the potential order                                        │ fem/nonlininteg_mixed.cpp        │ 7891820205 │
│ 5   │ AssembleHDGFaceGrad had a live upwinding branch whose residual counterpart is commented out — Jacobian exactly 2× the residual derivative at any velocity                             │ fem/nonlininteg_mixed.cpp        │ 7891820205 │
│ 6   │ AssembleHDGGrad's Array2D<DenseMatrix*> never sized — elmats(0,0) = &elmat_A writes through a null pointer. Any BlockNonlinearFormIntegrator under hybridization segfaults            │ fem/darcy/darcyhybridization.cpp │ f1caec38f5 │

and I had to make two tiny build changes to make the miniapp build out of tree (so I can have parallel and serial both building and testing), commit 0c2db0d901

None of these bugs were actually error-inducing (none of the miniapp or tests went down those paths) but probably want fixing int he first merge.

So, currently gf-hdg-dev is some fixes and some extra harnesses to permit the extension beyond one system -- feel free to merge all of it if you want. 

Let me know! My next steps are making sure that this HDG framework is robust on coupled systems of nonlinear poisson equations (which is what the nonlinear collision operator for a plasma winds up needing), before stepping into the curved domain work.  

Ian

najlkin:
@ianabel , that is all really cool! 😎 

You can definitely do NPC-style hybridization and more! No need for MATLAB 😁 And there is also HBRT, EDG, AMR, TMOP,...

Curved boundaries and BEM sound intriguing and make a lot of sense in the context of MCF! I would need to dive deeper to reason about them though 😅 .

The targeted problems are also interesting generally in the context of HDG, but also for me personally as I have background in plasma physics and this development is motivated by some MCF applications. The projects are concluding though, so we are looking for new ones to sustain this development. If there some space for collaboration, I would be happy to discuss that... 😉 

I may have a look at the unit tests, those sound practical 😉 . However, the phase 1 PR is already very large, so on one hand it makes sense to complement it by unit tests, but on the other it is hitting the capacity limits of the human reviewers and most of the code paths are checked by the examples and miniapps, which we test internally beyond just the GitHub CI 🤷 But definitely it can useful for development in addition to the regression tests to detect bugs early. But you are right, the regression tests were not properly tested on different machines or anything. They are really basic in essence, but have proved to be useful for detection of various anomalies and bugs internally though 🐛 . They should be all green now, so you may re-baseline them on your machine at least 😉  (Note that suitesparse is now default for hybridization, but the tests were made when it was not, so you may comment out the preprocessor declaration `USE_DIRECT_SOLVER_HYBRIDIZATION` in `darcyop.cpp`).

Systems of equations should work through vdim of `DarcyForm`. However, it has been for me so far only elasticity (i.e., spatial components) and Euler hydro system (i.e., a single nonlinear flux). The `VectorDiagonalBlockIntegrator` class is ready for it taking different integrators for different components, but it is not much demonstrated. If you need really a heterogeneous system with nonlinear fluxes or so, you will need to make the glue by yourself as you have figured out already probably 🙂. What I should mention here is that we were internally discussing concept of a "block form" for such problems, which would generalize such constructions and would not require to form dense matrices of the full system (like the current construction) and would stay with just some sparse block representation, which would not require that much of memory. But it has not materialized so far. If you are interested in nonlinear systems primarily, it is a less of a concern maybe, but it appears in construction of the gradient for the classical Newton method.

Generality of the stabilization is another thing, which would deserve some research and development. That HDG-I and II schemes are mostly for comparison with the paper, where they are also used more for testing. They are not much practical options really. I used `RusanovFlux` in the Braginskii miniapp, which is the only reasonable option for systems like Euler at the moment. So it would be definitely worthwhile to look into these and implement something better, but it is highly non-trivial if the system requires positivity like Euler 😬  . Another thing is that you want something fast, making use of all cancellations happening in schemes like HDG-I/II or monotonicity simplifications of Rusanov, so (for both reasons) I would be a bit sceptical of some completely general interface 🤔 .

In terms of element size dependency, feel free to experiment 🙌 . I just implemented the "default" option, but you may find something else better in your case. I was mainly concerned how poorly it behaves with AMR and strong anistotropy causing loss of positivity, so I played with the scaling there (not in repo, I think).

I need to look into the bugs, thanks for checking that 👍 . The good news are that they seem to be all in the nonlinear parts and not in the linear, which is about to be merged 😅 . As you say, they do not appear in the test runs, so that is a bit puzzling here 🤔. For example, some runs use Newton solver for sure, so it is strange to me there would be some segfault 😮 .

Coupling sounds totally doable! 👍 I have an example of coupling with Maxwell equations in `scatter.cpp` (probably broken at this point due to changes, but just for an idea). It matters if you want an iterative coupling or a single monolithic system. The former is easy, but the latter is harder and you will need reconstruction of the quantities for every iteration most likely. One could think of how to optimize the process, because such reconstruction happens for construction of the gradient internally anyway...

ianabel:
I've wired a general HDGStabilization class (general tau(u_h,u_h-hat) in NPC notation), with cheap fallback if it's chosen constant (or if it's chosen to be the current estimate). Note, the current implementation estimates gradients as well as being h-dependent, so can require extra work.

Currently this sits alongside the HDGFlux mechanism that's already in the branch. 

Conceptually, I think the additive split tau = tau_diff + tau_conv with the diffusive tau represented by an HDGStabilization function and the tau_conv being computed together with F(u-hat) for a general convective flow F(u), and represented by a FluxFun is fine. 

This does give two ways to implement effectively the same thing, but I can't speak to how this looks if the same interface has to work for RT / Broken-RT as well as classic HDG. 

I'm not sure exactly what you meant when you said there's cancellation and monotonicity simplifications for HDG-1/HDG-2/Rusanov, nothing looked like those cancellations would be ill-conditioned or near round-off [if one turned out to be an issue, making the general stabilization function return tau * (u - u_hat) to allow the simplification to be done is probably the way I'd go]. 

One point I'd like to discuss -- currently, for nonlinear problems, the way the DarcyForms work ends up with a local per-element newton solve. Classically, HDG linearizes for a global newton iteration *then* does the static condensation. Do you have a short explanation on why this was the path chosen (it results in some stability issues in using this for a grad-shafranov solver, so I would like to know why before I go and also implement the linearize-then-condense pathway as a parallel track).

in other news, if you want to play around with using all your hard work on curved domains, extended from a subdomain, then [gf-subdomains branch](https://github.com/mfem/mfem/tree/gf-hdg-dev) works and is tested (including postprocessing).

najlkin:
I was thinking about something like HDGStabilization initially too, but it is better in the end to bake it to the HDG flux/matrix logic. For linear problems, the thing is you might want to have exact cancellation in the upwinding and not to perform the calculation twice for this reason and also speed. Following the paper for nonlinear problems, the general form of the numerical flux ends up with an integral term, which you really do not want to evaluate during computation. Two special cases are the HDG-I/II schemes, but that is more for testing, in my opinion. That Rusanov is fine mostly, but not the best and you have to make some monotonicity assumptions there for simplification of the terms to remove the integral.

About the second thing, let me explain the design. You solve the global system for traces, so for mult or gradient you need to reconstruct the flux and/or potential locally from the known values of traces. This uses a local Newton solver (not necessarily, but that is usually the fastest option). There are three cases based on that if you have linear flux equation and nonlinear potential equation, vice versa, or everything nonlinear. With the local reconstruction, you can proceed and calculate the numerical flux and then the residuum of the trace equation or gradient. It is true that gradient and mult share a lot of computation, if you do them together, you can optimize them as I mentioned for the coupling. However, that is the prize for the general interface in MFEM, which works for both, Jacobian and Jacobian-free methods. But I am not sure you are talking about that, could you elaborate? Usually the global Newton/Jacobian-free + local Newton work fine together, but it is true you might tune the relative tolerances beyond the default values so that the local error does not cause destabilization on the global level. You may play also with adaptive values as the initial iterations of the global solver do not need high precision from the local one.

ianabel:
@najlkin -- on HDGStabilization, understood for exact matching of upwinding, ok to keep it around for (nonlinear) diffusion problems? Those parts of the stabilization function don't translate into numerical fluxes as neatly as the convective bits. Or would you rather I tried to work the ones I happen to need into subclasses of HDGDiffusionIntegrator? 

For the nonlinear solves, my issue is simpler -- local newton solves on some of my cells don't converge*, and if one proceeds exactly as in NPC, and eliminates the trace variables from the linearized equations, rather than from the original nonlinear ones, a local newton solve (which for a decent number of cells starts to get more expensive) is no longer needed.  In addition, running 10^4-10^5 per-element decoupled linear solves can be efficiently punted onto a GPU (or elsewhere) using the methods in linalg/batched. Variable-size local Newton solves won't be efficiently split. 

Is there a nonlinear problem that works better if you eliminate first using the local Newton solves? Happy to try and run the tests and figure out what the most streamlined path forwards is. 

*in this particular case, I'm thinking about a steady-state solve so there isn't the damping on the Newton solve that a time-march would give. I'm using Jacobians, rather than a Jacobian free approach for the outer loop, but the outer loop is fine. 

[no prizes for spotting that I've been working on diffusion-dominated regimes where more powerful numerical fluxes for shocks etc aren't needed]

najlkin:
@ianabel , I see, you are talking about nonlinear *diffusion* flux stabilization, that is a different story 😅 . As it is not hyperbolic, but elliptic/parabolic, you do not need to care about entropy production and numerical Riemann solvers, and the whole thing is much simpler. I think I would prefer some function/coefficient passed to `HDGDiffusionIntegrator` if it can be reduced to that. I would not generalize it too much, because things are more complicated for the convective part.

About the nonlinear solve, what you are saying is that you are solving the nonlinear problem without static condensation and the linear one with it? That makes sense as you do just one reconstruction instead of two (in mult and gradient), as I was talking about. But on top of that, you do linear reconstruction instead of nonlinear? That also saves time, but it is not compatible with Jacobian-free methods, I believe.  Anyway, this is a pathway I cannot mainstream easily, because gradient has to have the same size as the operator itself (which is hybridized) and you would need extra steps in the solvers etc. However, it is an appealing option, so we should think about how to wrap it 🤔 .

ianabel:
So, HDGStabilization is just something you add to a HDGDiffusionIntegrator with SetStabilization. So shouldn't intrude on other paths. Feel free to poke around in the branch to check the API is to your liking.

For the nonlinear, yes, you solve the weak-form nonlinear problem by Newton then you use the condensation on the linear Newton iterates to turn it into one global face-coupled system S along with many element-local solves. I don't know that this ends up being 'linear reconstruction' (partly because I need more cofffee and to remind myself what Finite Volume schemes mean by that) -- at each newton iteration one is always solving a linear problem, but the global flux / values relationship is converged to by the outer Newton solve. This would all happen inside one timestep for a time-dependent problem, so consistency of the iterates before convergence isn't an issue. 

If (because of a sqrt() or similar in a flux or source you need the iterates to always be positivity preserving (as well as whatever newton converges to) there may be something that needs doing. I don't have a good reference for positivity preservation in HDG schemes, so if you know one, that would be great.

Moving to Jacobian-free methods; this ordering always avoids constructing a true full Jacobian on all the degrees of freedom. Beyond that there are three levels of extension towards JFNK methods:

0) Assemble S, direct sparse solve
1) Assemble S, krylov solve
2) Don't assemble S, solve S x = y with an iterative krylov method

My current scheme supports all of those, tested on the convdiff bit of the miniapp (although we could do with something that takes more than one outer Newton step to converge!). 

Yes, there's some nastiness around fitting this to the MFEM API shape, so the DarcyHybridization class winds up carrying a bunch of state; also if we want clean JFNK through KINSolver that needs a couple of line addition to the KINSolver class.

Have a look in gf-hdg-linearise-first and see what you think. I'm trying to add the parallelism/device offload to the per-element solves right now, which is the real performance win for doing it this way

najlkin:
Sounds good with HDGStabilization 😉 I will have a look (but I am occupied by other things right now tbh).

No finite volumes here 😁 By linear reconstruction I simply mean back-solving of the linearized equations. My concern for JF methods is that: you start with the state u_n, p_n, x_n (=flux, potential, trace). The solver gives me x_n+1, but how do I get u_n+1 and p_n+1? With the linearization, I have to use grad(u_n, p_n, x_n) to reconstruct these. That makes sense with Newton, which follows this gradient, but not that much for JF methods, where the step is whatever and u_n+1, p_n+1 should be considered independent in principle. If not, I am worried this kind of retardation can pollute the dictionary of these methods. But if you say it works, it works 😄. Just keep it in mind maybe. Anyway, sounds cool 😎 , but wrapping will be less cool - I have thought of a few options, but none of them is completely clean ...well, rather hacky😬 .

ianabel:
Ahah. I begin to see :) Yes, we use a linear solve to take x_n+1 and u_n/p_n to u_n+1 / p_n+1 [Equation 17a of Nguyen Peraire Cockburn 2009] 

and yes I don't currently know what to do with a JF method that isn't JF-Newton / JFNK (although BFGS approximates the Jacobian, so perhaps there's an obvious way to extend there?)

I will leave your ordering of 'CondenseThenLinearise' in the codebase (set via SetNLOrdering) for someone who wishes to do some sort of method with a non-Newton outer iteration such as Anderson-accelerated Picard. There you condense, use nonlinear reconstruction, and only ever evaluate the residual. No Jacobians needed. That's out of my wheelhouse, so I don't know if that's ever the optimal way of doing things, but happy to leave it! 

Will try and find a grim convection problem (sod shock? livermore-classified-shock? ;) ) and make sure that is in the miniapp and doesn't break

PR #4371, opened 2024-06-25
General NCMesh spacing

dylan-copeland:
This PR generalizes the previously hard-coded bisection used for refinement in `NCMesh`, allowing for arbitrary spacing of the refined elements. This generalization also enables 3:1 refinement.
Screenshot 2024-06-25 at 4 03 02 PM
To illustrate scaled refinements, the miniapp `miniapps/shifted/distance` is modified with a simple, crude method for anisotropic AMR based on the distance field computed by the miniapp. Some examples are shown below, with (left) the distance plotted on the initial mesh and (right) the `ex1` solution on the adaptively refined mesh.
P0
P1

The changes to the `distance` miniapp are for illustration and testing purposes in the review. I plan to revert those changes before merging this PR.

TODO:
- [x] Revert changes to `distance` miniapp.

| [#4371](https://github.com/mfem/mfem/pull/4371) | @dylan-copeland | @tzanio | @acfisher + @cjvogl + @hughcars + @jakubcerveny | 9/21/24 | 12/17/24 | 12/18/24 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:
Wow! ping: @jakubcerveny

jakubcerveny:
(exploding head emoji :-)

This is really interesting, I'm wondering if there are any side effects and if it generalizes to more complex cases. If this really works it will be impressive.

tzanio:

mesh/ncmesh.hpp
```diff
@@ -32,17 +32,35 @@ namespace mfem
...
    int index; ///< Mesh element number
    char ref_type; ///< refinement XYZ bit mask (7 = full isotropic)
+   real_t scale_x, scale_y, scale_z; ///< Scaling in each direction
 
    Refinement() = default;
 
-   Refinement(int index, int type = Refinement::XYZ)
-      : index(index), ref_type(type) {}
+   Refinement(int index, int type = Refinement::XYZ, real_t s_x = 0.5,
+              real_t s_y = 0.5, real_t s_z = 0.5)
```
cjvogl, line 50: If the user wants to refine in Z only, will they have to specify dummy values for `s_x` and `s_y`?

dylan-copeland: Yes, unfortunately the user will have to specify dummy values for directions in which there is no refinement, and I agree that this is awkward. Your suggestion would be convenient for the 3 cases of refinement in one direction, but there are also the 3 refinement types `XY`, `XZ`, and `YZ`, which also require a dummy value and 2 real values. For those types, `std::pair` would not work. We can make functions like `SetX`, `SetXY`, etc., but I think this constructor is still more convenient, despite the dummy values. Another idea is to make a `struct RefinementType` to replace the `enum`. We could have all types derive from that base struct, e.g. `RefinementX`, `RefinementXY`, etc. That way, this constructor for `Refinement` would take a `RefinementType` and then get the values `s_x`, `s_y`, `s_z` from the derived struct, which would set the dummy values automatically.

cjvogl: > Your suggestion would be convenient for the 3 cases of refinement in one direction, but there are also the 3 refinement types XY, XZ, and YZ, which also require a dummy value and 2 real values. For those types, std::pair would not work.

FWIW, I thought the XY, XZ, and YZ refinements could be constructed with `Refinement(index, {{'X',0.5},{'Y",0.5}})`, `Refinement(index, {{'X',0.5},{'Z",0.5}})`, and `Refinement(index, {{'Y',0.5},{'Z",0.5}})`, respectively.  Or perhaps there's an issue I don't yet see.  That's not to say introducing a hierarchy of structs isn't better in some way... I think either my suggestion or the hierarchy of structs will provide a cleaner API to the user.

dylan-copeland: Right, we could use a list of refinements in each direction, instead of the `enum`, as the type can be inferred from the directions included in the list. I agree that (i) the list of refinements in each direction and (ii) the derived structs are both good options, better than the current version. Such API questions often get lots of input, so let's see if anyone else has opinions.

miniapps/shifted/distance.cpp
```diff
@@ -115,6 +115,7 @@ real_t sphere_ls(const Vector &x)
    const real_t r = sqrt(xc*xc + yc*yc + zc*zc);
 
    return (r >= radius) ? -1.0 : 1.0;
+   //return xc >= 0.0 ? -1.0 : 1.0;
```
cjvogl, line 118: Is this here intentionally or is this perhaps left-over debug code?

hughcars:
Question: "This generalization also enables 3:1 refinement." -- is this achieved by refining with spacing 1/3, then in the 2/3 side, using a 1/2? Or does an edge node now have multiple children? I.e. if splitting a segment in this 3/1 fashion, does it have 3 direct children, or 2 children, then one of those children have two children?

hughcars (commented):
High level question: how can this approach reconcile a conformal interface?

I.e. given two squares that share an edge, left is refined 3:1, by refining 1/3, then 1/2. the other is refined 3:1 by 2/3 then 1/2. Now there'll what should be a conformal face between the two macro elements, but both sides will think that the interface is actually nonconformal. This would be pretty deleterious for higher order spaces given how an NC face results in massive reduction in the basis. The same issue I think happens with two hex sharing a face refined with a different spacing. 

I'm not sure how you could keep track of such a scenario using a `real_t scale` without a floating point comparison trap door. Hypothetically an integer arithmetic trick might help? I.e. This split is defined in a locally specified fixed precision. This would look something like `int numerator, int denominator` rather then `real_t scale`, then adjacent elements splits of faces could be compared by checking the integer arithmetic from the rational numbers. This gets pretty involved pretty fast, but keeps the topological relations exact like the `1/2` that is implicitly used right now.

Would be an easy check with two 3:1 refined in Y squares with P2. The resolution for P2 would be lower than for the conformal mesh of the same vertices.

dylan-copeland:
> High level question: how can this approach reconcile a conformal interface?
> 
> I.e. given two squares that share an edge, left is refined 3:1, by refining 1/3, then 1/2. the other is refined 3:1 by 2/3 then 1/2. Now there'll what should be a conformal face between the two macro elements, but both sides will think that the interface is actually nonconformal. This would be pretty deleterious for higher order spaces given how an NC face results in massive reduction in the basis. The same issue I think happens with two hex sharing a face refined with a different spacing.
> 
> I'm not sure how you could keep track of such a scenario using a `real_t scale` without a floating point comparison trap door. Hypothetically an integer arithmetic trick might help? I.e. This split is defined in a locally specified fixed precision. This would look something like `int numerator, int denominator` rather then `real_t scale`, then adjacent elements splits of faces could be compared by checking the integer arithmetic from the rational numbers. This gets pretty involved pretty fast, but keeps the topological relations exact like the `1/2` that is implicitly used right now.
> 
> Would be an easy check with two 3:1 refined in Y squares with P2. The resolution for P2 would be lower than for the conformal mesh of the same vertices.

@hughcars this is a good question, which also applies to the usual case of 2:1 refinement with scaling 0.5 (even in the current `master` branch). I experimented with some hard-coded refinements in 2D, both 2:1 and 3:1, and found that the `NCMesh` class creates `master`/`slave` edges only in the truly non-conforming case. That is, `NCMesh` is already smart enough to recognize conforming interfaces between refined elements, even when the elements are refined sequentially, in different calls to `Mesh::GeneralRefinement`. For 2D edges, it looks like line 3572 of ncmesh.cpp `if (sb < se)` checks for whether an edge is conforming. For example, two 3:1 refinements in Y of a 2x1 coarse mesh results in the same FE space on an `NCMesh`, with high order, as a 2x3 conforming mesh. So it seems MFEM already determines conforming interfaces between refined elements, avoiding interpolation of master mesh entities and giving the correct conforming FE space on the interface instead of an interpolated subspace. This logic is unaffected by the generalization of spacing in this PR.

mesh/ncmesh.hpp
dylan-copeland: I changed the interface in a926841cf9eeb307f0cdaa48adc51b8f2f096926 similarly to the code suggested by @hughcars. One difference is that I had to use `real_t s[3]` to keep `Refinement` as a trivial type for the `Array` template type. The default refinement, using an empty list, requires `, {}`, which is a little inconvenient but not as bad as alternatives.

dylan-copeland: Commit https://github.com/mfem/mfem/pull/4371#:~:text=the%20interface%20in-,a926841,-similarly%20to%20the breaks backwards compatibility in 2 ways. One is the `, {}` for an empty list to get the default type, and the other is the enum name `Type`. We may be able to avoid the `, {}` by rearranging arguments in constructors, or by using a set function instead of a constructor with a list, but the enum name `Type` seems necessary in order to use the pair of type and scale. For example, `Refinement::Type::X` is necessary now, instead of just `Refinement::X`. I'm not sure we should make this breaking change here. Maybe it is something to keep in mind for MFEM 5.0. As the editor, what do you think @tzanio?

hughcars: The need for `Type::` comes from `enum` becoming `enum class`. Generally `enum class` is preferred as it prevents accidental aliasing of the underlying type into an `enum` by accident. That can be removed though if need be, but I would suggest it's good practice to use it. Adding another constructor that takes only the integer value and dispatches to that constructor with `, {}` would fix that issue: `Refinement(int i) : Refinement(i, {}) {}` in the class body should suffice.

hughcars:
> > High level question: how can this approach reconcile a conformal interface?
> > I.e. given two squares that share an edge, left is refined 3:1, by refining 1/3, then 1/2. the other is refined 3:1 by 2/3 then 1/2. Now there'll what should be a conformal face between the two macro elements, but both sides will think that the interface is actually nonconformal. This would be pretty deleterious for higher order spaces given how an NC face results in massive reduction in the basis. The same issue I think happens with two hex sharing a face refined with a different spacing.
> > I'm not sure how you could keep track of such a scenario using a `real_t scale` without a floating point comparison trap door. Hypothetically an integer arithmetic trick might help? I.e. This split is defined in a locally specified fixed precision. This would look something like `int numerator, int denominator` rather then `real_t scale`, then adjacent elements splits of faces could be compared by checking the integer arithmetic from the rational numbers. This gets pretty involved pretty fast, but keeps the topological relations exact like the `1/2` that is implicitly used right now.
> > Would be an easy check with two 3:1 refined in Y squares with P2. The resolution for P2 would be lower than for the conformal mesh of the same vertices.
> 
> @hughcars this is a good question, which also applies to the usual case of 2:1 refinement with scaling 0.5 (even in the current `master` branch). I experimented with some hard-coded refinements in 2D, both 2:1 and 3:1, and found that the `NCMesh` class creates `master`/`slave` edges only in the truly non-conforming case. That is, `NCMesh` is already smart enough to recognize conforming interfaces between refined elements, even when the elements are refined sequentially, in different calls to `Mesh::GeneralRefinement`. For 2D edges, it looks like line 3572 of ncmesh.cpp `if (sb < se)` checks for whether an edge is conforming. For example, two 3:1 refinements in Y of a 2x1 coarse mesh results in the same FE space on an `NCMesh`, with high order, as a 2x3 conforming mesh. So it seems MFEM already determines conforming interfaces between refined elements, avoiding interpolation of master mesh entities and giving the correct conforming FE space on the interface instead of an interpolated subspace. This logic is unaffected by the generalization of spacing in this PR.

In the 2:1 refinement scenario, for two elements the left being {0,1,2,3} and the right being {3,2,4,5}, when the left is split 1/3, it places node 6 @ 1/3 of the way from 2->3, then the right is split it places node 7 @ 2/3 of the way from 2->3. The left side now has 2 -> 6 -> 3, and the right 2 -> 7 -> 3. Then when splitting 6->3 in half, I am pretty sure that will introduce node 8, rather than discovering node 7. Similarly for splitting 2->7 in half. The resulting tree should have 2 -> 6 -> 8 -> 3, and 2->7->9->3, but nodes 6 and 9 would have the same vertex, and nodes 7 and 8 would have the same vertex. 

In the case of a split with ratio half, node 6 would be rediscovered in splitting the right hand face, because the nodes structure looks for the node associated with the edge 2->3, and there's only 1, whereas once the ratio used in the split is no longer 1/2, there are now two descendents of 2->3: node 6 and node 7. Having any split ratio that isn't 1/2, makes the nodal descendence ambiguous, if two nodes are provided, there can now be two different nodes associated to that edge (corresponding to the two different vertex locations), which need to be selected between.

mesh/ncmesh.hpp
dylan-copeland: I was thinking that `Type` would be necessary in the `std:pair`, but I was able to replace it with `char` in e5de10c9d91b1b2f6d8a8915bbf117baa5f96d7b which restores backwards compatibility.

dylan-copeland:
@hughcars The series of refinements you described is actually not possible, which is a different problem. I modified `ref321` to perform these refinements, and the 1/3 Y-refinement of the right element does not introduce a new vertex. Instead, it identifies the vertices for the left and right refinements, putting them at the same location. I will have to think about how to handle this case.

hughcars:
> @hughcars The series of refinements you described is actually not possible, which is a different problem. I modified `ref321` to perform these refinements, and the Y-refinement of the right element does not introduce a new vertex. Instead, it identifies the vertices for the left and right refinements, putting them at the same location. I will have to think about how to handle this case.

There was some brief discussion of this within the meeting today, and the possibility of having the spacing being defined on a per edge (or more specifically per pair of nodes) basis makes some sense. That would then ensure that there is only a single node per edge, so topologically changing the spacing doesn't change, but the geometry generated would. This would then also introduce the possibility of some pretty weird looking refinements, (and even the possibility of some r-adaptivity, if these spacings can be changed).

mesh/mesh.hpp
```diff
@@ -2261,9 +2261,18 @@ class Mesh
...
+   /// Return a bool indicating whether this mesh is conforming.
    bool Conforming() const { return ncmesh == NULL; }
+   /// Return a bool indicating whether this mesh is nonconforming.
    bool Nonconforming() const { return ncmesh != NULL; }
 
+   /** Designate this mesh for output as "NC mesh v1.1", meaning it is
+       nonconforming with nonuniform refinement spacings. */
+   void SetScaledNCMesh()
+   {
+      ncmesh->usingScaling = true;
```
hughcars, line 2273: `snake_case` for member variables, `using_scaling`, `reparent_scale`.

mesh/ncmesh.cpp
```diff
@@ -487,6 +502,111 @@ int NCMesh::Face::GetSingleElement() const
...
+
+   return t;
+}
+
+void Refinement::Set(int element, int type, real_t scale)
+{
+   index = element;
+   for (int i=0; i<3; ++i) { s[i] = 0.0; }
+   SetScale(ScaledType((char) type, scale));
+}
```
hughcars, line 602: Given there's no backwards compat issue, might as well just have `char type` in the arguments and not cast. If the int need to be kept, mark every method which uses it `MFEM_DEPRECATED` and use the char versions only internally to mfem.

mesh/ncmesh.cpp
```diff
@@ -487,6 +502,111 @@ int NCMesh::Face::GetSingleElement() const
...
+   for (int i=0; i<3; ++i) { s[i] = 0.0; }
+   SetScale(ScaledType(type, scale));
+}
+
+Refinement::Refinement(int index, int type, real_t scale)
+   : index(index)
+{
+   for (int i=0; i<3; ++i) { s[i] = 0.0; }
+   SetScale(ScaledType((char) type, scale));
+}
```
hughcars, line 578: Can delete and have `Refinement(int index, int type, real_t scale):Refinement(index, char(type), scale) {}` in the class definition instead. Better to just eliminate the `int` interface entirely though.

mesh/ncmesh.cpp
```diff
@@ -487,6 +502,111 @@ int NCMesh::Face::GetSingleElement() const
...
+
+int Refinement::GetType() const
```
hughcars, line 580: Don't need to be converting back and forth between integer and char:
```
char Refinement::GetType() const
{
   char t{0}; // Set the X, Y or Z bit
   for (int i = 0; i < 3; ++i)
      if (s[i] > real_t{0})
      {
         t |= (1 << i);
      }
   return t;
}
```
Then all the integer interfaces for setting the type can be removed and only use `char`. If the need for backing the enum with char is so strong, then should be consistent in the interface and force usage of char. I'd rather an enum class, to avoid all these type conversion traps, but that was ruled out due to backwards compat.

mesh/ncmesh.hpp
```diff
@@ -32,17 +32,38 @@ namespace mfem
...
-
-   Refinement(int index, int type = Refinement::XYZ)
-      : index(index), ref_type(type) {}
+   /// Refinement type XYZ, with scale 0.5.
+   Refinement(int index);
+   /// Default case of empty list @a refs is XYZ with scale 0.5.
+   Refinement(int index, const std::initializer_list<ScaledType> &refs);
+   /// Refine element with a single type and scale in all dimensions.
+   Refinement(int index, char type, real_t scale = 0.5);
+   Refinement(int index, int type, real_t scale = 0.5);
```
hughcars, line 56: All the `int` arged interfaces shouldn't exist, they're a type conversion issue waiting to happen. If back the enum with char is important enough to do in the first place, then avoiding all the type conversions back and forth should be too.

mesh/ncmesh.cpp
```diff
@@ -950,8 +1120,14 @@ void NCMesh::CheckIsoFace(int vn1, int vn2, int vn3, int vn4,
    }
 }
 
+void NCMesh::SetNodeScale(int p0, int p1, real_t scale)
+{
+   Node* node = nodes.Find(p0, p1);
+   if (node) { node->SetScale(GetScale(scale, p0 > p1)); }
```
hughcars, line 1126: If setting scale on a node, and its not found, that would probably be an assertable error? Or is speculatively setting the scale helpful?

mesh/ncmesh.cpp
```diff
@@ -1651,8 +1940,9 @@ void NCMesh::Refine(const Array<Refinement>& refinements)
    ref_stack.Reserve(refinements.Size());
    for (int i = refinements.Size()-1; i >= 0; i--)
    {
-      const Refinement& ref = refinements[i];
-      ref_stack.Append(Refinement(leaf_elements[ref.index], ref.ref_type));
+      Refinement ref = refinements[i];  // Copy
+      ref.index = leaf_elements[ref.index];
+      ref_stack.Append(ref);
```
hughcars, line 1945: Why the change to an extra line? I think this also gives up the possibility of move semantics (though I doubt Array uses them).

mesh/ncmesh.cpp
```diff
@@ -2751,6 +3043,8 @@ int NCMesh::QuadFaceSplitType(int v1, int v2, int v3, int v4,
    if (e1 >= 0 && e3 >= 0) { midf1 = FindMidEdgeNode(e1, e3); }
    if (e2 >= 0 && e4 >= 0) { midf2 = FindMidEdgeNode(e2, e4); }
 
+   s = 0.5;
```
hughcars, line 3046: This seems to only get used on the non-split branch, but if the face is not split, then surely the scale is supposed to be 0? Why not delete this and place `s=0.0` in the non-split branch below.

mesh/ncmesh.cpp
```diff
@@ -295,16 +296,30 @@ NCMesh::~NCMesh()
 #endif
 }
 
+void NCMesh::Node::SetScale(real_t s, bool overwrite)
```
hughcars, line 299: Probably worth checking that `0 < s < 1` so that you don't have coincident parent and child vertices, not sure what that would do to the mesh, but wouldn't be good.

hughcars (commented):
Tried the ref321 miniapp the resulting meshes look good, they also interact with the cut plane feature correctly. 

Minor issue with the usage of `int` in the interface for `Refinement`, I would suggest that `int` be deprecated and we use `char` consistently rather than C casts all over the place. If the ints have to remain, write them as dispatching in terms of the char version, then performing the cast on the input/outputs, and put `MFEM_DEPRECATED` around them.

hughcars:
The patch I am think of is:
```diff
diff --git a/mesh/ncmesh.cpp b/mesh/ncmesh.cpp
index fedbf02c3..6cb65d72c 100644
--- a/mesh/ncmesh.cpp
+++ b/mesh/ncmesh.cpp
@@ -570,41 +570,28 @@ Refinement::Refinement(int index, char type, real_t scale)
    SetScale(ScaledType(type, scale));
 }
 
-Refinement::Refinement(int index, int type, real_t scale)
-   : index(index)
-{
-   for (int i=0; i<3; ++i) { s[i] = 0.0; }
-   SetScale(ScaledType((char) type, scale));
-}
-
-int Refinement::GetType() const
+char Refinement::GetType() const
 {
-   int t = 0;
-   int d = 1;
-   for (int i=0; i<3; ++i)
-   {
-      if (s[i] > 0.0)
+   char t{0}; // Set the X, Y or Z bit
+   for (int i = 0; i < 3; ++i)
+      if (s[i] > real_t{0})
       {
-         t += d;
+         t |= (1 << i);
       }
-
-      d *= 2;
-   }
-
    return t;
 }
 
-void Refinement::Set(int element, int type, real_t scale)
+void Refinement::Set(int element, char type, real_t scale)
 {
    index = element;
    for (int i=0; i<3; ++i) { s[i] = 0.0; }
-   SetScale(ScaledType((char) type, scale));
+   SetScale(ScaledType(type, scale));
 }
 
-void Refinement::SetType(int type, real_t scale)
+void Refinement::SetType(char type, real_t scale)
 {
    for (int i=0; i<3; ++i) { s[i] = 0.0; }
-   SetScale(ScaledType((char) type, scale));
+   SetScale(ScaledType(type, scale));
 }
 
 NCMesh::Element::Element(Geometry::Type geom, int attr)
diff --git a/mesh/ncmesh.hpp b/mesh/ncmesh.hpp
index beed4df42..e86ab0180 100644
--- a/mesh/ncmesh.hpp
+++ b/mesh/ncmesh.hpp
@@ -53,15 +53,14 @@ struct Refinement
    Refinement(int index, const std::initializer_list<ScaledType> &refs);
    /// Refine element with a single type and scale in all dimensions.
    Refinement(int index, char type, real_t scale = 0.5);
-   Refinement(int index, int type, real_t scale = 0.5);
-   /// Return the type as an integer.
-   int GetType() const;
+   /// Return the type as char.
+   char GetType() const;
    /// Set the element, type, and scale.
-   void Set(int element, int type,
-            real_t scale = 0.5);  /// Uses @a scale in all dimensions
+   /// Uses @a scale in all dimensions.
+   void Set(int element, char type, real_t scale = 0.5);
    /// Set the type and scale, assuming the element is already set.
-   void SetType(int type,
-                real_t scale = 0.5);  /// Uses @a scale in all dimensions
+   /// Uses @a scale in all dimensions.
+   void SetType(char type, real_t scale = 0.5);
 private :
    void SetScale(const ScaledType &ref);
 };
```

mesh/ncmesh.cpp
dylan-copeland: It is definitely possible to create a disaster, although I expect this function to be called only after a Node is found (that is the usage so far). I added an MFEM_ASSERT in https://github.com/mfem/mfem/commit/e5cfc35f87969a015c88f892f8927267bd9ef4c5 so that performance won't be affected in an optimized build.

mesh/ncmesh.cpp
dylan-copeland: The purpose is to copy the 3 scaling factors in the array `s`. The constructor does not take this array. Originally, I wrote a constructor with optional arguments `sx, sy, sz`, which we decided was too awkward.

mesh/ncmesh.cpp
dylan-copeland: This value is not used in the non-split case, so I set it to the invalid value 0 as you suggested in e5cfc35f87969a015c88f892f8927267bd9ef4c5

mesh/ncmesh.hpp
dylan-copeland: Removed in e5cfc35f87969a015c88f892f8927267bd9ef4c5

mesh/mesh.hpp
dylan-copeland: Done in e5cfc35f87969a015c88f892f8927267bd9ef4c5

mesh/ncmesh.cpp
dylan-copeland: The assert fails in some cases, so I removed it in 95a5c503b10a44d919ad498575311606924c5539. It fails when called for certain pairs of midpoints, e.g. `ncmesh.cpp:1456`: `SetNodeScale(mid01, mid23, scale_y);`. It is not always known what pairs of midpoints have a `Node` defined, but when the `Node` does exist, its scale should be set.

hughcars (approved):
LGTM

cjvogl (commented):
Everything looks good from a code execution standpoint; however, I add some inline comments/suggestions to consider.  

Also, I believe reverting the changes to the distance miniapp are still on the to-do list?

mesh/mesh.cpp
```diff
@@ -10722,7 +10723,7 @@ void Mesh::RefineAtVertex(const Vertex& vert, real_t eps, int nonconforming)
       }
       if (refine)
       {
-         refs.Append(Refinement(i));
+	refs.Append(Refinement(i));
```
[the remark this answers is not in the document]
dylan-copeland: The indentation changes were not intended, and I thought `astyle` should not allow such changes. I guess it allows different indentations to be set at the beginning of a section enclosed by braces? Anyway, I reverted those changes.

mesh/ncmesh.hpp
```diff
@@ -516,6 +534,14 @@ class NCMesh
...
       bool UnrefVertex() { --vert_refc; return vert_refc || edge_refc; }
       bool UnrefEdge()   { --edge_refc; return vert_refc || edge_refc; }
+
+      real_t GetScale() const { return scale; }
+      void SetScale(real_t s, bool overwrite = false);
+
+   private:
+      real_t scale;  ///< Scale from struct Refinement, default 0.5
+      bool scaleSet; ///< Indicates whether scale is set and cannot be changed
+      static constexpr real_t scaleTol = 1.0e-8; ///< Scale comparison tolerance
```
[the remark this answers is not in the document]
dylan-copeland: I set this to another hard-coded value in the not-double case, see 181030cc291a3dd2e783eb773775df902d91fee1. These hard-coded values are questionable and not established from rigorous testing, but this tolerance is used just to check for consistency in scalings, and I think it should be safe to use a rather large tolerance as I have set.

miniapps/meshing/ref321.cpp
```diff
@@ -0,0 +1,267 @@
...
+      const int p = cf.embeddings[i].parent;
+      if (p == elem)
+      {
+         children.Append(i);
+      }
+   }
+}
+
+// Refine 3:1 via 2 refinements with scalings 2/3 and 1/2.
+void Refine31(Mesh & mesh, int elem, int type)
```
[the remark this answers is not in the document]
dylan-copeland: Right, it was casting from `char` to `int` to `char`. This is fixed in 181030cc291a3dd2e783eb773775df902d91fee1.

mesh/ncmesh.hpp
```diff
@@ -673,17 +701,22 @@ class NCMesh
...
+   Array<real_t> reparent_scale;  ///< scale associated with reparents (tmp)
 
    Table derefinements; ///< possible derefinements, see GetDerefinementTable
 
    /** Refine the element @a elem with the refinement @a ref_type
-       (c.f. Refinement::enum) */
-   void RefineElement(int elem, char ref_type);
+       (c.f. Refinement::enum) and scale in each direction. */
+   void RefineElement(int elem, char ref_type, real_t scale_x = 0.5,
+                      real_t scale_y = 0.5, real_t scale_z = 0.5);
```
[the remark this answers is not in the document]
dylan-copeland: The reason for the awkward input of 3 scalings is to preserve backward compatibility. I think the most sensible interface would be
`void RefineElement(const Refinement & ref)`
since `struct Refinement` now contains the element index, refinement type, and 3 scalings. This could be added as a new function, but the old function needs to be maintained. We can have the old and new functions, but some function has to contain the actual code, while the other just calls that function. The code in `RefineElement` actually uses the 5 input variables explicitly, so it is convenient to keep them as separate variables. For example, if we only input `Refinement & ref`, we would still want things like `scale_z = ref.s[2]`. So we could add a new function with the single argument `Refinement & ref`, but it would simply wrap the existing function, and I don't see that simplifying much. The API would be cleaner if we just replace the 5 arguments with `Refinement & ref`, but that breaking change would need to be done in MFEM 5.0. I think you have a good suggestion, but I'm not sure it can really help until 5.0.

mesh/ncmesh.hpp
cjvogl: Even though I wasn't very active in this thread, I suppose it's on me as the starter of the thread to resolve it... please feel free to re-open if neeeded

mesh/mesh.cpp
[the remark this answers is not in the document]
cjvogl: I also thought `astyle` would catch such things... but perhaps you're right.  Another oddity is that in 181030c, I do see changes on the corresponding indentation lines, but the indentation still looks off in the diff.  I'll add inline suggestions to highlight where.

mesh/mesh.cpp
```diff
@@ -10782,7 +10783,7 @@ void Mesh::RefineAtVertex(const Vertex& vert, real_t eps, int nonconforming)
       }
       if (refine)
       {
-         refs.Append(Refinement(i));
+	 refs.Append(Refinement(i));
```
cjvogl, line 10786:
```suggestion
	       refs.Append(Refinement(i));
```

mesh/mesh.cpp
```diff
@@ -10797,7 +10798,7 @@ bool Mesh::RefineByError(const Array<real_t> &elem_error, real_t threshold,
    {
       if (elem_error[i] > threshold)
       {
-         refs.Append(Refinement(i));
+	 refs.Append(Refinement(i));
```
cjvogl, line 10801:
```suggestion
	       refs.Append(Refinement(i));
```

mesh/ncmesh.hpp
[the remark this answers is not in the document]
cjvogl: > So we could add a new function with the single argument Refinement & ref, but it would simply wrap the existing function, and I don't see that simplifying much.

I see your point about effectively wrapping this function with `RefineElement(const Refinement &ref)` and then having `RefineElement(int elem, char ref_type, real_t scale=0.5)` call the former to keep backwards compatibility, but I personally prefer that API over having to, for example, specify dummy values for `scale_x` and `scale_y` for Z refinement... I find it nicely provides separate functions for isotropic and anisotropic refinement, even if one simply calls the other.

All that said, this is my subjective opinion and there's already one approval for this PR, so I don't want to hold it this important functionality over this (i.e., I'll approve once the indentation issues are fixed).

cjvogl (approved):

mesh/ncmesh.hpp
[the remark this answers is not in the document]
dylan-copeland: I added this wrapping in c03257233f58c2c78cfea963eaa520c627bf2dc1. For this PR, it only simplifies one line (ncmesh.cpp:1943), but I agree it is good to add the new function for future use.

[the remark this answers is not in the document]
dylan-copeland: Unfortunately, even this new function does not eliminate the 3 scaling parameters that you would like to avoid. I don't see a good way to do that without breaking backward compatibility. My hope is that in the future, we can just remove that function and keep the new one, which is a breaking change.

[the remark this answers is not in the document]
dylan-copeland: I restored the old interface, even without a single `scale` argument, in 4b87236e738c15b867faeb2277911f1efa50424e. The impact was much less than I expected, so the code remains very simple with this change.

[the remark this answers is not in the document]
cjvogl: This looks great!  Thanks for doing all the iterations!

tzanio:

tzanio:

tzanio:
Some modern art thanks to the new `ref321` miniapp

Screenshot 2024-12-17 at 5 15 45 PM

tzanio (approved):
Thanks for adding this @dylan-copeland. 🙏

Really impressive work! 🚀

tzanio:

dylan-copeland:
> Some modern art thanks to the new `ref321` miniapp
> 
Nice picture. To be even more artistic, one could use the golden ratio `0.6180339887498948` instead 2/3 for the first refinement.

tzanio (approved):
Thanks @dylan-copeland for adding another awesome capability to our AMR toolbelt! 🏆 

@cjvogl and @hughcars -- thank you for the careful review! 👍

PR #4416, opened 2024-07-19
Add `FiniteElementSpace::BuildDofToBdrArrays()` and Minor Doc Clarification

j-signorelli:
This PR implements the same functionality as `FiniteElementSpace::BuildDofToArrays()`, `FiniteElementSpace::GetElementForDof()`, and `FiniteElementSpace::GetLocalDofForDof()` but for boundary elements. This is useful particularly if needing to get the first boundary element's `ElementTransformation` and the `IntegrationPoint` for a given ldof.

Further, the documentation is slightly clarified for these and the newly-implemented member functions, as (at least for me) the phrasing `GetLocalDofForDof` was confusing (it accepts an [l(ocal)dof index](https://docs.mfem.org/html/classmfem_1_1FiniteElementSpace.html#ldof)).

| [#4416](https://github.com/mfem/mfem/pull/4416) | @j-signorelli | @tzanio | @najlkin + @pazner | 8/3/24 | 8/30/24 | 9/10/24 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:

pazner (commented):
The new member data arrays `dof_bdr_elem_array` and `dof_bdr_ldof_array` should be deleted in `FiniteElementSpace::Destroy`.

I wonder if it's better to build the arrays lazily on first call to either `GetBdrElementForDof` or `GetBdrLocalDofForDof` (we could do the same thing for `BuildDofToArrays`).

j-signorelli:
> The new member data arrays `dof_bdr_elem_array` and `dof_bdr_ldof_array` should be deleted in `FiniteElementSpace::Destroy`.
> 
> I wonder if it's better to build the arrays lazily on first call to either `GetBdrElementForDof` or `GetBdrLocalDofForDof` (we could do the same thing for `BuildDofToArrays`).

That sounds like a better approach, but the only issue is that `Get____` are all `const`, while `Build___` are not, so we cannot do this unless we define the arrays as `mutable`

pazner:
Yes, the arrays would have to be made mutable. I don't have a strong preference one way or the other.

pazner (approved):

najlkin:
I would support to build them on-demand in the getters, because: 1) it reduces the size of the public interface, which is already large, as you may remove the build methods from it completely 2) it eliminates the risk of getting segfaults when users are not aware of the build methods

j-signorelli:
@najlkin Got it, I added the calls to build the arrays to the getters in the last commit 70941f93f0559db43bb33af42942bfafe632966d, but would you also suggest moving the current implementation of the public build functions to be private, and deprecating the public ones with `MFEM_DEPRECATED` (and having them just call the private ones)?

najlkin:
Yes, I would do that if @pazner agrees (?) 😉

pazner:
Yes, that sounds fine to me. `BuildDofToBdrArrays` would be made private, `BuildDofToArrays` remains public but deprecated, and an additional "internal" `BuildDofToArrays_` would be added (called by `GetElementForDof` and `GetLocalDofForDof`).

pazner (approved):

fem/fespace.hpp
```diff
@@ -1163,18 +1174,21 @@ class FiniteElementSpace
...
-
-   /// Return the index of the first element that contains dof @a i.
-   /** This method can be called only after setup is performed using the method
-       BuildDofToArrays(). */
-   int GetElementForDof(int i) const { return dof_elem_array[i]; }
-   /// Return the local dof index in the first element that contains dof @a i.
-   /** This method can be called only after setup is performed using the method
-       BuildDofToArrays(). */
-   int GetLocalDofForDof(int i) const { return dof_ldof_array[i]; }
+   /** @brief Deprecated. See @a BuildDofToArrays_. */
```
pazner, line 1177:
```suggestion
   /// Deprecated. This function is not required to be called by the user.
```

fem/fespace.hpp
```diff
@@ -1163,18 +1174,21 @@ class FiniteElementSpace
...
-   /// Return the index of the first element that contains dof @a i.
-   /** This method can be called only after setup is performed using the method
-       BuildDofToArrays(). */
-   int GetElementForDof(int i) const { return dof_elem_array[i]; }
-   /// Return the local dof index in the first element that contains dof @a i.
-   /** This method can be called only after setup is performed using the method
-       BuildDofToArrays(). */
-   int GetLocalDofForDof(int i) const { return dof_ldof_array[i]; }
+   /// Deprecated. This function is not required to be called by the user.
+   MFEM_DEPRECATED void BuildDofToArrays() const { BuildDofToArrays_();};
```
najlkin, line 1178:
```suggestion
   MFEM_DEPRECATED void BuildDofToArrays() const { BuildDofToArrays_(); }
```

tests/unit/fem/test_build_dof_to_arrays.cpp
```diff
@@ -110,8 +110,6 @@ TEST_CASE("Build Dof To Arrays",
             FiniteElementSpace fespace(mesh, fec);
             int size = fespace.GetTrueVSize();
 
-            fespace.BuildDofToArrays();
-
             for (int i = 0; i<size; i++)
```
najlkin, line 113: Not super important, but if you have such a nice test, how about just adding here a loop over boundaries to check the new method? 😉

j-signorelli: @najlkin Apologies for the late reply - I have not been available to work on this.

I could but I think I would also have to update the `GetMesh` function at the bottom of this file to include boundary elements for all the different meshes which will take a good deal of time - it does not seem that boundary elements are automatically included?

najlkin: It is easy, just add `mesh->GenerateBoundaryElements()` at the end of `GetMesh()` 😉 .

najlkin (approved):
LGTM 😉

tzanio:

tzanio:
This PR is failing in our nightly regression testing with errors like:

```
Build Dof To Arrays (Parallel)
  Mesh Type: 4, Basis Type: 0
mfem/tests/unit/fem/test_build_dof_to_arrays.cpp:225
mfem/tests/unit/fem/test_build_dof_to_arrays.cpp:225: FAILED:
  {Unknown expression after the reported line}
```

and

```
Assertion failed: (i >= 0 && i < size) is false:
 --> Row index -1 is out of range [0,1)
 ... in function: void mfem::Table::GetRow(int, mfem::Array<int>&) const
 ... in file: general/table.cpp:261
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 1
make[1]: *** [makefile:264: punit_tests-test-par] Error 137
```

tests/unit/fem/test_build_dof_to_arrays.cpp
```diff
@@ -212,7 +244,10 @@ TEST_CASE("Build Dof To Arrays (Parallel)",
             ParFiniteElementSpace fespace(&pmesh, fec);
             HYPRE_Int size = fespace.GetTrueVSize();
 
-            fespace.BuildDofToArrays();
+            // Get all boundary ldofs
+            Array<int> bdr(1); bdr = 1;
+            fespace.GetEssentialVDofs(bdr, all_bdr_ldofs_marked);
+            FiniteElementSpace::MarkerToList(all_bdr_ldofs_marked, all_bdr_ldofs);
```
najlkin, line 250: Replace this by `GetEssentialTrueDofs()`, which returns a list directly, but I am not sure if that is the problem really. The split by METIS should be element-wise, so boundary should be always own 🤔 .

tests/unit/fem/test_build_dof_to_arrays.cpp
```diff
@@ -225,15 +260,35 @@ TEST_CASE("Build Dof To Arrays (Parallel)",
...
+               int elemdof = (dofs[l] >= 0) ? dofs[l] : (-1 - dofs[l]);
+
+               if (i != elemdof) { num_ldof_fails++; }
+
+               int bdr_e = fespace.GetBdrElementForDof(i);
+               int bdr_l = fespace.GetBdrLocalDofForDof(i);
+
+               if (all_bdr_ldofs.Find(i) >= 0) // if this is a bdr ldof
+               {
+                  if (bdr_e < 0 || bdr_e >= pmesh.GetNBE()) { num_bdr_elem_fails++; }
```
najlkin, line 272: Would be better to add `continue` here and elsewhere to not get into segfaults if it fails.

tests/unit/fem/test_build_dof_to_arrays.cpp
j-signorelli: Only issue however is that I need the ldofs, not the tdofs. I am looking into this myself now as well -- I think that iit *might* be a negative `int bel` being sent to `GetBdrElementDofs`

najlkin: Yes, that is probably the reason, because `all_bdr_ldofs` includes -1 probably, so it might come from here. There is no difference between TDof and VDof, they are just "owned", while VDofs might be ghosts (i.e. neighbor's).

j-signorelli:
For a local branch i have with this one merged with next, I am seeing that this error occurs when running on 3 or more procs. A DOF is being interpreted as  a bdr one (`all_bdr_ldofs.Find(i) >= 0` is `true'), while `GetBdrElementForDof` is returning -1 (essentially indicating that this DOF is not a boundary one).

It may be an issue with my retrieval of all boundary ldofs in `all_bdr_ldofs`

tests/unit/fem/test_build_dof_to_arrays.cpp
najlkin: Btw, it would be better to not use the name "LDofs" here as it might be confused with the local dofs in the sense below.

najlkin:
Try what I suggested above ☝️ .

tests/unit/fem/test_build_dof_to_arrays.cpp
najlkin: LDof is normally used in the sense of the local index within an *element*, so it is below the index to the array of the element Dofs. Please rename it here to TDofs 😉 .

j-signorelli: @najlkin This confusion was actually part of the doc change I clarified in the `GetLocalDofForDof` and `GetBdrLocalDofForDof` earlier in this PR --> I'm using the ldof definition [here](https://docs.mfem.org/html/classmfem_1_1FiniteElementSpace.html#ldof), where "local" refers to local on a given MPI proc. I had been quite confused when interpreting `GetLocalDofForDof` until digging in and seeing it took as an argument an ldof in the context of local to a rank as defined in the link above.

I've always treated ldof indices and tdof indices as generally disjoint (in that a tdof index and an ldof index for a given non-ghost DOF may be entirely different) - is this not the case?

I did try implementing the changes you suggested using `GetEssentialTrueDofs` but am still getting similar errors.

najlkin: I do not know who wrote that new documentation, but it is completely misleading 🤯 . We should fix it, unless there is a shift in the terminology, but obviously `GetLocalDofForDof()` does not follow that. @tzanio , do you know where this terminology comes from?

najlkin: In any case, @j-signorelli , have you confirmed that you have `-1` in the list returned by `GetEssentialTrueDofs()`?

najlkin: Ohh, actually it can be, because you take just the first boundary, but it might be neighbors's if you are in a corner, i.e. there are more boundaries. You must overwrite the array if the associated boundary element is `-1` 😮 .

fem/fespace.cpp
```diff
@@ -503,6 +503,31 @@ void FiniteElementSpace::BuildDofToArrays()
...
+   dof_bdr_ldof_array.SetSize (ndofs);
+   dof_bdr_elem_array = -1;
+   for (int i = 0; i < mesh -> GetNBE(); i++)
+   {
+      const int *dofs = bdr_elem_dof -> GetRow(i);
+      const int n = bdr_elem_dof -> RowSize(i);
+      for (int j = 0; j < n; j++)
+      {
+         int dof = DecodeDof(dofs[j]);
+         if (dof_bdr_elem_array[dof] < 0)
```
najlkin, line 522: Here, check if `dof_bdr_elem_array[dof] < 0` and then overwrite it too.

tests/unit/fem/test_build_dof_to_arrays.cpp
najlkin: Regarding the terminology, I just spoke with @tzanio and "local dof" should be primarily used in the sense of the MPI rank local. Unfortunately, the other use is also relatively common. It would be better to use a different name like element index or dof (but element dof is usually understood as a dof for an element), but it is how it is.

tzanio: Sorry for the late response, @najkin is correct in his description above ☝️ 

We have a better/more clear way to think about this, but that hasn't made it into the code.

Specifically, in the [operator decomposition](https://mfem.org/performance/#finite-element-operator-decomposition) where we have `T`-, `L`-, `E`-vectors we can talk about their entries as `TDofs`, `LDofs` and `EDofs` and each case we can refer to either global number of the dof or the local number inside the appropriate object -- MPI ranks for `TDofs` and `LDofs` and element for `EDofs`.

That being said, @j-signorelli -- do you understand what is causing the issue and how to fix it? Could it be that some of the MPI ranks are empty and that is not handled?

j-signorelli: > In any case, @j-signorelli , have you confirmed that you have `bdr_el==-1` for DOFs from the list returned by `GetEssentialTrueDofs()`?

Yes I can confirm that this is occurring when I use both `GetEssentialVDofs()` with `MarkerToList()` (LDofs) or `GetEssentialTrueDofs()` (tdofs). The `GetBdrElementForDof()` method is returning -1 for an index that is, according to those two methods, a bdr dof. When I add in the `continue` statement as you suggested below, the test still is not completing due to this error:

      Abort(134391310) on node 1 (rank 1 in comm 0): Fatal error in internal_Waitany: Message truncated, error stack:
      internal_Waitany(108)....: MPI_Waitany(count=2, array_of_requests=0x62edb1b1bd80, indx=0x7ffc72e5dacc, status=0x1) failed
      MPIR_Waitany(1281).......: 
      MPIDIG_recv_type_init(72): Message from rank 0 and tag 40822 truncated; 16 bytes received but buffer size is 36

This seems to be occurring during mesh decomposition. Notably I'm also getting these warnings:

    MFEM Warning: Non-positive attributes on the boundary!
     ... in function: virtual void mfem::Mesh::SetAttributes()
     ... in file: /home/j-signorelli/software/MFEM/git_repo/mesh/mesh.cpp:1619

but I am not really sure why because I believe I am properly setting all boundary elements to have an attribute of 1 at the end of `GetMesh()`:

    for (int i = 0; i < mesh->GetNBE(); i++) { mesh->SetBdrAttribute(i, 1); }


When I comment out the above line, I no longer get the fatal error. So just to summarize for completeness:

1. Including `continue` to prevent the seg fault yields another error I pasted above -- this new error doesn't occur if I comment out the `SetBdrAttribute` loop.
2. The non-positive attributes warning is showing when it really should not be as I am specifying them all to be 1.
3. The original error persists regardless and I am still unsure why - it may have something to do with the above problems.

fem/fespace.cpp
j-signorelli: Sorry I'm not sure what you mean here -- this conditional is to make sure that for a given bdr dof, only the first boundary element index `i` when looping from 0 to `GetNBE()` for that dof is associated with it.

najkin:
Приветствую. Мне по ошибке ,скорее всего, приходит ваше обсуждение проекта. Исправьте адресата на правильный, человек, который с вами работает, не получает информацию.
--
Отправлено из Mail.ru для Android четверг, 22 августа 2024г., 16:42 +03:00 от Joseph Signorelli  ***@***.*** :

tests/unit/fem/test_build_dof_to_arrays.cpp
j-signorelli: Only `MeshType::SEGMENT` is failing to properly set the bdr attributes. Instead of 1, it is being set to -1.

najlkin:
> Приветствую. Мне по ошибке ,скорее всего, приходит ваше обсуждение проекта. Исправьте адресата на правильный, человек, который с вами работает, не получает информацию. -- Отправлено из Mail.ru для Android четверг, 22 августа 2024г., 16:42 +03:00 от Joseph Signorelli ***@***.*** :
> […](#)
> @j-signorelli commented on this pull request. ---------------------------------------------------------------------- In fem/fespace.cpp : +{ + if (dof_bdr_elem_array.Size()) { return; } + + BuildBdrElementToDofTable(); + + dof_bdr_elem_array.SetSize (ndofs); + dof_bdr_ldof_array.SetSize (ndofs); + dof_bdr_elem_array = -1; + for (int i = 0; i < mesh -> GetNBE(); i++) + { + const int *dofs = bdr_elem_dof -> GetRow(i); + const int n = bdr_elem_dof -> RowSize(i); + for (int j = 0; j < n; j++) + { + int dof = DecodeDof(dofs[j]); + if (dof_bdr_elem_array[dof] < 0) Sorry I'm not sure what you mean here -- this conditional is to make sure that for a given bdr dof, only the first boundary element index i when looping from 0 to GetNBE() for that dof is associated with it. — Reply to this email directly, view it on GitHub , or unsubscribe . You are receiving this because you were mentioned. Message ID: @ github . com>

@najkin , we are sorry, there was a typo in the name. It was not addressed to you.

fem/fespace.cpp
najlkin: Ahh, right, that is a nonsense. I thought the loop goes differently.

tests/unit/fem/test_build_dof_to_arrays.cpp
najlkin: Hmm, this is interesting, it seems there is a bug somewhere in `GenerateBoundaryElements()`, `SetBdrAttribute()` or `FinalizeTopology()` 🤔 .

j-signorelli: @najkin I am not sure why only segment was failing, but [61a49b8](https://github.com/mfem/mfem/pull/4416/commits/61a49b8ca080a4522fb5db431542043f7a04f69b) seems to have resolved this problem.

najkin:
Ребят, пожалуйста, удалите меня из вашей рассылки писем.
--
Отправлено из Mail.ru для Android четверг, 22 августа 2024г., 19:29 +03:00 от Joseph Signorelli  ***@***.*** :

tests/unit/fem/test_build_dof_to_arrays.cpp
najlkin: But `FinalizeTopology()` has the default argument set to `true`, so I do not see how that can help 🤔 .

najlkin: Btw, please mind the spelling, it is `najlkin`, not `najkin`. That other guy got a message instead again 😄 .

najlkin: Ok, so what I have tested so far:

- `mesh->FinalizeTopology()` generates boundaries correctly, so I do not understand why there was a need to add anything as [you mentioned before](https://github.com/mfem/mfem/pull/4416#discussion_r1722117704) ?
- `mesh->GenerateBoundaryElements()` *before* `mesh->FinalizeTopology()` does something strange, I need to analyze that... 🤔 
- `mesh->GenerateBoundaryElements()` *after* `mesh->FinalizeTopology()` (`mesh->FinalizeTopology(false)`) works correctly though 😉

j-signorelli: @najlkin My apologies on the spelling! Oh I missed that regarding the default argument already being true... Then `GenerateBoundaryElements` must be problematic in some way with `MeshType::SEGMENT`.

j-signorelli: @najlkin Interesting, thank you for your assistance -- I hadn't realized that bdr elements would be automatically included.

I'm currently looking into why this MPI Waitany error is arising.

najlkin: In any case, are you getting any errors with the current version of the unit test? If not, then we should try to remerge it to `next`, because it generates massive error logs 😄 .

j-signorelli: I am still getting errors whenever I run on >= 3 mpi procs.

At this point, I have determined that the MPI error occurs ONLY after a `bdr_e==-1` situation occurs, where:

    if (bdr_e < 0 || bdr_e >= pmesh.GetNBE()) { num_bdr_elem_fails++; continue; }

evaluates to true on a given rank. When this happens, that MPI error occurs on that same rank for the *next* mesh decomposition here:

    ParMesh pmesh(MPI_COMM_WORLD, *mesh);

 I have no clue why this is occurring.

j-signorelli: I actually had thought that too but I got rid of the `continue` and instead changed it to:

    bdr_e = fespace.GetBdrElementForDof(all_bdr_ldofs[0]);
    bdr_l = fespace.GetBdrLocalDofForDof(all_bdr_ldofs[0]);
    
inside of the conditional and this same error was still occurring! I also didn't see any communications going on there either.

najlkin: About that `GenerateBoundaryElements()`, it should not be called before finalization of topology, because it actually just deletes the boundary structures as it cannot generate the boundaries correctly yet. Than it was actually `FinalizeTopology()` generating the boundary, but because of that deleted boundary structures, it copies the face elements from inside, which have -1 attribute for segment for some reason, so it becomes the boundary attribute. I do not know, maybe something more is going on, but it is a minor bug only.

najlkin: @termi-official , do you know why `Point` has the default attribute -1 unlike everything else (`Segement`, `Triangle`,...)? (@tzanio said you might know 😅 )

termi-official: From comparing the git blames I think it might just have been forgotten on the update to 1.2 https://github.com/mfem/mfem/commit/29ae8fb6ac9fce927fdfd01ef796f9df38f76e97

najlkin: Hmm, in other primitives than `Segment` it seems to be from version 1.0. Just that one was changed from -1 to 1 🤔 . But @tzanio does not know, so it is long buried in history why it was/is like that. I will open a separate PR for that, thanks 😉 .

najlkin: Btw, that is why there is: `mesh->AddBdrPoint(0);` , which is/was a workaround for the boundary attributes.

j-signorelli:
Found something interesting -- I believe that there may be a bug with either `ParFiniteElementSpace::Synchronize`, or something where the `Table *bdr_elem_dof` does not include all LDofs in parallel appropriately.

Specifically for `MeshType::Triangle2C`, `BasisType::H1` run on 3 MPI procs, here is the info for Rank 1, where the code is crashing first.

If I loop over all boundary elements  (BEs) (0 to `mesh->GetNBE()`) and print their individual ELDofs (the Dofs associated for ONE element) and coordinates, I get:
BE 0 | elemdof 0 | (3.14159, 1.11072)
BE 0 | elemdof 1 | (3.14159, 2.22144)
BE 0 | elemdof 2 | (3.14159, 1.41772) 
BE 0 | elemdof 3 | (3.14159, 1.91445)

If I print all LDof indices that do not return `-1` from `GetBdrElementForDof` + their associated BE, ELDof, coords:
BE 0 | elemdof 1 | LDof 0 | (3.14159, 2.22144) 
BE 0 | elemdof 0 | LDof 3 | (3.14159, 1.11072) 
BE 0 | elemdof 3 | LDof 8 | (3.14159, 1.91445) 
BE 0 | elemdof 2 | LDof 9 | (3.14159, 1.41772) 


The boundary LDof Indices from GetEssentialVDofs:
0 **2** 3 8 9  (<-- 2 should not be here! it is not associated with any BE!)

Just for completeness, the boundary TDof Indices from GetEssentialTrueDofs:
0 1 6 7 

The location of LDof index 2 is (1.5708, 0) and matches a boundary LDof location that exists on another rank.


To get `all_bdr_ldofs`, if I instead call:

    static_cast<FiniteElementSpace>(fespace).GetEssentialVDofs(bdr,all_bdr_ldofs);
    // Only difference between this and ParFiniteElementSpace is an additional call to Synchronize()

then everything works perfectly fine across all MPI procs and the test passes. **Using `FiniteElementSpace::GetEssentialVDofs()` instead of `ParFiniteElementSpace::GetEssentialVDofs()` ensures that the output from `GetEssentialVDofs` matches perfectly with indices that do not return `-1` from `GetBdrElementForDof`.**

najlkin:
That is still strange, because as I wrote earlier, boundaries should be always "our", i.e., they cannot be shared as the split is element-based. I cannot be reproduce it, it always works for me. Perhaps it depends on METIS version, where METIS does some strange cut in your case and also in the CI 🤔 . Upload that static cast and we will see if autotest will confirm this theory 😉 .

tests/unit/fem/test_build_dof_to_arrays.cpp
```diff
@@ -210,30 +247,59 @@ TEST_CASE("Build Dof To Arrays (Parallel)",
...
+               int elemdof = (dofs[l] >= 0) ? dofs[l] : (-1 - dofs[l]);
+
+               if (i != elemdof) { num_elemdof_fails++; }
+            }
+
+            // Check bdr elements
+
+            // Get all boundary ldofs
+            Array<int> bdr(1); bdr = 1;
+            static_cast<FiniteElementSpace>(fespace).GetEssentialVDofs(bdr,
```
najlkin, line 273:
```suggestion
            static_cast<FiniteElementSpace&>(fespace).GetEssentialVDofs(bdr,
```
You forgot reference here, it calls constructor then 😮

j-signorelli: Thank you!

najlkin:
@tzanio , could you remerge it in `next` please? It seems to be fixed now.

j-signorelli:
Was thinking more about it, I think that this is the likely scenario of what was causing the issues:
image

`ParFiniteElementSpace::Synchronize` includes that additional DOF without an associated BE.

najlkin:
Yes, that might be, thanks for the picture 👍 . It does not happen for me for some reason. Maybe my METIS tries to avoid this somehow 🤔 . Anyway, it is correct what you did with the static cast 😉 . It is basically what I meant with that TDofs, but it works only in basic scenarios, otherwise the indexing of TDofs is a bit different.

j-signorelli:
> Yes, that might be, thanks for the picture 👍 . It does not happen for me for some reason. Maybe my METIS tries to avoid this somehow 🤔 . Anyway, it is correct what you did with the static cast 😉 . It is basically what I meant with that TDofs, but it works only in basic scenarios, otherwise the indexing of TDofs is a bit different.

Yeah that was another interesting thing I observed taking a deep dive into this -- the TDof indices were very much not a subset of the LDof indices.

Regardless thank you _so much_ for all of your help  @najlkin !

tzanio:

tzanio:
> It seems to be fixed now.

Autotest is still giving errors :(

j-signorelli:
> Was thinking more about it, I think that something like this is the likely scenario of what was causing the issues in an H1-context: image
> 
> `ParFiniteElementSpace::Synchronize` includes that additional DOF without an associated BE.

@najlkin It may be the same situation I outlined here above, `Synchronize()` might just help avert this situation more frequently.

I have a stash locally when I was debugging that specifically checks whether non-`-1` entries from `GetBdrElementForDof()` are of the same size and one-to-one with the results from `ParFiniteElementSpace::GetEssentialVDofs()`, I think that it would be useful to check this same thing for the above code (`FiniteElementSpace::GetEssentialVDofs()` as a debug commit.

najlkin:
We have eliminated the problem with neighbor's essential dofs by bypassing `Synchronize` with the static cast. Locally, this cannot happen for conforming meshes as `GetEssentialVDofs()` identifies essential dofs by boundary attributes associated with boundary elements. However, something like that could happen for non-conforming meshes, see [the implementation](https://docs.mfem.org/html/fespace_8cpp_source.html#l00514). So first of all check maybe if there are non-conforming meshes. I do not see them, but just in case...
For debugging, you may print everything only when `num_bdr_elem_fails==1` just before `continue`.

tzanio:
If that's useful, when the following run fails for me:

```
cd tests/unit
mpirun -np 4 ./punit_tests "Build Dof To Arrays (Parallel)"
```

the parallel mesh looks as follows:

Screenshot 2024-08-27 at 11 01 16 PM

j-signorelli:
I am able to get a failing test for specifically 4 MPI procs as well, and I am seeing the same issue of an additional LDof index indicated by `FiniteElementSpace::GetEssentialVDofs` not associated with any boundary element. So maybe there are still situations where this issue can occur even without using `Synchronize()` @najlkin ?

If this is the case then it may be best to just restructure the test - there aren't any issues with the implemented functions in this PR.

najlkin:
Great, if you can reproduce that it will be much easier, I cannot, all tests pass for me. Probably because I use METIS 5. @j-signorelli , I agree with you that it might not be related to this PR, but implementation `GetEssentialVDofs()` is very simple. Nothing else than what you would do by hand, i.e., iteration over boundary elements and marking dofs if the attribute matches. Unless the mesh is non-conforming, have you checked that? (by `Nonconforming()`). Otherwise it is a bit of mystery 🤔

j-signorelli:
@najlkin I am also using METIS 5. Forgot to check nonconforming, thanks for the reminder! Just did: the failing case has `fespace.Nonconforming() == false` so it can't be that; the debug commit also just loops over the boundary elements but isn't picking up on the additional DOF I am getting for some reason from `GetEssentialVDofs()` -- I'll take a closer look sometime today at the failing case on my end

najlkin:
This is really strange, maybe check if you are still not getting to the parallel version of `GetEssentialVDofs` somehow? Could be a compiler error with the default argument maybe? It is a wild theory, but does `GetEssentialVDofs(bdr, all_bdr_ldofs_marked, -1)` make difference?

j-signorelli:
> This is really strange, maybe check if you are still not getting to the parallel version of `GetEssentialVDofs` somehow? Could be a compiler error with the default argument maybe? It is a wild theory, but does `GetEssentialVDofs(bdr, all_bdr_ldofs_marked, -1)` make difference?

Yes this seems to be the problem! Even after the cast, it seems sometimes that the parallel version of `GetEssentialVDofs` is called. Explicitly ensuring that the base one is called in the last commit is fixing the issue on my end (and hopefully will with autotest).

tzanio:
I can confirm that with https://github.com/mfem/mfem/pull/4416/commits/a80d35b3871c8692d3d158f412536d2b8ba34572  the unit test pass for me

tzanio:

hughcars:
> > This is really strange, maybe check if you are still not getting to the parallel version of `GetEssentialVDofs` somehow? Could be a compiler error with the default argument maybe? It is a wild theory, but does `GetEssentialVDofs(bdr, all_bdr_ldofs_marked, -1)` make difference?
> 
> Yes this seems to be the problem! Even after the cast, it seems sometimes that the parallel version of `GetEssentialVDofs` is called. Explicitly ensuring that the base one is called in the last commit is fixing the issue on my end (and hopefully will with autotest).

Pretty sure that's no bug, that's polymorphism. Casting to a reference to the base `FiniteElementSpace` does nothing to alter the virtual function call: that cast should do nothing. `fespace.FiniteElementSpace::GetEssentialVDofs(...)` is the syntax to forcibly use the base class method instead.

najlkin:
@hughcars , you are right, I was not thinking that way much, because it was passing for me. We are perhaps missing something here, but if it works in the autotest it is fine with me :wink:

tests/unit/fem/test_build_dof_to_arrays.cpp
```diff
@@ -464,7 +531,8 @@ Mesh * GetMesh(MeshType type)
          mesh->AddTet(8, 2, 7, 5);
          break;
    }
-   mesh->FinalizeTopology();
+   mesh->FinalizeTopology(true);
+   for (int i = 0; i < mesh->GetNBE(); i++) { mesh->SetBdrAttribute(i, 1); }
```
najlkin, line 535:
```suggestion
   mesh->FinalizeTopology();
```
This could be reverted as `Point` has been fixed, but it is not very important 😉

tzanio:

PR #4464, opened 2024-08-22
Fixed Point default attribute.

najlkin:
This PR fixes `Point`'s default attribute, which was -1 and not 1 like for other geometrical primitives. This causes problems when the boundaries are automatically generated in `Mesh::GenerateBoundaryElement()` and the attribute is copied to them, which triggers warning in `Mesh::FinalizeTopology()` as boundary attributes are not supposed to be negative.
This seems to a forgotten change in version 1.2 ( https://github.com/mfem/mfem/commit/29ae8fb6ac9fce927fdfd01ef796f9df38f76e97 ). It is a slightly breaking change, but hopefully not too much 🙂 .

❔ This was found in #4416 .

### DIFF
--- a/mesh/point.hpp
+++ b/mesh/point.hpp
@@ -30,7 +30,7 @@ public:
    Point() : Element(Geometry::POINT) {}
 
    /// Constructs point by specifying the indices and the attribute.
-   Point( const int *ind, int attr = -1 );
+   Point(const int *ind, int attr = 1);
 
    /// Return element's type.
    Type GetType() const override { return Element::POINT; }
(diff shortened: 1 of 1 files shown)


pazner (approved):

tzanio:

termi-official (commented):
I know I am always asking this, but can we add some kind of regression test for this which fails on master but works on this PR? This could also be useful in the future when implementing new geometries (e.g. 4D stuff).

najlkin:
We may remove this workaround for the unit test for dof-to-elements arrays:
https://github.com/mfem/mfem/blob/5f373e8c2a70acb413e0fb2a7c1be8054ab6079e/tests/unit/fem/test_build_dof_to_arrays.cpp#L255-L256
It triggers a warning on master as I described above. The workaround is not there for other geometries, so it is a nice check 😉 .

termi-official (approved):

tzanio:

PR #4483, opened 2024-09-03
Allow IterativeSolverMonitor to indicate early convergence

jandrej:
| [#4483](https://github.com/mfem/mfem/pull/4483) | @jandrej | @tzanio | @najlkin + @acfisher | 10/15/24 | 3/11/25 | 3/11/25 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:
@jandrej, can you resolve the conflicts before we start the review?

tzanio:
ping: @jandrej

jandrej:
pong @tzanio

tzanio:

najlkin:
@jandrej , I will have a closer look later 🧐 , but a general question - what is it good for? 😅 You want to have some way how to stop the solver not only based on the norm of the residual? It looks to me a bit hacky way to use `IterativeSolverMonitor` for it 🤔 . As the name says, it is a monitor, so it is supposed to track the progress (and print it, log it, or so), but I would not expect it to interfere with the process...

jandrej:
> For me, a much cleaner solution would be to make some virtual method of `IterativeSolver`, let us say `CheckConvergence()`, which would do the classical check of the norm, but the user could override it in a derived class from the solver to include some custom stopping criterion 👍 .

I tend to agree. @victor-decaria-nnl what do you think about this route?

victor-decaria-nnl:
I prefer the composition approach as it's currently proposed vs a virtual convergence check. Then we just have to maintain one custom class instead of a custom class of every solver we want to use a specific check for.

@najlkin An example for wanting to override the stopping criteria is if you're solving a block system of pdes with a large scale separation between the different fields, and you want to make sure the fields individually satisfy a tolerance. Using a global l2 critieria as things currently work may only effectively check the convergence on the largest magnitude fields if the scales are very separated.

jandrej:
> @jandrej I prefer the composition approach of using `IterativeSolverMonitor` as currently proposed vs a virtual convergence check. Then we just have to maintain one custom class instead of a custom class of every solver we want to use a specific check for.
> 
> @najlkin An example for wanting to override the stopping criteria is if you're solving a block system of pdes with a large scale separation between the different fields, and you want to make sure the fields individually satisfy a tolerance. Using a global l2 critieria as things currently work may only effectively check the convergence on the largest magnitude fields if the scales are very separated.

All very good points.

pazner:
One more consideration: if the monitor is used to indicate convergence, then the same convergence criterion can be used across different solvers. You don't need to create separate derived classes for SLI, CG, GMRES, etc., just pass "CustomCon vergenceMonitor" to the standard classes.

najlkin:
> @jandrej I prefer the composition approach of using `IterativeSolverMonitor` as currently proposed vs a virtual convergence check. Then we just have to maintain one custom class instead of a custom class of every solver we want to use a specific check for.

The code diff will be about the same in MFEM and the custom classes on the user side are as easy as maybe three lines of code, but it is much harder to maintain wrongly designed OOP (i.e., a class intended for something else having an extra effect). If you really want, there could be some different class for this purpose, but it looks to me as an overkill 🪓.

> @najlkin An example for wanting to override the stopping criteria is if you're solving a block system of pdes with a large scale separation between the different fields, and you want to make sure the fields individually satisfy a tolerance. Using a global l2 critieria as things currently work may only effectively check the convergence on the largest magnitude fields if the scales are very separated.

This can be actually already handled by `Norm()`, using something like `max(x.GetBlock(0).Norml2(), x.GetBlock(1).Norml2())` 🤔 . If you want to simplify this, we could think of some template wrapper for solvers, so something like `BlockSolver<GMRESSolver>` 👍 .

najlkin:
Ok folks, I tried the idea of a redefined `IterativeSolver::Dot()` (as `Norm()` itself is not virtual for consistency, see #4156 ). I made this wrapper for block systems 👇 . It works, but the convergence rate seems to be lower than with the normal solvers, at least what I quickly tested. Maybe try it by yourself, but it does not seem to be a good option. So that pathway with some method for checking only the stopping criterion, let us say `CheckConvergence()`, which would be similarly overridden, sounds more feasible 😉 .
```cpp
template<class S> class BlockSolver : public S
{
   const Array<int> &offsets;

   real_t Dot(const Vector &x, const Vector &y) const override
   {
      real_t dot = 0.;
      const BlockVector bx(const_cast<Vector&>(x), offsets);
      const BlockVector by(const_cast<Vector&>(y), offsets);
      for(int i = 0; i < offsets.Size()-1; i++)
      {
         const real_t d = S::Dot(bx.GetBlock(i), by.GetBlock(i));
         //dot += d;
         if(fabs(d) > fabs(dot)) { dot = d; }
      }
      return dot;
   }
public:
   BlockSolver(const Array<int> &offsets) : offsets(offsets) { }
};
```

tzanio:
We decided to rename `Monitor` to `Controller` and make the former an alias to the latter.

linalg/solvers.cpp
```diff
@@ -188,14 +188,16 @@ void IterativeSolver::SetOperator(const Operator &op)
    }
 }
 
-void IterativeSolver::Monitor(int it, real_t norm, const Vector& r,
+bool IterativeSolver::Monitor(int it, real_t norm, const Vector& r,
                               const Vector& x, bool final) const
 {
```
v-dobrev, line 193: When an `IterativeSolver` with a monitor is used multiple times, we need to make sure the monitor state is reset at the start. We can automate this by adding a `Reset` method to the monitor and calling it here when `it == 0 && !final` -- the assumption is that `IterativeSolver`s will always first call the monitor with `it = 0` and `final = false`. Another option is to just add the monitor `Reset` call at the beginning of the `Mult` method of all iterative solvers.

Since it is possible for user monitor classes to add their own state that needs to be reset, the `Reset` method should be either (a) virtual or (b) call a separate virtual method (e.g. `ResetDerived`) after resetting the monitor base class state.

linalg/solvers.cpp
```diff
@@ -766,7 +767,7 @@ void CGSolver::Mult(const Vector &b, Vector &x) const
       return;
    }
    r0 = std::max(nom*rel_tol*rel_tol, abs_tol*abs_tol);
-   if (nom <= r0)
+   if (Monitor(0, nom, r, x) || nom <= r0)
```
v-dobrev, line 770: I think we always should call the monitor with the initial state even when `nom < 0`. To this end, we can keep the `Monitor(...)` call where it is now, and check for monitor convergence here in the `if` statement with an inline method like `bool MonitorConverged() const { return monitor && monitor->HasConverged(); }`.

Also, the when we hit the final iteration (even if that final iteration is 0) we should call the monitor a second time with the same parameters except `final = true`. This allows monitors to properly format output at the end of a solve. I think we need to ensure this behavior not only here but in all iterative solvers, at all the places when they return from `Mult`.

So, in particular, in this context, we should call `Monitor(...)` with `final = false` in both the above case `if (nom < 0.0)` and here in the `if (...)` body.

Finally: the second parameter of `Monitor` is called `norm` and here we pass-in `nom` which is the square of a norm -- we should take a square root of `nom` and if it is negative, we can use either `-sqrt(-nom)` or `-1.0` to indicate an error in the `norm` computation. We should also do something similar to initialize `initial_norm` when `nom < 0.0`.

jandrej:
Please review the current changes and refrain from commits/merges that aren't moderated.

najlkin:
> Please review the current changes and refrain from commits/merges that aren't moderated.

@jandrej , I just merged master as https://github.com/mfem/mfem/pull/4483/commits/861d3d1422361183f77663a070f9fc507f1eecaa was failing due to #4650 .

najlkin:
https://github.com/mfem/mfem/blob/2d998865beb26ffa0eddc14da3b9fd793145e2e5/linalg/solvers.cpp#L1489 should go to 👇 https://github.com/mfem/mfem/blob/2d998865beb26ffa0eddc14da3b9fd793145e2e5/linalg/solvers.cpp#L1467

linalg/solvers.cpp
najlkin: Maybe we would not need to call `Monitor()` twice if it indicates convergence by itself, but the usual construction here is `if(Monitor(...) || norm < goal)`, so we do not know what indicated the convergence. I think the easiest way is to call it twice with `final = true` the second time as proposed by @v-dobrev 😉 . (I guess it should be `final = true` in the third paragraph).
I agree with the `norm`/`nom`. That `-sqrt(-nom)` sounds more logical and conforming with other places in the code to me.

linalg/solvers.cpp
```diff
@@ -1189,10 +1189,9 @@ void FGMRESSolver::Mult(const Vector &b, Vector &x) const
 
    final_norm = std::max(rel_tol*beta, abs_tol);
 
-   converged = false;
-
-   if (beta <= final_norm)
+   if (Monitor(0, beta, r, x) || beta <= final_norm)
    {
+      converged = false;
```
acfisher, line 1194: It seems odd to set this one way only to set it the other in  a couple of lines.  Is there a side effect that I am missing here?

jandrej: This must be a merge oversight. It shouldn't be in the if statement body.

acfisher (approved):
Looks good, with only minor edits needed from comments.

jandrej:
Added changes requested in comments and meeting. I did not change the norm computation as I'm not confident to know how different solvers are supposed to report their norms to be consistent. Can someone else propose those changes please?

linalg/solvers.cpp
```diff
@@ -2149,6 +2161,8 @@ void LBFGSSolver::Mult(const Vector &b, Vector &x) const
    {
       mfem::out << "LBFGS: No convergence!\n";
    }
+
+   Monitor(it, norm, r, x, true);
```
najlkin, line 2165: LBFGS does not have monitoring, either add monitor to the main loop or remove this.

linalg/solvers.cpp
```diff
@@ -1759,7 +1770,7 @@ void MINRESSolver::Mult(const Vector &b, Vector &x) const
       mfem::out << "MINRES: Number of iterations: " << setw(3) << final_iter << '\n';
    }
 
-   Monitor(final_iter, final_norm, *z, x, true);
+   converged = converged || Monitor(final_iter, final_norm, *z, x, true);
```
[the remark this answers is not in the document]
jandrej: Is that a question?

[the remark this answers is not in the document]
najlkin: No, suggestion. If you are ok with it, just hit "commit suggestion".

linalg/solvers.cpp
najlkin: Wrong fix, `converged = false` should be outside (otherwise it is uninitialized) and `converged = true` below.

najlkin (approved):
I think we are good here 👍 . MINRES has a bit different logic than other solvers, where the monitor can change the convergence flag at the end, while it is ignored in other solvers, but it is just an unimportant edge case probably.
Besides, just for reference, that square of the norm should be fixed in CG later.

tzanio:

tzanio:
@jandrej, should we mention this in `CHANGELOG`?

najlkin:
I would say the renaming should go to API changes. About the new functionality, I am not so sure, it could be briefly mentioned there as the reason maybe 🤔 .

najlkin:
@tzanio , we may try to remerge this to `next` 🔁

tzanio:

PR #4496, opened 2024-09-10
Added CMake dependencies on individual data files

cjvogl:
This PR is motivated by issues with the original `copy_data` target not properly capturing the dependencies on the contents of the `data` directory.  Specifically, modification to any of the data files (e.g., when fast-forwarding `main`) were ignored.  Granted, this approach requires any additions or renaming to be reflected in the new `data/CMakeLists.txt`, but I still see it as an improvement over the original approach.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -673,7 +673,7 @@ if (NOT ("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}"))
 #include \"${PROJECT_SOURCE_DIR}/${Header}\"
 ")
 
-    execute_process(COMMAND ${CMAKE_COMMAND} -E copy_if_different 
+    execute_process(COMMAND ${CMAKE_COMMAND} -E copy_if_different
       "${PROJECT_BINARY_DIR}/${Header}.tmp"
       "${PROJECT_BINARY_DIR}/${Header}"
     )
@@ -687,7 +687,7 @@ if (NOT ("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}"))
 #include \"mfem/${Header}\"
 ")
 
-    execute_process(COMMAND ${CMAKE_COMMAND} -E copy_if_different 
+    execute_process(COMMAND ${CMAKE_COMMAND} -E copy_if_different
       "${PROJECT_BINARY_DIR}/InstallHeaders/${Header}.tmp"
       "${PROJECT_BINARY_DIR}/InstallHeaders/${Header}"
     )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -748,15 +748,11 @@ add_dependencies(exec
 #  - https://gitlab.kitware.com/cmake/cmake/issues/8438
 #  - https://cmake.org/Bug/view.php?id=8438
 
-# Add a target to copy the mfem data directory to the build directory
-add_custom_command(OUTPUT data_is_copied
-  COMMAND ${CMAKE_COMMAND} -E copy_directory ${PROJECT_SOURCE_DIR}/data data
-  COMMAND ${CMAKE_COMMAND} -E touch data_is_copied
-  COMMENT "Copying the data directory ...")
-add_custom_target(copy_data DEPENDS data_is_copied)
-# Add 'copy_data' as a prerequisite for all executables, if the source and the
-# build directories are not the same.
+# If the source and the build directories are not the same, add data directory
+# that defines 'copy_data' target to copy data directory contents to the build
+# directory
 if (NOT ("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}"))
+  add_subdirectory(data)
   add_dependencies(${MFEM_EXEC_PREREQUISITES_TARGET_NAME} copy_data)
 endif()
 
--- a/data/CMakeLists.txt
+++ b/data/CMakeLists.txt
@@ -1,0 +1,118 @@
+list(APPEND DATA_FILES
+  amr-hex.mesh
+  amr-quad.mesh
+  ball-nurbs.mesh
+  beam-hex-nurbs.mesh
+  beam-hex.mesh
+  beam-hex.vtk
+  beam-quad-amr.mesh
+  beam-quad-nurbs-sf.mesh
+  beam-quad-nurbs.mesh
+  beam-quad.mesh
+  beam-quad.vtk
+  beam-tet.mesh
+  beam-tet.vtk
+  beam-tri.mesh
+  beam-tri.vtk
+  beam-wedge.mesh
+  beam-wedge.vtk
+  compass.geo
+  compass.mesh
+  compass.msh
+  cube-nurbs.mesh
+  diag-segment-2d.mesh
+  diag-segment-3d.mesh
+  disc-nurbs.mesh
+  escher-p2.mesh
+  escher-p2.vtk
+  escher-p3.mesh
+  escher.mesh
+  escher.vtk
+  fichera-amr.mesh
+  fichera-mixed-p2.mesh
+  fichera-mixed-p2.vtk
+  fichera-mixed.mesh
+  fichera-q2.mesh
+  fichera-q2.vtk
+  fichera-q3.mesh
+  fichera-quad-mixed.mesh
+  fichera-quad.mesh
+  fichera.mesh
+  fichera.vtk
+  hexagon.mesh
+  inline-hex.mesh
+  inline-pyramid.mesh
+  inline-quad.mesh
+  inline-segment.mesh
+  inline-tet.mesh
+  inline-tri.mesh
+  inline-wedge.mesh
+  klein-bottle.mesh
+  klein-donut.mesh
+  l-shape.mesh
+  llnl-p3.mesh
+  mobius-strip.mesh
+  octahedron.mesh
+  periodic-annulus-sector.geo
+  periodic-annulus-sector.msh
+  periodic-cube.geo
+  periodic-cube.mesh
+  periodic-cube.msh
+  periodic-hexagon.mesh
+  periodic-segment.mesh
+  periodic-square.geo
+  periodic-square.mesh
+  periodic-square.msh
+  periodic-torus-sector.geo
+  periodic-torus-sector.msh
+  pipe-nurbs-2d.mesh
+  pipe-nurbs-log.mesh
+  pipe-nurbs.mesh
+  ref-cube.mesh
+  ref-prism.mesh
+  ref-pyramid.mesh
+  ref-segment.mesh
+  ref-square.mesh
+  ref-tetrahedron.mesh
+  ref-triangle.mesh
+  rt-2d-p4-tri.mesh
+  rt-2d-q3.mesh
+  segment-nurbs.mesh
+  square-disc-nurbs-patch.mesh
+  square-disc-nurbs.mesh
+  square-disc-p2.mesh
+  square-disc-p2.vtk
+  square-disc-p3.mesh
+  square-disc-surf.mesh
+  square-disc.mesh
+  square-disc.vtk
+  square-mixed.mesh
+  square-nurbs-pw.mesh
+  square-nurbs.mesh
+  star-hilbert.mesh
+  star-mixed-p2.mesh
+  star-mixed-p2.vtk
+  star-mixed.mesh
+  star-q2.mesh
+  star-q2.vtk
+  star-q3.fms
+  star-q3.mesh
+  star-surf.mesh
+  star.mesh
+  star.vtk
+  tinyzoo-3d.mesh
+  toroid-hex.mesh
+  toroid-wedge.mesh
+)
+
+foreach(DATA_FILE ${DATA_FILES})
+  set(SRC_DATA_FILE ${CMAKE_CURRENT_SOURCE_DIR}/${DATA_FILE})
+  set(GEN_DATA_FILE ${CMAKE_CURRENT_BINARY_DIR}/${DATA_FILE})
+  add_custom_command(
+    OUTPUT ${GEN_DATA_FILE}
+    DEPENDS ${SRC_DATA_FILE}
+    COMMAND ${CMAKE_COMMAND} -E copy ${SRC_DATA_FILE} ${GEN_DATA_FILE})
+  list(APPEND GENERATED_DATA_FILES ${GEN_DATA_FILE})
+endforeach()
+
+add_custom_target(copy_data DEPENDS ${GENERATED_DATA_FILES})


tzanio:

nmnobre:
Hi Chris,

_Conceptually_ (we can always improve the syntax) would the following achieve the same result while also automatically covering the case of new added files?

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 00d50283b..9410d758c 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -749,11 +749,11 @@ add_dependencies(exec
 #  - https://cmake.org/Bug/view.php?id=8438
 
 # Add a target to copy the mfem data directory to the build directory
-add_custom_command(OUTPUT data_is_copied
+add_custom_command(OUTPUT data
+  DEPENDS ${PROJECT_SOURCE_DIR}/data/*
   COMMAND ${CMAKE_COMMAND} -E copy_directory ${PROJECT_SOURCE_DIR}/data data
-  COMMAND ${CMAKE_COMMAND} -E touch data_is_copied
   COMMENT "Copying the data directory ...")
-add_custom_target(copy_data DEPENDS data_is_copied)
+add_custom_target(copy_data DEPENDS data)
 # Add 'copy_data' as a prerequisite for all executables, if the source and the
 # build directories are not the same.
 if (NOT ("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}"))
 ```

**Comment by cjvogl** [comment]
> Hi Chris,
> 
> _Conceptually_ (we can always improve the syntax) would the following achieve the same result while also automatically covering the case of new added files?
> 
> ```diff
> diff --git a/CMakeLists.txt b/CMakeLists.txt
> index 00d50283b..9410d758c 100644
> --- a/CMakeLists.txt
> +++ b/CMakeLists.txt
> @@ -749,11 +749,11 @@ add_dependencies(exec
>  #  - https://cmake.org/Bug/view.php?id=8438
>  
>  # Add a target to copy the mfem data directory to the build directory
> -add_custom_command(OUTPUT data_is_copied
> +add_custom_command(OUTPUT data
> +  DEPENDS ${PROJECT_SOURCE_DIR}/data/*
>    COMMAND ${CMAKE_COMMAND} -E copy_directory ${PROJECT_SOURCE_DIR}/data data
> -  COMMAND ${CMAKE_COMMAND} -E touch data_is_copied
>    COMMENT "Copying the data directory ...")
> -add_custom_target(copy_data DEPENDS data_is_copied)
> +add_custom_target(copy_data DEPENDS data)
>  # Add 'copy_data' as a prerequisite for all executables, if the source and the
>  # build directories are not the same.
>  if (NOT ("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}"))
> ```

Hi Nuno. Thanks for your suggestion.  I believe you approach might cover the case where an existing file in the `data` directory is updated (have you tested this?); however, I'm less clear on whether this would cover when a new file is added to the `data` directory.  In other words, I imagine CMake will create a dependency tree based on the `data` contents at configure time but not update it at compile time.  I'm happy to be proven wrong though!

**Comment by nmnobre** [comment]
> Hi Nuno. Thanks for your suggestion! I believe your approach might cover the case where an existing file in the `data` directory is updated (have you tested this?); however, I'm less clear on whether this would cover when a new file is added to the `data` directory.

I've tested both cases, and it seems to work, but I was basically asking you to test on your machine as well. :)
I've tested on GNU/Linux and macOS.

**Comment by cjvogl** [comment]
> > Hi Nuno. Thanks for your suggestion! I believe your approach might cover the case where an existing file in the `data` directory is updated (have you tested this?); however, I'm less clear on whether this would cover when a new file is added to the `data` directory.
> 
> I've tested both cases, and it seems to work, but I was basically asking you to test on your machine as well. :) I've tested on GNU/Linux and macOS.

Ah thank you for clarifying: I took conceptually to mean the idea hadn't yet been tested. I'll try it on my machine, but it sounds promising if you've already tested it... and would be a great alternative to specifying the individual contents of `data`!

**Comment by cjvogl** [comment]
@nmnobre , I've tested [your suggestion](https://github.com/mfem/mfem/pull/4496#issuecomment-2594162399) using the scenarios I've added to the PR description, namely 

1. a file in the source `data` folder is modified
2. a file in the binary `data` folder is deleted
3. the binary `data` folder itself is deleted
4. a new file is added to the source `data` folder

As you noted, your approach nicely addresses 1 and 4 (with the handling of 4 being better in my opinion than my approach). Your approach also addresses 3.  The one scenario your approach does not handle is 2.

Personally, I keep going back and forth between voting for the compactness of your approach and voting for an approach that handles 2.  I'm curious to hear what you and others think.

**Comment by nmnobre** [comment]
> I'm curious to hear what you [...] think

I think we're being naive 😆. This does all of the four points above and adds a fifth: it also copies if you modify a file in the binary `data` folder:

```diff

diff --git a/CMakeLists.txt b/CMakeLists.txt
index 00d50283b..b34791956 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -749,11 +749,9 @@ add_dependencies(exec
 #  - https://cmake.org/Bug/view.php?id=8438
 
 # Add a target to copy the mfem data directory to the build directory
-add_custom_command(OUTPUT data_is_copied
-  COMMAND ${CMAKE_COMMAND} -E copy_directory ${PROJECT_SOURCE_DIR}/data data
-  COMMAND ${CMAKE_COMMAND} -E touch data_is_copied
-  COMMENT "Copying the data directory ...")
-add_custom_target(copy_data DEPENDS data_is_copied)
+add_custom_target(copy_data
+  COMMAND ${CMAKE_COMMAND} -E copy_directory_if_different ${PROJECT_SOURCE_DIR}/data data)
+
 # Add 'copy_data' as a prerequisite for all executables, if the source and the
 # build directories are not the same.
 if (NOT ("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}"))
```

There's a caveat, this relies on CMake 3.26 (which is ~2 years old now), but we could conditionally execute this if version >=3.26 and keep the old behaviour (or my slight modification above) if <3.26.

cjvogl:
> this relies on CMake 3.26

Unfortunately, that requirement is effectively a non-starter: many MFEM users are running on machines that don't provide v3.26. 

>  we could conditionally execute this if version >=3.26 and keep the old behaviour (or my slight modification above) if <3.26.

I would instead say the three options are

1. use your suggestion that ignores if files in the binary `data` directory are individually deleted
2. use the approach in the PR  that requires new files in the source `data` directory be added to `data/CMakeLists.txt`
3. a different approach that won't put out users running on machines with older CMake versions

I assume your vote is (1) ... let's see what others (including the assigned reviewers) think.

helloworld922:
You can use `configure_file(src dst COPY_ONLY)` to have CMake automatically track edits for you.

cjvogl:
> You can use `configure_file(src dst COPY_ONLY)` to have CMake automatically track edits for you.

Thanks for the suggestion, @helloworld922.  Can you elaborate on what you see as the benefits / drawbacks over options 1 and 2 in https://github.com/mfem/mfem/pull/4496#issuecomment-2596831575, in particular with respect to the scenarios in the PR description?

data/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+  star-q3.mesh
+  star-surf.mesh
+  star.mesh
+  star.vtk
+  tinyzoo-3d.mesh
+  toroid-hex.mesh
+  toroid-wedge.mesh
+)
+
+foreach(DATA_FILE ${DATA_FILES})
```
nmnobre, line 108:
```suggestion
file(GLOB DATA_FILES ${CMAKE_CURRENT_SOURCE_DIR}/*)

foreach(DATA_PATH ${DATA_FILES})
  get_filename_component(DATA_FILE ${DATA_PATH} NAME)
```

cjvogl: Right, I did try something similar but abandoned it because the reason you highlighted (which is why I think such loops are avoided in source directories in general).

justinlaughlin:
Just to throw out another potential solution - I added a `DEPENDS` declaration to the main cmake file and it seems to be tracking changes okay.

```diff
add_custom_command(OUTPUT data_is_copied
  COMMAND ${CMAKE_COMMAND} -E copy_directory ${PROJECT_SOURCE_DIR}/data data
+ DEPENDS ${PROJECT_SOURCE_DIR}/data
  COMMAND ${CMAKE_COMMAND} -E touch data_is_copied
  COMMENT "Copying the data directory ...")
add_custom_target(copy_data DEPENDS data_is_copied)

```

data/CMakeLists.txt
nmnobre: I see, I've edited my suggestion above to trigger a reconfiguration if the `data` source directory is modified (which happens if you create a new file).

cjvogl: Thanks @nmnobre , this looks promising: I'll test it out.

cjvogl: This checks all the boxes, thanks @nmnobre! I have minor syntax things to tweak, so I implemented it in [9672921](https://github.com/mfem/mfem/pull/4496/commits/96729213ef37b69b6f8273e2d8e87f00e0ac4f20) instead of applying this suggestion directly. As such, I'll leave it to you to officially resolve this thread if there isn't more discussion.

justinlaughlin:
> > Just to throw out another potential solution - I added a `DEPENDS` declaration to the main cmake file and it seems to be tracking changes okay.
> > ```diff
> > add_custom_command(OUTPUT data_is_copied
> >   COMMAND ${CMAKE_COMMAND} -E copy_directory ${PROJECT_SOURCE_DIR}/data data
> > + DEPENDS ${PROJECT_SOURCE_DIR}/data
> >   COMMAND ${CMAKE_COMMAND} -E touch data_is_copied
> >   COMMENT "Copying the data directory ...")
> > add_custom_target(copy_data DEPENDS data_is_copied)
> > ```
> 
> Thanks for the suggestion @justinlaughlin . This seems similar to what @nmnobre suggested in [#4496 (comment)](https://github.com/mfem/mfem/pull/4496#issuecomment-2594162399) (modulo the renaming changes)... or perhaps I'm missing something?

Oops 🤦‍♂️ I saw the `copy_directory_if_different` suggestion and missed the first one (too bad we can't use a higher CMake version, because that function seems to be *exactly* what we want).

After going back and forth a bit I think I prefer the original solution in this PR. I like the compactness of @nmnobre's suggestions but I think the more robust solution is better. A couple of other reasons is that in several miniapps [the solution] https://github.com/mfem/mfem/blob/master/miniapps/meshing/CMakeLists.txt#L12-L34) the solution is similar. Also, `mfem/data` doesn't seem to change very frequently. Maybe we can add a comment that this could be replaced with `copy_directory_if_different` if we ever enforce `CMake >= 3.26` in the future (mfem 5.0?)?

cjvogl:
> > > Just to throw out another potential solution - I added a `DEPENDS` declaration to the main cmake file and it seems to be tracking changes okay.
> > > ```diff
> > > add_custom_command(OUTPUT data_is_copied
> > >   COMMAND ${CMAKE_COMMAND} -E copy_directory ${PROJECT_SOURCE_DIR}/data data
> > > + DEPENDS ${PROJECT_SOURCE_DIR}/data
> > >   COMMAND ${CMAKE_COMMAND} -E touch data_is_copied
> > >   COMMENT "Copying the data directory ...")
> > > add_custom_target(copy_data DEPENDS data_is_copied)
> > > ```
> > 
> > 
> > Thanks for the suggestion @justinlaughlin . This seems similar to what @nmnobre suggested in [#4496 (comment)](https://github.com/mfem/mfem/pull/4496#issuecomment-2594162399) (modulo the renaming changes)... or perhaps I'm missing something?
> 
> Oops 🤦‍♂️ I saw the `copy_directory_if_different` suggestion and missed the first one (too bad we can't use a higher CMake version, because that function seems to be _exactly_ what we want).
> 
> After going back and forth a bit I think I prefer the original solution in this PR. I like the compactness of @nmnobre's suggestions but I think the more robust solution is better. A couple of other reasons is that in several miniapps [the solution] https://github.com/mfem/mfem/blob/master/miniapps/meshing/CMakeLists.txt#L12-L34) the solution is similar. Also, `mfem/data` doesn't seem to change very frequently. Maybe we can add a comment that this could be replaced with `copy_directory_if_different` if we ever enforce `CMake >= 3.26` in the future (mfem 5.0?)?

I appreciate your input and agree on the robust solution.  @nmnobre had a separate suggestion, which now appears in the PR via [9672921](https://github.com/mfem/mfem/pull/4496/commits/96729213ef37b69b6f8273e2d8e87f00e0ac4f20)), that I personally think balances the robustness of the original solution with the compactness of `copy_directory_if_different` (i.e., its not a one-line change like `copy_directory_if_different`, but it's fairly compact).  Let me know what you think.

data/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+# set property that will initiate a configure step if data directory is modified
```
justinlaughlin, line 1:
```suggestion
# Copyright (c) 2010-2024, Lawrence Livermore National Security, LLC. Produced
# at the Lawrence Livermore National Laboratory. All Rights reserved. See files
# LICENSE and NOTICE for details. LLNL-CODE-806117.
#
# This file is part of the MFEM library. For more information and source code
# availability visit https://mfem.org.
#
# MFEM is free software; you can redistribute it and/or modify it under the
# terms of the BSD-3 license. We welcome feedback and contributions, see file
# CONTRIBUTING.md for details.

# If using `CMake >= 3.26` this functionality of this file can be replaced with `copy_directory_if_different`

# set property that will initiate a configure step if data directory is modified
```

justinlaughlin: Something like this is what I had in mind. Just a note for future developers if/when we enforce a higher version.

justinlaughlin (approved):

justinlaughlin:
I approved since my only suggestion is to add a small comment; I tested 9672921 and it works as expected. The only case they are not synchronized is if a modification is made to the target file, but I think that is okay.

data/CMakeLists.txt
cjvogl: Much appreciated (and thanks for adding the license statement)!

data/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# LICENSE and NOTICE for details. LLNL-CODE-806117.
+#
+# This file is part of the MFEM library. For more information and source code
+# availability visit https://mfem.org.
+#
+# MFEM is free software; you can redistribute it and/or modify it under the
+# terms of the BSD-3 license. We welcome feedback and contributions, see file
+# CONTRIBUTING.md for details.
+
+# If using `CMake >= 3.26` this functionality of this file can be replaced with `copy_directory_if_different`
```
nmnobre, line 12:
```suggestion
# If using `CMake >= 3.26` the functionality of this file can be replaced with `copy_directory_if_different`
```

data/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+
+# set property that will initiate a configure step if data directory is modified
+set_property(DIRECTORY APPEND
+  PROPERTY CMAKE_CONFIGURE_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR})
+
+set(COPIED_DATA_FILES)
```
nmnobre, line 18:
```suggestion
```

data/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+  get_filename_component(DATA_FILENAME ${DATA_FILE} NAME)
+  set(COPIED_DATA_FILE ${CMAKE_CURRENT_BINARY_DIR}/${DATA_FILENAME})
+  add_custom_command(OUTPUT ${COPIED_DATA_FILE} DEPENDS ${DATA_FILE}
+    COMMAND ${CMAKE_COMMAND} -E copy ${DATA_FILE} ${COPIED_DATA_FILE}
+    COMMENT "")
+  list(APPEND COPIED_DATA_FILES ${COPIED_DATA_FILE})
+endforeach()
+
+add_custom_target(copy_data DEPENDS ${COPIED_DATA_FILES}
+  COMMENT "Synching the data directory ...")
```
nmnobre, line 30:
```suggestion
  COMMENT "Syncing the data directory ...")
```

nmnobre, line 30:
```suggestion
# If using `CMake >= 3.26`, the functionality of this file can be replaced with `copy_directory_if_different`

# set property that will initiate a configure step if data directory is modified
set_property(DIRECTORY APPEND
  PROPERTY CMAKE_CONFIGURE_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR})

file(GLOB DATA_FILES ${CMAKE_CURRENT_SOURCE_DIR}/*)
foreach(DATA_FILE ${DATA_FILES})
  add_custom_command(OUTPUT ${DATA_FILE}_CMD
    COMMAND ${CMAKE_COMMAND} -E copy_if_different ${DATA_FILE} .
    COMMENT "")
  list(APPEND DATA_CMDS ${DATA_FILE}_CMD)
endforeach()

add_custom_target(copy_data DEPENDS ${DATA_CMDS}
  COMMENT "Syncing the data directory ...")
```

nmnobre: This does (1) to (4) and additionally copies if we modify a file in the build data directory.

nmnobre:
Hi Chris (@cjvogl),

I pushed to avoid all the email spam (sorry about that :/). We can always drop the commit if I've missed something.
I think the suggested changes cover (1) to (4) and also cover the case of a modified file in the build data directory.

Note the required functionality requires CMake >= 3.5 and that we already require >= 3.8.

data/CMakeLists.txt
nmnobre, line 12:
```suggestion
# If using `CMake >= 3.26`, the functionality of this file can be replaced with `copy_directory_if_different`
```

data/CMakeLists.txt
nmnobre, line 30:
```suggestion
  COMMENT "Syncing the data directory ...")
```

helloworld922:
I tried out the current state of this branch (23edd7525), and I noticed a few things:

1. If you don't build any of the executables which depend on `copy_data` (right now looks like just unit tests?) or manually specify the `copy_data` target it doesn't appear to try to keep the data directory in sync.
2. If the data directory changes it seems to force a full rebuild of everything, even if the source files didn't change?
3. It looks like this version keeps the file with the newest timestamp? If I modify source/data/amr[hex.mesh then run `touch build/data/amr-hex.mesh`, it will not update build/data/amr-hex.mesh. This could be a desirable feature in some situations, but I prefer to have these always in sync when make is run to minimize surprises.

This was tested on Ubuntu 24.04 with CMake 3.31.2 and GNU Make.

nmnobre:
Hi @helloworld922,

Do you mind trying https://github.com/farscape-project/mfem/tree/refactor-cmake-data-directory?

Covers all of:

1. a file in the source data folder is modified
2. a file in the binary data folder is deleted
3. the binary data folder itself is deleted
4. a new file is added to the source data folder

And:

5. a file in the binary data folder is modified (irrespective of timestamps)
6. requires no updates to CMake
7. if the data directory changes, CMake reconfigures, but no unnecessary builds happen

helloworld922:
> Hi @helloworld922,
> 
> Do you mind trying https://github.com/farscape-project/mfem/tree/refactor-cmake-data-directory? I'd push here, but I regret doing that earlier, and I'd rather @cjvogl say he's okay with that beforehand. @cjvogl feel free to cherry-pick and amend as needed. :)
> 
> Covers all of:
> 
> 1. a file in the source data folder is modified
> 2. a file in the binary data folder is deleted
> 3. the binary data folder itself is deleted
> 4. a new file is added to the source data folder
> 
> And:
> 
> 5. a file in the binary data folder is modified (irrespective of timestamps)
> 6. requires no updates to CMake
> 7. if the data directory changes, CMake reconfigures, but no unnecessary builds happen

Here are my results:

1. I need to run either `make copy_data` or `make exec` to trigger any synchronization (even from a clean build directory with no build/data dir). A plain `make` will not do anything about the data directory. I think this is because I configured without examples/miniapps enabled? If I enable these, then plain `make` works.
2. modifying/deleting a tracked file in `build/data` triggers a copy, No rebuild is required.
3. adding a new untracked file to `build/data` does not trigger any response. No rebuild is required.
4. deleting build/data triggers a copy. No rebuild is required.
5. Adding/modifying/deleting a file in `src/data` triggers a copy and also sometimes triggers a rebuild? If I modify src/data/amr-hex.mesh it triggers a rebuild. However, the following do not trigger a rebuild:
- copying src/data/amr-hex.mesh to a new file src/data/amr-test.mesh
- modifying src/data/amr-test.mesh
- modifying src/data/amr-quad.mesh
- deleting src/data/amr-test.mesh

I couldn't figure out a pattern for when rebuild is or isn't required, but for some reason amr-hex.mesh is special? There might be other files which trigger rebuilds.

cjvogl:
> Hi @helloworld922,
> 
> Do you mind trying https://github.com/farscape-project/mfem/tree/refactor-cmake-data-directory? I'd push here, but I regret doing that earlier, and I'd rather @cjvogl say he's okay with that beforehand. @cjvogl feel free to cherry-pick and amend as needed. :)
> 
> Covers all of:
> 
>     1. a file in the source data folder is modified
> 
>     2. a file in the binary data folder is deleted
> 
>     3. the binary data folder itself is deleted
> 
>     4. a new file is added to the source data folder
> 
> 
> And:
> 
>     5. a file in the binary data folder is modified (irrespective of timestamps)
> 
>     6. requires no updates to CMake
> 
>     7. if the data directory changes, CMake reconfigures, but no unnecessary builds happen

@nmnobre , can you open a separate PR from your branch into this one? I'm concerned that having comments for two separate branches in the same PR will get very confusing.

CMakeLists.txt
```diff
@@ -756,15 +756,11 @@ add_dependencies(exec
...
-add_custom_target(copy_data DEPENDS data_is_copied)
-# Add 'copy_data' as a prerequisite for all executables, if the source and the
-# build directories are not the same.
+# If the source and the build directories are not the same, add data directory
+# that defines 'copy_data' target to copy data directory contents to the build
+# directory
 if (NOT ("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}"))
+  add_subdirectory(data)
   add_dependencies(${MFEM_EXEC_PREREQUISITES_TARGET_NAME} copy_data)
 endif()
```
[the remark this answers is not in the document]
cjvogl: Thanks Andrew for elaborating. I've tested your suggestion out and it indeed checks all the boxes in the PR description with less changes than the current approach (23edd7525)! As you alluded to in [a separate comment](https://github.com/mfem/mfem/pull/4496#issuecomment-2598792399), the one behavioral change that this will introduce is the data files are copied when no unit tests are created (e.g. building only `libmfem.a` and/or plus examples). I'm ok with this change but we should check with @tzanio (as editor of this PR) and @v-dobrev (as the original author of a majority the CMake code).

> There is a potential warning about this not working for all generators, but it seems to work fine for GNU Make.

I did not see the warning; however, I do want to avoid introducing a change that will produce warnings for users... any thoughts on why you're seeing a warning and I am not? I'm using CMake 3.26.4 if that is of use.

> This requires at least CMake 3.12 (Released Nov. 2018) for the CONFIGURE_DEPENDS option. 

I think changing the required CMake version from 3.8 to 3.12 is reasonable (i.e., won't affect many users), but we should check with @tzanio .

In short, I'm leaning towards your approach provided we
- [ ] track down when a CMake warning is generated
- [ ] get the ok to change behavior of the CMake build to always copy the data files (vs copying the data files only when executables are created)
- [ ] get the ok to increase the required CMake version from 3.8 to 3.12

cjvogl:
> I tried out the current state of this branch ([23edd75](https://github.com/mfem/mfem/commit/23edd75253db3f21e39907b3ed419a167f41d1d9)), and I noticed a few things:
> 
>     1. If you don't build any of the executables which depend on `copy_data` (right now looks like just unit tests?) or manually specify the `copy_data` target it doesn't appear to try to keep the data directory in sync.
> 
>     2. If the data directory changes it seems to force a full rebuild of everything, even if the source files didn't change?
> 
>     3. It looks like this version keeps the file with the newest timestamp? If I modify source/data/amr[hex.mesh then run `touch build/data/amr-hex.mesh`, it will not update build/data/amr-hex.mesh. This could be a desirable feature in some situations, but I prefer to have these always in sync when make is run to minimize surprises.
> 
> 
> This was tested on Ubuntu 24.04 with CMake 3.31.2 and GNU Make.

Thank you for testing this out!

For (1), it's tricky: the example executables depend on `exec_prerequisites` via the macro `add_mfem_examples`, which in turn depends on `copy_data`.  I believe the intended behavior is that if no executables are created, the `data` directory is not copied (by intended, I mean that the behavior is inherited from `master`)

For (2), I dont see this behavior... can you elaborate?

For (3), I agree with you on that the file in `build` should match that in `source` even if that means overwriting changes made in `build`.  I see that [your suggestion](https://github.com/mfem/mfem/pull/4496#discussion_r1919336764) would do exactly that, but if we cannot address the concerns noted there, I would say this PR is still an improvement over `master` in that it will at least address if the `source` file is "newer" than the `build`.

helloworld922:
> > I tried out the current state of this branch ([23edd75](https://github.com/mfem/mfem/commit/23edd75253db3f21e39907b3ed419a167f41d1d9)), and I noticed a few things:
> > ```
> > 1. If you don't build any of the executables which depend on `copy_data` (right now looks like just unit tests?) or manually specify the `copy_data` target it doesn't appear to try to keep the data directory in sync.
> > 
> > 2. If the data directory changes it seems to force a full rebuild of everything, even if the source files didn't change?
> > 
> > 3. It looks like this version keeps the file with the newest timestamp? If I modify source/data/amr[hex.mesh then run `touch build/data/amr-hex.mesh`, it will not update build/data/amr-hex.mesh. This could be a desirable feature in some situations, but I prefer to have these always in sync when make is run to minimize surprises.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This was tested on Ubuntu 24.04 with CMake 3.31.2 and GNU Make.
> 
> Thank you for testing this out!
> 
> For (1), it's tricky: the example executables depend on `exec_prerequisites` via the macro `add_mfem_examples`, which in turn depends on `copy_data`. I believe the intended behavior is that if no executables are created, the `data` directory is not copied (by intended, I mean that the behavior is inherited from `master`)
> 
> For (2), I dont see this behavior... can you elaborate?
> 
> For (3), I agree with you on that the file in `build` should match that in `source` even if that means overwriting changes made in `build`. I see that [your suggestion](https://github.com/mfem/mfem/pull/4496#discussion_r1919336764) would do exactly that, but if we cannot address the concerns noted there, I would say this PR is still an improvement over `master` in that it will at least address if the `source` file is "newer" than the `build`.

I suspect (2) might be due to the same reason I got source rebuilds when I tried @nmnobre's branch. I only tested modifying amr-hex.mesh because it was the first file listed lexicographically. I can try modifying other files as well as adding/removing files to see if it leads to occasional rebuilds. This may just be a bug or some strange corner case in CMake, and possibly version dependent? What version of CMake are you testing with?

CMakeLists.txt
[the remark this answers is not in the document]
helloworld922: The warning I was referring to is in the [Cmake documentation](https://cmake.org/cmake/help/latest/command/file.html#glob). When I actually tried it there was no warning or other indication of anything going wrong.

nmnobre:
> @nmnobre , can you open a separate PR from your branch into this one?

Would you mind just cherry-picking? I've just added a single commit on top (I didn't rewrite any history or anything) and it just improves what's currently on this PR by adding the update-on-modified-build-file feature. :) ( I can also push directly if that's easier?)

>  If I modify src/data/amr-hex.mesh it triggers a rebuild

Okay, I can reproduce this consistently if I 1) modify any file in `data` and then 2) `touch data`. I don't know why this happens though. Puzzling, but not very impactful I'd say.

cjvogl:
> Would you mind just cherry-picking? I've just added a single commit on top (I didn't rewrite any history or anything) and it just improves what's currently on this PR by adding the update-on-modified-build-file feature. :) ( I can also push directly if that's easier?)

Please do not push to this branch (especially force pushing). I suggest either provide your changes in a GitHub suggestion or open a new PR off `master`.

nmnobre:
I've had the chance to investigate the seemingly arbitrary rebuilds: they are not arbitrary and it's not a bug. They happen if `MFEM_GIT_STRING` changes, because this is injected into the configuration files, on which the normal build target depends on. So, if you had a clean working tree, then modify an existing file (the working tree is now dirty), and then change `data` in any way (just `touch` suffices, but in practice you need to add/remove a file) triggering a reconfigure on `make`, `MFEM_GIT_STRING` is recomputed and a rebuild happens. If your working tree is already dirty, making more changes to new/existing files in `data` won't trigger a rebuild. As soon as you commit or restore though (cleaning the working tree again), then a rebuild will happen if you add/remove a file in `data` and then `make`. Note adding a new file doesn't make the working tree dirty (because it'll be untracked), removing an existing one does - but again to trigger a rebuild the user would need to additionally add/remove a file. I hope that makes things clearer. :) TLDR: it'll very seldomly happen if at all.

So, given that, to get our list of data files we either use:
a)
```
file(GLOB DATA_FILES LIST_DIRECTORIES false CONFIGURE_DEPENDS ${PROJECT_SOURCE_DIR}/data/*)
```

b)
```
set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${PROJECT_SOURCE_DIR}/data)
file(GLOB DATA_FILES LIST_DIRECTORIES false ${PROJECT_SOURCE_DIR}/data/*)
```

Where the decision hangs on whether we care more about pushing users to CMake 3.12 - should go with b) - or the very rare rebuilds above - should go with a).

Once we decide that, we should decide on the semantics we want:

1) (either a) or b) will need adjusting for relative paths as per @helloworld922's code)
```
foreach(DATA_FILE ${DATA_FILES})
  configure_file(${DATA_FILE} ${DATA_FILE} COPYONLY)
endforeach()
```

2)
```
add_custom_target(copy_data
  COMMAND ${CMAKE_COMMAND} -E make_directory data
  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${DATA_FILES} data
  COMMENT "Syncing the data directory ...")
```

Where the decision is between copying every time (1) or copying when creating executables or explicitly requested (2).

Whatever the choice, I think for consistency we should also update the gslib, meshing and multidomain miniapps which have similar functionality at the top of their respective CMake files - I think this might be the only reason I might prefer (2) above.
Beyond that, I'm happy :)

cjvogl:
The funding that supported this PR has ended so I'm not sure I'll be able to commit effort to pushing things along.  Anyone else interested/funded to see this through?

cjvogl:
> Where the decision hangs on whether we care more about not pushing users to CMake 3.12 - should go with b) - or the very rare rebuilds above - should go with a).

@nmnobre , it looks like the minimum version was pushed to 3.12 in https://github.com/mfem/mfem/commit/8c0a137dae1a333f7c9286724ddf44414ff85746.  So that's a plus.

nmnobre:
I'm happy to see this through @cjvogl. Do I have your permission to commit and push to your branch? :)

cjvogl:
> I'm happy to see this through @cjvogl. Do I have your permission to commit and push to your branch? :)

Yes, please do!  Thank you for taking the reins on this!

nmnobre:
Okay, I've combined all of our best ideas into a super succinct implementation that covers all of the initial prerequisites:

1) a file in the source data folder is modified
2) a file in the binary data folder is deleted
3) the binary data folder itself is deleted
4) a new file is added to the source data folder 

Plus:

5) a file in the binary data folder is modified
6) full compatibility with the CMake version we currently require (3.12)
7) the ability to also install the data folder _if_ it has been copied to the build directory

With regards to my big post in January, I've gone with a), as initially suggested by @helloworld922 and which is now a no-brainer due to us moving to 3.12, and 2) to preserve the existing behaviour of copying only when creating executables or when explicitly requested by the user.

Hopefully the brevity and simplicity of the changes here make this trivial to (re)review.

tzanio:

cjvogl:
@tzanio, this should be ready for re-merging into `next`

tzanio (approved):
Thank you, @cjvogl, @nmnobre, @helloworld922 and @justinlaughlin!

PR #4513, opened 2024-09-19
Improvements of general hyperbolic framework

najlkin:
This PR improves the general hyperbolic framework from #3598 in several way:

- Renamed `RiemannSolver` to `NumericalFlux` for generality (with deprecation of the old name for compatibility).
- Added Jacobians to of the convection terms of `HyperbolicFormIntegrator` (`AssembleElementGrad()` and `AssembleFaceGrad()`), using also the underlying (new) method of `NumericalFlux` for gradient (`Grad()`)
- Added new methods to `NumericalFlux` for the average flux (`Average()`) and its Jacobian (`AverageGrad()`), which rely on the new methods `ComputeAvgFlux(DotN)()` of `FluxFunction`. This is for example used in HDG #4350 
- Added component-wise upwinded flux (`ComponentwiseUpwindFlux`) and extended Rusanov flux
- ~~Added a switch for Riemann solvers to example 18(p)~~
- Added the normal flux implementation to `AdvectionFlux` and `BurgersFlux` and implemented the new methods
- Added a sign parameter to `HyperbolicFormIntegrator`
- Other minor improvements of the code and documentation

❓ Most of the code has been tested in `darcy-hdg-dev` branch. The regression tests there can be run by from `examples` folder by `regression_test.py`

| [#4513](https://github.com/mfem/mfem/pull/4513) | @najlkin | @tzanio | @cjvogl + @tarikdzanic + @pazner + @mlstowell | 9/21/24 | 1/7/25 | 1/15/25 | |

<details>
    
- [x] Code builds.
- [x] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
   - [x] All new public, protected, and private classes, methods, data members, and functions have full Doxygen-style documentation in source comments. Documentation should include descriptions of member data, function arguments and return values, template parameters, and prerequisites for calling new functions.
</details>


dohyun-cse:
Hi @najlkin,

Thank you for the great PR.
Including additional numerical flux and average flux is a great improvement.
I took a quick look and it looks good to me. I'll look into more detail early next week.
Would you please include the usage of other fluxes in `ex18.cpp` and `ex18p.cpp` to make new fluxes more accessible and for easier testing?

Best, Dohyun

najlkin:
A good point @dohyun-cse :wink: , I added the new Riemann solvers to `ex18` and `ex18p`, so there is a command line switch for them now. Rusanov and Godunov work fine. However, I am getting negative pressure with Engquist-Osher 😕 , any ideas?

najlkin:
Oh, I maybe see it. If u- > u+, then the integral maybe should be interpreted as negative. However, that would mean that there is no difference between Godunov and Engquist-Osher for monotonous fluxes... 🤔

najlkin:
Yes, I think it must be like that to satisfy symmetry when reversing the sweep/axis, i.e., when you swap u- and u+ and the sign of the normal 🤔 . My bad, I did not realize that. I will remove Engquist-Osher as it should coincide with Godunov.

tzanio:

fem/hyperbolic.hpp
```diff
@@ -61,58 +61,102 @@ class FluxFunction
...
-    * @brief Compute flux Jacobian. Optionally overloaded in the derived class
+    * @brief Compute average flux for over the given interval of states.
+    * Optionally overloaded in a derived class.
+    *
+    * @param[in] state1 state of the beggining of the interval (num_equations)
+    * @param[in] state2 state of the end of the interval (num_equations)
+    * @param[in] Tr element transformation
+    * @param[out] flux_ average flux from the given element at the current
+    * integration point (num_equations, dim)
+    * @return real_t maximum characteristic speed
```
cjvogl, line 105: Perhaps this is a copy-paste from `ComputeFlux` that needs to be updated?

fem/hyperbolic.hpp
```diff
@@ -61,58 +61,102 @@ class FluxFunction
...
+    * @param[in] normal normal vector, see mfem::CalcOrtho() (dim)
+    * @param[in] Tr face transformation
+    * @param[out] fluxDotN average normal flux from the given element at the
+    * current integration point (num_equations)
+    * @return real_t maximum (normal) characteristic velocity
+    */
+   virtual real_t ComputeAvgFluxDotN(const Vector &state1, const Vector &state2,
+                                     const Vector &normal,
+                                     FaceElementTransformations &Tr,
+                                     Vector &fluxDotN) const;
```
cjvogl, line 127: I feel that how one evaluates the average is less a property of the numerical flux function (i.e., it's more a decision to be made by the calling code via the existing `ComputeFlux` function).  Thoughts? Perhaps this it is better and/or necessary to have this method and I just need some clarification as to why.

cjvogl:
Sorry for adding and then deleting in-line comments: I thought it perhaps better to ask the question as a general comment. I understand the `FluxFunction` interface to be for the evaluation of the hyperbolic flux function.  This motivates the `FluxFunction::ComputeFlux` function, with the added property of the maximum characteristic speed.  Perhaps this is merely due to my lack of experience with the motivating HDG method, but I don't see computing the average as a property intrinsic to the hyperbolic flux function (i.e., why `FluxFunction::ComputeAvgFlux`).  Is it typical for there to be an analytic average expression specific to a given flux function that cannot or should not otherwise be evaluated by the `RiemannSolver`?

fem/hyperbolic.hpp
```diff
@@ -61,58 +61,102 @@ class FluxFunction
...
    virtual real_t ComputeFluxDotN(const Vector &state, const Vector &normal,
                                   FaceElementTransformations &Tr,
                                   Vector &fluxDotN) const;
 
    /**
-    * @brief Compute flux Jacobian. Optionally overloaded in the derived class
+    * @brief Compute average flux for over the given interval of states.
+    * Optionally overloaded in a derived class.
+    *
+    * @param[in] state1 state of the beggining of the interval (num_equations)
```
cjvogl, line 100:
```suggestion
    * @param[in] state1 state of the beginning of the interval (num_equations)
```

fem/hyperbolic.hpp
cjvogl, line 105: Is this now the maximum characteristic speed 
- across the interval
- between the two provided states
?

najlkin: With the assumption of monotonicity, these two definitions coincide, so I do not know if to specify. It might get more unclear than clear 😄

cjvogl: Maybe add that the flux is assumed monotonic for this method?

najlkin:
@cjvogl , I clarified the definition of the average flux, which is just $\int_{u_1}^{u_2}F(u)du / (u_2-u_1)$. I do not want to evaluate this integral numerically, which would be expensive to do at every integration point in space. Instead, you may evaluate it analytically for simple fluxes like those provided in `hyperbolic.hpp`, or decide to provide some "good enough approximation" for your custom fluxes 😉 .

cjvogl:
> I clarified the definition of the average flux, which is just ∫ u 1 u 2 F ( u ) d u / ( u 2 − u 1 ) . 

Ah ok, are the `ComputeAvgFlux` and `ComputAvgFluxDotN` methods only applicable to scale equations (i.e., $u_1, u_2$ but not $\vec{u}_1, \vec{u}_2$)?

najlkin:
In principle, it can be for vectors. There is the parameter `num_equations` of `FluxFunction`, so it is up to you, but the advection and Burgers fluxes, which I use as examples, are scalar. Also it should be mentioned that Jacobians are dense matrices, so you may have cross-component dependencies 😉 .

cjvogl:
> In principle, it can be for vectors.

Can you update the documentation to indicate how it would be for vectors/systems?

najlkin:
Ok, I have not thought about that question before, but I defined that average component-wise, because only partial averages make sense to me for non-homogeneous things like Euler equations. On the other hand, some alternatives exist when you have something homogeneous like components of momentum 🤔 But that is on the user to abuse the framework 😄

najlkin:
To clarify on the situation with systems, I have not got to that on the side of HDG, so I am not completely sure what will be needed there. I plan to make then another PR and provide examples for the shallow water or Euler fluxes :wink:. This PR is large already, so I did not want to wait for that...

cjvogl:
> To clarify on the situation with systems, I have not got to that on the side of HDG, so I am not completely sure what will be needed there. I plan to make then another PR and provide examples for the shallow water or Euler fluxes 😉. This PR is large already, so I did not want to wait for that...

That makes sense.  My main concern is that the documentation is sufficient so that it is clear what is currently expected in a derived class. I suppose another option would be to put in a verify statement that checks something like `num_equations == 1` is 1 with an error message noting that systems are not yet supported (which would then be removed in a future PR when there are examples for systems).

najlkin:
@cjvogl , that makes sense, but where would you put it? It is just an interface, so `FluxFunction::ComputeAvgFlux()` has just `MFEM_ABORT()`. The only implementations are in `AdvectionFlux` and `BurgersFlux`, which are both scalar by construction 🤔 .

cjvogl:
> @cjvogl , that makes sense, but where would you put it? It is just an interface, so `FluxFunction::ComputeAvgFlux()` has just `MFEM_ABORT()`. The only implementations are in `AdvectionFlux` and `BurgersFlux`, which are both scalar by construction 🤔 .

Ah, good point... maybe just note in the documentation that the average is as you have it with some additional wording to note that systems are not yet support.

najlkin:
@cjvogl , I added some notes to Rusanov and Godunov flux, as the simplification for systems is not really on the side of flux function, where I am just defining a general interface, but rather the implementation in the Riemann solvers, which treat systems by components. It was the case already before for classical `Eval()` and it is just extended now to `Average()`, where the flux function is assumed to comply with that.

fem/hyperbolic.hpp
najlkin: It is not assumed here, so I specified that it should be maximum over the interval in very general 😉 .

fem/hyperbolic.cpp
```diff
@@ -212,15 +448,290 @@ real_t RusanovFlux::Eval(const Vector &state1, const Vector &state2,
...
+      fluxFunction.ComputeAvgFluxDotN(state1, state2, nor, Tr, fluxN2);
+      fluxFunction.ComputeFluxJacobianDotN(state1, nor, Tr, JDotN);
+
+      grad = 0.;
+
+      for (int i = 0; i < fluxFunction.num_equations; i++)
+      {
+         // Only diagonal terms of J are considered
+         // lim_{u → u⁻} (F̄(u⁻,u)n - F(u⁻)n) / (u - u⁻) = ½J(u⁻)n
+         const real_t gr12 = (state1(i) != state2(i))?
```
tarikdzanic, line 693: Do you need a numerical tolerance here? The reciprocal term might be an issue

fem/hyperbolic.cpp
```diff
@@ -212,15 +448,290 @@ real_t RusanovFlux::Eval(const Vector &state1, const Vector &state2,
...
+         fluxFunction.ComputeFluxJacobianDotN(state1, nor, Tr, JDotN);
+      }
+
+      grad = 0.;
+
+      for (int i = 0; i < fluxFunction.num_equations; i++)
+      {
+         // Only diagonal terms of J are considered
+         // lim_{u → u⁻} (F(u)n - F̄(u⁻,u)n) / (u - u⁻) = ½J(u⁻)n
+         const real_t gr12 = (state1(i) != state2(i))?
```
tarikdzanic, line 728: Same here

tarikdzanic:
I might be wrong here but I think this implementation of the Godunov Riemann solver is valid only for scalar conservation laws. For the Euler equations in ex18 you'd need an exact Godunov-type Riemann solver.

fem/hyperbolic.cpp
najlkin: This is mainly for the IC or initial guess, where you have exactly the same values. Maybe one would need some tolerance for propagation of some infinitesimal waves, but "physical" infinitesimal waves are usually well above the floating point precision, so it might be rather a concern for propagation of some spurious waves and stability of the solution. It is just a simple finite difference, so I do not think it is a big issue, but I am open to adding something. Do you have a suggestion for some robust definition of such tolerance?

najlkin:
> I might be wrong here but I think this implementation of the Godunov Riemann solver is valid only for scalar conservation laws. For the Euler equations in ex18 you'd need an exact Godunov-type Riemann solver.

Well, I was not much concerned much about systems of equations in this PR and, [as said above](https://github.com/mfem/mfem/pull/4513#issuecomment-2371732095), I plan to make another PR when I get to them. Currently, Godunov is implemented the same way as Rusanov, working component-by-component. I just added it to example 18 on [the request above](https://github.com/mfem/mfem/pull/4513#issuecomment-2361327710), so please take it as it is for now. It seems to work fine in the example like Rusanov does, but I am open to some documentation changes for the Riemann solvers, for example? Because you are right that their current implementation is very simplistic. I am already mentioning that the Jacobians are only diagonal and fluxes component-wise calculated in the doxygen of their methods.

fem/hyperbolic.cpp
tarikdzanic: I'm more worried about u2-u1 being on the order of machine precision (but nonzero), it could return values that are unfeasibly large. If you're sure that it's always exactly the same, that's fine (although maybe add a comment). Otherwise, you could just do the switch based on `abs(state1(i) - state2(i)) > eps` instead of `state1(i) != state2(i)` and have `eps` be a few orders of magnitude larger than machine precision (same for the Rusanov flux).

tarikdzanic:
The example case there is very mild so I wouldn't expect it to fail there, but it would likely start having some issues with more difficult problems. Up to you if you want to leave it in (but maybe add a comment in the example mentioning something like it's not really meant for the Euler equations in its current form). My only worry is that people might take the example case to indicate that it's a Godunov-type Riemann solver for the Euler equations without knowing and use it in their codes which might cause them issues.

najlkin:
@tarikdzanic , I added a note to the description of the switch for Riemann solvers and also a runtime warning to console when Godunov is selected 😉 .

tarikdzanic:
Looks good to me!

tarikdzanic (approved):

cjvogl:
> @tarikdzanic , I added a note to the description of the switch for Riemann solvers and also a runtime warning to console when Godunov is selected 😉 .

I share @tarikdzanic 's [concern](https://github.com/mfem/mfem/pull/4513#issuecomment-2389660056) that including this `RiemannSolver` that is intended only for scalar laws in examples that are systems of conservation laws is problematic, mainly because the examples serve as documentation for how to use the `HyperbolicFormIntegrator` and related classes.  The `RusanovFlux` is intended for systems of conservation, so I don't think the `GodunovSolver` should be added to example 18 even if it happens works ok.

fem/hyperbolic.cpp
```diff
@@ -163,28 +255,138 @@ void HyperbolicFormIntegrator::AssembleFaceVector(
...
-   const RiemannSolver &rsolver,
-   const int IntOrderOffset)
-   : NonlinearFormIntegrator(),
-     rsolver(rsolver),
-     fluxFunction(rsolver.GetFluxFunction()),
-     IntOrderOffset(IntOrderOffset),
-     num_equations(fluxFunction.num_equations)
+void HyperbolicFormIntegrator::AssembleFaceGrad(
+   const FiniteElement &el1, const FiniteElement &el2,
+   FaceElementTransformations &Tr, const Vector &elfun, DenseMatrix &elmat)
```
cjvogl, line 265:
```suggestion
   FaceElementTransformations &Tr, const Vector &elfun, DenseMatrix &elmat) override
```

cjvogl (changes requested):
I see that `RiemannSolver::Grad` and `FluxFunction::ComputeFluxJacobianDotN` are added so that `HyperbolicFormIntegrator::AssembleFaceGrad` and `HyperbolicFormIntegrator::AssembleElementGrad` can be implemented... and presumably used in #4350.

I believe the other methods `RiemannSolver::Average`, `RiemannSolver::AverageGrad`, `FluxFunction::ComputeAvgFlux` and `FluxFunction::ComputeAvgFluxDotN` are included here so that additional methods added to `NonlinearFormIntegrator` class in #4350 can be implemented in `HyperbolicFormIntegrator` (also in #4350) .  The issue with including the other methods here is that 
- the design choice of adding them to `RiemannSolver` and `FluxFunction` 
- the assumptions of their implementation (e.g., scalar laws vs systems of laws, monotonicity)

cannot be reviewed without the context of the changes to the `NonlinearFormIntegrator` in #4350 (which is a WIP).  I suggest removing them from this PR so they can be reviewed alongside the context in which they are being added in #4350 .

najlkin:
> I share @tarikdzanic 's [concern](https://github.com/mfem/mfem/pull/4513#issuecomment-2389660056) that it is problematic to include the `GodunovSolver` that is intended only for scalar laws in examples that are systems of conservation, mainly because example 18 serves as documentation for how to use the `HyperbolicFormIntegrator` and related classes. The `RusanovFlux` is intended for systems of conservation laws, so I don't think the `GodunovSolver` should be added to example 18 even if it happens works ok.

That was the reason why I did not include it in the first place. On the second thought, it might be the reason to include it to show its usage with  `HyperbolicFormIntegrator` and related classes, just the application is not perfectly suited for it, but the user is clearly informed about that. It is also not like Godunov flux was somehow horrendously wrong. They do nearly the same thing, where Godunov can be translated to Rusanov with $\lambda=\left|\frac{F(u^+)n-F(u^-)n}{u^+-u^-}\right|$, which is just a finite difference approximation of the velocity. The thing is this velocity is taken for each equation independently in Godunov instead of the maximal one over the whole system for Rusanov, which is a crime, but not that bad 😁 . I am fine with either option, i.e., this small demo without sample runs and with a clear warning, or deferring example 18 to the PR for systems. What is your opinion @dohyun-cse ?

najlkin:
> I believe the other methods `RiemannSolver::Average`, `RiemannSolver::AverageGrad`, `FluxFunction::ComputeAvgFlux` and `FluxFunction::ComputeAvgFluxDotN` are included here so that additional methods added to `NonlinearFormIntegrator` class in #4350 can be implemented in `HyperbolicFormIntegrator` (also in #4350) . The issue with including the other methods here is that
>     * the design choice of adding them to `RiemannSolver` and `FluxFunction`
>     * the assumptions of their implementation (e.g., scalar laws vs systems of laws, monotonicity)
> cannot be reviewed without the context of the changes to the `NonlinearFormIntegrator` in #4350 (which is a WIP). I suggest removing them from this PR so they can be reviewed alongside the context in which they are being added in #4350 .

@cjvogl , the thing with #4350 is that it is huge already and still far from merging. It is not in anyone's capacity to review that PR as is (and considering how reviewers were assigned so far, you might become one of them actually 😄 ). What I am trying to do is to spin off smaller PRs focused on certain topics and without much of specific connection of HDG/RT. This is also the case here, I have stripped the code of the specific parts for HDG and the only extra feature here is that calculation of the average flux over an interval, which is a fairly general thing and can be potentially useful for someone else. It is also nothing fancy and just follows the lines of other methods, so I believe it can be easily reviewed here 😉 . If you want to see it in action, [I suggested the regression tests in `darcy-hdg-dev` in the header ☝️](https://github.com/mfem/mfem/pull/4513#issue-2534989074) .

cjvogl:
> > I believe the other methods `RiemannSolver::Average`, `RiemannSolver::AverageGrad`, `FluxFunction::ComputeAvgFlux` and `FluxFunction::ComputeAvgFluxDotN` are included here so that additional methods added to `NonlinearFormIntegrator` class in #4350 can be implemented in `HyperbolicFormIntegrator` (also in #4350) . The issue with including the other methods here is that
> > * the design choice of adding them to `RiemannSolver` and `FluxFunction`
> > * the assumptions of their implementation (e.g., scalar laws vs systems of laws, monotonicity)
> > cannot be reviewed without the context of the changes to the `NonlinearFormIntegrator` in #4350 (which is a WIP). I suggest removing them from this PR so they can be reviewed alongside the context in which they are being added in #4350 .
> 
> @cjvogl , the thing with #4350 is that it is huge already and still far from merging. It is not in anyone's capacity to review that PR as is (and considering how reviewers were assigned so far, you might become one of them actually 😄 ). What I am trying to do is to spin off smaller PRs focused on certain topics and without much of specific connection of HDG/RT. This is also the case here, I have stripped the code of the specific parts for HDG and the only extra feature here is that calculation of the average flux over an interval, which is a fairly general thing and can be potentially useful for someone else. It is also nothing fancy and just follows the lines of other methods, so I believe it can be easily reviewed here 😉 . If you want to see it in action, [I suggested the regression tests in `darcy-hdg-dev` in the header ☝️](https://github.com/mfem/mfem/pull/4513#issue-2534989074) .

>  I believe it can be easily reviewed here

I unfortunately disagree.  As a user, I find abstract classes with numerous methods defined with failures very difficult to work with as a user, especially if there is no code that uses those methods.  Specifically, it then becomes unclear to me what I actually need to implement if I want to derive from that abstract class (e.g., how many methods of `RiemannSolver` do I need to implement if I want to define a new approximate Riemann solver for the Euler equations... it is clear to me in `main` and becomes unclear to me in this branch).  As a maintainer, I find introducing "not yet" used methods bad practice: what if someone sees those methods and uses them in a way that contradicts what you are doing in #4350 ?

That all said, there are more reviewers on this PR that will perhaps disagree with my philosophy and provide that second approval to bypass me.

najlkin:
> As a user, I find abstract classes with numerous methods calls to MFEM_ABORT (or similar) very difficult to work with as a user, especially if there is no code that uses those methods. Specifically, it then becomes unclear to me what I actually need to implement if I want to derive from that abstract class (e.g., how many methods of RiemannSolver do I need to implement if I want to define a new approximate Riemann solver for the Euler equations...

@cjvogl  I agree with you it can get very unclear and convoluted, that is why I have improved the doxygen, so methods of `HyperbolicFormIntegrator` mention what they are using exactly, so what needs to be implemented.

> As a code maintainer, I find introducing "not yet" used methods bad practice

They are not hidden or anything. They are offered to the users, but like many other things they do not have a specific example. What is more, there is a clear intention to deliver that example in the future 👍 .

> what if someone sees those methods and uses them in a way that contradicts what you are doing in https://github.com/mfem/mfem/pull/4350 ?

That cannot happen, it is clearly defined here what they do and I am bound by the definition to use them that way as other users/developers.

> What if I'm asked to fix or extend those methods before https://github.com/mfem/mfem/pull/4350 is merged?

Please do so, I plan to rebase `darcy-hdg-dev` branch or merge later changes. In the case of extending them, it is a good practice to discuss the changes with the original author, so I should know about them and we may coordinate them.

Ok, what are the opinions of others?

cjvogl:
> > what if someone sees those methods and uses them in a way that contradicts what you are doing in #4350 ?
> 
> That cannot happen, it is clearly defined here what they do and I am bound by the definition to use them that way as other users/developers.
> 
It is dangerous to tempt the users in this way.... just saying.

najlkin:
@cjvogl , I added some notes about usage to `FluxFunction` and `RiemannSolver`. That is the maximum I can do, as removing the average-related functions would make this PR nearly useless for the HDG development. I need to chop the beast 🥩 😄 . The functions are not much specific and no worries, the plan is to merge HDG this fiscal year and I am here to oversee it to that end.

cjvogl:
To avoid the look that this PR is being neglected by me, I'll note that Jan and I have discussed this offline last week and will meet again this meet to clarify the connection between
- `RiemannSolver::Average`
- `FluxFuntion::ComputeAvgFlux`
- $\frac{1}{u_2 - u_1} \int_{u_1}^{u_2} F(u) du$
- $\frac{1}{u_2 - u_1} \int_{u_1}^{u_2} \hat{F}(u_1, u) du$
(and the related set of `RiemannSolver::Grad`, etc.)

cjvogl:
Documenting some takeaways from our discussion from today:
- `RiemannSolver::Average` is currently implemented to be a particular average of the numerical flux $\int_{u1}^{u2} \hat{F}(u1,u,x) du / (u2-u1)$ where $\hat{F}$ is a numerical flux that may or may not be based on the solution to the Riemann problem corresponding to $u_t + \frac{\partial}{\partial x_n} (F.\hat{n}) = 0$
- this concept of an average in terms of a Riemann solver may also refer to something like $\int_0^1 F'(u(\xi)) d\xi$ where $u(\xi) = u1 + (u2 - u1) \xi$ (see, e.g., Roe linearization).
- the `RiemannSolver` interface can be used to derive general numerical fluxes (i.e., ones not necessarily based on the solution to a Riemann problem)
- to avoid the conflict in terminology of an average, a suitable solution is to rename the `RiemannSolver` class to `NumericalFlux` (or something similar), with an optional deprecated typedef if backwards compatibility is required

@najlkin , please feel free to suggest changes to this summary

fem/hyperbolic.hpp
```diff
@@ -222,28 +361,55 @@ class HyperbolicFormIntegrator : public NonlinearFormIntegrator
...
+                         const Vector &elfun, DenseMatrix &elmat) override;
 };
 
 
 /**
  * @brief Rusanov flux, also known as local Lax-Friedrichs,
  *    F̂ n = ½(F(u⁺,x)n + F(u⁻,x)n) - ½λ(u⁺ - u⁻)
  * where λ is the maximum characteristic velocity
- *
+ * @note The implementation assumes monotonous F(u,x) in u
```
cjvogl, line 412: This statement doesn't generalize to systems... perhaps it can be moved to the particular part of the implementation that assumes monotonicity with scalar equations?

fem/hyperbolic.hpp
```diff
@@ -258,26 +424,215 @@ class RusanovFlux : public RiemannSolver
...
+   mutable DenseMatrix JDotN;
+#endif
+};
+
+/**
+ * @brief Godunov flux, a special case of Engquist-Osher flux,
+ *    F̂ n = min F(u)n    for u⁻ ≤ u⁺, u ∈ [u⁻,u⁺]
+ *    F̂ n = max F(u)n    for u⁻ > u⁺, u ∈ [u⁺,u⁻]
+ * @note The implementation assumes monotonous F(u,x) in u
+ */
```
cjvogl, line 525: I'm not familiar with Godunov flux defined this way.  I've seen it widely used as being the choice for the numerical flux $\hat{F}(\vec{u_-},\vec{u_+},x)$ such that $\hat{F}(\vec{u_-},\vec{u_+},x) = F(\tilde{u}(\vec{u_-},\vec{u_+},x))$, where $\tilde{u}(\vec{u_-},\vec{u_+},x)$ is the Riemann solution evaluated at $x$ to the corresponding Riemann problem
$$\frac{\partial \vec{u}}{\partial t} + \frac{\partial}{\partial \tilde{x}} \vec{F}(\vec{u}) = \vec{0}$$
with initial condition $\vec{u} = \vec{u_-}$ at $\tilde{x} < x$ and $\vec{u} = \vec{u_+}$ for $\tilde{x} > x$.  This implementation looks to be specific for scalar equations with monotonic flux functions... perhaps this should be named `EnquistOsherFlux` to be consistent with terminology in, e.g., https://www.sciencedirect.com/science/article/pii/S009630031630621X (which is consistent with other works I'm familiar with)?

cjvogl (changes requested):
After the naming change of `RiemannSolver` to `NumericalFlux` [(or something similar)](https://github.com/mfem/mfem/pull/4513#issuecomment-2492618555) and the issue noted inline with `GodunovFlux` I think this is good to go (I've also suggested some comment changes along the way to be more consistent with systems of equations)

fem/hyperbolic.hpp
```diff
@@ -129,21 +195,87 @@ class RiemannSolver
...
    virtual real_t Eval(const Vector &state1, const Vector &state2,
                        const Vector &nor, FaceElementTransformations &Tr,
                        Vector &flux) const = 0;
 
+   /**
+    * @brief Evaluates Jacobian of the normal numerical flux for the given
+    * states and normal. Optionally overloaded in a derived class.
+    *
+    * Used in HyperbolicFormIntegrator::AssembleFaceGrad() for Jacobian
+    * of the convective term at the face.
```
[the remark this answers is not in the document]
najlkin: The notation is a bit tricky, because the derivative is over both sides, so I took the middle path saying it is Jacobian of the integral term.

fem/hyperbolic.hpp
najlkin: It is the same definition (eq. 2.17 in the paper), just assuming monotonicity similarly to `RusanovFlux` discussed above, so min/max over the interval reduces to min/max of the endpoints. I am not sure what we will do for systems, so I would leave it as it is for now.

najlkin:
@dohyun-cse , technically, it is not necessary, but it would be very nice if you could make a review of the PR as you are the original author. Your feedback would be appreciated.

examples/ex18.cpp
```diff
@@ -89,6 +90,8 @@ int main(int argc, char *argv[])
                   "Number of times to refine the mesh uniformly.");
    args.AddOption(&order, "-o", "--order",
                   "Order (degree) of the finite elements.");
+   args.AddOption(&rsolver_type, "-rs", "--riemann-solver",
```
dohyun-cse, line 93: Since we already changed `RiemannSolver` to `NumericalFlux`, would it be better to rename variable `rsolver_type` and option `rs` to `numer_flux_type` and `nf`?

najlkin: Maybe, but the notion of Riemann solver makes sense in the context of hyperbolic equations. The concern was mainly generalization to other things 🤔 .

fem/hyperbolic.cpp
```diff
@@ -33,15 +56,15 @@ void HyperbolicFormIntegrator::AssembleElementVector(const FiniteElement &el,
    // shape function value at an integration point
    Vector shape(dof);
    // derivative of shape function at an integration point
-   DenseMatrix dshape(dof, el.GetDim());
+   DenseMatrix dshape(dof, Tr.GetSpaceDim());
```
dohyun-cse, line 59: I don't believe that we can generalize this class to a general manifold without significant modifications. In that case, is it necessary to use `GetSpaceDim()` instead of `el.GetDim()`? It certainly improves readability, but this might give the impression that it works for general manifold as we distinguish `dim` and `sdim`.

fem/hyperbolic.cpp
```diff
@@ -33,15 +56,15 @@ void HyperbolicFormIntegrator::AssembleElementVector(const FiniteElement &el,
...
    // state value at an integration point
    Vector state(num_equations);
    // flux value at an integration point
    DenseMatrix flux(num_equations, el.GetDim());
```
dohyun-cse, line 63: Also, this should be `Tr.GetSpaceDim()` if we want to use the physical derivative with `sdim != dim`.

fem/hyperbolic.cpp
```diff
@@ -77,7 +100,77 @@ void HyperbolicFormIntegrator::AssembleElementVector(const FiniteElement &el,
...
+
+      // integrate (J(u,x), grad v)
+      const real_t w = ip.weight * Tr.Weight() * sign;
+      for (int di = 0; di < num_equations; di++)
+         for (int dj = 0; dj < num_equations; dj++)
+            for (int i = 0; i < dof; i++)
+               for (int j = 0; j < dof; j++)
+                  for (int d = 0; d < fluxFunction.dim; d++)
+                  {
+                     grad(di*dof+i, dj*dof+j) += w * dshape(i,d) * shape(j) * J(di,dj,d);
```
dohyun-cse, line 172: I think we need `sign` here

fem/hyperbolic.cpp
```diff
@@ -163,28 +256,138 @@ void HyperbolicFormIntegrator::AssembleFaceVector(
...
+         for (int dj = 0; dj < fluxFunction.num_equations; dj++)
+         {
+            // pre-multiply integration weight to Jacobian
+            const real_t w = -ip.weight * sign * JDotN(di,dj);
+            for (int j = 0; j < dof1; j++)
+            {
+               // Test side 1
+               for (int i = 0; i < dof1; i++)
+               {
+                  elmat(i+dof1*di, j+dof1*dj) += w * shape1(i) * shape1(j);
```
dohyun-cse, line 352: Also `sign` here

fem/hyperbolic.cpp
najlkin: There is in `w` 😉

fem/hyperbolic.cpp
najlkin: There is in w 😉

fem/hyperbolic.cpp
najlkin: It is more a psychological question, I think 🙂 . I just did not want to keep around some obvious mistakes. Impression or not, the code will crash, but not here but at the point where the implementation is deficient for manifolds, i.e., multiplication of the flux and gradient. Anyway, not very important, I may change it back if you insist.

fem/hyperbolic.cpp
najlkin: Not really, the velocity must be tangential to the manifold (to stay on the manifold). It should be this integrator, which handles the tangential integration or there could be something like `FluxFunction::ComputePhysFlux()`. Anyway, I would leave it for now.

fem/hyperbolic.hpp
cjvogl: It is not the same definition: Equation 2.17 is the Godunov flux *specifically* for a scalar system, which does not generalize to systems (also mentioned in the paper). This is similar to implementing a Newton solver that can only specifically for scalar equations and naming it `NewtonSolver`: a user will likely think they can use `NewtonSolver` for a nonlinear system of equations. If you are opposed to naming it `ScalarGodunovFlux` then you need to safeguard against users assuming they can apply it to systems (being explicit in the documentation that it's only for scalar equations, adding a verify or assert statement to check the dimension is 1, etc.).

Also, I'm not sure you need to assume monotonicity, i.e., this is the Godunov flux for a scalar conservation flux regardless if the flux is monotonic?  I could be wrong.

fem/hyperbolic.hpp
cjvogl: Sounds like what you want to say is that the flux is "convex" instead of "monotonous"?  Technically, the proper way to document this would be to just update the documentation to say that lambda is `max(|F(u-,x)n|, |F(u+,x)n|)` and not that it's the maximum characteristic velocity... because those two are not generally equivalent.  This would avoid trying to determine under what conditions `max(|F(u-,x)n|, |F(u+,x)n|)` is the maximum characteristic speed.

najlkin: Ok, I put the exact definition to the brief description and explanation to the note. However, it is important to explain to the user under what assumptions it is Rusanov flux actually and when it is provably stable, because you do not want to deduce that by yourself from the implementation (or its description). So it is Rusanov flux simplified under the assumption of a monotonous flux (as the endpoints are arbitrary, only a monotonous function is convex in all cases).

fem/hyperbolic.hpp
najlkin: The thing is that we do not have any Riemann solver for systems, as both assume monotonicity, yet there is example 18(p) using (originally one of) them for a system and working just fine and this framework has made it to a release of MFEM, so maneuvering is very limited. Moreover, it does not make sense to rename one of them when both apply the same assumption. Secondly, we may decide to extend these classes for systems of equations instead of using a different name, but it is hard to predict now. The best I can do is to put warning somewhere, I think 🤔 .
Without the assumption of monotonicity, you would have to scan somehow the whole interval and still you would have no guarantee, because flux is not necessarily polynomial, so just theoretically it could be a delta function (or its closest representable continuous projection) and you would need a (nearly) infinite number of evaluation points to find it. In other words, people always use some simplifying assumptions, which can be weaker for sure, but this is fine for my purposes and the basic examples of linear advection and Burgers, which are offered.

najlkin: Ok, I moved the warning from example 18 directly to the constructor of `GodunovFlux`.

fem/hyperbolic.hpp
```diff
@@ -414,20 +418,18 @@ class HyperbolicFormIntegrator : public NonlinearFormIntegrator
 /**
  * @brief Rusanov flux, also known as local Lax-Friedrichs,
  *    F̂ n = ½(F(u⁺,x)n + F(u⁻,x)n) - ½λ(u⁺ - u⁻)
- * where λ is the maximum characteristic velocity
- * @note The implementation assumes monotonous F(u,x) in u
+ * where λ = max(|F(u⁺,x)n|, |F(u⁻,x)n|).
+ * @note This construction assumes monotonous F(u,x) in u,
+ * implying λ is the maximum characteristic velocity.
```
cjvogl, line 423: This does not make sense for systems.  I've given suggestions how to remedy this.

fem/hyperbolic.hpp
```diff
@@ -19,17 +19,17 @@ namespace mfem
...
+//     ∫_K F(u):∇v,   -∫_f F̂(u)⋅n[v]
 //
-// Here, T is an element, e is an edge, and [[⋅]] is jump. This form integrator
-// is coupled with RiemannSolver that implements the numerical flux F̂. For
-// RiemannSolver, the Rusanov flux, also known as local Lax-Friedrichs flux, is
-// provided.
+// Here, K is an element, f is a face, n normal and [⋅] is jump. This form
+// integrator is coupled with NumericalFlux that implements the numerical flux
+// F̂. For NumericalFlux, the Rusanov flux, also known as local Lax-Friedrichs
+// flux, or Godunov flux are provided.
```
cjvogl, line 32: Godunov flux only provided for scalar conservation laws.

fem/hyperbolic.hpp
cjvogl: Example 18 was using the Rusanov flux as a numerical flux for a system in a very valid way.  This PR is introducing a class that implements the Godunov flux for a scalar system and then applies it component-wise to a system, which is invalid because that is not a Godunov flux, i.e., the flux function from the conservation law evaluated at the Riemann solution. I will not approve until this is addressed in the manner in which I have already requested.

fem/hyperbolic.hpp
cjvogl: Burgers flux $f(u) = \frac{1}{2} u^2$ is not monotonic.  As I suggested in another comment, perhaps "convex" is the appropriate property?

examples/ex18.cpp
cjvogl: I agree with this suggestion: in this context, the Rusanov flux is used a numerical flux and not a Riemann solver (i.e., it's not providing the value of the Riemann solution at x/t = 0).

fem/hyperbolic.hpp
cjvogl: I took another look through the implementation and the statement

>  The reason is that lambda is taken as max(|F(u-,x)n|, |F(u+,x)n|)

is not correctly capturing how lambda is currently implemented.  Given this thread is on an outdated version of the documentation, I'll elaborated on the updated version.

fem/hyperbolic.hpp
cjvogl: @dohyun-cse should correct me if I'm wrong, but what the current implementation of Rusanov / (local) Lax Friedrichs flux does assume is that the maximum characteristic velocity is given by $max( |\lambda(u_-)|, |\lambda(u_+)|)$ or in other words, that the maximum characteristic velocity overall (or between $u_-$ and $u_+$) is found by evaluating the maximum characteristic velocity at $u_-$ and $u_+$.  Note the maximum characteristic velocity at $u_\pm$ is **not** $F(u_\pm)$** as introduced in 5895c08.  This needs to be remedied.

I suggest simply stating that the $\max |\frac{df}{du}(u)|$ quantity in the Rusanov / (local) Lax Friedrichs flux is assumed to be $\max_{u = u_\pm} |\frac{df}{du}(u)|$

fem/hyperbolic.hpp
cjvogl: For what it's worth, the way Godunov flux is typically implemented is that it leverages an abstract Riemann solver, which is typically problem specific, to provide the Riemann solution at x/t = 0 to be used as the numerical flux, i.e., the $\vec{u}(x/t)$ in $\vec{f}(\vec{u}(x/t))$.  Here, you've hard-coded a Riemann solver for scalar conservation laws, effectively that $u(x/t)$ is either $u_-$ (right-moving Riemann solution), $u_+$ (left-moving Riemann solution), or such that $f'(u) = 0$ (rarefaction Riemann solution). The particular form you've chosen leverages the particular properties that determine whether the Riemann solution is right-moving, left-moving, or a rarefaction fan to get the form you're using.

For linear systems, exact Riemann solvers are typically supplied (see, e.g., [my work showing a modification to standard linear elasticity Riemann solver](https://link.springer.com/article/10.1007/s10915-017-0459-y)).  For nonlinear systems, approximate Riemann solvers have been carefully constructed for common nonlinear systems (with that work continuing).  Using the scalar conservation law Riemann solution for general nonlinear systems is widely considered **not valid** except perhaps some corner cases, otherwise there wouldn't be need for more work on approximate Riemann solvers. I believe this invalid application of a Riemann solver for scalar equations to the Euler equations (which has a rich history of carefully constructed & tailored Riemann solvers) is what @tarikdzanic was also worried about in [this comment](https://github.com/mfem/mfem/pull/4513#issuecomment-2386745754) and [this comment](https://github.com/mfem/mfem/pull/4513#issuecomment-2389660056).

fem/hyperbolic.hpp
najlkin: Ahh, yes, we are talking about the (characteristic) velocity, so it is the derivative. I just copied that over from elsewhere. I need to fix that everywhere.
Regarding systems, it makes some sense, because it should be the maximal velocity (but what it really is depends on the `FluxFunction`) and a scalar function can be monotonous in multiple parameters.

examples/ex18.cpp
najlkin: Ok, done.

fem/hyperbolic.hpp
najlkin: Mentioned it is component-wise.

fem/hyperbolic.hpp
najlkin: The flux F(u,x) must be monotonous in u, so it is guaranteed that max_u (F(u,x)) for u in [u-,u+] is the same as max(F(u-,x),F(u+,x)). In different notations it might be called differently, but here it makes sense this way.
You are right it does not work perfectly for Burgers, only works when the without a change of sign. That is limiting, but documented and with the current framework I do not see other options.
I am not saying this works fully generally for systems and that is why I moved the warning from example 18  (introduced based on the mentioned requests) directly to the constructor, warning it is a component-wise implementation.

cjvogl: > The flux F(u,x) must be monotonous in u, so it is guaranteed that max_u (F(u,x)) for u in [u-,u+] is the same as max(F(u-,x),F(u+,x)).

Whether or not max_u (F(u,x)) for u in [u-,u+] is the same as max(F(u-,x),F(u+,x)) does not address the issue that you've hardcoded a Riemann solution into `GodunovSolver` that is only valid for scalar conservation laws.

> You are right it does not work perfectly for Burgers, only works without a change of the sign. 

No, I said the opposite: the Riemann solution that you've hardcoded into `GodunovSolver` is indeed the Riemann solution for Burgers equation, because Burgers equation has a flux that is convex.

> I am not saying this works fully generally for systems and that is why I moved the warning from example 18 (introduced based on the mentioned requests) directly to the constructor, warning it is a component-wise implementation.

This does not address my concern that a class titled `GodunovSolver` indicates that you've implemented a Godunov solver for system when the hardcoded Riemann solution is not valid for use on systems.

My recommendation remains that this PR not be merged until this issue is adequately addressed (suggestions for how to do so are found above).

fem/hyperbolic.hpp
cjvogl: Marking thread as resolved as the line referenced is now outdated.

examples/ex18.cpp
```diff
@@ -89,6 +90,8 @@ int main(int argc, char *argv[])
                   "Number of times to refine the mesh uniformly.");
    args.AddOption(&order, "-o", "--order",
                   "Order (degree) of the finite elements.");
+   args.AddOption(&num_flux_type, "-nf", "--numerical-flux",
+                  "Numerical flux: 1 - Rusanov, 2 - Godunov (component-wise).");
```
cjvogl, line 94:
```suggestion
                  "Numerical flux: 1 - Rusanov, 2 - Godunov (using scalar equation Riemann solver).");
```

CHANGELOG
```diff
@@ -22,6 +22,13 @@ Discretization improvements
 
 - Added support for external boundary submeshes with nonconformal mesh adaptation.
 
+- Added assembly of Jacobians to `HyperbolicFormIntegrator`.
+
+- Added average fluxes to `RiemannSolver` and `FluxFunction`.
+
+- Added Godunov flux (`GodunovFlux`) and added it as an option to examples 18
+  and 18p.
```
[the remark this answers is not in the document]
najlkin: Done.

fem/hyperbolic.hpp
najlkin: Ok, to unblock this, I generalized `RusanovFlux::Grad()` for systems, so we have at least one `NumericalFlux` for them. Hence, it makes sense to rename Godunov to `ScalarGodunovFlux` (https://github.com/mfem/mfem/pull/4513/commits/1e4f2b5d2bc5da72b24241670d613b1f9a63ba46).

examples/ex18.cpp
najlkin: I do not find this description very clear honestly. First, it is not clear how something scalar is applied to vectors. Secondly, Godunov is an approximate Riemann solver itself, so it is confusing what is using what. While "component-wise" says it relatively clearly within the limited space. Maybe "component-wise scalar"?

dohyun-cse: How about `Scalar Godunov (component-wise)`?

dohyun-cse (approved):

examples/ex18.cpp
cjvogl: >Godunov is an approximate Riemann solver itself, so it is confusing what is using what. 

No, the Godunov method can use either an exact Riemann solution or an approximate Riemann solution. As I've noted in other comments, consider the Godunov method not as Riemann solver (in that it does not provide a Riemann solution) but instead the Godunov method **uses** a Riemann solution (typically provided by a Riemann solver) to provide a numerical flux. Other methods exist to provide a numerical flux given a Riemann solution, e.g., the wave propagation algorithm by LeVeque.

> it is not clear how something scalar is applied to vectors. 

I agree and would approve not using here for the Euler equations

> Maybe "component-wise scalar"?

If your suggestion is "2 - Godunov (component-wise scalar)", then that works for me.

cjvogl: I'll note that whatever change is decided on, ex18 should be updated to match.

fem/hyperbolic.hpp
```diff
@@ -19,17 +19,17 @@ namespace mfem
...
+//     ∫_K F(u):∇v,   -∫_f F̂(u)⋅n[v]
 //
-// Here, T is an element, e is an edge, and [[⋅]] is jump. This form integrator
-// is coupled with RiemannSolver that implements the numerical flux F̂. For
-// RiemannSolver, the Rusanov flux, also known as local Lax-Friedrichs flux, is
-// provided.
+// Here, K is an element, f is a face, n normal and [⋅] is jump. This form
+// integrator is coupled with NumericalFlux that implements the numerical flux
+// F̂. For NumericalFlux, the Rusanov flux, also known as local Lax-Friedrichs
+// flux, or component-wise Godunov flux are provided.
```
cjvogl, line 32:
```suggestion
// flux, or Godunov flux for scalar conservation laws are provided.
```

fem/hyperbolic.hpp
```diff
@@ -222,62 +373,269 @@ class HyperbolicFormIntegrator : public NonlinearFormIntegrator
...
+};
 
+/**
+ * @brief Godunov flux for scalar equations
+ *
+ * Godunov flux for scalar equations, a special case of Engquist-Osher flux,
+ * is defined as follows:
+ *    F̂ n = min F(u)n    for u⁻ ≤ u⁺, u ∈ [u⁻,u⁺]
+ *    F̂ n = max F(u)n    for u⁻ > u⁺, u ∈ [u⁺,u⁻]
+ * @note The implementation assumes monotonous F(u,x) in u
```
cjvogl, line 535: This is incorrect: it assumes **convex** F(u,x) in u.

examples/ex18.cpp
cjvogl: Apologies for any confusion from emails: I had accidentally left comments "pending" that required some cleanup.

> Godunov is an approximate Riemann solver itself

No, the Godunov method can use an exact Riemann solution or an approximate Riemann solution.  As I've noted elsewhere, consider the Godunov method to **use** a Riemann solution / solver to provide a numerical flux instead of being a Riemann solver (i.e., the Godunov method itself does not supply a Riemann solution and there are other methods for providing a numerical flux from a Riemann solution, e.g., the wave propagation algorithm by LeVeque). Such a view should clear up the confusion on what is using what.

> Maybe "component-wise scalar"?

If the suggestion is "2 - Godunov (component-wise scalar)", then that would address the original concern behind this thread.  Given, as you noted, it is not clear that applying something scalar to a vector is valid (and in this case, I don't think it's valid to use a scalar Riemann solution for the Euler equations), I would also support not providing this option at all.

fem/hyperbolic.hpp
```diff
@@ -222,62 +373,273 @@ class HyperbolicFormIntegrator : public NonlinearFormIntegrator
...
 
 
 /**
  * @brief Rusanov flux, also known as local Lax-Friedrichs,
  *    F̂ n = ½(F(u⁺,x)n + F(u⁻,x)n) - ½λ(u⁺ - u⁻)
- * where λ is the maximum characteristic velocity
- *
+ * where λ = max(|dF(u⁺,x)/du⁺⋅n|, |dF(u⁻,x)/du⁻⋅n|).
+ * @note This construction assumes monotonous |dF(u,x)/du⋅n| in u,
+ * implying λ is the maximum characteristic speed.
```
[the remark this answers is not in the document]
najlkin: This is not very instructive for the users and does not say anything more than the code and requires some deduction to understand what that means practically. For smooth functions (implicitly assumed), only monotonous functions satisfy this property for arbitrary intervals (for which the numerical flux must work).

examples/ex18.cpp
najlkin: As I see it, the current framework is structured such way that `NumericalFlux` (formerly `RiemannSolver`) is based on:

- approximate Riemann solver - Both currently implemented classes usable with any `FluxFunction` under some necessary assumptions on monotonicity/convexity.
- exact Riemann solver - Could be some class directly tailored to some problem and solving the Riemann problem analytically, which would _not_ take any `FluxFunction` as input, because some logic for the flux would have to be "baked" into the class, going beyond the interface of a `FluxFunction`. Currently not implemented.

Anyway, it seems we have the result 2 vs. 1 reviewer about example 18. It was not my original intention anyway, so I will remove it. When we implement systems, we may return to this switching code.

najlkin: Removed in https://github.com/mfem/mfem/pull/4513/commits/9943072dcaba138792ccb3f0962a042b33f828e2 .

fem/hyperbolic.hpp
najlkin: Sorry for the mix up with Rusanov (due to that copied over description), the assumptions are different. Here, we need just monotonicity, i.e., min/max F(u) over some interval [u-,u+] be min/max(F(u-), F(u+)). So such function can be, for example, convex (not crossing min), concave (not crossing max) or inflex.

cjvogl: No, this is the numerical flux that results when Godunov's method is used with the exact Riemann solution for a scalar conservation law with convex flux, with *convex* being used here in the context as defined by [Wikipedia](https://en.wikipedia.org/wiki/Convex_function) (and elsewhere).

Again, consider that Burgers flux is **not** monotonic, yet this is still the Godunov using the **exact Riemann solution** for Burgers equation (characteristic flow to the right or left, a shock moving left or right, or a rarefaction fan).

fem/hyperbolic.hpp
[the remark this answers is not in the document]
cjvogl: What is the underlying definition of the Rusanov flux? I know it as $\hat{F}(u^-,u^+) = \frac{1}{2}(F(u^-) + F(u^+)) - \max \| F'(u) \| (u^+ - u^-)$, or in the case of the local Lax Friedrichs version, $\hat{F}(u^-,u^+) = \frac{1}{2}(F(u^-) + F(u^+)) - \max_{u \in R(u^-,u^+)} \| F'(u) \| (u^+ - u^-)$, $R$ represents the region between $u^-$ and $u^+$. Do you disagree?

If you do not disagree, then the assumption of monotonicity is too restrictive.  Consider a flux where $dF/du$ is not monotonic but has extrema at the endpoints for particular $u^-$ and $u^+$, e.g., $F(u) = \frac{1}{2} u^2 (\frac{1}{6} u^2 -1)$ with $u^- = -10$ and $u^+ = 5$... the extrema of $|dF/du| = |\frac{1}{3}u^3 - u|$ for $u \in [-10,5]$  is still at $u^-=-10$ despite $|dF/du|$ not being monotonic.

examples/ex18.cpp
cjvogl: As I see it, there are numerical fluxes that do not use Riemann solutions at all, e.g., the Rusanov / Local Lax Friedrichs flux.

fem/hyperbolic.hpp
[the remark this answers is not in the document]
najlkin: There is nothing like particular `u-` and `u+`, that is the thing 🙂 . The class must work for arbitrary ones as they are input parameters of the method, not chosen by the class or set in the constructor of the class at least. Practically, they are typically coming from a solver, which may randomly oscillate during the solution procedure.

fem/hyperbolic.hpp
najlkin: Consider this case: Burgers flux `F(u)=1/2u^2` (in 1D) and `u⁻=-1` and `u⁺=+1`. Then the definition takes the case `u- ≤ u+` and `min F(u)` over the interval `[u⁻,u⁺]` is 0, while the current implementation based on the *monotonicity* assumption gives the result `min(1/2*(-1)^2,1/2*(+1)^2)=1/2`. So it is not fully valid for Burgers when `u` changes the sign, but fine for the cases of positive or negative `u` (which is also how I tested that).

fem/hyperbolic.hpp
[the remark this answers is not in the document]
cjvogl: > The class must work for arbitrary ones

What about the implementation of the Rusanov flux in this class does not work for arbitrary values?

Your response also did not address my the question I posed, so I will repeat it: What is the underlying definition of the Rusanov flux? In other words, your comment that "this construction assumes ..." implies there's a construction that does not make the assumption... what does that construction look like?

fem/hyperbolic.hpp
cjvogl: Ah, ok: the comments are misleading and implied (to me at least) that the implementation is actually using the max and min of the flux, as to properly implement the Godunov flux for scalar conservation laws. With this knowledge, I do not find value in a class that is designed only for scalar conservation laws with monotonic fluxes, and therefore, I do not think it should be in `hyperbolic.hpp` and `hyperbolic.cpp`.  Instead perhaps the class can be included in the specific code where it is useful.

cjvogl: Upon further inspection, the numerical flux as implement here, i.e.,
- for $u^- \leq u^+$
  - $\hat{F} = F(u^+)$ if $F'(u) < 0$
  - $\hat{F} = F(u^-)$ if $F'(u) > 0$
- for $u^- > u^+$
  - $\hat{F} = F(u^+)$ if $F'(u) < 0$
  - $\hat{F} = F(u^-)$ if $F'(u) > 0$

can be more compactly described as
- $\hat{F} = F(u^+)$ if $F'(u) < 0$
- $\hat{F} = F(u^-)$ if $F'(u) > 0$

which could be described as a component-wise upwinding. With an appropriate renaming, e.g., `ComponentwiseUpwind`, and updated comments to reflect the compact definition above, I can see the inclusion of the corresponding class into `hyperbolic.hpp`.  The key thing is to not imply to the user that a Godunov flux has been implemented in a general sense, as this implementation is only the Godunov flux for a limited set of scalar conservation laws (specifically, ones whose Riemann solutions are simple waves that do not include rarefaction fans).

fem/hyperbolic.hpp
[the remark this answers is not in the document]
cjvogl: If you find my suggest not very instructive whereas I find the original comment not very instructive, then I suggest they be combined to something along the lines of
```c++
 * @note The implementation assumes monotonous
     |dF(u,x)/du⋅n| in u so that the maximum 
     characteristic speed for u between u⁺ and u⁻ is
     given by max(|dF(u⁺,x)/du⋅n|, |dF(u⁻,x)/du⋅n|)
```

fem/hyperbolic.hpp
jandrej: I tend to agree with the naming. Gudonov describes a scheme and not a flux. The Gudonov scheme requires a choice of flux. `ComponentwiseUpwindFlux` seems appropriate for the method implemented.

najlkin: I am fine with renaming 👍 . Some people call it a scheme, some approximate Riemann solver, but in any case we cannot have a general Godunov or at least with the current framework. As [I tried to outline before](https://github.com/mfem/mfem/pull/4513#discussion_r1884283662), it would have to be tailored to the flux unless very strong assumptions are applied like here, which is also how I read @jandrej 's comment.

cjvogl: > in any case we cannot have a general Godunov or at least with the current framework

For a general scalar conservation laws with convex flux function, I've seen 
```
 *    F̂ n = min F(u)n    for u⁻ ≤ u⁺, u ∈ [u⁻,u⁺]
 *    F̂ n = max F(u)n    for u⁻ > u⁺, u ∈ **[u⁺,u⁻]
```

implemented using convex optimization routines.  

> it would have to be tailored to the flux 

I would agree that it would be more efficient to have an implementation for each flux function, which is why I was initially confused why this PR was introducing a `GodunovFlux` class.

> I am fine with renaming 👍 . 

That is good to hear.

najlkin: Renamed in 837333b88a7a003cd6b4e109b0f2ef8d7d673bbc .

fem/hyperbolic.hpp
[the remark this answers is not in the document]
najlkin: Rewritten in 72de1e5d50baaf5f311b8af91cef1d01300fb8df .

cjvogl (approved):

najlkin:
@jandrej , @mlstowell , @pazner , do you want to have a look? Otherwise let us move on to testing...

jandrej:
> @jandrej , @mlstowell , @pazner , do you want to have a look? Otherwise let us move on to testing...

👍

fem/hyperbolic.hpp
```diff
@@ -61,58 +61,126 @@ class FluxFunction
...
+    * @return real_t maximum (normal) characteristic speed |dF(u,x)/du⋅n|
     */
    virtual real_t ComputeFluxDotN(const Vector &state, const Vector &normal,
                                   FaceElementTransformations &Tr,
                                   Vector &fluxDotN) const;
 
    /**
-    * @brief Compute flux Jacobian. Optionally overloaded in the derived class
-    * when Jacobian is necessary (e.g. Newton iteration, flux limiter)
+    * @brief Compute average flux for over the given interval of states.
```
mlstowell, line 101: Typo: "for" seems superfluous

mlstowell (approved):
I'm particularly impressed with the detailed documentation. Truly excellent work, @najlkin!

pazner (approved):

fem/hyperbolic.hpp
najlkin: Thanks, a good catch 😉 .

najlkin:
@tzanio , it seems we are ready here! 🎉 Thanks everyone 👏 (https://github.com/mfem/mfem/pull/4513/commits/837333b88a7a003cd6b4e109b0f2ef8d7d673bbc passed and https://github.com/mfem/mfem/pull/4513/commits/6c2f6a9e7fd05585b4d64b691b871adb694077b9 is just a typo, the sanitizer fails due to #4650 )

tzanio:

tzanio (approved):
Thanks for the nice improvements @najlkin ! 👍 

@tarikdzanic, @dohyun-cse, @pazner, @cjvogl and @mlstowell -- thank you for the careful review! 🙏

PR #4518, opened 2024-09-24
Complex mumps

adam-sim-dev:
PR by @psocratis


psocratis:
@adam-sim-dev , I am assuming there is need for this? If so I can revive the branch and work on in a bit to get it ready

adam-sim-dev:
> @adam-sim-dev , I am assuming there is need for this? If so I can revive the branch and work on in a bit to get it ready

I work with complex FE problems frequently, and complex linags could simplify the codes. That is why I I searched existing mfem branches and opened this and the other PRs.

This branch is outdated compared with the current implementation of MUMPS solver. Maybe we can just make small modification to the MUMPSSover. So it needs further improvements, including the doc, make and cmake stuffs.

Complex FE support is quite limit now, though we can use equivalent real ones. For example, there is ComplexGridFunction, but we have call GetSubVector two times for the real and imaginary parts, respectively. Anyway, there is no harm to add more complex FE support. Just my two cents.

PR #4551, opened 2024-10-28
dFEM

jandrej:
This is the main development branch for all dFEM related things. Right now this PR only serves as a reference for easier discovery.

CI will fail because it hasn't been updated to C++17 and Enzyme integration


najlkin:
As I mentioned at the meeting, I think there should be some way how to specify what should be the interpretation of the items in `FieldDescriptor`. A finite element space can be used to obtain the physical space interpolation, which is the current behavior, but the same way one may want the reference space interpolation instead, for example. Ideally, there could be some customization available 😮 . Conceptually, these manually differentiated things could be outside of `DifferentiableOperator` maybe 🤔 .

jandrej:
Closing because this was merged through separate smaller PR's.

PR #4557, opened 2024-10-29
Integrator base class unifying logic for selecting IntegrationRule

david-kamensky:
The original motivation for this was to have integrators other than `DiffusionIntegrator` make use of `NonlinearFormIntegrator::patchRules` when choosing their `IntegrationRule`.  Directly copy-and-pasting the NURBS-specific logic from `DiffusionIntegrator::AssembleElementMatrix` to all other linear-form, bilinear-form, and nonlinear-form integrators' assembly methods seemed excessive, and my proposed solution in this pull request is to collect some common functionality in a new `Integrator` base class that all of these integrators inherit from.  

After some discussion over Slack, @dylan-copeland suggested I open a pull request tagged as WIP, where the proposed pattern has only been applied to some integrators, to illustrate the idea and collect feedback.

| [#4557](https://github.com/mfem/mfem/pull/4557) | @david-kamensky | @tzanio | @pazner + @dylan-copeland + @najlkin | 12/14/24 | 2/3/25 | 2/3/25 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


dylan-copeland:
Thanks @david-kamensky, this looks like a good improvement that will reduce a lot of duplicated code. One thing still gets repeated everywhere, which is the usage of `bool deleteRule`. Do you think this could further simplified, and made safer, by  using smart pointers `std::unique_ptr`, `std::shared_ptr`? Or would that be a large change that propagates through many lines of code and take too much work?
It's not clear to me that the base class `Integrator` is necessary, when `BilinearFormIntegrator` already inherits from `NonlinearFormIntegrator`. Is there a good reason to have this new base class, or can we keep the existing classes as they were? (The more changes, the harder it can be to get approval.)

david-kamensky:
@dylan-copeland Thanks for the quick feedback!

> One thing still gets repeated everywhere, which is the usage of bool deleteRule. Do you think this could be further simplified, and made safer, by using smart pointers `std::unique_ptr`, `std::shared_ptr`?

I was wondering this myself, although it doesn't seem like smart pointers are currently used extensively in MFEM.  (The contribution guidelines [here](https://github.com/mfem/mfem/blob/master/CONTRIBUTING.md) say contributions should favor C++03, although this sounds like it's out-of-date based on Tzanio's introductory remarks at the workshop last week.)  However, I think `shared_ptr` would be the typical "modern C++" solution to needing a pointer that may or may not have to free itself, depending on whether it's owned elsewhere.  I can take a look at what would be needed if C++11 is considered acceptable; it would need to propagate to wherever pointers to non-temporary quadrature rules are typically owned, but it probably wouldn't be so bad.

> It's not clear to me that the base class Integrator is necessary, when BilinearFormIntegrator already inherits from NonlinearFormIntegrator. Is there a good reason to have this new base class, or can we keep the existing classes as they were? (The more changes, the harder it can be to get approval.)

The reason I created the new base class was because we also want to be able to use the custom rules in `LinearFormIntegrator`, which does not inherit from `NonlinearFormIntegrator`.

dylan-copeland:
@david-kamensky Please feel free to add smart pointers, if it's not too much work. MFEM has been adopting smart pointers, e.g. `mesh/spacing.hpp`. No one should complain about that in 2024. The new base class makes sense to me now.

fem/integrator.hpp
```diff
@@ -0,0 +1,94 @@
...
+#define MFEM_INTEGRATOR
+
+#include "../config/config.hpp"
+#include "fe.hpp"
+
+namespace mfem
+{
+
+/** @brief This wraps a pointer to an integration rule, while allowing it to be designated as temporary, to be deleted after passing out of scope. */
+class IntegrationRulePtr
```
najlkin, line 11: Is this class really needed? I think we do not want ultimately a wrapper for a pointer of every type. That pointers from the standard library would be a better option and there can be one directly in `Integrator`, so something like `unique_ptr<const IntegrationRule> DefIntRule;` 😉 .

david-kamensky: It's not quite the right use-case for `unique_ptr`.  The returned pointer is usually to an integration rule owned by `IntRules`, which we do not want to go out-of-scope, but there's an exception where it may be a pointer to a temporary integration rule generated on-the-fly, which needs to be deleted by the calling scope.  So, it's really an occasionally-shared pointer, but going through and converting `IntRules` to store `shared_ptr`s to integration rules quickly grows in scope and also has some performance implications from the reference counting overhead (that is not necessary most of the time).  It is probably a design issue that this exception case requiring deletion by the caller exists, but I'm not sure what the best solution is.  The wrapper class here is slightly nicer than the previous pattern of taking a `bool&` that is updated to indicate whether the calling scope needs to `delete` the returned integration rule.  Would maybe a `std::variant<mfem::IntegrationRule*, std::unique_ptr<mfem::IntegrationRule>>` be a better way to get the desired conditional deletion without defining a new class?

david-kamensky: (Googling quickly, it looks like `std::variant` is only available since C++17, so I don't think it's permitted in MFEM yet.)

dylan-copeland: It sounds like having a global variable `IntRules` is preventing usage of `std` smart pointers. Maybe we should change this in a future PR, since it would be a major change beyond the scope of this PR.

najlkin: Well, this starts to be very similar discussion to what we had with @pazner in #4091 . I think the best way forward would be to generalize by template `OperatorHandle` (to just `Handle` perhaps? 🙂 ). That would provide exactly the needed capacity for optionally owned pointers. How far we will adopt it in MFEM and/or extend it for reusing identical objects we may leave for later. Do you want to do it or should I? 🚀

pazner: Yes, with regards to cleaning up ownership rules I think we should definitely start doing some modernization; my hope would be to this in a consistent way across the (very large) MFEM codebase, so it might take some discussion to land on an acceptable design.

I thought about this briefly, and also considered creating a `Handle<T>` class in MFEM that would _either_ contain a raw pointer (not owned) _or_ a `shared_ptr` (owned, of course).

Interfaces could be transitioned to taking `Handle<T>` instead of `T*` in a backwards-compatible way by having a non-owning constructor `Handle(T*)`.

This idea is sketched out here: https://godbolt.org/z/nnYxqe1ar

najlkin: That looks cool 😎 , but yes, that will require a bigger discussion, so for this PR, how about just a basic template replacing `unique_ptr` in the way the current implementation of `OperatorHandle` does? This could be extended to that `shared_ptr` later...

david-kamensky: It sounds like you're proposing a template generalization of `IntegrationRulePtr` that would work for other optionally-owned pointer types, and already has another possible use-case for operators.  I like that idea, although I'm less familiar with the other use case you're envisioning and not sure whether there are general conventions for where to put or how to implement these sorts of general "utility" types in MFEM, so maybe you can go ahead and implement the idea you already have if it's relatively easy.

pazner: My vote on this (and #4091) is to try to reach a consensus about a systematic way to handle these and other ownership issues consistently in the code, rather than introduce one-off solutions that will have to be refactored later (maintaining backwards compatibility, etc.).

najlkin: Ok folks, I created #4569 with a basic templated generalization of `OperatorHandle`. It can be used here instead of the one-off class `IntegrationRulePtr`. I think it is a good solution for now and we may extend later following to what @pazner posted 😉 .

david-kamensky: Thanks for drafting an implementation of the new generalization.  One potential issue I see in the context of the present PR is that the proposed `Handle` interface doesn't look like it will transfer ownership when returning a `Handle` from a function, which is the behavior needed from `IntegrationRulePtr` here.  The concern is that if a `Handle` is created in a function (e.g., the `GetIntegrationRule` method here), then returned from that function, the returned value would be a shallow copy, and the original owning object would delete the pointer when destructed at the end of the function's scope.

najlkin: It is scoped in the `Integrator` anyway, so the idea (similar to [what I wrote above](https://github.com/mfem/mfem/pull/4557#discussion_r1830025862)) was to just make something like `Handle<const IntegrationRule> DefIntRule;` attribute of `Integrator` for now 😉 .

pazner: @david-kamensky, is the approach I described at the beginning of https://github.com/mfem/mfem/pull/4569#issuecomment-2460971338 possible?

The idea would be that `NURBSMeshRules` would have an `IntegrationRules` data member. `NURBSMeshRules::GetElementRule` would return a reference to this data member; it would always be a non-owning reference (there would be no need for `deleteRule` — `deleteRule` would always be false). `NURBSMeshRules::GetElementRule` would resize/overwrite this internal integration rule if needed.

This way, all the `GetIntegrationRule` functions would return non-owning references/pointers, and there is no need for a "maybe-owning" pointer type.

Does that sound reasonable?

david-kamensky: Oh, okay, I missed that part of your suggestion.  My understanding of what you mean is that we'd just keep updating the value of this attribute with the `Handle::Reset` method when building temporary rules not owned by `IntRules`, which would resolve the issue I mentioned above.  In the case where a temporary rule is generated on each element, this would defer freeing the last temporary rule generated until the `Integrator` is destroyed (instead of until the assembly method completes), but that is probably not enough memory to be a significant concern.

najlkin: You are right, but it is the case only for the legacy OpenMP backend, which, as the name says, is not much used anymore and most of the integrators are not thread safe anyway (but I have not checked the specific integrators you modified). Though, as you are developing a general system, it would be better to support that, so there is preprocessor declaration `MFEM_THREAD_SAFE`, which is used to guard the shared attributes of the integrators (see e.g. `hyperbolic.cpp/hpp`). Maybe a solution is to pass the owning handle when `MFEM_THREAD_SAFE` is defined and non-owning one (pointing to the shared attribute) when it is not to keep some caching 😉 .

pazner: Hi @david-kamensky, does https://github.com/mfem/mfem/pull/4557#discussion_r1831825377 sound possible to you? (If it's not clear what was meant there, I can provide a patch with the changes needed for the proposed approach). This unfortunately has the same thread-safety issue as what Jan suggested. The advantage is that no specialized types (neither `IntegrationRulePtr` nor `Handle`) would be needed.

david-kamensky: @pazner Yes, that sounds possible.  However, if we're okay with
* Lack of thread safety
* Deferring deletion of the last-used temporary element rule until after the scope of the element-level assembly method (until either the `Integrator` or `NURBSMeshRules` is deleted)

then I think we could also just go with @najlkin's original proposal of just using a `unique_ptr<IntegrationRule>` member of `Integrator`, which would avoid modifying the public interface of `NURBSMeshRules`.  (We actually couldn't combine the temporary-rule pointer with the existing non-owned `IntegrationRule* ir` member of `Integrator` as a single `Handle` even if we wanted to; whether to override that with a temporary rule depends on the element argument to the assembly method, and is not an inherent attribute of the `Integrator`.)

`NURBSMeshRules` is probably a more logical owner for the temporary rule, but I don't know how strict we want to be about preserving the existing interface with the `deleteRule` argument (to avoid "breaking changes"), even if it's not ideal.

pazner: If the concern is changing the interface of `NURBSMeshRules`, the interface can remain the same, but `deleteRules` is just always set to false.

dylan-copeland: If the only breaking change is to `NURBSMeshRules::GetElementRule`, I would think that is ok, since it is a relatively new function not used by anyone other than you @david-kamensky, as far as I know.

najlkin: But that thread safety is a trouble if we want to apply this system of integration rules everywhere... 🤔

david-kamensky: I like the idea of making a small breaking change to fix this issue of integration rule ownership at its source, by eliminating the `deleteRule` behavior from `NURBSMeshRules::GetElementRule`.  

We could have a (very memory-intensive) thread-safe fallback option of just pre-computing all the element rules and adding them to `NURBSMeshRules::elementToRule` in the `NURBSMeshRules` constructor, or simply put an `MFEM_ABORT` inside `#ifdef MFEM_THREAD_SAFE`/`#endif` in the code-path for using a temporary rule, which would only be reached if using `NURBSMeshRules` and multi-threading together.

david-kamensky: I pushed a new commit just now following the second approach of just `MFEM_ABORT`-ing if trying to create a temporary rule on a NURBS element with `MFEM_THREAD_SAFE` defined.  Otherwise, `NURBSMeshRules` has a `private` member `std::unique_ptr<const IntegrationRule> temporaryElementRule` that is re-assigned each time a new temporary rule is formed on an element (automatically `delete`-ing the previous rule).  

This completely removes the `IntegrationRulePtr` class and does not require any analogous replacement (e.g., `Handle`).  I made the small breaking change of removing the now-obsolete `bool& deleteRule` argument from `NURBSMeshRules::GetElementRule`.  (But, as @pazner pointed out, we could always leave it in as a do-nothing parameter if we want to strictly enforce backward compatibility of interfaces.)

najlkin: I would say I more or less like this solution. It kills some pathways, but legacy OpenMP in combination with NURBS with non-reduced rules are a special case of a special case, so not very likely to appear, I guess 😄 . Also it was not much supported before anyway, if I am not wrong.
We could think of a thread-safe approach with objects (temporary rules) assigned to the thread ids, [as I proposed elsewhere](https://github.com/mfem/mfem/pull/4065#issuecomment-2377434398), but that would be an overkill to do just specifically here. However, we may think about it in the context something more general like `Handle`... 🤔

fem/integrator.hpp
```diff
@@ -0,0 +1,59 @@
...
+namespace mfem
+{
+/** @brief This base class implements some shared functionality between linear and nonlinear form integrators. */
+class Integrator
+{
+protected:
+   const IntegrationRule *IntRule;
+   NURBSMeshRules *patchRules = nullptr;
+
+   /** This will pick the NURBS rule's restriction to test_fe if given and applicable, and IntRule otherwise, prioritizing the NURBS rule if available.
```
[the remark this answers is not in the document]
david-kamensky: I modified the description to refer only to "the element" as suggested, and also started it with a more general first sentence that doesn't single-out the NURBS case.

fem/integrator.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+public:
+   /** Create a new Integrator, optionally providing a prescribed quadrature rule to use in assembly. */
+   Integrator(const IntegrationRule *ir = NULL) : IntRule(ir) {};
+
+   /** Prescribe a fixed IntegrationRule to use, or set to null to let the integrator choose an appropriate rule. */
+   virtual void SetIntRule(const IntegrationRule *ir) { IntRule = ir; }
+
+   /** Prescribe a fixed IntegrationRule to use. */
+   void SetIntegrationRule(const IntegrationRule &ir) { SetIntRule(&ir); }
```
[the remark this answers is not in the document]
david-kamensky: I moved the public interface up to the top, as suggested.

fem/integrator.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include "../config/config.hpp"
+#include "fe.hpp"
+
+namespace mfem
+{
+/** @brief This base class implements some shared functionality between linear and nonlinear form integrators. */
+class Integrator
+{
+protected:
```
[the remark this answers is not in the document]
david-kamensky: I moved the public interface up to the top, as suggested.

tzanio:

fem/integrator.hpp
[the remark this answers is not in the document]
dylan-copeland: The only places I see it being used are `fem/lor/lor.cpp`, `miniapps/dpg/pacoustics.cpp`, and some tests. It seems safe to remove as far as MFEM is concerned, but some applications may be using it. Even so, changing to `SetIntRule` would be easy. I would be in favor of removing it or deprecating it, to be safe. What do others think?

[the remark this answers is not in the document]
dylan-copeland: I meant it would be safer to deprecate, rather than remove. I also don't see a problem with having both versions, but the reference version is almost completely unused and seems unnecessary. It may be simpler to have just one version, since 2 functions with slightly different names that do the same thing could be confusing.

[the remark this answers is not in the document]
v-dobrev: For some context: this was added in #193. Originally, `SetIntRule` was in class `BilinearFormIntegrator`, then @vladotomov added `SetIntegrationRule` in `NonlinearFormIntegrator` (from which `BilinearFormIntegrator` is derived). Then `SetIntRule` was also moved up to `NonlinearFormIntegrator` and both versions were kept.

fem/bilininteg.hpp
```diff
@@ -2135,7 +2135,14 @@ class GradientIntegrator : public BilinearFormIntegrator
 
    static const IntegrationRule &GetRule(const FiniteElement &trial_fe,
                                          const FiniteElement &test_fe,
-                                         ElementTransformation &Trans);
+                                         const ElementTransformation &Trans);
+protected:
+   const IntegrationRule* GetDefaultIntegrationRule(const FiniteElement* trial_fe,
+                                                    const FiniteElement* test_fe,
+                                                    const ElementTransformation* trans) const override
```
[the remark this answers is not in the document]
david-kamensky: I've shortened this in the latest commit, and ran `make style` again for good measure.

fem/nonlininteg.hpp
```diff
@@ -353,6 +338,10 @@ class HyperelasticNLFIntegrator : public NonlinearFormIntegrator
    void AssembleElementGrad(const FiniteElement &el,
                             ElementTransformation &Ttr,
                             const Vector &elfun, DenseMatrix &elmat) override;
+protected:
+   const IntegrationRule* GetDefaultIntegrationRule(const FiniteElement* trial_fe,
+                                                    const FiniteElement* test_fe,
+                                                    const ElementTransformation* trans) const override;
```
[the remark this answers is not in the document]
david-kamensky: I've shortened this in the latest commit, and ran `make style` again for good measure.

fem/integrator.cpp
```diff
@@ -0,0 +1,46 @@
+#include "integrator.hpp"
+#include "fem.hpp"
+#include "intrules.hpp"
+
+namespace mfem
+{
+const IntegrationRule* Integrator::GetIntegrationRule(const FiniteElement*
+                                                      trial_fe, const FiniteElement* test_fe,
+                                                      const ElementTransformation* trans) const
```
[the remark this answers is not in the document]
david-kamensky: I've shortened this in the latest commit, and ran `make style` again for good measure.

fem/bilininteg.hpp
```diff
@@ -2315,6 +2323,13 @@ class DiffusionIntegrator: public BilinearFormIntegrator
       ApplyPAKernels::Specialization<DIM,D1D,Q1D>::Add();
       DiagonalPAKernels::Specialization<DIM,D1D,Q1D>::Add();
    }
+protected:
+   const IntegrationRule* GetDefaultIntegrationRule(const FiniteElement* trial_fe,
```
dylan-copeland, line 2327: Long lines.

fem/integrator.cpp
```diff
@@ -0,0 +1,48 @@
...
+   {
+      result = IntRule;
+   }
+   else
+   {
+      result = GetDefaultIntegrationRule(trial_fe, test_fe, trans);
+   }
+   return result;
+}
+const IntegrationRule* Integrator::GetIntegrationRule(
```
dylan-copeland, line 32: Add an empty line between functions.

dylan-copeland (approved):

fem/intrules.cpp
```diff
@@ -1912,10 +1914,9 @@ IntegrationRule& NURBSMeshRules::GetElementRule(const int elem,
       np *= npd[d];
    }
 
-   IntegrationRule *irp = new IntegrationRule(np);
-   deleteRule = true;
+   temporaryElementRule = IntegrationRule(np);
```
najlkin, line 1917: I wonder what this will do as `IntegrationRule` does not have a copy constructor or assignment operator 🤪 . It would be better to call just `SetSize(np)`.

pazner: The assignment operator should be inherited from `Array<T>`. But yes, `SetSize(np)` looks better 👍

fem/nonlininteg.cpp
```diff
@@ -757,7 +758,8 @@ void VectorConvectionNLFIntegrator::AssembleElementVector(
    ELV.UseExternalData(elvect.GetData(), nd, dim);
 
    Vector vec1(dim), vec2(dim);
-   const IntegrationRule *ir = IntRule ? IntRule : &GetRule(el, T);
+   const IntegrationRule* ir = GetIntegrationRule(&el, &T);
+   if (!ir) {ir = &GetRule(el, T);}
```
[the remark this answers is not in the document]
david-kamensky: The eventual desired pattern would be to only call `Integrator::GetIntegrationRule`, and absorb the behavior of `VectorConvectionNLFIntegrator::GetRule` into a subclass-specific `override` of the virtual method `Integrator::GetDefaultIntegrationRule`, which is called within `Integrator::GetIntegrationRule`.

The `if(!ir) {...}` case handled in the code highlighted above is a workaround because `Integrator::GetIntegrationRule` is not yet overridden by `VectorConvectionNLFIntegrator`.  (As we discussed some time ago, the plan was to wait until after an initial round of review to update all subclasses of `Integrator`.)  This is not really "extra" logic in the sense that it's really just the `else` branch of the original ternary statement in the deleted line, but the original logic is now spread over two lines to avoid evaluating `GetIntegrationRule` both before and after the `?` (since `GetIntegrationRule` generalizes to also check for NURBS rules).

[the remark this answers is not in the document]
najlkin: What I was saying is override `VectorConvectionNLFIntegrator::GetDefaultIntegrationRule()` and that of other integrators, where a separate method for the default integration rule rule already exists, so it is just a one-liner in the header instead of this extra line in the source code, so does not qualify, I believe, as the labor mentioned.

fem/integrator.hpp
```diff
@@ -0,0 +1,79 @@
...
+   /** For patchwise integration, SetNURBSPatelementchIntRule must be called.
+       This will override IntRule if both are non-null. */
+   void SetNURBSPatchIntRule(NURBSMeshRules *pr) { patchRules = pr; }
+
+   /** Check if a NURBS patch integration rule has been set. */
+   bool HasNURBSPatchIntRule() const { return patchRules != nullptr; }
+
+   /** Directly return the IntRule pointer (possibly null) without checking
+       for NURBS patch rules or falling back on a default. */
+   const IntegrationRule *GetIntegrationRule() const { return IntRule; }
```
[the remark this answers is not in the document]
david-kamensky: Searching the code with recursive `grep`, it looks like there are some class methods called `GetIntRule`, but they are not members of classes that inherit from `Integrator`, so we can't simplify by moving them up to `Integrator`.  The only zero-argument version I saw is a member of `GenericIntegrationRule`.

fem/lininteg.hpp
```diff
@@ -51,9 +51,6 @@ class LinearFormIntegrator
                                        FaceElementTransformations &Tr,
                                        Vector &elvect);
 
-   virtual void SetIntRule(const IntegrationRule *ir) { IntRule = ir; }
-   const IntegrationRule* GetIntRule() { return IntRule; }
```
najlkin, line 55: @david-kamensky , here ☝️

david-kamensky: Oh, I didn't realize I'd deleted it; that explains why I couldn't find it via `grep` in the branch!

david-kamensky: (Resolving in favor of original thread.)

fem/intrules.cpp
najlkin: Right, I thought it is not inherited, but it is and compounded with the default assignment operator, which can be a dangerous concept in some scenarios in general 🤔 . Anyway, `SetSize()` please 😉

fem/bilininteg.hpp
david-kamensky: Fixed in latest commits.

fem/integrator.cpp
david-kamensky: Fixed in latest commits.

fem/lininteg.cpp
```diff
@@ -86,7 +87,7 @@ void DomainLFGradIntegrator::AssembleRHSElementVect(
    elvect.SetSize(dof);
    elvect = 0.0;
 
-   const IntegrationRule *ir = IntRule;
+   const IntegrationRule* ir = GetIntegrationRule(&el, &Tr);
```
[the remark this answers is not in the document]
david-kamensky: I've updated this and similar instances to put the asterisk on the variable name.

fem/intrules.hpp
```diff
@@ -20,6 +20,7 @@
 
 #include <vector>
 #include <map>
+#include <memory>
```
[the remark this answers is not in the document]
david-kamensky: I don't know; the code compiles without it.  It was probably left over from an earlier implementation using `unique_ptr` (which I overwrote in the commit history with a force push before requesting review).

fem/integrator.hpp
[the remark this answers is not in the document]
david-kamensky: I've added the deleted linear form implementation of `GetIntRule` to `Integrator` for backward compatibility in the latest commits.  (This doesn't appear to be called anywhere else in the MFEM codebase, but maybe it is used in third-party applications.)

fem/integrator.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+protected:
+   const IntegrationRule *IntRule;
+   NURBSMeshRules *patchRules = nullptr;
+
+   /** This method selects an integration rule in a way that depends on the
```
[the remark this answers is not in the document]
david-kamensky: I've separated the `@brief` description from `@details` in this and another method in `integrator.hpp`.

fem/integrator.hpp
```diff
@@ -0,0 +1,79 @@
...
+   /** This method is intended to be overriden by subclasses to choose an
+       appropriate integration rule based on the finite element spaces and/or
+       element transformation. The trial_fe and test_fe should be equal for
+       linear forms. The default base-class implementation returns null, which
+       assumes that an appropriate rule is provided by another means, or that
+       null integration rules are handled appropriately by the caller. */
+   virtual const IntegrationRule* GetDefaultIntegrationRule(
+      const FiniteElement* trial_fe, const FiniteElement* test_fe,
+      const ElementTransformation* trans) const
+   {return NULL;}
```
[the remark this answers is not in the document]
david-kamensky: I've added the spaces in the latest commits.

fem/integrator.cpp
```diff
@@ -0,0 +1,48 @@
+#include "integrator.hpp"
```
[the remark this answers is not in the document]
david-kamensky: I've added the header to this file and the corresponding `.cpp`.  I left the copyright date as 2024 for consistency with other files, since those will presumably get updated all at once in some automated commit.

fem/intrules.cpp
david-kamensky: I've switched from assignment to `.SetSize(np)` in the latest commits.

fem/integrator.cpp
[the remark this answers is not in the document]
najlkin: Commit of automated @tzanio 😄 Thanks, we will check before merging.

fem/integrator.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+const IntegrationRule* Integrator::GetIntegrationRule(
+   const FiniteElement* el,
+   const ElementTransformation* trans) const
+{
+   return GetIntegrationRule(el, el, trans);
+}
+
+const IntegrationRule* Integrator::GetIntegrationRule(const FiniteElement* el)
+const
```
najlkin, line 58: This is ugly, break the line before please.

fem/integrator.hpp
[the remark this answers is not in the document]
najlkin: Just a cosmetic change, but to make it more clear and more centralized it would be better to call one method from the other like you do it with `SetInt(egration)Rule()`

fem/integrator.hpp
```diff
@@ -0,0 +1,104 @@
...
+                prioritizing the NURBS rule if available. If neither is
+                valid, the integrator will fall back on the virtual method
+                GetDefaultIntegrationRule to choose a default integration
+                rule, where subclasses can override this in a problem-specific
+                way.
+   */
+   const IntegrationRule* GetIntegrationRule(
+      const FiniteElement* trial_fe, const FiniteElement* test_fe,
+      const ElementTransformation* trans) const;
+   /** Overload that is equivalent to passing NULL for the
```
najlkin, line 72: Empty lines between methods here as well please.

fem/integrator.hpp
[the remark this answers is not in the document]
najlkin: Btw, that `@details` is not needed when you leave an empty line before, but does not hurt 😉

fem/bilininteg.cpp
```diff
@@ -599,7 +600,8 @@ void MixedVectorIntegrator::AssembleElementMatrix2(
 
    elmat.SetSize(test_nd, trial_nd);
 
-   const IntegrationRule *ir = IntRule;
+
```
najlkin, line 603: Some extra blank lines.

fem/nonlininteg.cpp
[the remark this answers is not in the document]
najlkin: Thanks, fine for now, I think 😉 .

fem/integrator.hpp
```diff
@@ -0,0 +1,104 @@
...
+      const FiniteElement* trial_fe,
+      const FiniteElement* test_fe) const;
+   /** Overload for cases where trial_fe and test_fe are the same. */
+   const IntegrationRule* GetIntegrationRule(
+      const FiniteElement* el,
+      const ElementTransformation* trans) const;
+   /** Overload that is equivalent to passing NULL for the
+       ElementTransformation pointer, if it is unavailable or unused in the
+       default behavior. */
+   const IntegrationRule* GetIntegrationRule(const FiniteElement* el) const;
```
najlkin, line 85: If it is equivalent to passing NULL, you may just use NULL default parameter value directly. This makes impression this method is somehow special, while it is not. It will also reduce the number of overloads.

fem/integrator.hpp
```diff
@@ -0,0 +1,104 @@
...
+   const IntegrationRule* GetIntegrationRule(
+      const FiniteElement* trial_fe, const FiniteElement* test_fe,
+      const ElementTransformation* trans) const;
+   /** Overload that is equivalent to passing NULL for the
+       ElementTransformation pointer, if it is unavailable or unused in the
+       default behavior. */
+   const IntegrationRule* GetIntegrationRule(
+      const FiniteElement* trial_fe,
+      const FiniteElement* test_fe) const;
+   /** Overload for cases where trial_fe and test_fe are the same. */
```
najlkin, line 78:
```suggestion
   /** @brief Selects an integration rule based on the the arguments and
              internal state. (Version for identical trial_fe and test_fe)
       @see  GetIntegrationRule(const FiniteElement*, const FiniteElement*,
             const ElementTransformation*)
   */
```

fem/integrator.cpp
david-kamensky: I've updated the formatting here to avoid the lone `const` in a line.

fem/integrator.hpp
david-kamensky: I've added empty lines here.

fem/integrator.hpp
david-kamensky: I've condensed this with default parameters for the cases where the element transform is null.

fem/integrator.hpp
david-kamensky: I made this documentation change, but just lumped it together with the other formatting changes, to keep the commit history concise and avoid possibly triggering an unnecessary pipeline.

fem/integrator.cpp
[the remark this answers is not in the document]
tzanio: Oh right, it is that time of the year again... Thanks for the reminder, I will add it to my list.

fem/integrator.hpp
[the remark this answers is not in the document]
david-kamensky: I've updated `GetIntRule` to call `GetIntegrationRule`.

fem/bilininteg.cpp
david-kamensky: I've cleaned out excess blank lines around `const IntegrationRule *ir = ...` in `bilininteg.cpp` in the latest commit.

fem/integrator.hpp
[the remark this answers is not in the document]
david-kamensky: After thinking about it for a bit, I decided to favor `GetIntRule` as the direct accessor for `IntRule`, and make the zero-argument public `GetIntegrationRule` the one that's kept around for backward compatibility, just to emphasize the distinction between directly getting `IntRule` vs. executing some more elaborate selection logic, like the protected virtual `GetIntegrationRule` method does.  I updated the handful of uses of `Integrator::GetIntegrationRule()` within MFEM itself to `Integrator::GetIntRule()` accordingly, but kept `Integrator::GetIntegrationRule()` for backward compatibility in any user applications that might call it.

There's still some "inconsistency" between the getters and setters, since the "short" and "long" versions of `SetInt(egration)Rule` also differ in pointer vs. reference for the argument, but this discrepancy was already part of the public interfaces of the various classes being unified under `Integrator`.

I'm not sure what the conventions/policies are around using `MFEM_DEPRECATED`; I can add that to the "long" versions of each if you think that is warranted.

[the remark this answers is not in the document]
najlkin: Perfect 👍 . With deprecation, we will wait some time probably and requires a consensus, so leave it as it is for now.

fem/lor/lor.cpp
```diff
@@ -29,7 +29,7 @@ void LORBase::AddIntegrators(BilinearForm &a_from,
    {
       BilinearFormIntegrator *integrator = (*integrators)[i];
       (a_to.*add_integrator)(integrator);
-      ir_map[integrator] = integrator->GetIntegrationRule();
+      ir_map[integrator] = integrator->GetIntRule();
```
pazner, line 32: Sorry, I know this has been already discussed in this PR. Is our preference for `IntRule` or fully spelled-out `IntegrationRule`?

Because the newly introduced `Integrator::GetIntegrationRule` and `Integrator::GetDefaultIntegrationRule` use the fully spelled-out version (which would be my preference, just for consistency with the class name `IntegrationRule`), but this change is moving in the other direction.

fem/integrator.hpp
```diff
@@ -0,0 +1,100 @@
...
+                element spaces and/or element transformation. The trial_fe
+                and test_fe should be equal for linear forms. The default
+                base-class implementation returns null, which assumes that
+                an appropriate rule is provided by another means, or that null
+                integration rules are handled appropriately by the caller.
+   */
+   virtual const IntegrationRule* GetDefaultIntegrationRule(
+      const FiniteElement* trial_fe, const FiniteElement* test_fe,
+      const ElementTransformation* trans) const
+   { return NULL; }
```
pazner, line 96: If the finite elements and element transformations passed to these functions are not supposed to be null, then it would be better to pass by const reference rather than pointer.

fem/integrator.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace mfem
+{
+/** @brief This base class implements some shared functionality between
+    linear and nonlinear form integrators. */
+class Integrator
+{
+public:
+   /** Create a new Integrator, optionally providing a prescribed quadrature
+       rule to use in assembly. */
```
pazner, line 26: Here and below the short Doxygen comments should be tagged with `@brief` (otherwise they won't appear at the top of the docs page).

fem/lor/lor.cpp
david-kamensky: My rationale here is:
* We have a new method to get an `IntegrationRule`, which takes some arguments and follows a more complex logic for selecting an appropriate rule than simply returning the `IntRule` member.  The most straightforward name for this seemed like `GetIntegrationRule(...)`.
* The most predictable convention for naming simple accessors methods of class members is `Get<MemberName>()`, which `GetIntRule()` is consistent with.

Following this logic, we should not have a zero-argument `GetIntegrationRule()` method that is a simple accessor for `IntRule` (which is why I replaced its usage within MFEM with `GetIntRule()`).  However, I retained the zero-argument `GetIntegrationRule()` method solely for backward compatibility in user applications, since it already existed in the public interface for one of the classes that now inherits from `Integrator`.

fem/integrator.hpp
david-kamensky: Not all `Integrator` subclasses depend on an `ElementTransformation` for `IntegrationRule` selection, so it should be optional, and there is no clear candidate for an `ElementTransformation` in many cases.  In C++17 or later, it would be possible to use something like `std::optional<std::reference_wrapper<const ElementTransformation>> = std::nullopt` to avoid relying on `nullptr` as a sentinel value, although this is not available in C++11.

fem/integrator.hpp
david-kamensky: In the latest commit, I've added `@brief` to the shorter Doxygen comments in `integrator.hpp`.

fem/lor/lor.cpp
pazner: Makes sense, but I think the name of a protected member is somewhat of an implementation detail, whereas the name of the accessor is part of the public API. For example, `Mesh::GetFaceToElementTable` returns the member `Mesh::face_elem` — certainly `Mesh::GetFaceElem` is not a better name.

I don't think that if you present the user with two functions `GetIntRule` and `GetIntegrationRule` the difference in naming will communicate very much useful information (especially since the `IntRule` data member is undocumented).

Thinking about this more, the logic (and what we're trying to communicate with these names) is a bit complicated. The user can set the `IntRule`, which will force the integrator to use the specified rule instead of the default, _except_ in the case of NURBS patch rules, in which case the user's choice will be overridden.

fem/integrator.hpp
pazner: I'm not sure I understand — this is used on a per-element basis to get the default integration rule for the given element, isn't the `ElementTransformation` just that of the current element?

pazner: To explain my concern with this: do integrators that override this function have to handle the case of any of the passed pointers being null? It doesn't look like any of the classes that override `GetDefaultIntegrationRule` handle any of these being null. If I implement a new integrator and override this, when is it safe to assume they're non-null and when is it not?

fem/integrator.hpp
```diff
@@ -0,0 +1,100 @@
...
+   /** @brief Equivalent to GetIntRule, but retained for backward
+       compatibility with applications. */
+   const IntegrationRule *GetIntegrationRule() const { return GetIntRule(); }
+
+protected:
+   const IntegrationRule *IntRule;
+   NURBSMeshRules *patchRules = nullptr;
+
+   /** @brief Selects an integration rule based on the the arguments and
+              internal state of the Integrator object.
```
pazner, line 56: I think "returns" is a better description than "selects" — calling this doesn't change the internal state of the integrator, which is a potential interpretation of "selects".

(Same for the Doxygen comment below).

fem/lor/lor.cpp
david-kamensky: I agree that the logic for which `IntegrationRule`s get priority over others is a bit complicated.  This is something I preserved from the existing rule selection logic that is hard-coded into `DiffusionIntegrator::AssembleElementMatrix` on `master`, with the original aim of being able to use `BilinearFormIntegrator::patchRules` in element-by-element assembly in other linear/nonlinear/bilinear integrator types.  Trying to avoid repetition of the logic from `DiffusionIntegrator::AssembleElementMatrix` then naturally led to introducing the `Integrator` base class.

Maybe the solution here is to rename `GetIntegrationRule(...)` (with arguments and nontrivial logic) to something like `SelectIntegrationRule(...)` to emphasize that it is doing some "thinking" to choose the right rule, and not just retrieving a protected class member.  Then we can preserve the redundant accessors `GetIntRule()` and `GetIntegrationRule()` for backward compatibility, but favor `GetIntegrationRule()` for clarity.

tzanio:

fem/integrator.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "intrules.hpp"
+
+namespace mfem
+{
+const IntegrationRule* Integrator::GetIntegrationRule(
+   const FiniteElement& trial_fe, const FiniteElement& test_fe,
+   const ElementTransformation& trans) const
+{
+   const NURBSFiniteElement *NURBSFE = dynamic_cast<const NURBSFiniteElement *>
+                                       (&test_fe);
```
v-dobrev, line 23: Dynamic casts are not cheap, so it's better to move this one inside the `if` statement:
```c++
   const NURBSFiniteElement *NURBSFE;
   if (patchRules &&
       (NURBSFE = dynamic_cast<const NURBSFiniteElement *>(&test_fe)))
```
This way, the `dynamic_cast` is performed only when needed.

fem/integrator.hpp
```diff
@@ -0,0 +1,100 @@
...
+namespace mfem
+{
+/** @brief This base class implements some shared functionality between
+    linear and nonlinear form integrators. */
+class Integrator
+{
+public:
+   /** @brief Create a new Integrator, optionally providing a prescribed
+       quadrature rule to use in assembly. */
+   Integrator(const IntegrationRule *ir = NULL) : IntRule(ir) {};
```
v-dobrev, line 27: Remove the extra `;` at the end of this line.

fem/nonlininteg.cpp
```diff
@@ -417,7 +419,6 @@ real_t HyperelasticNLFIntegrator::GetElementEnergy(const FiniteElement &el,
 
       energy += ip.weight * Ttr.Weight() * model->EvalW(Jpt);
    }
-
```
v-dobrev, line 420: Revert please.

fem/integrator.hpp
david-kamensky: I've updated "selects" to "returns" in the documentation comments for this method.

fem/integrator.cpp
david-kamensky: I've made this optimization in the latest commit.

fem/nonlininteg.cpp
david-kamensky: I reverted the removal of the blank line that is highlighted by the GitHub review interface; let me know if you meant to refer to a larger set of changes that should be reverted.

v-dobrev: I meant just this specific change.

Typically, PRs should not make changes like this, that are unnecessary.

fem/integrator.hpp
```diff
@@ -0,0 +1,100 @@
...
+   Integrator(const IntegrationRule *ir = NULL) : IntRule(ir) {}
+
+   /** @brief Prescribe a fixed IntegrationRule to use, or set to null to let
+       the integrator choose an appropriate rule. */
+   virtual void SetIntRule(const IntegrationRule *ir) { IntRule = ir; }
+
+   /** @brief Prescribe a fixed IntegrationRule to use. */
+   void SetIntegrationRule(const IntegrationRule &ir) { SetIntRule(&ir); }
+
+   /** @brief For patchwise integration, SetNURBSPatelementchIntRule must be
```
justinlaughlin, line 36:
```suggestion
   /** @brief For patchwise integration, SetNURBSPatchIntRule must be
```

fem/integrator.hpp
```diff
@@ -0,0 +1,100 @@
...
+   /** @brief Prescribe a fixed IntegrationRule to use, or set to null to let
+       the integrator choose an appropriate rule. */
+   virtual void SetIntRule(const IntegrationRule *ir) { IntRule = ir; }
+
+   /** @brief Prescribe a fixed IntegrationRule to use. */
+   void SetIntegrationRule(const IntegrationRule &ir) { SetIntRule(&ir); }
+
+   /** @brief For patchwise integration, SetNURBSPatchIntRule must be
+       called. This will override IntRule if both are non-null. */
+   void SetNURBSPatchIntRule(NURBSMeshRules *pr) { patchRules = pr; }
```
najlkin, line 38:
```suggestion
   void SetNURBSPatchIntRule(NURBSMeshRules *pr)
   { patchRules = pr; if (pr) { IntRule = nullptr; } }
```
Here this one plus documentation.

fem/integrator.hpp
```diff
@@ -0,0 +1,100 @@
...
+public:
+   /** @brief Create a new Integrator, optionally providing a prescribed
+       quadrature rule to use in assembly. */
+   Integrator(const IntegrationRule *ir = NULL) : IntRule(ir) {}
+
+   /** @brief Prescribe a fixed IntegrationRule to use, or set to null to let
+       the integrator choose an appropriate rule. */
+   virtual void SetIntRule(const IntegrationRule *ir) { IntRule = ir; }
+
+   /** @brief Prescribe a fixed IntegrationRule to use. */
```
najlkin, line 33: Please update the documentation also here.

fem/integrator.hpp
```diff
@@ -0,0 +1,100 @@
...
+class Integrator
+{
+public:
+   /** @brief Create a new Integrator, optionally providing a prescribed
+       quadrature rule to use in assembly. */
+   Integrator(const IntegrationRule *ir = NULL) : IntRule(ir) {}
+
+   /** @brief Prescribe a fixed IntegrationRule to use, or set to null to let
+       the integrator choose an appropriate rule. */
+   virtual void SetIntRule(const IntegrationRule *ir) { IntRule = ir; }
```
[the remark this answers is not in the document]
david-kamensky: I updated this in the latest commit (using the return value of assignment to shorten the suggested change very slightly) and explaining the behavior in a `@details` comment.

fem/integrator.hpp
david-kamensky: I've done this in the latest commit.

fem/integrator.hpp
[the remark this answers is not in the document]
david-kamensky: Modified to suggested code in a follow-up commit to fix the debug build, which is treats assignment within an `if(...)` as an error.

fem/integrator.hpp
najlkin: Looks good, maybe just add `@see SetIntRule()`, so the user is navigated to the more detailed explanation.

david-kamensky: I've added this `@see` note in the latest commit.

najlkin (approved):

pazner (approved):

tzanio:

tzanio:
Do we need to mention this in `CHANGELOG`?

najlkin:
> Do we need to mention this in `CHANGELOG`?

I think it is important not that much for users, but for developers to mention it, so the new methods for integration rules get adopted in new codes at least 😉 .

tzanio:
> > Do we need to mention this in `CHANGELOG`?
> 
> I think it is important not that much for users, but for developers to mention it, so the new methods for integration rules get adopted in new codes at least 😉 .

@david-kamensky: if you update `CHANGELOG`, I can merge this today

tzanio:
Thanks!

tzanio:

PR #4558, opened 2024-10-30
Revert "Batched mult transpose in LOR transfer"

artv3:
Reverts mfem/mfem#4555

Hi all, I believe I found a bug with this PR on the GPU. If the plan is to still merge into-next can we revert this change? Otherwise I think we need a bit more time for testing.


pazner:
It looks like the problem is device specific. Hopefully it should be easy to address. I would prefer a targeted fix (and suspect it is something minor).

pazner:
I don't think this is needed any more in light of the fix 1c90b1fd4fc358d7eeb900e50ab0202af7c8a87b.

PR #4567, opened 2024-11-01
[WIP] LorentzPM miniapp [lorentz-pm-dev]

mlstowell:
Adding a miniapp to demonstrate the use of `ParMesh::FindPoints` to compute particle trajectories due to Lorentz forces where the electric and magnetic fields are read from different `VisItDataCollection` objects.

The name of the miniapp comes from "Lorentz" and "Particle Mover". This is far from a true PIC code but it demonstrates one way to implement the particle mover portion of a PIC code.

Note: this miniapp stems from a conversation held during the MFEM 2024 Workshop.

### DIFF
--- a/miniapps/electromagnetics/CMakeLists.txt
+++ b/miniapps/electromagnetics/CMakeLists.txt
@@ -34,6 +34,11 @@ if (MFEM_USE_MPI)
    EXTRA_HEADERS maxwell_solver.hpp ${MFEM_MINIAPPS_COMMON_HEADERS}
    LIBRARIES mfem-common)
 
+ add_mfem_miniapp(lorentz_pm
+   MAIN lorentz_pm.cpp
+   EXTRA_HEADERS ${MFEM_MINIAPPS_COMMON_HEADERS}
+   LIBRARIES mfem-common)
+
   # Add the corresponding tests to the "test" target
   if (MFEM_ENABLE_TESTING)
     add_test(NAME tesla_np=4
--- a/miniapps/electromagnetics/lorentz_pm.cpp
+++ b/miniapps/electromagnetics/lorentz_pm.cpp
@@ -1,0 +1,489 @@
+// Copyright (c) 2010-2024, Lawrence Livermore National Security, LLC. Produced
+// at the Lawrence Livermore National Laboratory. All Rights reserved. See files
+// LICENSE and NOTICE for details. LLNL-CODE-806117.
+//
+// This file is part of the MFEM library. For more information and source code
+// availability visit https://mfem.org.
+//
+// MFEM is free software; you can redistribute it and/or modify it under the
+// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// CONTRIBUTING.md for details.
+//
+//          -------------------------------------------------------
+//          LorentzPM Miniapp:  Simple Lorentz Force Particle Mover
+//          -------------------------------------------------------
+//
+// This miniapp computes the trajectory of a single charged particle subject to
+// Lorentz forces.
+//
+//                           dp/dt = q (E + v x B)
+//
+// The method used is the explicit Boris algortihm which conserves phase space
+// volume for long term accuracy.
+//
+// The electric and magnetic fields are read from VisItDataCollection objects
+// such as those produced by the Volta and Tesla miniapps. It is notable that
+// these two fields do not need to be defined on the same mesh. Of course, the
+// particle trajectory can only be computed on the intersection of the two
+// domains. The starting point of the path must be chosen within in this
+// intersection and the trajectory will terminate when it leaves the
+// intersection or reaches a specified time duration.
+//
+// Note that the VisItDataCollection objects must have been stored using the
+// parallel format e.g. visit_dc.SetFormat(DataCollection::PARALLEL_FORMAT);.
+// Without this optional format specifier the vector field lookups will fail.
+//
+// Compile with: make lorentz_pm
+//
+// Sample runs:
+//
+// ./volta -m bcc_16.mesh -dbcs 1 -cs '0 0 0 0.1 2e-11' -rs 2
+// ./tesla -m hex_prism_3.mesh -rs 2 -bm '0 0 -0.1 0 0 0.1 0.1 1e8'
+// ./lorentz_pm -x0 '-0.5 0.1 0.0' -p0 '0 0 0' -q -10 -tf 8 -dt 1e-3 -rf 1e-6
+//
+// This miniapp demonstrates the use of the ParMesh::FindPoints functionality
+// to evaluate field data from stored DataCollection objects.  While this
+// miniapp is far from a full particle-in-cell (PIC) code it does demonstrate
+// some of the building blocks that might be used to construct the particle
+// mover portion of a PIC code.
+
+#include "mfem.hpp"
+#include "../common/fem_extras.hpp"
+#include "../common/pfem_extras.hpp"
+#include "electromagnetics.hpp"
+#include <fstream>
+#include <iostream>
+
+using namespace std;
+using namespace mfem;
+using namespace mfem::common;
+using namespace mfem::electromagnetics;
+
+typedef DataCollection::FieldMapType fields_t;
+
+// Background Electric field
+static Vector e_dir({1.0, 0.0, 0.0});
+static real_t e_mag = 1.0;
+void background_e(const Vector &x, Vector &E)
+{
+   E = e_dir;
+   E *= e_mag;
+}
+
+// Background Electric field
+static Vector b_dir({0.0, 0.0, 1.0});
+static real_t b_mag = 10.0;
+void background_b(const Vector &x, Vector &B)
+{
+   B = b_dir;
+   B *= b_mag;
+}
+
+class BorisAlgorithm
+{
+private:
+   real_t charge_;
+   real_t mass_;
+
+   //ParMesh &pmesh_;
+   //VectorCoefficient &ECoef_;
+   //VectorCoefficient &BCoef_;
+   /*
+   VisItDataCollection &E_dc_;
+   const char * E_field_;
+
+   VisItDataCollection &B_dc_;
+   const char * B_field_;
+   */
+   ParMesh *E_pmesh_;
+   ParGridFunction *E_field_;
+
+   ParMesh *B_pmesh_;
+   ParGridFunction *B_field_;
+
+   mutable Array<int> E_elem_id_;
+   mutable Array<IntegrationPoint> E_ip_;
+
+   mutable Array<int> B_elem_id_;
+   mutable Array<IntegrationPoint> B_ip_;
+
+   mutable Vector E_;
+   mutable Vector B_;
+   mutable Vector pxB_;
+   mutable Vector pm_;
+   mutable Vector pp_;
+
+public:
+   BorisAlgorithm(ParGridFunction *E_gf,
+                  ParGridFunction *B_gf,
+                  real_t charge, real_t mass)
+      : charge_(charge), mass_(mass),
+        E_field_(E_gf),
+        B_field_(B_gf),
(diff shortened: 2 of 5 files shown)


mlstowell:
Check out this image of the magnetic mirror effect!
mag_mirror

I'll use the start of this as the third sample run.

tzanio:
ping: @Zane-Hays

tzanio:

najlkin:
This is cool @mlstowell  , not only showcasing particle dynamics, but also tracking and visualization 👍 . However, I have one general point for discussion. I would say it is not much common to track and visualize a single particle, you rather want more of them for random sampling of the flow field. Also applications like PIC use many more than a single particle typically 🙂 . I understand the intention is to demonstrate the basic capability and not target performance, but the bar is still higher in MFEM maybe and miniapps are maybe closer to applications. So how about adding a dimension for particles, batching the operation, like queries to `FindPoints`, particle mover, and mesh construction? There could be then the associated option to randomly place a given number of particles around the domain 🤔 .

tzanio:
ping: @mlstowell

kmittal2 (approved):
Looks good to me! We will update the Lorentz miniapp to handle multiple particles using the new particle infrastructure in near future.

miniapps/electromagnetics/lorentz.cpp
```diff
@@ -0,0 +1,559 @@
...
+   {
+      Vector E_p_min(3); E_p_min = std::numeric_limits<real_t>::lowest();
+      Vector E_p_max(3); E_p_max = std::numeric_limits<real_t>::max();
+      if (E_dc != NULL)
+      {
+         ParMesh * E_pmesh = dynamic_cast<ParMesh*>(E_dc->GetMesh());
+         E_pmesh->GetBoundingBox(E_p_min, E_p_max);
+      }
+
+      Vector B_p_min(3); B_p_min = std::numeric_limits<real_t>::lowest();
```
najlkin: `infinity()` from MFEM would be easier to use and also detect

miniapps/electromagnetics/lorentz.cpp
```diff
@@ -0,0 +1,559 @@
...
+      step++;
+
+      pos_data.SetCol(step, pos);
+      mom_data.SetCol(step + 1, mom);
+   }
+   while (boris.Step(pos, mom, t, dt) && step < nsteps - 1);
+
+   if (Mpi::Root() && (visit || visualization))
+   {
+      Mesh trajectory(2, 2 * (step + 1), step, 0, 3);
```
najlkin: Reconstruction of the trajectory could be in a separate function for better readibility and reusability 😉 .

miniapps/electromagnetics/lorentz.cpp
```diff
@@ -0,0 +1,559 @@
...
+         v[3] = 2 * i + 1;
+
+         trajectory.AddQuad(v);
+      }
+
+      trajectory.FinalizeQuadMesh(1);
+
+      L2_FECollection    fec_l2(0, 2);
+      FiniteElementSpace fes_l2(&trajectory, &fec_l2);
+      GridFunction traj_time(&fes_l2);
```
najlkin, line 418: This could be modified a little bit to use the alpha channel colormaps in latest GLVis 😮 . There are two options:
- specify a transparent color palette - check share/palettes-variable-opacity.txt, the format is very simple. For black particles, you will need just something like:
```
palette 1-color-alpha RGBAf
 0  0  0  0
 0  0  0  1
```
- use the keys for transparency - send a few `k` or `K` keystrokes and shift it with `<` or `,` to the end of the range

miniapps/electromagnetics/lorentz.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+   bool Step(Vector &q, Vector &p, real_t &t, real_t &dt)
+   {
+      // Locate current point in each mesh, evaluate the fields, and collect
+      // field values on the root processor.
+      if (!GetValue(E_pmesh_, E_field_, q, E_)) { return false; }
+      if (!GetValue(B_pmesh_, B_field_, q, B_)) { return false; }
+
+      // Compute updated position and momentum using the Boris algorithm
+      if (Mpi::Root())
```
najlkin, line 171: Well, this is not really great for performance, but it is because of generality when the fields are from different meshes, right? There could be rank-local approach if the meshes are the same, which is a common case, but for a single or a few particles it does not matter much, I guess 😁 .

miniapps/electromagnetics/lorentz.cpp
```diff
@@ -0,0 +1,559 @@
...
+//
+// Sample runs:
+//
+//   Free particle moving with constant velocity
+//      mpirun -np 4 lorentz -p0 '1 1 1'
+//
+//   Particle accelerating in a constant electric field
+//      mpirun -np 4 volta -m ../../data/inline-hex.mesh -dbcs '1 6' -dbcv '0 1'
+//      mpirun -np 4 lorentz -er Volta-AMR-Parallel -x0 '0.5 0.5 0.9'
+//                           -p0 '1 0 0'
```
najlkin: Was not there some problem with multi-line sample runs? Should not it be protected somehow to not be picked by the script?🤔

miniapps/electromagnetics/lorentz.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+         // Update the position
+         q.Add(dt / mass_, p);
+      }
+
+      // Update the time
+      t += dt;
+
+      // Broadcast the updated position
+      MPI_Bcast(q.GetData(), 3, MPITypeMap<real_t>::mpi_type,
```
najlkin, line 209: We should really define `(MFEM_)MPI_REAL_T` globally, Tribol already does that in the miniapp 🤔

miniapps/electromagnetics/lorentz.cpp
mlstowell: Perhaps but I don't trust `-infinity()+infinity()` to be a usable number. Technically this sum is indeterminant.

miniapps/electromagnetics/lorentz.cpp
mlstowell: Done

miniapps/electromagnetics/lorentz.cpp
mlstowell: Done

miniapps/electromagnetics/lorentz.cpp
mlstowell: I agree that sounds useful. I'm reluctant to change anything in the main library for this mini app though.

miniapps/electromagnetics/lorentz.cpp
mlstowell: Yes, you are correct. However, I believe the work of @j-signorelli will make this irrelevant.

miniapps/electromagnetics/lorentz.cpp
mlstowell: I would be happy to add this but I'm not that familiar with these transparency keys. Exactly what "keys" string would you suggest? I also like your suggestion to use the dark background.

miniapps/electromagnetics/lorentz.cpp
najlkin: It is double (or float) infinity ([a special bit combination](https://en.wikipedia.org/wiki/Double-precision_floating-point_format)). It behaves correctly as you would expect, you may compare it with other numbers or do other operations (except the mathematically undefined). However, it is easier to check, because `isfinite()` is false 😉 .

miniapps/electromagnetics/lorentz.cpp
najlkin: Ok, we may leave it as it is.

miniapps/electromagnetics/lorentz.cpp
najlkin: Ok, let us keep this open for reference.

miniapps/electromagnetics/lorentz.cpp
mlstowell: Okay, let's try this:
```
   cout << "-infinity: " << -infinity() << endl;
   cout << "infinity:  " << infinity() << endl;
   cout << "-infinity+infinity:  " << -infinity()+infinity() << endl;
```
The output is:
```
-infinity: -inf
infinity:  inf
-infinity+infinity:  nan
```
This is not what I need. I want the sum in the last line to be zero. Now try it the other way:
```
   cout << "lowest: " << std::numeric_limits<real_t>::lowest() << endl;
   cout << "max: " << std::numeric_limits<real_t>::max() << endl;
   cout << "lowest+max: " << std::numeric_limits<real_t>::lowest()+std::numeric_limits<real_t>::max() << endl;
```
Which produces the following output:
```
lowest: -1.79769e+308
max: 1.79769e+308
lowest+max: 0
```
This is what I want. I could check for infinity using `isfinite()` and execute different expressions accordingly but this would only complicate things.

miniapps/electromagnetics/lorentz.cpp
najlkin: The first option gives much better control over the transparency, so with the palette:
```
palette 1-color-alpha RGBAf
 1  0  0  0
 1  0  0  1
```
you get something like this 🤩 :
GLVis_s01
However, you have to add the palette to a/the file, so we would need a PR for GLVis to mainstream it, and also you have to set it when starting GLVis through `-pal` argument. [I wanted that to be also a stream command](https://github.com/GLVis/glvis/pull/344#issuecomment-2940761242), but anyway it is currently only like this 🫤 .

Not quite it, but at least something similar you may get with keys `aaAckkkkk<<<<<` (plus `palette 2` command), but it is not going from 0 to 1 and behaves strange 🤪 .

miniapps/electromagnetics/lorentz.cpp
najlkin: That case `-infinity()+infinity()` should never happen as either E or B mesh exists (as checked above). If you consider that `GetBoundingBox()` can fail, it also uses `infinity()` internally, so it would probably return that.

miniapps/electromagnetics/lorentz.cpp
mlstowell: Well, I do like the idea but perhaps we should wait until GLVis more fully supports this. I'm also fairly certain that @j-signorelli has changed the way this trajectory is visualized so whatever we do here may be only temporary. Perhaps we can revisit this question in Joe's follow-up PR.

najlkin: But someone has to do it in GLVis 😁 . @justinlaughlin , would you consider adding that palette file commands to glvis/glvis#344 or do you want me to do it? The refactoring in glvis/glvis#335 will be merged soon hopefully, so it will be super easy adding new commands 😉

j-signorelli: Yeah if this helps: the way I implemented particle trajectory visualization in GLVis for N particles does not support any coloring - it's just a mesh of segments in 3D for each particle (while properly handling the possibility of particles being added/removed).

najlkin: We may circle back to it when it will be about to be merged into master and by that time a new release of GLVis might be out, but for the timeline we may leave it as it is here or use that `k` key (but it is not great honestly 🫤 ).

miniapps/electromagnetics/lorentz.cpp
mlstowell: You're not wrong but I honestly don't see why this is important. I'll change it in the interest of ending this discussion.

miniapps/electromagnetics/lorentz.cpp
justinlaughlin: Could you add those commands @najlkin and add me as a reviewer? I think 344 was fairly self-contained; I'd like to learn from you how we can now add commands "super easily" once 335 is merged 😅

tzanio:

PR #4569, opened 2024-11-06
[WIP] General handling of ownership

najlkin:


pazner:
Thanks for putting this together @najlkin. I imagine that coming to a consensus about the right design for this might take some further discussions.

I have a couple opinions on what I would like to see in a design:

1. I don't think we should base the design on `OperatorHandle`
    * Based on many past interactions with MFEM users and developers, the semantics of `OperatorHandle` are consistently brought up as a point of confusion; I don't think this class is well-liked.
    * `OperatorHandle` mixes a couple of concepts: it is both a partially managed ("half-smart") pointer, and performs a kind of "type erasure" to provide a more generic interface
    * For a resource ownership class like `Handle<T>`, the type erasure is irrelevant. The `As` and `Is` interfaces don't serve a useful purpose there. (I see you removed `pSet` in 1f06947, which is a step in the right direction).
2. I don't think the concept of a "shallow copy" is a good one. This means, for example, that you can't return an owning handle by value. It also means that assignment can have different semantics depending on whether the copy constructor is invoked or elided (RVO), which is dangerous.
3. I would personally argue for "smarter" ownership semantics (e.g. something like `shared_ptr`). This would help make a lot of the manual ownership juggling more automatic.
    * Of course, reference counting involves some overhead, so it should be opt-in, and used only in non-performance-critical parts of the code.
    * The main thing that the `OperatorHandle` interface can't express cleanly is _transferring_ or _sharing_ ownership. Because of the "shallow copy" semantics, passing an `OperatorHandle` is always a "borrow". Sharing ownership is impossible, and transferring ownership requires manually setting ownership flags.
4. In my opinion, we should aim for a design that will allow users to use it correctly without specifically providing ownership flags.

Those are some of the considerations that went into the design that I proposed here: https://godbolt.org/z/oP8szM741

najlkin:
@pazner , thanks for writing this down. First of all, let me clarify that this is not supposed to be the ultimate solution. It can be evolved to that if you want to take over this PR or there can be a follow up. Currently, this is an intermediate step to unblock #4557 , for example. The logic is following, `OperatorHandle` will have to be implemented to keep compatibility, so will be `Handle`, which makes that actually easier and is in a way future compatible.

About the points:

1. Not my choice,  `OperatorHandle` exists already and will have to be supported.
   - I do not know how liked  `OperatorHandle` is or not, but `Handle` does not provide much more than the basic interface for a pointer wrapper. There are a few extra things, which I do not like particularly (like `As()` or `Is()`), but does not make much sense to not generalize them and some folks like to do dangerous stuff (which reminds me of #4136 🙄 ).
   - There are no types in `Handle` and so no erasure.
   - `As()` or `Is()` are just shorthands for doing the static and dynamic cast as before...
2. Both, `OperatorHandle` and `Handle` have the default move constructor, so you can return them by value. It makes some sense for a kind of unique pointers. It is a risk, but again, it is in `OperatorHandle`, so there is not much choice. For the future, shared pointers and reference incrementing by copy constructor might be a better option potentially, but that needs more discussion.
3. For now, `Handle` offers something like `unique_ptr`, which is in line with the implementation of `OperatorHandle`. Feel free to make it smarter in the future 😉 . I am looking forward to it 👍 

So for now, I would either merge this to #4557 and you may continue in this PR, or merge this PR to master and you may make a follow up. If you are concerned about API, we may put `Handle` to an internal namespace for now 🤔 .

pazner:
Thanks @najlkin. I understand that you are introducing this as a "stop-gap" solution for PR #4557. My thoughts are that we should make a such an interface decision carefully and after consensus has been reached on the team, otherwise we can end up in a position where we support a suboptimal interface indefinitely. I don't know the details of #4557, but I very strongly suspect that a version of that PR could go through very easily without introducing more sweeping changes (For example, why can't `NURBSMeshRules` simply have internal `IntegrationRule` member data, a reference to which is returned by `NURBSMeshRules::GetElementRule`, obviating the need for any `deleteRule` logic? — I will comment on that PR to move the relevant discussion there).

To respond to your points,

1. `OperatorHandle` exists, and is not going anywhere, but if we introduce a smart pointer type, we don't need to be constrained by the design decisions of `OperatorHandle` — they can be separate. (This is also why we need to carefully design interfaces, so we're not locked into decisions that we may want to change later).
    * `As`, `Is`, and `Get` are all present in `OperatorHandle` because the wrapped pointer is always an abstract class. For a more general smart pointer type that would not be the case, and there is no real reason to provide these (users can always perform the casts themselves). `unique_ptr` and `shared_ptr` don't provide these.
2. `Handle` does not have a defaulted move constructor, instead the copy constructor will be called (a technicality, maybe). The problem is that the copy constructor does not transfer ownership, but returning by value with RVO _will_ transfer ownership, which is not consistent. Returning an owning `Handle` by value will result in a dangling pointer.
3. A "smarter" version (e.g. based on `shared_ptr`) will necessitate a different interface. For example, `SetOwner(false)` will not be possible in that case. Considerations like this are why I think we should settle on a design first, and then move forward with a PR.

najlkin:
@pazner , to the points:

1. I wanted from the beginning to defer the design choices for later and to implement only the existing design, which has to be supported. For that reason, I moved `Handle` to `internal` namespace to not be used by users (even when the chances are low).
   - I am no fan of `As()`, `Is()` or `Get()` and you are right they make even less sense for `Handle`, so I moved them back to not foster that design. But just saying that people like to do dirty stuff and might ask for that like [they did for `BlockOperator`](https://github.com/mfem/mfem/pull/4136#discussion_r1496697622) 😁 
2. I just tested that move constructor and returning by value works like a charm 🪄 🙂 , but you are right [it should not have technically the implicitly-defined one](https://en.cppreference.com/w/cpp/language/move_constructor) 🤔 . It is just copy elision and RVO probably. So I added the default one explicitly to make it consistent 😉 .
3. Sure, changes might be needed later, but that will require some discussion and some labor if it will be a breaking change for `OperatorHandle`. That is exactly what I am trying to avoid now and defer for later.

pazner:
Thanks @najlkin. I think a technical meeting with others on the team will be most productive to advance this discussion. I'd be happy to schedule something if you're interested.

1. Regarding `OperatorHandle`, my position is that we should design an interface for `Handle` (or whatever we decide to call it) _independently of `OperatorHandle`_. I don't think we should be constrained by the design decisions of `OperatorHandle`, and if that means that `OperatorHandle` cannot be expressed as `Handle<Operator>`, then so be it. To give a specific example, shared pointers cannot generally release ownership the way that unique pointers can. `OperatorHandle` has a mechanism to release ownership. It will be difficult (though perhaps not impossible) to resolve this.
2. Yes, it worked because of RVO, but it would not work with the previous implementation without RVO. See [here](https://godbolt.org/z/hhvz14KTG) for an example of this failing with the flag `-fno-elide-constructors`. Having the copied-into and moved-into objects in different states is not desirable.
3. One option is to keep `OperatorHandle` as it is without making breaking changes. The new interface can be distinct.

PR #4583, opened 2024-11-11
Install miniapp content

lindsayad:
The installation of the binaries I could take or leave especially because they will not work without also installing the mesh file dependencies. But installation of the headers and library I think are potentially important for downstream applications

| [#4583](https://github.com/mfem/mfem/pull/4583) | @lindsayad | @tzanio | @helloworld922 + @mlstowell + @v-dobrev + @jandrej | 11/12/24 | 12/17/24 | 12/18/24 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -636,8 +636,12 @@ mfem_cache_path(INSTALL_CMAKE_DIR "${INSTALL_LIB_DIR}/cmake/mfem"
 
 mfem_path_to_fullpath("${INSTALL_LIB_DIR}" "${CMAKE_INSTALL_PREFIX}" _lib_path)
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
-set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
-set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
+if(NOT DEFINED CMAKE_INSTALL_RPATH)
+  set(CMAKE_INSTALL_RPATH "${_lib_path}")
+endif()
+if(NOT DEFINED CMAKE_INSTALL_NAME_DIR)
+  set(CMAKE_INSTALL_NAME_DIR "${_lib_path}")
+endif()
 
 # Variables used when generating _config.hpp, and config.mk
 set(MFEM_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -615,27 +615,20 @@ set(MASTER_HEADERS
   ${PROJECT_SOURCE_DIR}/mfem.hpp
   ${PROJECT_SOURCE_DIR}/mfem-performance.hpp)
 
-# Installation options
-set(MFEM_USE_GNUINSTALLDIRS OFF
-  CACHE BOOL "Enable to use the GNUInstallDirs CMake module.")
-set(install_include_dir_default include)
-set(install_bin_dir_default bin)
-set(install_lib_dir_default lib)
-set(install_share_dir_default share)
-if (MFEM_USE_GNUINSTALLDIRS)
-  include(GNUInstallDirs)
-  set(install_include_dir_default "${CMAKE_INSTALL_INCLUDEDIR}")
-  set(install_bin_dir_default "${CMAKE_INSTALL_BINDIR}")
-  set(install_lib_dir_default "${CMAKE_INSTALL_LIBDIR}")
-  set(install_share_dir_default "${CMAKE_INSTALL_DATAROOTDIR}")
+# Installation options (we use GNUInstallDirs but prefer lib by default)
+set(MFEM_USE_GNUINSTALLDIRS OFF CACHE BOOL
+  "Use CMAKE_INSTALL_LIBDIR as defined by the GNUInstallDirs CMake module.")
+if (NOT MFEM_USE_GNUINSTALLDIRS)
+  mfem_cache_path(CMAKE_INSTALL_LIBDIR "lib" "Object code libraries (lib)")
 endif()
-mfem_cache_path(INSTALL_INCLUDE_DIR "${install_include_dir_default}"
+include(GNUInstallDirs)
+mfem_cache_path(INSTALL_INCLUDE_DIR "${CMAKE_INSTALL_INCLUDEDIR}"
   "Relative or absolute path for installing header files.")
-mfem_cache_path(INSTALL_BIN_DIR "${install_bin_dir_default}"
+mfem_cache_path(INSTALL_BIN_DIR "${CMAKE_INSTALL_BINDIR}"
   "Relative or absolute path for installing the binaries.")
-mfem_cache_path(INSTALL_LIB_DIR "${install_lib_dir_default}"
+mfem_cache_path(INSTALL_LIB_DIR "${CMAKE_INSTALL_LIBDIR}"
   "Relative or absolute path for installing the library.")
-mfem_cache_path(INSTALL_SHARE_DIR "${install_share_dir_default}"
+mfem_cache_path(INSTALL_SHARE_DIR "${CMAKE_INSTALL_DATAROOTDIR}"
   "Relative or absolute path for installing shared data.")
 # other options: "share/mfem/cmake", "lib/mfem/cmake"
 mfem_cache_path(INSTALL_CMAKE_DIR "${INSTALL_LIB_DIR}/cmake/mfem"
@@ -646,10 +639,9 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
 set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
 set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
 
-set(MFEM_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR} CACHE PATH
-    "The MFEM source directory" FORCE)
-set(MFEM_INSTALL_DIR ${CMAKE_INSTALL_PREFIX} CACHE PATH
-    "The MFEM install directory" FORCE)
+# Variables used when generating _config.hpp, and config.mk
+set(MFEM_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+set(MFEM_INSTALL_DIR ${CMAKE_INSTALL_PREFIX})
 
 # Declaring the library
 mfem_add_library(mfem ${SOURCES} ${HEADERS} ${MASTER_HEADERS})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -633,7 +633,7 @@ mfem_cache_path(INSTALL_INCLUDE_DIR "${install_include_dir_default}"
   "Relative or absolute path for installing header files.")
 mfem_cache_path(INSTALL_BIN_DIR "${install_bin_dir_default}"
   "Relative or absolute path for installing the binaries.")
-mfem_cache_path(INSTALL_LIB_DIR ${install_lib_dir_default}
+mfem_cache_path(INSTALL_LIB_DIR "${install_lib_dir_default}"
   "Relative or absolute path for installing the library.")
 mfem_cache_path(INSTALL_SHARE_DIR "${install_share_dir_default}"
   "Relative or absolute path for installing shared data.")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -615,20 +615,37 @@ set(MASTER_HEADERS
   ${PROJECT_SOURCE_DIR}/mfem.hpp
   ${PROJECT_SOURCE_DIR}/mfem-performance.hpp)
 
-set(_lib_path "${CMAKE_INSTALL_PREFIX}/lib")
-set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
-set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
-set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
-
-set(INSTALL_INCLUDE_DIR include
-  CACHE PATH "Relative path for installing header files.")
-set(INSTALL_BIN_DIR bin
-  CACHE PATH "Relative path for installing the binaries.")
-set(INSTALL_LIB_DIR lib
-  CACHE PATH "Relative path for installing the library.")
+# Installation options
+set(MFEM_USE_GNUINSTALLDIRS OFF
+  CACHE BOOL "Enable to use the GNUInstallDirs CMake module.")
+if (NOT ${MFEM_USE_GNUINSTALLDIRS})
+  set(INSTALL_INCLUDE_DIR include
+    CACHE PATH "Relative path for installing header files.")
+  set(INSTALL_BIN_DIR bin
+    CACHE PATH "Relative path for installing the binaries.")
+  set(INSTALL_LIB_DIR lib
+    CACHE PATH "Relative path for installing the library.")
+  set(INSTALL_SHARE_DIR share
+    CACHE PATH "Relative path for installing shared data.")
+else()
+  include(GNUInstallDirs)
+  set(INSTALL_INCLUDE_DIR ${CMAKE_INSTALL_INCLUDEDIR}
+    CACHE PATH "Relative path for installing header files.")
+  set(INSTALL_BIN_DIR ${CMAKE_INSTALL_BINDIR}
+    CACHE PATH "Relative path for installing the binaries.")
+  set(INSTALL_LIB_DIR ${CMAKE_INSTALL_LIBDIR}
+    CACHE PATH "Relative path for installing the library.")
+  set(INSTALL_SHARE_DIR ${CMAKE_INSTALL_DATAROOTDIR}
+    CACHE PATH "Relative path for installing shared data.")
+endif()
 # other options: "share/mfem/cmake", "lib/mfem/cmake"
 set(INSTALL_CMAKE_DIR lib/cmake/mfem
   CACHE PATH "Relative path for installing cmake config files.")
+
+set(_lib_path "${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR}")
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
+set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
+set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
 
 set(MFEM_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR} CACHE PATH
     "The MFEM source directory" FORCE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -618,31 +618,30 @@ set(MASTER_HEADERS
 # Installation options
 set(MFEM_USE_GNUINSTALLDIRS OFF
   CACHE BOOL "Enable to use the GNUInstallDirs CMake module.")
-if (NOT ${MFEM_USE_GNUINSTALLDIRS})
-  set(INSTALL_INCLUDE_DIR include
-    CACHE PATH "Relative path for installing header files.")
-  set(INSTALL_BIN_DIR bin
-    CACHE PATH "Relative path for installing the binaries.")
-  set(INSTALL_LIB_DIR lib
-    CACHE PATH "Relative path for installing the library.")
-  set(INSTALL_SHARE_DIR share
-    CACHE PATH "Relative path for installing shared data.")
-else()
+set(install_include_dir_default include)
+set(install_bin_dir_default bin)
+set(install_lib_dir_default lib)
+set(install_share_dir_default share)
+if (MFEM_USE_GNUINSTALLDIRS)
   include(GNUInstallDirs)
-  set(INSTALL_INCLUDE_DIR ${CMAKE_INSTALL_INCLUDEDIR}
-    CACHE PATH "Relative path for installing header files.")
-  set(INSTALL_BIN_DIR ${CMAKE_INSTALL_BINDIR}
-    CACHE PATH "Relative path for installing the binaries.")
-  set(INSTALL_LIB_DIR ${CMAKE_INSTALL_LIBDIR}
-    CACHE PATH "Relative path for installing the library.")
-  set(INSTALL_SHARE_DIR ${CMAKE_INSTALL_DATAROOTDIR}
-    CACHE PATH "Relative path for installing shared data.")
+  set(install_include_dir_default "${CMAKE_INSTALL_INCLUDEDIR}")
+  set(install_bin_dir_default "${CMAKE_INSTALL_BINDIR}")
+  set(install_lib_dir_default "${CMAKE_INSTALL_LIBDIR}")
+  set(install_share_dir_default "${CMAKE_INSTALL_DATAROOTDIR}")
 endif()
+mfem_cache_path(INSTALL_INCLUDE_DIR "${install_include_dir_default}"
+  "Relative or absolute path for installing header files.")
+mfem_cache_path(INSTALL_BIN_DIR "${install_bin_dir_default}"
+  "Relative or absolute path for installing the binaries.")
+mfem_cache_path(INSTALL_LIB_DIR ${install_lib_dir_default}
+  "Relative or absolute path for installing the library.")
+mfem_cache_path(INSTALL_SHARE_DIR "${install_share_dir_default}"
+  "Relative or absolute path for installing shared data.")
 # other options: "share/mfem/cmake", "lib/mfem/cmake"
-set(INSTALL_CMAKE_DIR lib/cmake/mfem
-  CACHE PATH "Relative path for installing cmake config files.")
+mfem_cache_path(INSTALL_CMAKE_DIR ${INSTALL_LIB_DIR}/cmake/mfem
+  "Relative or absolute path for installing cmake config files.")
 
-set(_lib_path "${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR}")
+mfem_path_to_fullpath("${INSTALL_LIB_DIR}" "${CMAKE_INSTALL_PREFIX}" _lib_path)
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
 set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
 set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -638,7 +638,7 @@ mfem_cache_path(INSTALL_LIB_DIR "${install_lib_dir_default}"
 mfem_cache_path(INSTALL_SHARE_DIR "${install_share_dir_default}"
   "Relative or absolute path for installing shared data.")
 # other options: "share/mfem/cmake", "lib/mfem/cmake"
-mfem_cache_path(INSTALL_CMAKE_DIR ${INSTALL_LIB_DIR}/cmake/mfem
+mfem_cache_path(INSTALL_CMAKE_DIR "${INSTALL_LIB_DIR}/cmake/mfem"
   "Relative or absolute path for installing cmake config files.")
 
 mfem_path_to_fullpath("${INSTALL_LIB_DIR}" "${CMAKE_INSTALL_PREFIX}" _lib_path)
@@ -846,8 +846,7 @@ foreach(Header mfem.hpp mfem-performance.hpp)
 endforeach()
 install(FILES ${MASTER_HEADERS} DESTINATION ${INSTALL_INCLUDE_DIR}/mfem)
 
-# Install the headers. Note that miniapp header installation is dictated from
-# its subdirectory
+# Install the headers (except common miniapp which is installed from its subdir)
 install(DIRECTORY ${MFEM_SOURCE_DIRS}
   DESTINATION ${INSTALL_INCLUDE_DIR}/mfem
   FILES_MATCHING PATTERN "*.hpp")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -620,6 +620,16 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
 set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
 set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
 
+set(INSTALL_INCLUDE_DIR include
+  CACHE PATH "Relative path for installing header files.")
+set(INSTALL_BIN_DIR bin
+  CACHE PATH "Relative path for installing the binaries.")
+set(INSTALL_LIB_DIR lib
+  CACHE PATH "Relative path for installing the library.")
+# other options: "share/mfem/cmake", "lib/mfem/cmake"
+set(INSTALL_CMAKE_DIR lib/cmake/mfem
+  CACHE PATH "Relative path for installing cmake config files.")
+
 set(MFEM_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR} CACHE PATH
     "The MFEM source directory" FORCE)
 set(MFEM_INSTALL_DIR ${CMAKE_INSTALL_PREFIX} CACHE PATH
@@ -742,7 +752,11 @@ endif()
 # Create a target for all miniapps and, optionally, enable it.
 set(MFEM_ALL_MINIAPPS_TARGET_NAME miniapps)
 add_mfem_target(${MFEM_ALL_MINIAPPS_TARGET_NAME} ${MFEM_ENABLE_MINIAPPS})
-add_subdirectory(miniapps EXCLUDE_FROM_ALL)
+if (MFEM_ENABLE_MINIAPPS)
+  add_subdirectory(miniapps) #install miniapps if enabled
+else()
+  add_subdirectory(miniapps EXCLUDE_FROM_ALL)
+endif()
 
 # Target to build all executables, i.e. everything.
 add_custom_target(exec)
@@ -792,15 +806,6 @@ add_subdirectory(doc)
 #-------------------------------------------------------------------------------
 
 message(STATUS "CMAKE_INSTALL_PREFIX = ${CMAKE_INSTALL_PREFIX}")
-set(INSTALL_INCLUDE_DIR include
-  CACHE PATH "Relative path for installing header files.")
-set(INSTALL_BIN_DIR bin
-  CACHE PATH "Relative path for installing the binaries.")
-set(INSTALL_LIB_DIR lib
-  CACHE PATH "Relative path for installing the library.")
-# other options: "share/mfem/cmake", "lib/mfem/cmake"
-set(INSTALL_CMAKE_DIR lib/cmake/mfem
-  CACHE PATH "Relative path for installing cmake config files.")
 
 target_include_directories(mfem BEFORE
   PUBLIC
@@ -825,7 +830,8 @@ foreach(Header mfem.hpp mfem-performance.hpp)
 endforeach()
 install(FILES ${MASTER_HEADERS} DESTINATION ${INSTALL_INCLUDE_DIR}/mfem)
 
-# Install the headers; currently, the miniapps headers are excluded
+# Install the headers. Note that miniapp header installation is dictated from
+# its subdirectory
 install(DIRECTORY ${MFEM_SOURCE_DIRS}
   DESTINATION ${INSTALL_INCLUDE_DIR}/mfem
   FILES_MATCHING PATTERN "*.hpp")
--- a/config/cmake/modules/MfemCmakeUtilities.cmake
+++ b/config/cmake/modules/MfemCmakeUtilities.cmake
@@ -138,6 +138,8 @@ macro(add_mfem_miniapp MFEM_EXE_NAME)
   # Actually add the executable
   mfem_add_executable(${MFEM_EXE_NAME} ${MAIN_LIST}
       ${EXTRA_SOURCES_LIST} ${EXTRA_HEADERS_LIST})
+  install(TARGETS ${MFEM_EXE_NAME}
+          RUNTIME DESTINATION miniapps)
   add_dependencies(${MFEM_ALL_MINIAPPS_TARGET_NAME} ${MFEM_EXE_NAME})
   add_dependencies(${MFEM_EXE_NAME} ${MFEM_EXEC_PREREQUISITES_TARGET_NAME})
 
--- a/config/cmake/modules/MfemCmakeUtilities.cmake
+++ b/config/cmake/modules/MfemCmakeUtilities.cmake
@@ -828,6 +828,20 @@ endfunction(mfem_get_target_options)
 
 
 #
+# If ${Path} is not an absolute path, assign ${Prefix}/${Path} to the variable
+# ${OutVar}. If ${Path} is an absolute path, assign ${Path} to the variable
+# ${OutVar}.
+#
+function(mfem_path_to_fullpath Path Prefix OutVar)
+  if(IS_ABSOLUTE "${Path}")
+    set(${OutVar} "${Path}" PARENT_SCOPE)
+  else()
+    set(${OutVar} "${Prefix}/${Path}" PARENT_SCOPE)
+  endif()
+endfunction()
+
+
+#
 #   Function that creates 'config.mk' from 'config.mk.in' for the both the
 #   build- and the install-locations and define install rules for 'config.mk'
 #   and 'test.mk'.
@@ -993,9 +1007,12 @@ function(mfem_export_mk_files)
     "${PROJECT_BINARY_DIR}/config/test.mk" COPYONLY)
 
   # Update variables for the install-tree version of 'config.mk'
-  set(MFEM_INC_DIR "${CMAKE_INSTALL_PREFIX}/${INSTALL_INCLUDE_DIR}")
-  set(MFEM_LIB_DIR "${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR}")
-  set(MFEM_TEST_MK "${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/test.mk")
+  mfem_path_to_fullpath(
+    "${INSTALL_INCLUDE_DIR}" "${CMAKE_INSTALL_PREFIX}" MFEM_INC_DIR)
+  mfem_path_to_fullpath(
+    "${INSTALL_LIB_DIR}" "${CMAKE_INSTALL_PREFIX}" MFEM_LIB_DIR)
+  mfem_path_to_fullpath(
+    "${INSTALL_SHARE_DIR}/mfem/test.mk" "${CMAKE_INSTALL_PREFIX}" MFEM_TEST_MK)
   set(MFEM_CONFIG_EXTRA "")
 
   # Create the install-tree version of 'config.mk'
@@ -1005,9 +1022,30 @@ function(mfem_export_mk_files)
 
   # Install rules for 'config.mk' and 'test.mk'
   install(FILES ${PROJECT_SOURCE_DIR}/config/test.mk
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/)
+    DESTINATION ${INSTALL_SHARE_DIR}/mfem/)
   install(FILES ${PROJECT_BINARY_DIR}/config/config-install.mk
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/
+    DESTINATION ${INSTALL_SHARE_DIR}/mfem/
     RENAME config.mk)
 
+endfunction()
+
+
+#
+# Function similar to the macro _GNUInstallDirs_cache_path from the module
+# GNUInstallDirs. Used to process variables like INSTALL_LIB_DIR if they are
+# set on the cmake command line without specifying type: -DINSTALL_LIB_DIR=lib.
+# Without this special treatment, relative paths are expanded to full paths
+# and we want to avoid that.
+#
+function(mfem_cache_path PathVar DefaultPath HelpStr)
+  if(NOT DEFINED ${PathVar})
+    set(${PathVar} "${DefaultPath}" CACHE PATH "${HelpStr}")
+  endif()
+  get_property(cache_type CACHE ${PathVar} PROPERTY TYPE)
+  if(cache_type STREQUAL "UNINITIALIZED")
+    file(TO_CMAKE_PATH "${${PathVar}}" cmakepath)
+    set_property(CACHE ${PathVar} PROPERTY TYPE PATH)
+    set_property(CACHE ${PathVar} PROPERTY VALUE "${cmakepath}")
+    set_property(CACHE ${PathVar} PROPERTY HELPSTRING "${HelpStr}")
+  endif()
 endfunction()
--- a/config/cmake/modules/MfemCmakeUtilities.cmake
+++ b/config/cmake/modules/MfemCmakeUtilities.cmake
@@ -993,9 +993,9 @@ function(mfem_export_mk_files)
     "${PROJECT_BINARY_DIR}/config/test.mk" COPYONLY)
 
   # Update variables for the install-tree version of 'config.mk'
-  set(MFEM_INC_DIR "${CMAKE_INSTALL_PREFIX}/include")
-  set(MFEM_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib")
-  set(MFEM_TEST_MK "${CMAKE_INSTALL_PREFIX}/share/mfem/test.mk")
+  set(MFEM_INC_DIR "${CMAKE_INSTALL_PREFIX}/${INSTALL_INCLUDE_DIR}")
+  set(MFEM_LIB_DIR "${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR}")
+  set(MFEM_TEST_MK "${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/test.mk")
   set(MFEM_CONFIG_EXTRA "")
 
   # Create the install-tree version of 'config.mk'
@@ -1005,8 +1005,9 @@ function(mfem_export_mk_files)
 
   # Install rules for 'config.mk' and 'test.mk'
   install(FILES ${PROJECT_SOURCE_DIR}/config/test.mk
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/share/mfem/)
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/)
   install(FILES ${PROJECT_BINARY_DIR}/config/config-install.mk
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/share/mfem/ RENAME config.mk)
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/
+    RENAME config.mk)
 
 endfunction()
--- a/makefile
+++ b/makefile
@@ -629,15 +629,6 @@ install: $(if $(static),$(BLD)libmfem.a) $(if $(shared),$(BLD)libmfem.$(SO_EXT))
 	   ( $(MKINSTALLDIR) $(PREFIX_INC)/mfem/$$dir ) && \
 	   $(INSTALLDEF) $(SRC)$$dir/*.hpp $(PREFIX_INC)/mfem/$$dir; \
 	done
-# install miniapp common libraries and headers if built
-	if ls $(BLD)miniapps/common/libmfem-common.* &> /dev/null; then \
-	   $(if $(static),$(INSTALLDEF) $(BLD)miniapps/common/libmfem-common.a $(PREFIX_LIB) &&) \
-	   $(if $(shared),$(INSTALLDEF) $(BLD)miniapps/common/libmfem-common.$(SO_VER) $(PREFIX_LIB) \
-	      && chmod $(INSTALL_BIN_PERM) $(PREFIX_LIB)/libmfem-common.$(SO_VER) \
-	      && ln -sf libmfem-common.$(SO_VER) $(PREFIX_LIB)/libmfem-common.$(SO_EXT) &&) \
-	   $(MKINSTALLDIR) $(PREFIX_INC)/mfem/miniapps/common && \
-	   $(INSTALLDEF) $(SRC)miniapps/common/*.hpp $(PREFIX_INC)/mfem/miniapps/common; \
-	fi
 # install *.okl files
 	for dir in $(OKL_DIRS); do \
 	   ( $(MKINSTALLDIR) $(PREFIX_INC)/mfem/$$dir ) && \
--- a/makefile
+++ b/makefile
@@ -629,6 +629,15 @@ install: $(if $(static),$(BLD)libmfem.a) $(if $(shared),$(BLD)libmfem.$(SO_EXT))
 	   ( $(MKINSTALLDIR) $(PREFIX_INC)/mfem/$$dir ) && \
 	   $(INSTALLDEF) $(SRC)$$dir/*.hpp $(PREFIX_INC)/mfem/$$dir; \
 	done
+# install miniapp common libraries and headers if built
+	if ls $(BLD)miniapps/common/libmfem-common.* &> /dev/null; then \
+	   $(if $(static),$(INSTALLDEF) $(BLD)miniapps/common/libmfem-common.a $(PREFIX_LIB) &&) \
+	   $(if $(shared),$(INSTALLDEF) $(BLD)miniapps/common/libmfem-common.$(SO_VER) $(PREFIX_LIB) \
+	      && chmod $(INSTALL_BIN_PERM) $(PREFIX_LIB)/libmfem-common.$(SO_VER) \
+	      && ln -sf libmfem-common.$(SO_VER) $(PREFIX_LIB)/libmfem-common.$(SO_EXT) &&) \
+	   $(MKINSTALLDIR) $(PREFIX_INC)/mfem/miniapps/common && \
+	   $(INSTALLDEF) $(SRC)miniapps/common/*.hpp $(PREFIX_INC)/mfem/miniapps/common; \
+	fi
 # install *.okl files
 	for dir in $(OKL_DIRS); do \
 	   ( $(MKINSTALLDIR) $(PREFIX_INC)/mfem/$$dir ) && \
--- a/miniapps/common/CMakeLists.txt
+++ b/miniapps/common/CMakeLists.txt
@@ -43,5 +43,12 @@ set(MINIAPPS_COMMON_HEADERS ${MINIAPPS_COMMON_HEADERS} ${HDRS} PARENT_SCOPE)
 
 mfem_add_library(mfem-common ${SRCS} ${HDRS})
 target_link_libraries(mfem-common mfem)
-install(TARGETS mfem-common ARCHIVE DESTINATION ${INSTALL_LIB_DIR} LIBRARY DESTINATION ${INSTALL_LIB_DIR})
-install(DIRECTORY ${CMAKE_SOURCE_DIR}/miniapps/common/ DESTINATION ${INSTALL_INCLUDE_DIR}/mfem/miniapps/common FILES_MATCHING PATTERN "*.hpp")
+set_target_properties(mfem-common PROPERTIES VERSION "${mfem_VERSION}")
+set_target_properties(mfem-common PROPERTIES SOVERSION "${mfem_VERSION}")
+install(TARGETS mfem-common
+  ARCHIVE DESTINATION ${INSTALL_LIB_DIR}
+  LIBRARY DESTINATION ${INSTALL_LIB_DIR}
+  RUNTIME DESTINATION ${INSTALL_BIN_DIR})
+install(DIRECTORY ${CMAKE_SOURCE_DIR}/miniapps/common
+  DESTINATION ${INSTALL_INCLUDE_DIR}/mfem/miniapps
+  FILES_MATCHING PATTERN "*.hpp")
--- a/miniapps/common/CMakeLists.txt
+++ b/miniapps/common/CMakeLists.txt
@@ -43,3 +43,5 @@ set(MINIAPPS_COMMON_HEADERS ${MINIAPPS_COMMON_HEADERS} ${HDRS} PARENT_SCOPE)
 
 mfem_add_library(mfem-common ${SRCS} ${HDRS})
 target_link_libraries(mfem-common mfem)
+install(TARGETS mfem-common ARCHIVE DESTINATION ${INSTALL_LIB_DIR} LIBRARY DESTINATION ${INSTALL_LIB_DIR})
+install(DIRECTORY ${CMAKE_SOURCE_DIR}/miniapps/common/ DESTINATION ${INSTALL_INCLUDE_DIR}/mfem/miniapps/common FILES_MATCHING PATTERN "*.hpp")
--- a/.gitignore
+++ b/.gitignore
@@ -427,3 +427,6 @@ pkg.gitcommit
 
 # Jupyter Notebook Checkpoints
 .ipynb_checkpoints
+
+# emacs tag file
+TAGS
--- a/INSTALL
+++ b/INSTALL
@@ -971,7 +971,11 @@ or
 Note that running all the tests may take a while.
 
 Installation prefix can be configured by setting the standard CMake variable
-CMAKE_INSTALL_PREFIX. To install the library, use
+CMAKE_INSTALL_PREFIX; for fine-tuning the installation directories the following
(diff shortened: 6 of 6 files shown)


v-dobrev:
@lindsayad, to start the review process for the PR, please add the `ready-for-review` label if/when ready. Thanks!

lindsayad:
HI @v-dobrev. I don't believe I have permissions to apply labels

v-dobrev:
@lindsayad, I invited you to the MFEM Contributors team -- that should let you assign labels, I think.

nmnobre:
Thanks for doing this Alex, I had it on my stack for precisely the same reasons. 😛

When I looked at what would be entailed to implement this, I spotted this line, so maybe move things around or delete it:
https://github.com/mfem/mfem/blob/a24370a2ed97856a736eb4cee4a603ee903ac2ae/CMakeLists.txt#L828

tzanio:

nmnobre (commented):
Other than my comment above, I'm happy with this since it just uses the same code structure we already do for the examples :)

I have also made the required changes to the GNU makefile for feature parity in a separate fork, https://github.com/mfem/mfem/compare/master...farscape-project:mfem:make_miniapp. I'm happy to either cherry-pick that commit here or to open a subsequent PR.

miniapps/common/CMakeLists.txt
```diff
@@ -43,3 +43,5 @@ set(MINIAPPS_COMMON_HEADERS ${MINIAPPS_COMMON_HEADERS} ${HDRS} PARENT_SCOPE)
 
 mfem_add_library(mfem-common ${SRCS} ${HDRS})
 target_link_libraries(mfem-common mfem)
+install(TARGETS mfem-common ARCHIVE DESTINATION lib LIBRARY DESTINATION lib)
+install(DIRECTORY ${CMAKE_SOURCE_DIR}/miniapps/common/ DESTINATION include/mfem/miniapps/common FILES_MATCHING PATTERN "*.hpp")
```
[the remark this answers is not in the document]
lindsayad: Alright @nmnobre I tried implementing 2. See what you think

CMakeLists.txt
```diff
@@ -158,13 +158,20 @@ if (XSDK_ENABLE_Fortran)
...
+if(POLICY CMP0068)
+    cmake_policy(SET CMP0068 NEW)
+endif()
 
 # CMake needs to know where to find things
 set(MFEM_CMAKE_PATH ${PROJECT_SOURCE_DIR}/config)
 set(CMAKE_MODULE_PATH ${MFEM_CMAKE_PATH}/cmake/modules)
 
+# Load standard install paths
+include(GNUInstallDirs)
```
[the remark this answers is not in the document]
lindsayad: Just implementing suggestion 2. from https://github.com/mfem/mfem/pull/4583#discussion_r1857486014. I'm fine going with suggestion 3. instead

CMakeLists.txt
```diff
@@ -653,8 +664,14 @@ if (MFEM_USE_MPI)
   endif()
 endif()
 
+file(RELATIVE_PATH REL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}" "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
```
nmnobre, line 667:
```suggestion
file(RELATIVE_PATH REL_LIB_DIR "${CMAKE_INSTALL_FULL_BINDIR}" "${CMAKE_INSTALL_FULL_LIBDIR}")
```

CMakeLists.txt
```diff
@@ -615,7 +626,7 @@ set(MASTER_HEADERS
   ${PROJECT_SOURCE_DIR}/mfem.hpp
   ${PROJECT_SOURCE_DIR}/mfem-performance.hpp)
 
-set(_lib_path "${CMAKE_INSTALL_PREFIX}/lib")
+set(_lib_path "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
 set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
 set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
```
nmnobre, line 632:
```suggestion
set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_FULL_LIBDIR}" CACHE PATH "")
set(CMAKE_INSTALL_NAME_DIR "${CMAKE_INSTALL_FULL_LIBDIR}" CACHE PATH "")
```

CMakeLists.txt
```diff
@@ -158,13 +158,24 @@ if (XSDK_ENABLE_Fortran)
...
-set(CMAKE_MACOSX_RPATH OFF CACHE BOOL "")
+set(CMAKE_MACOSX_RPATH ON CACHE BOOL "")
+
+if (APPLE)
+   # Set policy CMP0068 to NEW so that install_name will include RPATH settings
+   if(POLICY CMP0068)
+       cmake_policy(SET CMP0068 NEW)
+   endif()
+
+   set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
```
nmnobre, line 169: If we do this, do we need to set `BUILD_WITH_INSTALL_RPATH` later on?
And, again, if we're doing this, shouldn't we also set `CMAKE_INSTALL_RPATH` and `CMAKE_INSTALL_NAME_DIR` instead?

nmnobre (changes requested):
Hi Alex,

Since I think I'm the only who has laid eyes on this so far, can you squash the commits since my last comments into just 3 commits:
1) changes related to installing the miniapps
2) changes related to rpath issues on macOS (still haven't tested these)
3) .gitignore for emacs

CMakeLists.txt
```diff
@@ -158,13 +158,24 @@ if (XSDK_ENABLE_Fortran)
...
+endif()
 
 # CMake needs to know where to find things
 set(MFEM_CMAKE_PATH ${PROJECT_SOURCE_DIR}/config)
 set(CMAKE_MODULE_PATH ${MFEM_CMAKE_PATH}/cmake/modules)
 
+# Load standard install paths
```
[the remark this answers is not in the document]
lindsayad: Shouldn't this line come after `include(GNUInstallDirs)`? So that we ensure that our "custom" path overrides the default value set by GNUInstallDirs?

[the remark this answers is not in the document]
nmnobre: No, because this is a cache variable, so we need to leapfrog GNUInstallDirs before it sets it, at which point it'll be too late for us to override it, in other words: doing set on a cache variable doesn't override it, just sets it if undefined.

CMakeLists.txt
lindsayad: followed your suggestion in https://github.com/mfem/mfem/pull/4583/commits/ddfc59f3c321b7ccbed0e9fcffb6f0983ab49739

CMakeLists.txt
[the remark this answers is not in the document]
lindsayad: We could use you to train INL's AI tool

CMakeLists.txt
nmnobre: Check Slack.

CMakeLists.txt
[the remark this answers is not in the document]
v-dobrev: Yes, I don't want to use `GNUInstallDirs`. We have our own options for what's installed where -- can't we just use them? I'm talking about the following variables: https://github.com/mfem/mfem/blob/899a96b7699096c994bf64a375421c9084915141/CMakeLists.txt#L798-L806

If you really want to use `GNUInstallDirs`, we can make it an option but I prefer to NOT use it as the default.

CMakeLists.txt
```diff
@@ -158,13 +158,29 @@ if (XSDK_ENABLE_Fortran)
...
 set(MFEM_CMAKE_PATH ${PROJECT_SOURCE_DIR}/config)
 set(CMAKE_MODULE_PATH ${MFEM_CMAKE_PATH}/cmake/modules)
 
+# Load standard install paths (but keep lib as the default on all platforms)
+set(CMAKE_INSTALL_LIBDIR lib CACHE PATH "Object code libraries (lib)")
+include(GNUInstallDirs)
+
+# Use RPATH on MacOS to avoid re-linking which invalidates code signatures
+file(RELATIVE_PATH REL_LIB_DIR "${CMAKE_INSTALL_FULL_BINDIR}" "${CMAKE_INSTALL_FULL_LIBDIR}")
+if (APPLE)
```
v-dobrev, line 173: I don't quite follow -- what issue is this solving?

CMakeLists.txt
```diff
@@ -158,13 +158,29 @@ if (XSDK_ENABLE_Fortran)
    enable_language(Fortran)
 endif()
 
-# Suppress warnings about MACOSX_RPATH
-set(CMAKE_MACOSX_RPATH OFF CACHE BOOL "")
+set(CMAKE_MACOSX_RPATH ON CACHE BOOL "")
```
v-dobrev, line 161: Why this change?

CMakeLists.txt
```diff
@@ -615,10 +631,9 @@ set(MASTER_HEADERS
   ${PROJECT_SOURCE_DIR}/mfem.hpp
   ${PROJECT_SOURCE_DIR}/mfem-performance.hpp)
 
-set(_lib_path "${CMAKE_INSTALL_PREFIX}/lib")
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
-set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
-set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
+set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_FULL_LIBDIR}" CACHE PATH "")
+set(CMAKE_INSTALL_NAME_DIR "${CMAKE_INSTALL_FULL_LIBDIR}" CACHE PATH "")
```
v-dobrev, line 636: This should use `INSTALL_LIB_DIR` and not assume `GNUInstallDirs` is always used. This will require the defaults for `INSTALL_LIB_DIR` and similar to be moved up. If we want to have an option to use `GNUInstallDirs` then `INSTALL_LIB_DIR` can be defined based on what `GNUInstallDirs` defines.

CMakeLists.txt
```diff
@@ -792,19 +811,13 @@ add_subdirectory(doc)
 #-------------------------------------------------------------------------------
 
 message(STATUS "CMAKE_INSTALL_PREFIX = ${CMAKE_INSTALL_PREFIX}")
-set(INSTALL_INCLUDE_DIR include
-  CACHE PATH "Relative path for installing header files.")
-set(INSTALL_BIN_DIR bin
-  CACHE PATH "Relative path for installing the binaries.")
-set(INSTALL_LIB_DIR lib
-  CACHE PATH "Relative path for installing the library.")
```
v-dobrev, line 800: These should be restored and moved up as noted in my comment above.

CMakeLists.txt
```diff
@@ -792,19 +811,13 @@ add_subdirectory(doc)
...
 # other options: "share/mfem/cmake", "lib/mfem/cmake"
-set(INSTALL_CMAKE_DIR lib/cmake/mfem
+set(INSTALL_CMAKE_DIR ${CMAKE_INSTALL_LIBDIR}/cmake/mfem
```
v-dobrev, line 815: Use `INSTALL_LIB_DIR` instead of `GNUInstallDirs`'s `CMAKE_INSTALL_LIBDIR`. This should be updated in other places too.

config/cmake/modules/MfemCmakeUtilities.cmake
```diff
@@ -101,7 +101,6 @@ macro(add_mfem_examples EXE_SRCS)
     endif()
     add_dependencies(${EXE_NAME}
       ${MFEM_EXEC_PREREQUISITES_TARGET_NAME} ${EXE_PREREQUISITE})
-
```
v-dobrev, line 104: Revert please.

config/cmake/modules/MfemCmakeUtilities.cmake
```diff
@@ -138,6 +137,8 @@ macro(add_mfem_miniapp MFEM_EXE_NAME)
   # Actually add the executable
   mfem_add_executable(${MFEM_EXE_NAME} ${MAIN_LIST}
       ${EXTRA_SOURCES_LIST} ${EXTRA_HEADERS_LIST})
+  install(TARGETS ${MFEM_EXE_NAME}
+          RUNTIME DESTINATION miniapps)
```
v-dobrev, line 141: This should be done only if `MFEM_ENABLE_MINIAPPS` is true. I'm not sure if using `add_subdirectory(miniapps EXCLUDE_FROM_ALL)` excludes also the `install` commands or just the targets.

makefile
```diff
@@ -629,6 +629,15 @@ install: $(if $(static),$(BLD)libmfem.a) $(if $(shared),$(BLD)libmfem.$(SO_EXT))
...
 	done
+# install miniapp common libraries and headers if built
+	if ls $(BLD)miniapps/common/libmfem-common.* &> /dev/null; then \
+	   $(if $(static),$(INSTALLDEF) $(BLD)miniapps/common/libmfem-common.a $(PREFIX_LIB) &&) \
+	   $(if $(shared),$(INSTALLDEF) $(BLD)miniapps/common/libmfem-common.$(SO_VER) $(PREFIX_LIB) \
+	      && chmod $(INSTALL_BIN_PERM) $(PREFIX_LIB)/libmfem-common.$(SO_VER) \
+	      && ln -sf libmfem-common.$(SO_VER) $(PREFIX_LIB)/libmfem-common.$(SO_EXT) &&) \
+	   $(MKINSTALLDIR) $(PREFIX_INC)/mfem/miniapps/common && \
+	   $(INSTALLDEF) $(SRC)miniapps/common/*.hpp $(PREFIX_INC)/mfem/miniapps/common; \
+	fi
```
v-dobrev, line 640: In the GNU make build system, we're not installing examples, so I prefer to not install miniapps either. Unless, of course, you actually want to use this feature.

nmnobre: This wouldn't be installing binaries, only libraries and headers for the common miniapp, so I'd argue it's a bit different. If you are open to the idea of merging this, I'll open another PR though, I don't want to overload the discussion here, I already feel guilty enough for suggesting `GNUInstallDirs`.

config/cmake/modules/MfemCmakeUtilities.cmake
nmnobre: It is, from the manual:

"If the EXCLUDE_FROM_ALL argument is provided, it has the following effects: [...]
Any install rules defined in the subdirectory or below will be ignored when installing the parent directory."

makefile
v-dobrev: If you think you are going to use this, it's not a problem to include it. I just thought you are primarily interested in the CMake build system.

One suggestion: I think it is better to trigger this installation explicitly via a makefile variable, e.g. `MFEM_INSTALL_MINIAPPS_COMMON`, set to `YES/NO` with a default setting in `config/defaults.mk`. The variable can be set during config time and exported to `config.mk` similar to other configuration options. I can make this change, if you are not familiar with all the details on how to do this.

nmnobre: I'm happy to live with CMake personally and for our purposes it does everything we need.

I was just trying to go for feature parity (just like #4252 was).
I guess I perhaps don't understand why we are maintaining two build systems, I thought it was for the convenience of the user, to allow them to use whatever they are most comfortable with. For instance, the "Getting Started/Building MFEM" manual page uses GNU make, and doesn't even mention CMake, but then the user is locked out of some functionality if they don't migrate to CMake.

CMakeLists.txt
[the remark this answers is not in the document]
nmnobre: No, we don't need `GNUInstallDirs`. I suggested it precisely as a way of avoiding or own options in favour of standard names, just like we use the standard `CMAKE_INSTALL_PREFIX` for example. I'm really sorry we now have to revert Alex's changes and for any time wasted for all people involved.

makefile
v-dobrev: > I perhaps don't understand why we are maintaining two build systems

Different users use different build systems, so we don't want to force them to switch.

> "Getting Started/Building MFEM" manual page uses GNU make, and doesn't even mention CMake

This will be addressed in https://github.com/mfem/web/pull/291. 😄 

> the user is locked out of some functionality if they don't migrate to CMake

There are only minor differences between GNU make and CMake functionality -- one example of that being that installing the MFEM examples is not supported by GNU make.

Regarding the current PR: I'm fine either way: add or don't add the GNU make support for installing the `mfem-common` library and its headers.

CMakeLists.txt
lindsayad: On Mac arm64, there are tons of warnings during the install phase, e.g.
```
-- Installing: /Users/lindad/projects/mfem-install-miniapp-content/installed/examples/ex27
/Users/lindad/miniforge/envs/moose-mpich/bin/install_name_tool: warning: changes being made to the file will invalidate the code signature in: /Users/lindad/projects/mfem-install-miniapp-content/installed/examples/ex27
[cctools-port]: generating fake signature for '/Users/lindad/projects/mfem-install-miniapp-content/installed/examples/ex27'
/Users/lindad/miniforge/envs/moose-mpich/bin/install_name_tool: warning: changes being made to the file will invalidate the code signature in: /Users/lindad/projects/mfem-install-miniapp-content/installed/examples/ex27
[cctools-port]: generating fake signature for '/Users/lindad/projects/mfem-install-miniapp-content/installed/examples/ex27'
```
By using the install rpath for the build binaries, then I could avoid re-linking and avoid these warnings ... of course this would have consequences for `make check`

lindsayad: However, motivated by @nmnobre digging into this with his own Mac ARM machine and not seeing any warnings, I built an entire stack from scratch using my Mac's system tools instead of conda tools ... and then I also saw no warnings. So conda seems to be the foul player here. I will revert this work

makefile
nmnobre: > I think it is better to trigger this installation explicitly via a makefile variable, e.g. MFEM_INSTALL_MINIAPPS_COMMON, set to YES/NO with a default setting in config/defaults.mk.

I agree doing a raw `ls` to check for existence is a bit rough. And I think defining such a variable goes a bit against the philosophy of our GNU makefiles as you've hinted at. I think the cleanest, most natural solution would be to define a new rule such as `install-miniapps: install miniapps` which would only work in practice for people who actually open and inspect the makefile (in the same way as, e.g., the `miniapps` or `test-miniapps` rules which we don't publicise to the user). So, if the team prefers to just not have the option, I'm happy to drop it.

lindsayad:
Ok I think this is ready for further review. I've left in the GNUInstallDirs commit plus its reversion for historical purposes, but am happy to cut those out

v-dobrev:
Here is a patch with some additional tweaks:
* Add a new option to use `GNUInstallDirs`: `MFEM_USE_GNUINSTALLDIRS` (`OFF` by default).
* Remove miniapp installation from GNU make -- if someone really prefers to keep it, we can work on an improved installation procedure.
* Install `miniapps/common` headers in `<prefix>/include/mfem/miniapps/common` and not in `<prefix>/include/mfem/miniapps/common/common` (`common` was repeated twice).
* Install `libmfem-common.*` with a version suffix and link -- similar to how we install `libmfem.*`.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 67f37a9b5e..fe83c83afc 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -615,21 +615,38 @@ set(MASTER_HEADERS
   ${PROJECT_SOURCE_DIR}/mfem.hpp
   ${PROJECT_SOURCE_DIR}/mfem-performance.hpp)
 
-set(_lib_path "${CMAKE_INSTALL_PREFIX}/lib")
-set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
-set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
-set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
-
-set(INSTALL_INCLUDE_DIR include
-  CACHE PATH "Relative path for installing header files.")
-set(INSTALL_BIN_DIR bin
-  CACHE PATH "Relative path for installing the binaries.")
-set(INSTALL_LIB_DIR lib
-  CACHE PATH "Relative path for installing the library.")
+# Installation options
+set(MFEM_USE_GNUINSTALLDIRS OFF
+  CACHE BOOL "Enable to use the GNUInstallDirs CMake module.")
+if (NOT ${MFEM_USE_GNUINSTALLDIRS})
+  set(INSTALL_INCLUDE_DIR include
+    CACHE PATH "Relative path for installing header files.")
+  set(INSTALL_BIN_DIR bin
+    CACHE PATH "Relative path for installing the binaries.")
+  set(INSTALL_LIB_DIR lib
+    CACHE PATH "Relative path for installing the library.")
+  set(INSTALL_SHARE_DIR share
+    CACHE PATH "Relative path for installing shared data.")
+else()
+  include(GNUInstallDirs)
+  set(INSTALL_INCLUDE_DIR ${CMAKE_INSTALL_INCLUDEDIR}
+    CACHE PATH "Relative path for installing header files.")
+  set(INSTALL_BIN_DIR ${CMAKE_INSTALL_BINDIR}
+    CACHE PATH "Relative path for installing the binaries.")
+  set(INSTALL_LIB_DIR ${CMAKE_INSTALL_LIBDIR}
+    CACHE PATH "Relative path for installing the library.")
+  set(INSTALL_SHARE_DIR ${CMAKE_INSTALL_DATAROOTDIR}
+    CACHE PATH "Relative path for installing shared data.")
+endif()
 # other options: "share/mfem/cmake", "lib/mfem/cmake"
 set(INSTALL_CMAKE_DIR lib/cmake/mfem
   CACHE PATH "Relative path for installing cmake config files.")
 
+set(_lib_path "${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR}")
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
+set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
+set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
+
 set(MFEM_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR} CACHE PATH
     "The MFEM source directory" FORCE)
 set(MFEM_INSTALL_DIR ${CMAKE_INSTALL_PREFIX} CACHE PATH
diff --git a/config/cmake/modules/MfemCmakeUtilities.cmake b/config/cmake/modules/MfemCmakeUtilities.cmake
index cb2d35641f..f7ecc9e6c9 100644
--- a/config/cmake/modules/MfemCmakeUtilities.cmake
+++ b/config/cmake/modules/MfemCmakeUtilities.cmake
@@ -993,9 +993,9 @@ function(mfem_export_mk_files)
     "${PROJECT_BINARY_DIR}/config/test.mk" COPYONLY)
 
   # Update variables for the install-tree version of 'config.mk'
-  set(MFEM_INC_DIR "${CMAKE_INSTALL_PREFIX}/include")
-  set(MFEM_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib")
-  set(MFEM_TEST_MK "${CMAKE_INSTALL_PREFIX}/share/mfem/test.mk")
+  set(MFEM_INC_DIR "${CMAKE_INSTALL_PREFIX}/${INSTALL_INCLUDE_DIR}")
+  set(MFEM_LIB_DIR "${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR}")
+  set(MFEM_TEST_MK "${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/test.mk")
   set(MFEM_CONFIG_EXTRA "")
 
   # Create the install-tree version of 'config.mk'
@@ -1005,8 +1005,9 @@ function(mfem_export_mk_files)
 
   # Install rules for 'config.mk' and 'test.mk'
   install(FILES ${PROJECT_SOURCE_DIR}/config/test.mk
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/share/mfem/)
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/)
   install(FILES ${PROJECT_BINARY_DIR}/config/config-install.mk
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/share/mfem/ RENAME config.mk)
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/
+    RENAME config.mk)
 
 endfunction()
diff --git a/makefile b/makefile
index d6cbcad77c..43587c09f1 100644
--- a/makefile
+++ b/makefile
@@ -629,15 +629,6 @@ install: $(if $(static),$(BLD)libmfem.a) $(if $(shared),$(BLD)libmfem.$(SO_EXT))
 	   ( $(MKINSTALLDIR) $(PREFIX_INC)/mfem/$$dir ) && \
 	   $(INSTALLDEF) $(SRC)$$dir/*.hpp $(PREFIX_INC)/mfem/$$dir; \
 	done
-# install miniapp common libraries and headers if built
-	if ls $(BLD)miniapps/common/libmfem-common.* &> /dev/null; then \
-	   $(if $(static),$(INSTALLDEF) $(BLD)miniapps/common/libmfem-common.a $(PREFIX_LIB) &&) \
-	   $(if $(shared),$(INSTALLDEF) $(BLD)miniapps/common/libmfem-common.$(SO_VER) $(PREFIX_LIB) \
-	      && chmod $(INSTALL_BIN_PERM) $(PREFIX_LIB)/libmfem-common.$(SO_VER) \
-	      && ln -sf libmfem-common.$(SO_VER) $(PREFIX_LIB)/libmfem-common.$(SO_EXT) &&) \
-	   $(MKINSTALLDIR) $(PREFIX_INC)/mfem/miniapps/common && \
-	   $(INSTALLDEF) $(SRC)miniapps/common/*.hpp $(PREFIX_INC)/mfem/miniapps/common; \
-	fi
 # install *.okl files
 	for dir in $(OKL_DIRS); do \
 	   ( $(MKINSTALLDIR) $(PREFIX_INC)/mfem/$$dir ) && \
diff --git a/miniapps/common/CMakeLists.txt b/miniapps/common/CMakeLists.txt
index 5d8f086427..2875711104 100644
--- a/miniapps/common/CMakeLists.txt
+++ b/miniapps/common/CMakeLists.txt
@@ -43,5 +43,12 @@ set(MINIAPPS_COMMON_HEADERS ${MINIAPPS_COMMON_HEADERS} ${HDRS} PARENT_SCOPE)
 
 mfem_add_library(mfem-common ${SRCS} ${HDRS})
 target_link_libraries(mfem-common mfem)
-install(TARGETS mfem-common ARCHIVE DESTINATION ${INSTALL_LIB_DIR} LIBRARY DESTINATION ${INSTALL_LIB_DIR})
-install(DIRECTORY ${CMAKE_SOURCE_DIR}/miniapps/common/ DESTINATION ${INSTALL_INCLUDE_DIR}/mfem/miniapps/common FILES_MATCHING PATTERN "*.hpp")
+set_target_properties(mfem-common PROPERTIES VERSION "${mfem_VERSION}")
+set_target_properties(mfem-common PROPERTIES SOVERSION "${mfem_VERSION}")
+install(TARGETS mfem-common
+  ARCHIVE DESTINATION ${INSTALL_LIB_DIR}
+  LIBRARY DESTINATION ${INSTALL_LIB_DIR}
+  RUNTIME DESTINATION ${INSTALL_BIN_DIR})
+install(DIRECTORY ${CMAKE_SOURCE_DIR}/miniapps/common
+  DESTINATION ${INSTALL_INCLUDE_DIR}/mfem/miniapps
+  FILES_MATCHING PATTERN "*.hpp")
```

lindsayad:
> * Install `miniapps/common` headers in `<prefix>/include/mfem/miniapps/common` and not in `<prefix>/include/mfem/miniapps/common/common` (`common` was repeated twice).

Are you sure about this? I do not see this on my machine
```
(base) lindad@INL429698:~/projects/mfem/installed/include/mfem/miniapps/common(alex-dev)$ ls
dist_solver.hpp fem_extras.hpp  mesh_extras.hpp mfem-common.hpp pfem_extras.hpp
```

v-dobrev:
> > * Install `miniapps/common` headers in `<prefix>/include/mfem/miniapps/common` and not in `<prefix>/include/mfem/miniapps/common/common` (`common` was repeated twice).
> 
> Are you sure about this? I do not see this on my machine
> 
> ```
> (base) lindad@INL429698:~/projects/mfem/installed/include/mfem/miniapps/common(alex-dev)$ ls
> dist_solver.hpp fem_extras.hpp  mesh_extras.hpp mfem-common.hpp pfem_extras.hpp
> ```

I saw this behavior (`common` repeated twice) after I removed the last `/` in `${CMAKE_SOURCE_DIR}/miniapps/common/` -- maybe that is the difference?

lindsayad:
> I saw this behavior (`common` repeated twice) after I removed the last `/` in `${CMAKE_SOURCE_DIR}/miniapps/common/` -- maybe that is the difference?

Yea I verified that making this change does lead to the repeated `common`

nmnobre:
I've finally had some time to study the situation thoroughly, let me try to explain it as best as I can.

First, @v-dobrev, I now understand we never really intended for `INSTALL_{INCLUDE,BIN,LIB}_DIR` to be set by the user in the command line, did we? We've always had hard code paths (for instance the rpath which would never respect what the user would pass in `INSTALL_LIB_DIR` anyway). So I think it's my bad I mislead Alex and I think yourself in this regard. _The variables shouldn't be marked `CACHE`._

Second, the script above assumes these variables hold relative paths, which isn't true if the user tries to set them. Indeed from the manual for `set()`:
"Furthermore, if the <type> is PATH or FILEPATH and the <value> provided on the command line is a relative path, then the set command will treat the path as relative to the current working directory and convert it to an absolute path."

So, I think this is enough evidence that we can rest assured there's nobody out there setting any of `INSTALL_{INCLUDE,BIN,LIB}_DIR`. Thus, if we wanted to allow the user set their preference from now on, it is indeed safe to rename these variable freely. `GNUInstallDirs` takes care to preserve relative paths passed on the command line, so it's not affected by the issue above.

v-dobrev:
Hi @nmnobre,

Thanks for taking a closer look at the changes I proposed. Some comments:

> ... we never really intended for `INSTALL_{INCLUDE,BIN,LIB}_DIR` to be set by the user in the command line, did we?

We did intent for these to be user-configurable, however, as you noticed they were not used in some places, so that was a bug.

As you point out, setting e.g. `-DINSTALL_LIB_DIR=lib99` will not behave as one might expect (I did not realize this limitation), however, one can use either `-DINSTALL_LIB_DIR:PATH=lib99` or `-DINSTALL_LIB_DIR:STRING=lib99` to get the desired result. We can add logic to handle the case of `-DINSTALL_LIB_DIR=lib99` as well.

The `INSTALL_*_DIR` variables are actually used, for example in the Fedora MFEM package, see https://src.fedoraproject.org/rpms/mfem/blob/main/f/mfem.spec. There, these variables are set to absolute paths which we should continue to support and not assume that the paths are always relative. This will require some additional tweaks which I'm looking into making -- I'll post a suggested patch for this in a little bit.

Regarding the `SOVERSION`: I think I have both `VERSION` and `SOVERSION` just as reminder that there are separate options that can be set independently.

nmnobre:
> We did intent for these to be user-configurable, however, as you noticed they were not used in some places, so that was a bug.

Okay, I'm glad that's the case. :)

> As you point out, setting e.g. `-DINSTALL_LIB_DIR=lib99` will not behave as one might expect (I did not realize this limitation), however, one can use either `-DINSTALL_LIB_DIR:PATH=lib99` or `-DINSTALL_LIB_DIR:STRING=lib99` to get the desired result. We can add logic to handle the case of `-DINSTALL_LIB_DIR=lib99` as well.

Yes, if the user specifies the type, everything works as expected (although, in practice, I rarely see it?). So to workaround this, I suggest we have a look at the source for `GNUInstallDirs`, see around this line https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/GNUInstallDirs.cmake#L172. But then, it does feel a bit like we are re~inventing~writing the wheel if you know what I mean? :) But, in light of the below, I do see _why_ we might want to do this.

> The `INSTALL_*_DIR` variables are actually used, for example in the Fedora MFEM package

-----

> Regarding the `SOVERSION`: I think I have both `VERSION` and `SOVERSION` just as reminder that there are separate options that can be set independently.

Fair enough.

v-dobrev:
Here's the patch I came up with to address the issues noted by @nmnobre. I also added some documentation in the `INSTALL` file about the fine-tuning of the installation directories with CMake.
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index fe83c83afc..da42b1528c 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -618,31 +618,30 @@ set(MASTER_HEADERS
 # Installation options
 set(MFEM_USE_GNUINSTALLDIRS OFF
   CACHE BOOL "Enable to use the GNUInstallDirs CMake module.")
-if (NOT ${MFEM_USE_GNUINSTALLDIRS})
-  set(INSTALL_INCLUDE_DIR include
-    CACHE PATH "Relative path for installing header files.")
-  set(INSTALL_BIN_DIR bin
-    CACHE PATH "Relative path for installing the binaries.")
-  set(INSTALL_LIB_DIR lib
-    CACHE PATH "Relative path for installing the library.")
-  set(INSTALL_SHARE_DIR share
-    CACHE PATH "Relative path for installing shared data.")
-else()
+set(install_include_dir_default include)
+set(install_bin_dir_default bin)
+set(install_lib_dir_default lib)
+set(install_share_dir_default share)
+if (MFEM_USE_GNUINSTALLDIRS)
   include(GNUInstallDirs)
-  set(INSTALL_INCLUDE_DIR ${CMAKE_INSTALL_INCLUDEDIR}
-    CACHE PATH "Relative path for installing header files.")
-  set(INSTALL_BIN_DIR ${CMAKE_INSTALL_BINDIR}
-    CACHE PATH "Relative path for installing the binaries.")
-  set(INSTALL_LIB_DIR ${CMAKE_INSTALL_LIBDIR}
-    CACHE PATH "Relative path for installing the library.")
-  set(INSTALL_SHARE_DIR ${CMAKE_INSTALL_DATAROOTDIR}
-    CACHE PATH "Relative path for installing shared data.")
-endif()
+  set(install_include_dir_default "${CMAKE_INSTALL_INCLUDEDIR}")
+  set(install_bin_dir_default "${CMAKE_INSTALL_BINDIR}")
+  set(install_lib_dir_default "${CMAKE_INSTALL_LIBDIR}")
+  set(install_share_dir_default "${CMAKE_INSTALL_DATAROOTDIR}")
+endif()
+mfem_cache_path(INSTALL_INCLUDE_DIR "${install_include_dir_default}"
+  "Relative or absolute path for installing header files.")
+mfem_cache_path(INSTALL_BIN_DIR "${install_bin_dir_default}"
+  "Relative or absolute path for installing the binaries.")
+mfem_cache_path(INSTALL_LIB_DIR ${install_lib_dir_default}
+  "Relative or absolute path for installing the library.")
+mfem_cache_path(INSTALL_SHARE_DIR "${install_share_dir_default}"
+  "Relative or absolute path for installing shared data.")
 # other options: "share/mfem/cmake", "lib/mfem/cmake"
-set(INSTALL_CMAKE_DIR lib/cmake/mfem
-  CACHE PATH "Relative path for installing cmake config files.")
+mfem_cache_path(INSTALL_CMAKE_DIR ${INSTALL_LIB_DIR}/cmake/mfem
+  "Relative or absolute path for installing cmake config files.")
 
-set(_lib_path "${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR}")
+mfem_path_to_fullpath("${INSTALL_LIB_DIR}" "${CMAKE_INSTALL_PREFIX}" _lib_path)
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
 set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
 set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
diff --git a/INSTALL b/INSTALL
index 962e386894..d95b10c005 100644
--- a/INSTALL
+++ b/INSTALL
@@ -971,7 +971,11 @@ or
 Note that running all the tests may take a while.
 
 Installation prefix can be configured by setting the standard CMake variable
-CMAKE_INSTALL_PREFIX. To install the library, use
+CMAKE_INSTALL_PREFIX; for fine-tuning the installation directories the following
+variables can be used: INSTALL_INCLUDE_DIR, INSTALL_LIB_DIR, INSTALL_BIN_DIR,
+INSTALL_SHARE_DIR, and INSTALL_CMAKE_DIR; alternatively, fine-tuning can be done
+by setting MFEM_USE_GNUINSTALLDIRS=ON to use the CMake module GNUInstallDirs.
+To install the library, use
 
    make install
 or
diff --git a/config/cmake/modules/MfemCmakeUtilities.cmake b/config/cmake/modules/MfemCmakeUtilities.cmake
index f7ecc9e6c9..7869dbef32 100644
--- a/config/cmake/modules/MfemCmakeUtilities.cmake
+++ b/config/cmake/modules/MfemCmakeUtilities.cmake
@@ -827,6 +827,20 @@ function(mfem_get_target_options Target CompileOptsVar LinkOptsVar)
 endfunction(mfem_get_target_options)
 
 
+#
+# If ${Path} is not an absolute path, assign ${Prefix}/${Path} to the variable
+# ${OutVar}. If ${Path} is an absolute path, assign ${Path} to the variable
+# ${OutVar}.
+#
+function(mfem_path_to_fullpath Path Prefix OutVar)
+  if(IS_ABSOLUTE "${Path}")
+    set(${OutVar} "${Path}" PARENT_SCOPE)
+  else()
+    set(${OutVar} "${Prefix}/${Path}" PARENT_SCOPE)
+  endif()
+endfunction()
+
+
 #
 #   Function that creates 'config.mk' from 'config.mk.in' for the both the
 #   build- and the install-locations and define install rules for 'config.mk'
@@ -993,9 +1007,12 @@ function(mfem_export_mk_files)
     "${PROJECT_BINARY_DIR}/config/test.mk" COPYONLY)
 
   # Update variables for the install-tree version of 'config.mk'
-  set(MFEM_INC_DIR "${CMAKE_INSTALL_PREFIX}/${INSTALL_INCLUDE_DIR}")
-  set(MFEM_LIB_DIR "${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR}")
-  set(MFEM_TEST_MK "${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/test.mk")
+  mfem_path_to_fullpath(
+    "${INSTALL_INCLUDE_DIR}" "${CMAKE_INSTALL_PREFIX}" MFEM_INC_DIR)
+  mfem_path_to_fullpath(
+    "${INSTALL_LIB_DIR}" "${CMAKE_INSTALL_PREFIX}" MFEM_LIB_DIR)
+  mfem_path_to_fullpath(
+    "${INSTALL_SHARE_DIR}/mfem/test.mk" "${CMAKE_INSTALL_PREFIX}" MFEM_TEST_MK)
   set(MFEM_CONFIG_EXTRA "")
 
   # Create the install-tree version of 'config.mk'
@@ -1005,9 +1022,30 @@ function(mfem_export_mk_files)
 
   # Install rules for 'config.mk' and 'test.mk'
   install(FILES ${PROJECT_SOURCE_DIR}/config/test.mk
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/)
+    DESTINATION ${INSTALL_SHARE_DIR}/mfem/)
   install(FILES ${PROJECT_BINARY_DIR}/config/config-install.mk
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/${INSTALL_SHARE_DIR}/mfem/
+    DESTINATION ${INSTALL_SHARE_DIR}/mfem/
     RENAME config.mk)
 
 endfunction()
+
+
+#
+# Function similar to the macro _GNUInstallDirs_cache_path from the module
+# GNUInstallDirs. Used to process variables like INSTALL_LIB_DIR if they are
+# set on the cmake command line without specifying type: -DINSTALL_LIB_DIR=lib.
+# Without this special treatment, relative paths are expanded to full paths
+# and we want to avoid that.
+#
+function(mfem_cache_path PathVar DefaultPath HelpStr)
+  if(NOT DEFINED ${PathVar})
+    set(${PathVar} "${DefaultPath}" CACHE PATH "${HelpStr}")
+  endif()
+  get_property(cache_type CACHE ${PathVar} PROPERTY TYPE)
+  if(cache_type STREQUAL "UNINITIALIZED")
+    file(TO_CMAKE_PATH "${${PathVar}}" cmakepath)
+    set_property(CACHE ${PathVar} PROPERTY TYPE PATH)
+    set_property(CACHE ${PathVar} PROPERTY VALUE "${cmakepath}")
+    set_property(CACHE ${PathVar} PROPERTY HELPSTRING "${HelpStr}")
+  endif()
+endfunction()
```

v-dobrev (approved):

tzanio:
@nmnobre, @mlstowell, @helloworld922 and @jandrej -- Can you take another look when you get a chance?

nmnobre (approved):
I think it's looking good, I've suggested a few minor changes for consistency throughout, including the below (which Github didn't allow me to suggest inline -.-):

```diff
diff --git a/INSTALL b/INSTALL
index d95b10c00..6b7fc3c86 100644
--- a/INSTALL
+++ b/INSTALL
@@ -220,9 +220,9 @@ An optional installation of the library and the headers can be performed with
 
    make install [PREFIX=<dir>]
 
-The library will be installed in $(PREFIX)/lib, the headers in
-$(PREFIX)/include, and the configuration makefile (config.mk) in
-$(PREFIX)/share/mfem. The PREFIX option can also be set during configuration.
+The library will be installed in <PREFIX>/lib, the headers in
+<PREFIX>/include, and the configuration makefile (config.mk) in
+<PREFIX>/share/mfem. The PREFIX option can also be set during configuration.
 
 Information about the current build configuration can be viewed using
 
@@ -271,8 +271,8 @@ Build options:
 
 Installation options:
    PREFIX  - Specify the installation directory. The library (libmfem.a) will be
-             installed in $(PREFIX)/lib, the headers in $(PREFIX)/include, and
-             the configuration makefile (config.mk) in $(PREFIX)/share/mfem.
+             installed in <PREFIX>/lib, the headers in <PREFIX>/include, and
+             the configuration makefile (config.mk) in <PREFIX>/share/mfem.
    INSTALL - Specify the install program, default = /usr/bin/install
    INSTALL_DEF_PERM - Specify the default install permissions. This affects
                       headers and configuration makefiles, default = 644
@@ -984,8 +984,9 @@ or
    cmake --build . --config Release --target install  [Xcode]
    cmake --build . --config Release --target INSTALL  [Visual Studio]
 
-The library will be installed in <PREFIX>/lib, the headers in <PREFIX>/include,
-and the configuration CMake files in <PREFIX>/lib/cmake/mfem.
+By default, the library will be installed in <CMAKE_INSTALL_PREFIX>/lib, the
+headers in <CMAKE_INSTALL_PREFIX>/include, and the configuration CMake files in
+<CMAKE_INSTALL_PREFIX>/lib/cmake/mfem.
 
 
 Configuration variables (CMake)
```

------------------------------------
------------------------------------

Now, ideally, I'd prefer if both our variables and GNUInstallDirs worked interchangeably, so the user doesn't have to think too much or doesn't have to enable yet another flag. So we can say "It just works (TM)":

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 436541102..d79e66e5a 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -615,20 +615,13 @@ set(MASTER_HEADERS
   ${PROJECT_SOURCE_DIR}/mfem.hpp
   ${PROJECT_SOURCE_DIR}/mfem-performance.hpp)
 
-# Installation options
-set(MFEM_USE_GNUINSTALLDIRS OFF
-  CACHE BOOL "Enable to use the GNUInstallDirs CMake module.")
-set(install_include_dir_default include)
-set(install_bin_dir_default bin)
-set(install_lib_dir_default lib)
-set(install_share_dir_default share)
-if (MFEM_USE_GNUINSTALLDIRS)
-  include(GNUInstallDirs)
-  set(install_include_dir_default "${CMAKE_INSTALL_INCLUDEDIR}")
-  set(install_bin_dir_default "${CMAKE_INSTALL_BINDIR}")
-  set(install_lib_dir_default "${CMAKE_INSTALL_LIBDIR}")
-  set(install_share_dir_default "${CMAKE_INSTALL_DATAROOTDIR}")
-endif()
+# Installation options (we use GNUInstallDirs but prefer lib on all platforms)
+mfem_cache_path(CMAKE_INSTALL_LIBDIR "lib" "Object code libraries (lib)")
+include(GNUInstallDirs)
+set(install_include_dir_default "${CMAKE_INSTALL_INCLUDEDIR}")
+set(install_bin_dir_default "${CMAKE_INSTALL_BINDIR}")
+set(install_lib_dir_default "${CMAKE_INSTALL_LIBDIR}")
+set(install_share_dir_default "${CMAKE_INSTALL_DATAROOTDIR}")
 mfem_cache_path(INSTALL_INCLUDE_DIR "${install_include_dir_default}"
   "Relative or absolute path for installing header files.")
 mfem_cache_path(INSTALL_BIN_DIR "${install_bin_dir_default}"
diff --git a/INSTALL b/INSTALL
index 6b7fc3c86..5ad01cd1c 100644
--- a/INSTALL
+++ b/INSTALL
@@ -973,8 +973,9 @@ Note that running all the tests may take a while.
 Installation prefix can be configured by setting the standard CMake variable
 CMAKE_INSTALL_PREFIX; for fine-tuning the installation directories the following
 variables can be used: INSTALL_INCLUDE_DIR, INSTALL_LIB_DIR, INSTALL_BIN_DIR,
-INSTALL_SHARE_DIR, and INSTALL_CMAKE_DIR; alternatively, fine-tuning can be done
-by setting MFEM_USE_GNUINSTALLDIRS=ON to use the CMake module GNUInstallDirs.
+INSTALL_SHARE_DIR, and INSTALL_CMAKE_DIR; alternatively, with lower precedence,
+the CMake GNUInstallDirs variables can also be used: CMAKE_INSTALL_INCLUDEDIR,
+CMAKE_INSTALL_LIBDIR, CMAKE_INSTALL_BINDIR and CMAKE_INSTALL_DATAROOTDIR.
 To install the library, use
 
    make install
```

CMakeLists.txt
```diff
@@ -615,7 +615,33 @@ set(MASTER_HEADERS
...
+mfem_cache_path(INSTALL_SHARE_DIR "${install_share_dir_default}"
+  "Relative or absolute path for installing shared data.")
+# other options: "share/mfem/cmake", "lib/mfem/cmake"
+mfem_cache_path(INSTALL_CMAKE_DIR ${INSTALL_LIB_DIR}/cmake/mfem
+  "Relative or absolute path for installing cmake config files.")
+
+mfem_path_to_fullpath("${INSTALL_LIB_DIR}" "${CMAKE_INSTALL_PREFIX}" _lib_path)
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON CACHE BOOL "")
 set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")
 set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE PATH "")
```
helloworld922, line 647:
```suggestion
set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE INTERNAL "" FORCE)
set(CMAKE_INSTALL_NAME_DIR "${_lib_path}" CACHE INTERNAL "" FORCE)
```

Is there anytime these should be different from `${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR}`? Without FORCE (alternative: remove the `CACHE`) changing `CMAKE_INSTALL_PREFIX` won't update these without the user deleting these two entries from the cache.

helloworld922: Similar to this, `MFEM_SOURCE_DIR` and `MFEM_INSTALL_DIR` should either be marked `CACHE INTERNAL FORCE` or just not put into `CACHE` at all.

helloworld922 (changes requested):

CMakeLists.txt
```diff
@@ -825,7 +846,8 @@ foreach(Header mfem.hpp mfem-performance.hpp)
 endforeach()
 install(FILES ${MASTER_HEADERS} DESTINATION ${INSTALL_INCLUDE_DIR}/mfem)
 
-# Install the headers; currently, the miniapps headers are excluded
+# Install the headers. Note that miniapp header installation is dictated from
+# its subdirectory
```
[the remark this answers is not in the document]
v-dobrev: "note common" -> "note that common"

CMakeLists.txt
nmnobre: > Is there anytime these should be different from ${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR}?

Yes, e.g. `INSTALL_LIB_DIR` can hold an absolute path, completely unrelated to `CMAKE_INSTALL_PREFIX`.

v-dobrev:
@nmnobre, your suggested patch for the main `CMakeLists.txt` removes the flag `MFEM_USE_GNUINSTALLDIRS` -- I want to have that flag do what it does in my version -- use the `CMAKE_INSTALL_LIBDIR` default value for `INSTALL_LIB_DIR` when `MFEM_USE_GNUINSTALLDIRS=ON`. This also affects your suggested change in `INSTALL` that removes `MFEM_USE_GNUINSTALLDIRS` -- I suppose it was not clear what this flag was meant to do, so we can add an explanation.

Also, I don't think we want to change `$(PREFIX)`  to `<PREFIX>` in the `INSTALL` file -- I don't think this change makes things clearer.

CMakeLists.txt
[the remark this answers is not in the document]
v-dobrev: I guess a shorter option is "note common" --> "note: common" 😄

nmnobre:
> use the `CMAKE_INSTALL_LIBDIR` default value for `INSTALL_LIB_DIR` when `MFEM_USE_GNUINSTALLDIRS=ON`

What do we gain from that though? :)
I think you see `GNUInstallDirs` as a way of defining some defaults, whereas I see it as a way of getting 'standard' variable aliases?

> Also, I don't think we want to change `$(PREFIX)` to `<PREFIX>` in the `INSTALL` file -- I don't think this change makes things clearer.

There's nowhere else we use the `$(...)` notation in that file, it's always `<...>`, it was just for consistency. Indeed, for GNU Make we used `$(PREFIX)` but for CMake we had `<PREFIX>` (which I changed to `<CMAKE_INSTALL_PREFIX>`).

CMakeLists.txt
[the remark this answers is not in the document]
nmnobre: Rephrased.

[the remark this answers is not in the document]
v-dobrev: I'm fine with both versions (the current and the suggested).

nmnobre:
> The goal is to allow the user to reuse the logic in `GNUInstallDirs` for `CMAKE_INSTALL_LIBDIR` instead having to re-implement that themselves. This logic is not exactly trivial.

I can see that; it's just like this I don't have much motivation to use the functionality (I wonder if others will), because to do the same thing I now need an extra flag. I would switch to `CMAKE_INSTALL_LIBDIR` if it was just as easy, but not otherwise. In my mind, this would become the standard way of doing things, and the previous variables would become as if 'deprecated'. So, if we're adding this just because I suggested it and you don't know any other users who would immediately benefit, we can drop it, no problem. :)

> My reasoning is as follows: `<name>` refers to any string whose contents is described by "name" and "name" is typically not the name of a variable, e.g. we use `<mfem-source-dir>`; on the other hand `$(PREFIX)` refers specifically to the contents of the variable `PREFIX` defined in the GNU make build system (typically on the command line or in `config/user.mk`). In the context of CMake, `<PREFIX>` was mean to be "the path that specifies the install prefix" which is given by `${CMAKE_INSTALL_PREFIX}`. Does this clarify the distinction I'm making? We don't need to enforce this but this is my reasoning for using `<name>` vs using `$(varname)` in GNU make context vs using `${varname}` in CMake context.

Okay, I understand that now, I think I might've been misled by the capitalisation and that for GNU Make we say `PREFIX=<dir>`; I now understand that `<dir>` for GNU Make and `<PREFIX>` for CMake  are equivalent in your logic. :)

CMakeLists.txt
lindsayad: > Regarding MFEM_SOURCE_DIR and MFEM_INSTALL_DIR: I agree, I don't think there's a reason for them to not be made INTERNAL or just not stored in the cache.

You want me to do this here in this PR? I'm a little sad that I did not just use the existing variables in the beginning to accomplish the goal of this PR, which is installing the common headers and library. This PR has really exploded in terms of discussion beyond the scope I intended

v-dobrev:
> So, if we're adding this just because I suggested it and you don't know any other users who would immediately benefit, we can drop it, no problem. :)

The use `GNUInstallDirs` has come up on other occasions before you brought it up, so I think having the option to use `GNUInstallDirs` will be useful to others. The price for users to pay is that they'll need to enable `MFEM_USE_GNUINSTALLDIRS` which is much simpler than reproducing the logic to set `CMAKE_INSTALL_LIBDIR` (or `INSTALL_LIB_DIR`) the same way as the default in `GNUInstallDirs`.

All I'm suggesting is to make the following line in your suggestion:
```cmake
mfem_cache_path(CMAKE_INSTALL_LIBDIR "lib" "Object code libraries (lib)")
```
optional putting it in an `if` statement:
```cmake
if(NOT MFEM_USE_GNUINSTALLDIRS)
  mfem_cache_path(CMAKE_INSTALL_LIBDIR "lib" "Object code libraries (lib)")
endif()
```
Also, since `GNUInstallDirs` is always included in your suggestion, we can get rid of all the `install_*_default` variables to simplify the code a bit.

CMakeLists.txt
v-dobrev: @lindsayad, I apologize, you are right -- the scope of the changes that we (the reviewers) started to ask for grew far beyond the original scope.

In order to minimize any additional effort on your side, I can take care of adding any final changes, if that is okay with you? Since the PR branch is in a fork, you may need to give me permission to write to your brach -- there should be a checkbox for that in the right panel of the PR page.

lindsayad: that works great for me, thanks @v-dobrev. Yes, as a maintainer myself, I always grant access to maintainers 😄

nmnobre:
> ```cmake
> if(NOT MFEM_USE_GNUINSTALLDIRS)
>   mfem_cache_path(CMAKE_INSTALL_LIBDIR "lib" "Object code libraries (lib)")
> endif()
> ```
> 
> Also, since `GNUInstallDirs` is always included in your suggestion, we can get rid of all the `install_*_default` variables to simplify the code a bit.

I love this, if you're happy, I'm (very) happy!

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 32cdcd01d..cd828def7 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -615,27 +615,20 @@ set(MASTER_HEADERS
   ${PROJECT_SOURCE_DIR}/mfem.hpp
   ${PROJECT_SOURCE_DIR}/mfem-performance.hpp)
 
-# Installation options
+# Installation options (we use GNUInstallDirs but prefer lib by default)
 set(MFEM_USE_GNUINSTALLDIRS OFF
-  CACHE BOOL "Enable to use the GNUInstallDirs CMake module.")
-set(install_include_dir_default include)
-set(install_bin_dir_default bin)
-set(install_lib_dir_default lib)
-set(install_share_dir_default share)
-if (MFEM_USE_GNUINSTALLDIRS)
-  include(GNUInstallDirs)
-  set(install_include_dir_default "${CMAKE_INSTALL_INCLUDEDIR}")
-  set(install_bin_dir_default "${CMAKE_INSTALL_BINDIR}")
-  set(install_lib_dir_default "${CMAKE_INSTALL_LIBDIR}")
-  set(install_share_dir_default "${CMAKE_INSTALL_DATAROOTDIR}")
-endif()
-mfem_cache_path(INSTALL_INCLUDE_DIR "${install_include_dir_default}"
+  CACHE BOOL "Enable to use the GNUInstallDirs CMake module defaults.")
+if(NOT MFEM_USE_GNUINSTALLDIRS)
+  mfem_cache_path(CMAKE_INSTALL_LIBDIR "lib" "Object code libraries (lib)")
+endif()
+include(GNUInstallDirs)
+mfem_cache_path(INSTALL_INCLUDE_DIR "${CMAKE_INSTALL_INCLUDEDIR}"
   "Relative or absolute path for installing header files.")
-mfem_cache_path(INSTALL_BIN_DIR "${install_bin_dir_default}"
+mfem_cache_path(INSTALL_BIN_DIR "${CMAKE_INSTALL_BINDIR}"
   "Relative or absolute path for installing the binaries.")
-mfem_cache_path(INSTALL_LIB_DIR "${install_lib_dir_default}"
+mfem_cache_path(INSTALL_LIB_DIR "${CMAKE_INSTALL_LIBDIR}"
   "Relative or absolute path for installing the library.")
-mfem_cache_path(INSTALL_SHARE_DIR "${install_share_dir_default}"
+mfem_cache_path(INSTALL_SHARE_DIR "${CMAKE_INSTALL_DATAROOTDIR}"
   "Relative or absolute path for installing shared data.")
 # other options: "share/mfem/cmake", "lib/mfem/cmake"
 mfem_cache_path(INSTALL_CMAKE_DIR "${INSTALL_LIB_DIR}/cmake/mfem"
diff --git a/INSTALL b/INSTALL
index d95b10c00..96c0f17f6 100644
--- a/INSTALL
+++ b/INSTALL
@@ -973,8 +973,13 @@ Note that running all the tests may take a while.
 Installation prefix can be configured by setting the standard CMake variable
 CMAKE_INSTALL_PREFIX; for fine-tuning the installation directories the following
 variables can be used: INSTALL_INCLUDE_DIR, INSTALL_LIB_DIR, INSTALL_BIN_DIR,
-INSTALL_SHARE_DIR, and INSTALL_CMAKE_DIR; alternatively, fine-tuning can be done
-by setting MFEM_USE_GNUINSTALLDIRS=ON to use the CMake module GNUInstallDirs.
+INSTALL_SHARE_DIR, and INSTALL_CMAKE_DIR; alternatively, with lower precedence,
+the CMake GNUInstallDirs variables can also be used: CMAKE_INSTALL_INCLUDEDIR,
+CMAKE_INSTALL_LIBDIR, CMAKE_INSTALL_BINDIR and CMAKE_INSTALL_DATAROOTDIR.
+Note that the default for INSTALL_LIB_DIR and CMAKE_INSTALL_LIBDIR is lib. This
+is in contrast to the default behavior for GNUInstallDirs, which defines a
+platform-dependent default value for CMAKE_INSTALL_LIBDIR (lib or lib64). To
+restore this behavior, the user can set MFEM_USE_GNUINSTALLDIRS=YES.
 To install the library, use
 
    make install
```

nmnobre:
> My reasoning is as follows: `<name>` refers to any string whose contents is described by "name" and "name" is typically not the name of a variable, e.g. we use `<mfem-source-dir>`; on the other hand `$(PREFIX)` refers specifically to the contents of the variable `PREFIX` defined in the GNU make build system (typically on the command line or in `config/user.mk`). In the context of CMake, `<PREFIX>` was mean to be "the path that specifies the install prefix" which is given by `${CMAKE_INSTALL_PREFIX}`. Does this clarify the distinction I'm making? We don't need to enforce this but this is my reasoning for using `<name>` vs using `$(varname)` in GNU make context vs using `${varname}` in CMake context.

Here I'd suggest the following since the `${...}` is (luckily!) valid for both GNU make and CMake:

```diff
diff --git a/INSTALL b/INSTALL
index 96c0f17f6..80e642a7d 100644
--- a/INSTALL
+++ b/INSTALL
@@ -220,9 +220,9 @@ An optional installation of the library and the headers can be performed with
 
    make install [PREFIX=<dir>]
 
-The library will be installed in $(PREFIX)/lib, the headers in
-$(PREFIX)/include, and the configuration makefile (config.mk) in
-$(PREFIX)/share/mfem. The PREFIX option can also be set during configuration.
+The library will be installed in ${PREFIX}/lib, the headers in
+${PREFIX}/include, and the configuration makefile (config.mk) in
+${PREFIX}/share/mfem. The PREFIX option can also be set during configuration.
 
 Information about the current build configuration can be viewed using
 
@@ -271,8 +271,8 @@ Build options:
 
 Installation options:
    PREFIX  - Specify the installation directory. The library (libmfem.a) will be
-             installed in $(PREFIX)/lib, the headers in $(PREFIX)/include, and
-             the configuration makefile (config.mk) in $(PREFIX)/share/mfem.
+             installed in ${PREFIX}/lib, the headers in ${PREFIX}/include, and
+             the configuration makefile (config.mk) in ${PREFIX}/share/mfem.
    INSTALL - Specify the install program, default = /usr/bin/install
    INSTALL_DEF_PERM - Specify the default install permissions. This affects
                       headers and configuration makefiles, default = 644
@@ -989,8 +989,9 @@ or
    cmake --build . --config Release --target install  [Xcode]
    cmake --build . --config Release --target INSTALL  [Visual Studio]
 
-The library will be installed in <PREFIX>/lib, the headers in <PREFIX>/include,
-and the configuration CMake files in <PREFIX>/lib/cmake/mfem.
+By default, the library will be installed in ${CMAKE_INSTALL_PREFIX}/lib, the
+headers in ${CMAKE_INSTALL_PREFIX}/include, and the configuration CMake files
+in ${CMAKE_INSTALL_PREFIX}/lib/cmake/mfem.
 
 
 Configuration variables (CMake)
```

v-dobrev:
@nmnobre, @helloworld922, I think I added all changes based on the discussions above and a few other small changes. Let me know if I missed something or if you notice anything else worth addressing here.

CMakeLists.txt
helloworld922: If we want these to be optionally user configurable, then I think we should use this:

```suggestion
if(NOT DEFINED CMAKE_INSTALL_RPATH)
  set(CMAKE_INSTALL_RPATH "${_lib_path}")
endif()
if(NOT DEFINED CMAKE_INSTALL_NAME_DIR)
  set(CMAKE_INSTALL_NAME_DIR "${_lib_path}")
endif()
```

This way, if the user doesn't manually specify one or both of these CMake will automatically track them to `${_lib_path}`, and if the user does set them CMake will use the user-defined values.

One time I can think it makes sense to not use `${_lib_path}` is to allow the installed binary (and shared libmfem) to be relocated after install (say as part of an installer or dependency of another package), in which case maybe it makes sense to default these to `$ORIGIN/../${INSTALL_LIB_DIR}`?

v-dobrev: > This way, if the user doesn't manually specify one or both of these CMake will automatically track them to `${_lib_path}`, and if the user does set them CMake will use the user-defined values.

Value set on the command line with `-DCMAKE_INSTALL_RPATH=...` are cached and will not be overwritten by `set(CMAKE_INSTALL_RPATH "${_lib_path}" CACHE PATH "")` -- this way we do not need to explicitly check if `CMAKE_INSTALL_RPATH` is defined. This will also work if one adds their own setting to `config/user.cmake` with `set(CMAKE_INSTALL_RPATH "${my_rpath}" CACHE PATH "")`. Checking if `CMAKE_INSTALL_RPATH` is defined makes sense only if we want to support the case where `CMAKE_INSTALL_RPATH` is defined as a non-cache variable. Is that the reason you want to have the explicit `if (NOT DEFINED ...)` check?

> One time I can think it makes sense to not use `${_lib_path}` is to allow the installed binary (and shared libmfem) to be relocated after install (say as part of an installer or dependency of another package), in which case maybe it makes sense to default these to `$ORIGIN/../${INSTALL_LIB_DIR}`?

Yes, I think this may be one use case. Another case could be when a user wants to add more paths to the rpath for some reason.

mlstowell (approved):

CMakeLists.txt
nmnobre: @v-dobrev, I think @helloworld922 meant that if you run CMake once with a given `CMAKE_INSTALL_PATH` and then a subsequent time (without deleting the contents of the directory you're running CMake from) with a different `CMAKE_INSTALL_PATH`, then these two variables will not reflect that change and will be stuck with the path computed during the first CMake pass.

helloworld922: Yes, the situation I was envisioning is if you call cmake initially then later call `cmake -DCMAKE_INSTALL_PREFIX=...` (or use `ccmake`/`cmake-gui`to modify these) it won't reflect these changes in `CMAKE_INSTALL_RPATH` and `CMAKE_INSTALL_NAMED_DIR` even if the user never intended to set them initially.

v-dobrev: Okay, I see. I'm not sure we fully support re-configuration with CMake as you describe it. I know it is something CMake wants to support but it typically creates complications such as what you describe.

I never trust re-configuration (my impression is that most projects, including mfem, have issues with that), so I always delete everything in the CMake build directory before running `cmake` again.

That said, what you suggest seems a step in the right direction to support re-configuration, so let's add it -- I'll commit your suggestion.

nmnobre:
> @nmnobre, @helloworld922, I think I added all changes based on the discussions above and a few other small changes. Let me know if I missed something or if you notice anything else worth addressing here.

I think this is perfect! 🤩
Wasn't the easiest, but we got there in the end. Thanks for the patience/perseverance everybody.
I'd just squash to tidy it up a bit now.

helloworld922 (approved):

helloworld922:
Tested it out and everything appears to be configuring and building correctly.

tzanio:

tzanio (approved):
Thanks for the useful contribution @lindsayad! 👍 

@nmnobre, @v-dobrev, + @helloworld922, @mlstowell and @jandrej -- thank you for the feedback and the careful review! 🙏

lindsayad:
My contribution ended up just being the miniapp install idea ... I doubt any of my LOC ever made it in. But thank you 😄 Thanks @v-dobrev for getting it over the finish line

PR #4597, opened 2024-11-22
Move VectorDim, CurlDim to FiniteElementSpace

phdum-a:
Dear all,

[This is my first MFEM PR. I am in the same group as @hughcars and now working on code using MFEM].

This is a mini-PR that just moves existing functions. Specifically, `VectorDim` and `CurlDim` are defined in `GridFunction`, but only depend on the underlying `FiniteElementSpace`. We want to pre-allocate some memory that the depends on the actual vector size and where we know the FES, but not the gridfunction yet. 

Would it be ok to move those functions to `FiniteElementSpace`?

Two small points:
1. I've renamed these functions `GetVectorDim` and `GetCurlDim` to match the style of similar functions in `FiniteElementSpace`, but left them as is in `GridFunction` for backwards compatibility. Is this ok?
2. I've added short docstrings that were missing, so could you please check these are understandable? Especially to distinguish this function from `GetVDim`.

Thanks!

### DIFF
--- a/fem/fespace.cpp
+++ b/fem/fespace.cpp
@@ -1358,6 +1358,46 @@ int FiniteElementSpace::GetNConformingDofs() const
    return P ? (P->Width() / vdim) : ndofs;
 }
 
+int FiniteElementSpace::GetVectorDim() const
+{
+   const FiniteElement *fe;
+   if (!GetNE())
+   {
+      static const Geometry::Type geoms[3] =
+      { Geometry::SEGMENT, Geometry::TRIANGLE, Geometry::TETRAHEDRON };
+      fe = FEColl()->FiniteElementForGeometry(geoms[GetMesh()->Dimension()-1]);
+   }
+   else
+   {
+      fe = GetFE(0);
+   }
+   if (!fe || fe->GetRangeType() == FiniteElement::SCALAR)
+   {
+      return GetVDim();
+   }
+   return GetVDim()*std::max(GetMesh()->SpaceDimension(), fe->GetRangeDim());
+}
+
+int FiniteElementSpace::GetCurlDim() const
+{
+   const FiniteElement *fe;
+   if (!GetNE())
+   {
+      static const Geometry::Type geoms[3] =
+      { Geometry::SEGMENT, Geometry::TRIANGLE, Geometry::TETRAHEDRON };
+      fe = FEColl()->FiniteElementForGeometry(geoms[GetMesh()->Dimension()-1]);
+   }
+   else
+   {
+      fe = GetFE(0);
+   }
+   if (!fe || fe->GetRangeType() == FiniteElement::SCALAR)
+   {
+      return 2 * GetMesh()->SpaceDimension() - 3;
+   }
+   return GetVDim()*fe->GetCurlDim();
+}
+
 const ElementRestrictionOperator *FiniteElementSpace::GetElementRestriction(
    ElementDofOrdering e_ordering) const
 {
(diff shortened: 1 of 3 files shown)


phdum-a:
Hi @tzanio — thank you very much! I've now joined the MFEM organization. I will close this PR and reopen one from an mfem dev branch, as requested.

fem/fespace.hpp
```diff
@@ -427,6 +427,7 @@ class FiniteElementSpace
    /// Replicate 'mat' in the vector dimension, according to vdim ordering mode.
    void MakeVDimMatrix(SparseMatrix &mat) const;
 
+
```
pazner, line 430:
```suggestion
```

fem/fespace.hpp
```diff
@@ -748,6 +749,13 @@ class FiniteElementSpace
 
    int GetConformingVSize() const { return vdim * GetNConformingDofs(); }
 
+   /// Return the dimension of an vector in the space, including if elements
+   /// are vector-valued
```
pazner, line 753: This could use a bit more explanation. If I understand correctly, the finite elements themselves may be vector-valued (e.g. $H(\mathrm{div})$ or $`H(\mathrm{curl})`$), say with $n$ components, and the finite element space may have a "vdim", i.e. we may be on the space $[V_h]^m$ which is like $m$ copies of $V_h$. `GetVectorDim()` will return in this case $n \times m$. Is that right?

Maybe this can be spelled out explicitly and mention in the docs why this is different from `FiniteElementSpace::GetVDim`. Maybe the documentation for `FiniteElementSpace::GetVDim` should also be adjusted.

fem/fespace.hpp
```diff
@@ -748,6 +749,13 @@ class FiniteElementSpace
...
+   int GetVectorDim() const;
+
+   /// Return dimension of curl
```
pazner, line 756:
```suggestion
   /// Return the dimension of the curl of a GridFunction defined on this space.
```

fem/gridfunc.cpp
```diff
@@ -324,45 +324,12 @@ void GridFunction::ComputeFlux(BilinearFormIntegrator &blfi,
 
 int GridFunction::VectorDim() const
```
pazner, line 325: Can you add Doxygen comments for this and for `GridFunction::CurlDim` (something like "Returns FiniteElementSpace::GetVectorDim()" would be fine).

fem/fespace.cpp
```diff
@@ -1358,6 +1358,46 @@ int FiniteElementSpace::GetNConformingDofs() const
...
+   if (!GetNE())
+   {
+      static const Geometry::Type geoms[3] =
+      { Geometry::SEGMENT, Geometry::TRIANGLE, Geometry::TETRAHEDRON };
+      fe = FEColl()->FiniteElementForGeometry(geoms[GetMesh()->Dimension()-1]);
+   }
+   else
+   {
+      fe = GetFE(0);
+   }
```
pazner, line 1373: I believe this logic (and similarly in `GetCurlDim`) could be simplified using PR #4260. @v-dobrev, should we try to move that PR forward?

pazner (commented):
I think this is a good change, thanks @phdum-a. Some minor comments...

fem/fespace.hpp
phdum-a: Done in #4598

fem/gridfunc.cpp
phdum-a: Done in #4598

fem/fespace.hpp
phdum-a: Done in #4598

fem/fespace.hpp
phdum-a: See https://github.com/mfem/mfem/pull/4598#issuecomment-2524254283

fem/fespace.cpp
phdum-a: Moved to https://github.com/mfem/mfem/pull/4598#issuecomment-2524280095

PR #4629, opened 2024-12-12
Handles/smart pointers

pazner:
Based on the discussion in #4614, this is a proposed PR to introduce a `Handle<T>` smart pointer class. This class models "maybe-owning" semantics, where the handle can either represent shared ownership (using `std::shared_ptr`) or a non-owning pointer.

As an illustration, `ProductOperator`, `SumOperator`, and `TripleProductOperator` have been changed to use `Handle` instead of raw pointers. The interface is completely backwards-compatible (no other library code or user code needs to be changed). Because of these changes, those operators get proper move and copy semantics for free, which is a nice quality-of-life improvement.


general/handle.hpp
```diff
@@ -0,0 +1,200 @@
...
+/// It is an invariant of this class that **at most** one of the data members
+/// @a not_owned and @a owned will be non-null.
+template <typename T>
+class Handle
+{
+   /// If this is a non-owning handle, @a not_owned will point to the data.
+   T *not_owned = nullptr;
+
+   /// If this is an owning handle, @a owned will point to the data.
+   std::shared_ptr<T> owned = nullptr;
```
tepperly, line 48:
```suggestion
   std::shared_ptr<T> owned;
```
Initialization to `nullptr` is implicit.

general/handle.hpp
```diff
@@ -0,0 +1,200 @@
...
+   ///
+   /// If @a take_ownership is true, then the Handle assumes ownership over the
+   /// pointer, and it should not be deleted externally. Otherwise, the Handle
+   /// will be non-owning, and it is the user's responsibility to ensure the
+   /// correct lifetime of @a t.
+   Handle(T *t, bool take_ownership)
+   {
+      if (take_ownership) { owned.reset(t); }
+      else { not_owned = t; }
+   }
```
tepperly, line 69: What about basing whether it's owned by the type of the argument:
```suggestion
   /// @brief Create a Handle pointing to non-owned @a t.
   ///
   ///  Create an non-owning Handle; it is the user's responsibility to ensure the
   /// correct lifetime of @a t.
   Handle(T *t) : not_owned(t)
   {
   }
   
      /// @brief Create a Handle pointing to non-owned @a t.
   ///
   ///  Create an owning Handle to @a t.
   Handle(::std::unique_ptr<T> t) : owned(t)
   {
   }
```

general/handle.hpp
```diff
@@ -0,0 +1,200 @@
...
+   {
+      owned.reset();
+      not_owned = nullptr;
+   }
+
+   /// @brief Reset the Handle to point to @a t.
+   ///
+   /// The Handle may assume ownership of the pointer according to @a
+   /// take_ownership (see @ref Handle(T*, bool)).
+   void Reset(T *t, bool take_ownership)
```
tepperly, line 155: Suggest similar change to `T *t` and `std::unique_ptr<T> t`.

general/handle.hpp
pazner: The `Handle(std::unique_ptr<T> t)` overload could be useful (e.g. to create an owning handle directly from the return value of a function that returns `unique_ptr`), but does making it the only mechanism offer a concrete benefit? The type is essentially playing the role of a flag, since the `unique_ptr` is never used as such, it's just immediately released.

tepperly: For me, the meaning of a `T *t` argument is very ambiguous, and I can't know what is intended without reading the comments and/or the implementation. A raw pointer is ambiguous about ownership, and it's ambiguous about size. Is it a pointer to zero or one object, or is it the start of an array of objects. `std::unique_ptr` is clear about ownership transfer, and array or not.
If you always use `std::unique_ptr` for ownership transfer, users can infer that non-smart raw pointers indicates a borrowed reference.

pazner: I'm not sure the "all ownership transfers happen via `unique_ptr`" rule can be followed closely enough so that the inference that all raw pointers are non-owning borrows can be made reliably (and in any case, non-owning borrows can be modeled with references instead of pointers).

For example, the `shared_ptr::shared_ptr(T*)` and `shared_ptr::reset(T*)` assume ownership of the passed pointer. I'm really not sure that `shared_ptr::shared_ptr(unique_ptr<T>)` and `shared_ptr::reset(unique_ptr<T>)` would be a better interface (it introduces a certain amount of "ceremony" with IMO little benefit in clarity). But since we use the standard library (and many external libraries), we can't reliably assume that all uses of raw pointers are non-owning.

tepperly: I totally agree about using reference arguments instead of pointers when `nullptr` isn't an option.

My impression as a user is that inconsistent pointer use is one of the challenging things about using MFEM. It's pretty common for me to have to read a method's implementation to figure out how the pointer is being used. With other C/C++ libraries that I use, it's not common for me to have to read method implementations to figure things out. This is probably due to a combination of documentation and consistent interface design.

pazner: I agree completely. Inconsistent pointer usage is a pain point when using MFEM, and users shouldn't have to read the implementation to understand the ownership semantics (but often do). We should work towards addressing both of these issues in the next major release of MFEM.

PR #4660, opened 2025-01-09
Reduction framework

helloworld922:
Implements a general reduction framework and updates vector reduction operations like dot, min, max, sum, and norm to use this framework.

Users need to provide two functors: a reduction body and a reducer.
- The reduction body computes and updates a partial reduction `f(index, partial)`. For example, a dot product reduction body looks like `partial += v0[index] * v1[index];`.
- The reducer provides a `join` function for combining two partial results and `init_val` function for the default empty range reduction. A variety of built-in reducers can be found in `general/reducers.hpp`.

For right now `mfem::reduce` also requires as input a workspace buffer `Array` to be passed in. This will be resized as needed for the reduction implementation. This allows the workspace to be re-used for future reductions. In the future this should be replaced with an internal temporary buffer created using #4065.

This PR not only makes writing future reducers easier, but it improves the performance of existing reducers. Example throughput improvements of vector dot products on Lassen (Nvidia V100) and Tuolumne (AMD Mi300A):

lassen_dot_throughput
tuo_dot_throughput

| [#4660](https://github.com/mfem/mfem/pull/4660) | @helloworld922 | @tzanio | @pazner + @jandrej + @camierjs | 1/15/25 | 3/27/25 | 3/29/25 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


general/reducers.hpp
```diff
@@ -0,0 +1,1169 @@
...
+   {
+      if (b < a)
+      {
+         a = b;
+      }
+   }
+
+   MFEM_HOST_DEVICE void init_val(uint64_t &a) const
+   {
+      a = uint64_t(0xffffffffffffffffull);
```
jandrej, line 255: I think you can use `std::numeric_limits<T>::max()` etc. as they're defined as ```static constexpr T max() noexcept``` since C++11.

helloworld922: I changed these for the macros in `<climits>` since nvcc by default doesn't allow calling host-only constexpr functions in device code. I think adding the `--expt-relaxed-constexpr` flag should allow `std::numeric_limits<T>::max()` to compile and run, though it may be a surprise to user codes that this flag would be required if they just include `forall.hpp`.

general/reducers.hpp
```diff
@@ -0,0 +1,395 @@
...
+          a = b;                                                               \
+        }                                                                      \
+      }                                                                        \
+      MFEM_HOST_DEVICE void init_val(value_type &a) const {                    \
+        a = static_cast<type>(val);                                            \
+      }                                                                        \
+   }
+
+MFEM_STAMP_MIN_REDUCER(bool, true);
+MFEM_STAMP_MIN_REDUCER(char, CHAR_MAX);
```
jandrej, line 151: Why do we instantiate those explicitly? The symbols are now generated for each file that includes the reducer header, right?

helloworld922: I define a temporary macro so I don't have to copy-paste the explicit/partial specialization, so it does need to be instantiated per type.

tzanio:

general/reducers.hpp
```diff
@@ -0,0 +1,395 @@
...
+   static_assert(std::is_integral<T>::value, "Only works for integral types");
+   using value_type = T;
+   MFEM_HOST_DEVICE void join(value_type &a, const value_type &b) const
+   {
+      a &= b;
+   }
+
+   MFEM_HOST_DEVICE void init_val(value_type &a) const
+   {
+      // hopefully this will set all bits in a
```
camierjs, line 76: For `unsigned` integral types, we could use the bitwise NOT operator, 
for signed integral types, maybe we can comment that we are 'assuming two's complement' ?

general/reducers.hpp
camierjs: What does the `STAMP` stand for?

camierjs (commented):
Nice addition @helloworld922 for the reductions!

Would it be able to push a few numbers in the PR's description to motivate these changes ? That would be useful for comparison, even for later.

There are a lot new reducers with new associated types: it would be useful to have some non-regression tests to go over all of them: I can help if needed.

general/reducers.hpp
helloworld922: These are temporary macros which allow me to insert in full/partial template specializations for various primitive types, and I then use this macro to fully/partially specialize the reducer for all standard C++ primitive types. They are undefined after use so users/other developers should never see them outside of the specific regions in this file.

The reason to not leave a callable macro is the potential to end up with some sort of ODR violation.

general/reducers.hpp
helloworld922: I changed this to `~T(0)`, this is equivalent for two's complement representations, and I think it also handles the case just in case any other kind of binary representation is used.

helloworld922:
> Nice addition @helloworld922 for the reductions!
> 
> Would it be able to push a few numbers in the PR's description to motivate these changes ? That would be useful for comparison, even for later.
> 
> There are a lot new reducers with new associated types: it would be useful to have some non-regression tests to go over all of them: I can help if needed.

Added performance throughput numbers comparing the current MFEM master, this reduction framework, RAJAperf's implementations of dot product, as well as cublas/hipblas.

I can add unit tests for the reducers, though I'm a little unsure how the unit testing framework is setup, especially for GPU tests. Perhaps we can discuss this in a meeting/slack?

helloworld922:
@camierjs @jandrej @pazner I have all the reducers tested with CPU/CUDA with at least one built-in type so the PR is ready for another look now. I will add HIP/RAJA related testing in a different PR.

linalg/vector.cpp
```diff
@@ -28,6 +28,209 @@
...
+{
+   real_t res = infinity();
+   reduce(
+      size, res,
+   [=] MFEM_HOST_DEVICE(int i, real_t &r) { r = fmin(r, m_data[i]); },
+   MinReducer<real_t> {}, true, vector_workspace());
+   return res;
+}
+
+static real_t devVectorMax(int size, const real_t *m_data)
```
pazner, line 119: Maybe it would be cleaner if this was a lambda inside `Vector::Max`. It seems like the only reason to factor it out as a function is because it's called in both the CUDA and HIP cases.

Maybe we want a generic "using accelerator" check, and if that's true, call the "dev" version, rather than separate CUDA and HIP checks.

general/forall.hpp
```diff
@@ -783,6 +786,111 @@ inline void forall_3D_grid(int N, int X, int Y, int Z, int G, lambda &&body)
...
+ * @a bx x block width
+ * @a by y block width
+ * @a bz z block width
+ * @a smem_bytes amount of dynamic shared memory in bytes
+ * @a h_body host body
+ * @a d_body device body
+ * @tparam T pointer type for dynamic shared memory buffer
+ */
+template <class h_lambda, class d_lambda>
+inline void forall_smem(bool use_dev, int nx, int ny, int nz, int bx, int by,
```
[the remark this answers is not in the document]
helloworld922: The semantics of this function are a little different from the existing forall calls. This function is designed to be as close to raw cuda/hip as possible where your lambda body is entirely responsible for a single block (including bounds checking), while the forall<x>d functions have a 1D loop bound which is bounds checked.

This is beneficial for writing some GPU-specific algorithms like reduction and scan, but the lambda body almost certainly won't run correctly on the CPU since it relies on having access to blockIdx/threadIdx/blockDim/gridDim/etc.

I wasn't really sure what approach to take for this function since reduction at least has a separate host code path which skips calling this forall_smem entirely. Perhaps trying to call `forall_smem` with a CPU/Debug device should fail loudly at runtime since almost certainly this is expected to be a bug?

general/reducers.hpp
```diff
@@ -0,0 +1,400 @@
...
+   using value_type = double;
+   MFEM_HOST_DEVICE void join(value_type &a, value_type b) const
+   {
+      a = fmin(a, b);
+   }
+
+   MFEM_HOST_DEVICE void init_val(value_type &a) const { a = HUGE_VAL; }
+};
+
+#define MFEM_STAMP_MIN_REDUCER(type, val)                                      \
```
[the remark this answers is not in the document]
helloworld922: floating point types can't be template parameters before C++20

general/forall.hpp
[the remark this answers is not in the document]
pazner: Can you do something similar to what's done here, using MFEM's `MFEM_THREAD_ID` and `MFEM_THREAD_SIZE` macros? This reduction kernel works on device and CPU

https://github.com/mfem/mfem/blob/8447e6273535511f75da133c34c01394d2c88661/fem/dgmassinv_kernels.hpp#L122-L173

general/reducers.hpp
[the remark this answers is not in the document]
helloworld922: fmin/fmax has slightly different behavior when encountering nan's vs. the comparison/assign method. If one argument is nan, it returns the other value.

In theory fmin/fmax should perform equivalent or better, as they typically map to branchless intrinsics which the comparison/assign method doesn't (possibly can't because of different behavior involving nan?). In practice, when I tried to measure this on Lassen/Tuolumne I couldn't find any appreciable difference in performance.

Using `std::numeric_limits<T>::max()` or any other constexpr host-only function with Cuda technically requires an extra compiler flag (`--expt-relaxed-constexpr`) which we currently don't set. Newer versions of Cuda seem to just give a warning if this flag is missing, but some versions do issue a compiler error. We could enable this compiler flag, though I avoided this to minimize potential unintended impacts on other codes.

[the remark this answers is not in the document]
pazner: Thanks, that's useful information.

Good point about host-only function. I agree we don't necessarily want to force applications to use a specific compiler flag. Another option is something like this:

```cpp
template <typename T, T VAL>
struct GenericVal
{
   MFEM_HOST_DEVICE static constexpr T Get() { return VAL; }
};

template <typename T>
struct MaxVal : GenericVal<T, std::numeric_limits<T>::max()> { };

template <>
struct MaxVal<float> { MFEM_HOST_DEVICE static constexpr float Get() { return HUGE_VALF; } };

template <>
struct MaxVal<double> { MFEM_HOST_DEVICE static constexpr double Get() { return HUGE_VAL; } };
```

Then the generic version should work on device

```cpp
template <typename T>
struct MinReducer
{
   MFEM_HOST_DEVICE void join(T &a, T b) const
   {
      if (b < a)
      {
         a = b;
      }
   }
   MFEM_HOST_DEVICE void init_val(T &a) const
   {
      a = MaxVal<T>::Get();
   }
};
```

general/forall.hpp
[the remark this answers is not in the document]
helloworld922: I re-worked it into a fixed size shared memory buffer and called the standard forall_2D. I was unable to measure any significant performance differences. Perhaps we should have a separate discussion about a proper dynamic shared memory size interface for forall which can gracefully handle host code?

general/reducers.hpp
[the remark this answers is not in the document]
helloworld922: I ended up for integral types using `std::integral_constant<T, std::numeric_limits<T>::min()>::value` to circumvent the calling of a host-only constexpr function in a cuda kernel, and manually specializing on floating point types.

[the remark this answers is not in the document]
pazner: Cool :+1: TIL about `std::integral_value`

I think it's better now that we don't need to explicitly instantiate all the types using a macro

general/reducers.hpp
```diff
@@ -0,0 +1,436 @@
...
+#include <cmath>
+#include <cstdint>
+#include <limits>
+#include <type_traits>
+
+namespace mfem
+{
+
+/// Pair of values which can be used in device code
+template <class A, class B> struct DevicePair
```
jandrej, line 27: Nitpick but do we need to call this `DevicePair`? I think we can just call it `Pair`. Or can we even use `std::pair`? I don't see anything obvious that restricts device usage.

pazner: `std::pair` won't work in CUDA code. I think the analogy here is to `DeviceTensor`.

jandrej: Whose naming I despise as well 😮‍💨

pazner: I think renaming to `Pair` and documenting that it is device-compatible (e.g. if any member functions are added, they should be `MFEM_HOST_DEVICE`) could be a good idea 👍

helloworld922: There's already an existing `mfem::Pair` class (`general/sort_pairs.hpp`). It has some strange quirks, though, for example having a default less than comparison operator which only compares the `one` field.

helloworld922: I think if we enable `--expt-relaxed-constexpr` and switch to C++20 we can use `std::pair` in device code.

pazner: Good point, but not likely in the near term. Personally I'm fine with `DevicePair`, but maybe @jandrej has another idea.

jandrej: Keep as is for now. Thanks.

jandrej (approved):

tzanio:
Waiting for @camierjs to take another look and for @pazner to review

general/forall.hpp
```diff
@@ -850,6 +853,163 @@ inline MemoryClass GetHypreForallMemoryClass()
...
+            Array<T> &workspace)
+{
+   if (N == 0)
+   {
+      return;
+   }
+
+   if (!use_dev)
+   {
+      goto backend_cpu;
```
pazner, line 956: I know this pattern is used elsewhere, but probably better to avoid the `goto`, for example by changing the next if statement to `if (use_dev && mfem::Device::Allows(…`.

general/mem_manager.cpp
```diff
@@ -17,6 +17,7 @@
 #include <unordered_map>
 #include <algorithm> // std::max
 #include <cstdint>
+#include <memory>
```
pazner, line 20: Not needed anymore (without the addition of dynamic shared memory)

general/mem_manager.hpp
```diff
@@ -1346,7 +1346,6 @@ inline int Memory<T>::CompareHostAndDevice(int size) const
    return MemoryManager::CompareHostAndDevice_(h_ptr, size*sizeof(T), flags);
 }
 
-
```
pazner, line 1349: Can be reverted

linalg/vector.cpp
```diff
@@ -855,37 +934,148 @@ real_t Vector::Norml2() const
...
    // argument of each call to std::pow is <= 1 to avoid overflow.
-   if (0 == size)
+   if (size == 0)
    {
       return 0.0;
-   } // end if 0 == size
+   }
 
-   data.Read(MemoryClass::HOST, size);
+   if (UseDevice())
```
pazner, line 942: For this and other reduction functions, since your `reduce` function has a CPU version, can we just have a single-source implementation, rather than specialized CPU fallbacks? Is there a performance concern or some other consideration?

pazner (commented):
Really nice work @helloworld922, just a couple comments, but I think this is really close to being ready

linalg/vector.cpp
helloworld922: There shouldn't be any significant performance or correctness differences using the same code path and have reduce handle the logic for device selection. This definitely makes the code a lot easier to read, too.

general/forall.hpp
helloworld922: Removed all gotos from vector.cpp

pazner (approved):
Thanks for the changes @helloworld922! The latest version looks really good.

My only tiny nitpick is style for the function names `vector_workspace`, `Lpvector_workspace`, and `reduce`. (Typical MFEM style is to capitalize each word, on the other hand, we use lowercase for `mfem::forall`, I don't know why that is).

tzanio:

helloworld922:
😆 I was going to change the names to be camel case, but it looks like Tzanio was quicker to the draw than me. I suppose this is extra motivation to get #4065 merged so we can remove these temporary workspace vectors completely.

I'm not sure why `mfem::forall` is lowercase either.

pazner:
> 😆 I was going to change the names to be camel case, but it looks like Tzanio was quicker to the draw than me. I suppose this is extra motivation to get #4065 merged so we can remove these temporary workspace vectors completely.

Yes, agreed, I will try to prioritize #4065 ASAP.

FWIW you can still push to this branch, Tzanio can re-merge in next.

tzanio:
Yes, just let me know and I will remerge

camierjs (approved):

PR #4696, opened 2025-02-12
Update base dockerfile to radiuss-docker

justinlaughlin:
Looks like the old repo rse-ops is [deprecated](https://github.com/rse-ops/docker-images/issues/112)  and is now [llnl/radiuss-docker](https://github.com/LLNL/radiuss-docker). Updating since `Ubuntu 20.04` will not be supported by GitHub soon.

### DIFF
--- a/config/docker/Dockerfile.base
+++ b/config/docker/Dockerfile.base
@@ -1,4 +1,4 @@
-FROM ghcr.io/rse-ops/cuda-ubuntu-20.04:cuda-11.0.3
+FROM ghcr.io/llnl/radiuss:ubuntu-22.04-cuda-11-0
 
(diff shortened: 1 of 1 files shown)


justinlaughlin:
testing here https://github.com/mfem/mfem/actions/runs/13292792936

v-dobrev:
@justinlaughlin, is this going to fix the Docker failures in CI for `master`? If so, we should probably merge this.

v-dobrev:
Also, since now this is marked WIP, what else needs to be done in order for the PR to be ready?

justinlaughlin:
Closing since the new images do not install spack. If we want to keep supporting docker we should probably build ours based off of the AWS images.

vsoch:
+1 I would suggest a base image / mfem image that installs from a clean OS base without spack.

PR #4706, opened 2025-02-14
Address some subtle bugs in the changed Newmark integration method.

tupek2:
Changes to the Newmark integration broke a Serac test.  It looks like the new behavior never allocates state[0] (the acceleration), so the initialization solve is called every timestep.  Also, I don't think it makes sense to ever do MidPointStep(x, dxdt, t, dt) for the first step of Newmark?  Certainly its non-standard.  I'd suggest removing that option, but if there is a good reason, OK.  I presume we are the main users of this functionality at the moment.

| [#4706](https://github.com/mfem/mfem/pull/4706) | @tupek2 | @tzanio | @tzanio + @cjvogl | 2/17/25 | 2/25/25 | 2/26/25 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:
@tupek2, is this `ready-for-review`? (see [CONTRIBUTING.md](https://github.com/mfem/mfem/blob/master/CONTRIBUTING.md#quick-summary))

tupek2:
Yes, I'm pretty confident the old version had an issue, but I'm not sure on the correct fix, or, for that matter, how to create a small reproducer.

tzanio:

linalg/ode.cpp
```diff
@@ -1161,17 +1161,11 @@ void NewmarkSolver::Step(Vector &x, Vector &dxdt, real_t &t, real_t &dt)
...
-      if (no_mult)
-      {
-         MidPointStep(x, dxdt, t, dt);
-         return;
-      }
-      else
-      {
-         f->Mult(x, dxdt, state[0]);
-      }
+      state[0].SetSize(f->Width());
```
cjvogl, line 1166: I apologize for the (likely naive) confusion with the `NewmarkSolver`... when do we expect either `state.Size() == 0` or `state[0].Size() == 0`? My understanding is that after `SecondOrderODESolver::Init` is called, `NewmarkSolver.state` should have size `1` with `state[0]` having size `f->Width()`. Is the size of `state` or `state[0]` altered by another function call?

tupek2: There was apparently a recent refactor in mfem (late last fall, I think) to have the time integrators use this state array.  I don't fully understand it, but it broke our usage of Newmark, which previously had a simpler implementation.  We are trying to merge the latest mfem, but this is blocking us.  I was hoping maybe the new design was understood better internally, as a lot of the behavior is truly befuddling to me.  I can dig in more to answer these questions, and they are more or less the same questions I have.  The version I have up there does seem to fix our test, which is nice, but I'm not sure if its the 'right' solution.

cjvogl: Right, for what it's worth, I believe the relevant PR was https://github.com/mfem/mfem/pull/3480 which was author by @IdoAkkerman .  Instead of imposing the use of `ODEStateVector` by adding it as a protected member variable in the base class `SecondOrderODESolver`, I think it's better to let the derived classes of `SecondOrderODESolver` decide the best implementation.  I'll take a pass at what this would like like.

cjvogl: Ok, more specifically relevant to this PR is the commit https://github.com/mfem/mfem/commit/77d3c64a0dca68351484e7be1693f7d1ea7d63dc, which changed
1. refactoring of using of a boolean `first` that would have triggered the conditional statement this comment thread is attached to
2. changing from use of `Vector dxdt2` in `SecondOrderODESolver` to `ODEStateDataVector state` (which is effectively a `std::vector<Vector>`) ... I suppose this change was intended for future support of multi-step second-order method, but @IdoAkkerman would need to confirm that.  In the meantime, I'll propose a more minimal change than reverting to `Vector dxdt`.

cjvogl: Sorry for the many messages, but I think I now see what is going on.  In `master`, the conditional statement `if (state.Size() == 0)` will be triggered after `SecondOrderODESolver::Init` because even though `state[0]` is initialized, the number of states `ss` is set to 0.  The **bug** that I believe is causing the test failure is that `if (state.Size()) == 0` will **always** trigger, because `ss` is never changed.

I'll add a suggestion to fix this without modifying the current use of `ODEStateDataVector`, although I wouldn't be opposed to reverting `SecondOrderODESolver` to use `Vector` (which I find more clear) instead of `ODEStateDataVector`.

cjvogl: Sorry (yet again): GitHub is not letting me do a "suggestion", so here's the patch I propose to fix the bug with minimal changes to `master`, which is effective reverting the changes currently in the PR to be replaced by a call to `ODEStateDataVector::Increment`:
```bash
diff --git a/linalg/ode.cpp b/linalg/ode.cpp
index a4a3718827..14a2ec173c 100644
--- a/linalg/ode.cpp
+++ b/linalg/ode.cpp
@@ -1161,11 +1161,18 @@ void NewmarkSolver::Step(Vector &x, Vector &dxdt, real_t &t, real_t &dt)
    real_t fac4 = gamma;

    // In the first pass compute d2xdt2 directly from operator.
-   if (state[0].Size() == 0)
+   if (state.Size() == 0)
    {
-      state[0].SetSize(f->Width());
-      state[0] = 0.0;
-      f->Mult(x, dxdt, state[0]);
+      state.Increment();
+      if (no_mult)
+      {
+         MidPointStep(x, dxdt, t, dt);
+         return;
+      }
+      else
+      {
+         f->Mult(x, dxdt, state[0]);
+      }
    }
    f->SetTime(t + dt);

@@ -1232,6 +1239,7 @@ void GeneralizedAlpha2Solver::Step(Vector &x, Vector &dxdt,
    // In the first pass compute d2xdt2 directly from operator.
    if (state.Size() == 0)
    {
+      state.Increment();
       if (no_mult)
       {
          MidPointStep(x, dxdt, t, dt);
diff --git a/linalg/ode.hpp b/linalg/ode.hpp
index 7844cc157b..4cda534469 100644
--- a/linalg/ode.hpp
+++ b/linalg/ode.hpp
@@ -808,12 +808,14 @@ class NewmarkSolver : public SecondOrderODESolver
 {
 private:
    real_t beta, gamma;
+   bool no_mult;

 public:
    NewmarkSolver(real_t beta_ = 0.25, real_t gamma_ = 0.5, bool no_mult_ = false)
    {
       beta = beta_;
       gamma = gamma_;
+      no_mult = no_mult_;
    };

    void PrintProperties(std::ostream &os = mfem::out);
```

cjvogl (changes requested):

linalg/ode.cpp
cjvogl: Thanks for uncovering the bug!  I've applied the patch in PR #4706 as opposed to applying it, just in case I've actually broken things... if everything works on your end, you can merge it into this branch and I'll approve this PR (although I'm open to other workflows).

tupek2: The patch gets all our tests to pass, is simpler and more consistent with the new style going forward, and now I understand 90% of the issues I was having prior to the fix, which is an improvement (I still don't know how more of our tests in serac didn't catch this, but whatever).  Thanks again!

cjvogl (approved):

tzanio:
@IdoAkkerman, can you take a look when you have some time?

linalg/ode.cpp
```diff
@@ -1163,6 +1163,7 @@ void NewmarkSolver::Step(Vector &x, Vector &dxdt, real_t &t, real_t &dt)
    // In the first pass compute d2xdt2 directly from operator.
    if (state.Size() == 0)
    {
+      state.Increment();
```
cjvogl, line 1166: Should this be after the `if(no_mult)` conditional?  I ask because the control flow for `NewmarkSolver::Step` seems to be different than `GeneralizedAlpha2Solver::Step` when `no_mult==True`, i.e.,
- in `NewmarkSolver::Step` with `no_mult==True`, `state.Size() == 0` will evaluate to `True` only once because the call to `state.Increment()`
- in `GeneralizedAlpha2Solver::Step` with `no_mult==True`, `state.Size()==0` will always evaluate to `True` because the call to `state.Increment()` is never reached.

I realize there is some irony in my asking a question about this line as the author of the commit that added it.

cjvogl: After seeing ba36f8f, I'm marking this as resolved.

tupek2: OK, thanks.  I made the change as you suggested.  At the very least to keep the different integrators more consistent!

tzanio (approved):
Thanks @tupek2

tzanio:

PR #4723, opened 2025-02-23
Update benchmarks

### LINKED ISSUE #4709: Old version of google benchmark
MFEM has a dependency on an old version of google benchmark. The `error_message` member variable was removed from the `BenchmarkReporter` class (see [this issue](https://github.com/google/benchmark/issues/1603)), and as a result when trying to compile the benchmarks in the `tests/benchmarks` directory, it fails with an error:

`mfem/tests/benchmarks/bench.hpp(54): error: class "benchmark::BenchmarkReporter::Run" has no member "error_message"`

This is using the latest version of google benchmark (1.9.1)

camierjs:
Fix to use google benchmark to fix #4709

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -132,6 +132,10 @@ Miscellaneous
   assembly routines (although the latter is still possible, by leaving the new
   virtual method as its default base implementation of returning `NULL`).
 
+- Updated benchmarks to work with newest Google Benchmarks classes.
+
+- Renamed the 'MFEM_ENABLE_GOOGLE_BENCHMARKS' option to 'MFEM_ENABLE_BENCHMARKS'.
+
 API changes
 -----------
 - API change: 'TMOP_Metric_skew2D' has been marked as deprecated.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -744,11 +744,17 @@ if (MFEM_ENABLE_TESTING)
   add_mfem_target(${MFEM_ALL_TESTS_TARGET_NAME} OFF)
   add_subdirectory(tests EXCLUDE_FROM_ALL)
 
-  # Create a target for all benchmarks and, optionally, enable it.
-  set(MFEM_ALL_BENCHMARKS_TARGET_NAME benchmarks)
-  add_mfem_target(${MFEM_ALL_BENCHMARKS_TARGET_NAME}
-                  ${MFEM_ENABLE_GOOGLE_BENCHMARKS})
-  add_subdirectory(tests/benchmarks EXCLUDE_FROM_ALL)
+  if (MFEM_USE_BENCHMARK)
+    # Create a target for all benchmarks and, optionally, enable it.
+    set(MFEM_ALL_BENCHMARKS_TARGET_NAME benchmarks)
+    add_mfem_target(${MFEM_ALL_BENCHMARKS_TARGET_NAME}
+                    ${MFEM_ENABLE_BENCHMARKS})
+    if (MFEM_ENABLE_BENCHMARKS)
+      add_subdirectory(tests/benchmarks EXCLUDE_FROM_ALL)
+    else()
+      add_subdirectory(tests/benchmarks EXCLUDE_FROM_ALL)
+    endif()
+  endif()
 endif()
 
 # Define a target that all examples and miniapps will depend on.
--- a/config/defaults.cmake
+++ b/config/defaults.cmake
@@ -86,7 +86,7 @@ set(MFEM_MPI_NP 4 CACHE STRING "Number of processes used for MPI tests")
 option(MFEM_ENABLE_TESTING "Enable the ctest framework for testing" ON)
 option(MFEM_ENABLE_EXAMPLES "Build all of the examples" OFF)
 option(MFEM_ENABLE_MINIAPPS "Build all of the miniapps" OFF)
-option(MFEM_ENABLE_GOOGLE_BENCHMARKS "Build all of the Google benchmarks" OFF)
+option(MFEM_ENABLE_BENCHMARKS "Build all of the benchmarks" OFF)
 
 # Setting CXX/MPICXX on the command line or in user.cmake will overwrite the
 # autodetected C++ compiler.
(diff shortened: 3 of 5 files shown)


tzanio:
ping: @sean-baccas

tzanio:
No need for changes in the Makefile build system?

tzanio:

CHANGELOG
```diff
@@ -132,6 +132,10 @@ Miscellaneous
   assembly routines (although the latter is still possible, by leaving the new
   virtual method as its default base implementation of returning `NULL`).
 
+- Updated benchmarks to work with newest Google Benchmarks classes.
+
+- Renamed the 'MFEM_ENABLE_GOOGLE_BENCHMARKS' option to 'MFEM_ENABLE_BENCHMARKS'.
```
tzanio, line 137:
```suggestion
- Updated the benchmarks (in tests/benchmarks) to work with the latest Google
  Benchmarks classes (version 1.9.1). Renamed the MFEM_ENABLE_GOOGLE_BENCHMARKS
  CMake option to just MFEM_ENABLE_BENCHMARKS.
```

tzanio (approved):
Thanks for updating this, @camierjs 🙏

sean-baccas:
Looks good, but unfortunately I get linker errors to `benchmark::internal::global_context[abi:cxx11]` - is there a build option that resolves this?

CHANGELOG
camierjs: done in https://github.com/mfem/mfem/pull/4723/commits/a0723083c65ff1dc1712e79e1683c40268658055

camierjs:
> Looks good, but unfortunately I get linker errors to `benchmark::internal::global_context[abi:cxx11]` - is there a build option that resolves this?

Setting `MFEM_USE_BENCHMARK=YES` and `BENCHMARK_DIR=...` while using `make` seems enough.
Make sure to link against `-lbenchmark`, `make info` should show `MFEM_EXT_LIBS=-L.../google-benchmark/lib -lbenchmark`.
`cmake` builds with the `MFEM_USE_BENCHMARK` set `ON` and the right `include_directories` should also bring `libbenchmark.a` to the link stage.

sean-baccas:
> > Looks good, but unfortunately I get linker errors to `benchmark::internal::global_context[abi:cxx11]` - is there a build option that resolves this?
> 
> Setting `MFEM_USE_BENCHMARK=YES` and `BENCHMARK_DIR=...` while using `make` seems enough. Make sure to link against `-lbenchmark`, `make info` should show `MFEM_EXT_LIBS=-L.../google-benchmark/lib -lbenchmark`. `cmake` builds with the `MFEM_USE_BENCHMARK` set `ON` and the right `include_directories` should also bring `libbenchmark.a` to the link stage.

Managed to fix that - I had messed my environment up!

sean-baccas:
I also note that the macro `BENCHMARK_OVERRIDE` is now missing from google benchmark itself (see [here](https://github.com/mfem/mfem/blob/a0723083c65ff1dc1712e79e1683c40268658055/tests/benchmarks/bench_tmop.cpp#L132)). I can push a fix for this (just needs to change to `override`) - do I need permissions to do so?

sean-baccas:
Further, I've had a couple of other issues since:

- the benchmarks don't detect that I've got a gpu available to run on, but I can make the benchmarks run with the cuda/ceed-cuda backend if I change the device string in the source file and then recompile. Is that an issue with my google benchmark installation?
- I am trying to recreate the results seen in [this paper](https://arxiv.org/pdf/2402.15940), but I'm having issues running the benchmarks in bench_ceed.cpp. In particular, in BP2/1, I get an assertion error:

`Assertion failed: (dofs.Size() <= elemvect.Size()) is false:
 --> Size mismatch: length of dofs is 24, length of elemvect is 8
 ... in function: void mfem::Vector::AddElementVector(const Array<int> &, const mfem::Vector &)
 ... in file: /home/ir-bacc3/repos/mfem/linalg/vector.cpp:671
`

Have you seen errors like this before?

helloworld922:
What's the difference between `MFEM_ENABLE_BENCHMARKS` and `MFEM_USE_BENCHMARK`? Should these be merged into the same option?

helloworld922:
I'm getting a compiler error in `test/benchmarks/bench_virtuals.cpp`, it looks like Google Benchmark is using C++14 specific features (`std::make_unique`). The Google benchmark readme seems to say that to compile Google Benchmark itself requires C++14, but it should be compatible to link with C++03? Not sure if that has changed internally without the documentation being updated.

camierjs:
> I also note that the macro `BENCHMARK_OVERRIDE` is now missing from google benchmark

should be fixed in https://github.com/mfem/mfem/pull/4723/commits/1b83c640558323aec3ba73aa4d89ba0b5f942680

camierjs:
> * the benchmarks don't detect that I've got a gpu available to run on

You should be able to use the command line argument: `--benchmark_context=device=...` to set the MFEM device.

helloworld922 (approved):

tzanio:

config/defaults.mk
```diff
@@ -552,6 +552,9 @@ ALGOIM_OPT = -I$(ALGOIM_DIR)/src $(BLITZ_OPT)
 ALGOIM_LIB = $(BLITZ_LIB)
 
 # BENCHMARK library configuration
+ifeq ($(MFEM_USE_BENCHMARK),YES)
+   BASE_FLAGS = -std=c++14
+endif
```
tzanio, line 557: Note that this will have to be modified due to [this change](https://github.com/mfem/mfem/pull/4452/commits/4606b273d8ece2a33233cb40ff6a4f79a4d84656) in https://github.com/mfem/mfem/pull/4452

camierjs:
> * I am trying to recreate the results seen in [this paper](https://arxiv.org/pdf/2402.15940), but I'm having issues running the benchmarks in bench_ceed.cpp. In particular, in BP2/1, I get an assertion error

should be fixed with https://github.com/mfem/mfem/pull/4723/commits/6bcd0ca2080a76427d15a372dcf541806e32b81e

tzanio:

sean-baccas:
Thanks, this looks really nice. Managed to build and run the benchmarks without error

tzanio:
@sean-baccas, do you mind formally approving the PR if this fixes the issue?

sean-baccas (approved):

PR #4770, opened 2025-03-31
Use the <min>...<max> syntax for CMake

adam-sim-dev:
CMake Deprecation Warning at CMakeLists.txt:15 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.


tzanio:
Superseded by https://github.com/mfem/mfem/pull/4774

tzanio:
@adam-sim-dev, did you see https://github.com/mfem/web/pull/308#issuecomment-2637351762?

PR #4766, opened 2025-03-28
Remove many of the raw pointers from the new MMA solver

tepperly:
We're planning to experiment using mfem's MMA solver in LiDO. I asked @bslazarov about replacing the raw pointers with `::std::unique_ptr`, and he said that I could make suggestions.

Here are the changes I'm proposing with this PR:
* Replace all the raw pointers with `::std::unique_ptr` including those in LUSolve.
* Remove `#include`'s from the header file that aren't necessary.
* Change MMASubBase's constructor argument to a reference because it shouldn't be optional or changeable.
* Reran astyle on mma.hpp and mma.cpp

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#4766](https://github.com/mfem/mfem/pull/4766) | @tepperly | @tzanio | @bslazarov + @SchmidtMat | 4/1/25 | 4/1/25 | 4/2/25 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
[...]

PR #4771, opened 2025-03-31
Fix error in MFEM_ABORT in MMA with LAPACK

pazner:
Fixes compilation error when compiling with LAPACK enabled:
```
linalg/mma.cpp:709:92: error: too many arguments provided to function-like macro invocation
  709 |                   MFEM_ABORT("MMA: Argument %d in linear system solve has illegal value.", info);
```

### DIFF
--- a/linalg/mma.cpp
+++ b/linalg/mma.cpp
@@ -706,7 +706,8 @@ void MMA::MMASubSvanberg::Update(const real_t* dfdx,
                }
                else
                {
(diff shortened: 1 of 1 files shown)


vladotomov (approved):

tzanio:
See https://github.com/mfem/mfem/pull/4766#issuecomment-2766526020

pazner:
Ah, didn't see that. Looks like this is superseded by #4741.

bslazarov (approved):

tzanio:
The fix should be in `master` now

PR #4782, opened 2025-04-05
Add additional debug info to try and catch an illusive github CI bug

helloworld922:
Try to catch and debug the illusive failure in `punit_tests` removing `parallel_in_serial.mesh` in github CI.


pazner:
The cause of this elusive bug is actually the [broken MPICH package on Ubuntu 24.04](https://bugs.launchpad.net/ubuntu/+source/mpich/+bug/2072338). See #4783.

tzanio:
Wow 😱

PR #4811, opened 2025-04-16
AMR Matrix-free derefinement operator

helloworld922:
Implements a matrix-free operator for AMR derefinement. Construction of the operators is significantly faster than sparse matrix construction, and application is GPU accelerated.

Notes:
- Currently limited to FiniteElementSpace (ParFiniteElementSpace still WIP)
- There is an existing FiniteElementSpace::DerefinementOperator which is unrelated to this work, it seems to be used for InterpolationGridTransfer? Can the implementation somehow be merged with this?


tzanio:

fem/fespace.cpp
```diff
@@ -17,16 +17,299 @@
...
+      MFEM_VERIFY(old_ndofs, "Missing previous (finer) space.");
+      MFEM_VERIFY(fespace->GetNDofs() <= old_ndofs,
+                  "Previous space is not finer.");
+
+      const CoarseFineTransformations &dtrans =
+         fespace->GetMesh()->ncmesh->GetDerefinementTransforms();
+
+      MFEM_ASSERT(dtrans.embeddings.Size() == old_elem_dof->Size(), "");
+
+      bool is_dg = fespace->FEColl()->GetContType() ==
```
dylan-copeland, line 96: `const bool`. Isn't this what AI tools like Copilot should do, instead of just writing useless, verbose comments?

fem/fespace.cpp
```diff
@@ -17,16 +17,299 @@
...
+   Array<int> block_row_idcs_offsets;
+   /// offsets into col_idcs
+   Array<int> block_col_idcs_offsets;
+   /// mapping for row dofs, oob indicates the block row should be ignored.
+   /// negative means the row data should be negated.
+   Array<int> row_idcs;
+   /// mapping for col dofs, negative means the col data should be negated.
+   Array<int> col_idcs;
+   Vector block_storage;
+   int max_rows;
```
dylan-copeland, line 49: Please document.

dylan-copeland:
@helloworld922 how should this PR be tested, both by reviewers and CI? Should there be unit tests or sample runs in an example or miniapp?

fem/fespace.cpp
```diff
@@ -17,16 +17,299 @@
...
+   }
+
+   DerefineMatrixOp(FiniteElementSpace &fespace_, int old_ndofs,
+                    const Table *old_elem_dof, const Table *old_elem_fos)
+      : Operator(fespace_.GetVSize(), old_ndofs * fespace_.GetVDim()),
+        fespace(&fespace_)
+   {
+      static Kernels kernels;
+      constexpr int max_team_size = 256;
+      /// TODO: Implement DofTransformation support
```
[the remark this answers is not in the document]
helloworld922: This was copied from the original `FiniteElementSpace::DerefinementMatrix` function, and I noticed similar comments throughout FiniteElementSpace. I'm not entirely sure what `DofTransformation` does or is used for, though this implementation should match the original `DerefinementMatrix` in capabilities (or lack there-of).

fem/fespace.cpp
```diff
@@ -17,16 +17,299 @@
...
 
 using namespace std;
 
 namespace mfem
 {
+/// \cond DO_NOT_DOCUMENT
 
-template <> void Ordering::
-DofsToVDofs<Ordering::byNODES>(int ndofs, int vdim, Array<int> &dofs)
+class DerefineMatrixOp : public Operator
```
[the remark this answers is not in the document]
helloworld922: It could be a struct, I wasn't sure what convention MFEM prefers to use. The name `DerefineMatrixOp` was to try and distinguish it from the existing `FiniteElementSpace::DerefinementOperator`. It isn't obvious to me if `FiniteElementSpace::DerefinementOperator` can be modified or directly used in place of `DerefinementMatrix`, do you happen to know what operation this class performs vs. `DerefinementMatrix`?

fem/fespace.cpp
helloworld922: added

fem/pfespace.cpp
```diff
@@ -22,11 +22,648 @@
...
+
+   ParDerefineMatrixOp(ParFiniteElementSpace &fespace_, int old_ndofs,
+                       const Table *old_elem_dof, const Table *old_elem_fos)
+      : Operator(fespace_.GetVSize(), old_ndofs * fespace_.GetVDim()),
+        fespace(&fespace_)
+   {
+      static Kernels kernels;
+      constexpr int max_team_size = 256;
+
+      int NRanks = fespace->GetNRanks();
```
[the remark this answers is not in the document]
helloworld922: changed

fem/fes_kernels.hpp
```diff
@@ -0,0 +1,235 @@
...
+{
+/// \cond DO_NOT_DOCUMENT
+namespace internal
+{
+
+template <Ordering::Type Order, class Base, bool Diag = true>
+struct DerefineMatrixOpFunctorBase;
+
+template <class Base>
+struct DerefineMatrixOpFunctorBase<Ordering::byNODES, Base, true>
```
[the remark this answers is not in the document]
helloworld922: added

helloworld922:
@dylan-copeland any test which uses AMR derefinement should work close to identically (barring fp rounding) as it did before, though this may be difficult to test since by design the old DerefinementMatrix/ParallelDerefinementMatrix are never called.

`ex15`/`ex15p` run will exercise the derefinement code in serial and parallel, but only on the CPU. `unit/fem/test_derefine.cpp` also tests AMR coarsening, though I think these are also limited to CPU only? I can try to update them to also run on the GPU.

fem/fes_kernels.hpp
```diff
@@ -0,0 +1,242 @@
...
+/// and fine element are on our rank), false otherwise (coarse element is on our
+/// rank, fine element is on a different rank).
+///
+template <Ordering::Type Order, class Base, bool Diag = true>
+struct DerefineMatrixOpFunctorBase;
+
+template <class Base>
+struct DerefineMatrixOpFunctorBase<Ordering::byNODES, Base, true>
+{
+   /// block col idcs offsets
```
dylan-copeland, line 45: It would be better not to abbreviate in comments.

fem/fes_kernels.hpp
```diff
@@ -0,0 +1,242 @@
...
+/// base class can access child class fields without virtual functions
+/// @tparam Diag true if this corresponds to the diagonal block (coarse element
+/// and fine element are on our rank), false otherwise (coarse element is on our
+/// rank, fine element is on a different rank).
+///
+template <Ordering::Type Order, class Base, bool Diag = true>
+struct DerefineMatrixOpFunctorBase;
+
+template <class Base>
+struct DerefineMatrixOpFunctorBase<Ordering::byNODES, Base, true>
```
dylan-copeland, line 43: The only differences between byNodes and byVDIM are IndexX and IndexY. Can the rest be in a base class, to avoid duplication?

fem/fes_kernels.hpp
```diff
@@ -0,0 +1,242 @@
...
+/// Implements matrix-vector multiply $y = A x$ for a sparse matrix composed of
+/// a sum of smaller dense blocks. There is additional permutation/sign
+/// information associated with each block. The base class only implements
+/// helper routines such as computing block widths, index into x, index into y,
+/// and column in A given sub-block information.
+/// @sa DerefineMatrixOpMultFunctor
+///
+/// @tparam Order vdim ordering for x and y. Note that for Diag = false this is
+/// ignored for x as x has a special interleaved order.
+/// @tparam Base used for the curious recurring template pattern (CRTP) so the
```
dylan-copeland, line 33: Is CRTP faster than virtual functions for real tests? I assume this is an optimization that matters, but I haven't seen this elsewhere in MFEM.

fem/fes_kernels.hpp
```diff
@@ -0,0 +1,242 @@
...
+      return segptr[tmp] * static_cast<const Base *>(this)->vdims + col +
+             vdim * segwidth;
+   }
+   int MFEM_HOST_DEVICE IndexY(int row, int vdim) const
+   {
+      return vdim + row * static_cast<const Base *>(this)->vdims;
+   }
+};
+
+/// internally used to implementing the derefinement operator Mult diagonal
```
dylan-copeland, line 166: `implementing` -> `implement`

fem/fespace.cpp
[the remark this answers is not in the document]
dylan-copeland: I'm not aware of a convention for `struct` vs. `class`, so I would just go with what makes the most sense in each case. I don't have a strong preference. I see now that `DerefinementOperator` already exists, and I'm not sure what the difference is. The different, similar names and classes may get confusing, but that may be a larger problem than this PR.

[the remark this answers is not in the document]
dylan-copeland: Also, this file is getting long, so we may want to split it up. It may make sense to put this new code in a different file, but the splitting into more files may be a question for another PR.

fem/pfespace.cpp
```diff
@@ -22,11 +22,660 @@
...
+      block_col_idcs_offsets.SetSize(num_diagonal_blocks + 1);
+      block_col_idcs_offsets.HostWrite();
+      block_off_diag_row_idcs_offsets.SetSize(num_offdiagonal_blocks + 1);
+      block_off_diag_row_idcs_offsets.HostWrite();
+      block_off_diag_col_offsets.SetSize(num_offdiagonal_blocks);
+      block_off_diag_col_offsets.HostWrite();
+      block_off_diag_widths.SetSize(num_offdiagonal_blocks);
+      block_off_diag_widths.HostWrite();
+      pack_col_idcs.SetSize(send_len);
+      // memory manager doesn't appear to have a grace-full fallback for
```
dylan-copeland, line 285: `grace-full` -> `graceful`

fem/pfespace.cpp
```diff
@@ -4551,30 +5199,44 @@ ParFiniteElementSpace::ParallelDerefinementMatrix(int old_ndofs,
...
+   // assume embedding orders are consistent (i.e. what we expect to receive
+   // first from a given rank is sent first, etc.)
+   std::vector<MPI_Request> requests;
+   requests.reserve(to_send.size() + to_recv.size());
+   // enqueue recvs
+   for (auto &v : to_recv)
+   {
+      requests.emplace_back();
+      MPI_Irecv(v.second.data(), v.second.size(), HYPRE_MPI_BIG_INT, v.first,
+                291, MyComm, &requests.back());
```
dylan-copeland, line 5232: The tag 291 has already been in use, so this PR is not to blame, but I think it would be better to define this somewhere with a name. In PR 4038, I introduced `enum VarMessageTag` in `general/communication.hpp`, since those tags were used in multiple files. If 291 is limited to `ParFiniteElementSpace`, maybe it could be a `constexpr` in this class. Anyway, it's easier to manage the tags in a way like this, instead of hard-coded numbers.

fem/fes_kernels.hpp
helloworld922: Virtual functions can't be called from CUDA `__device__` functions (and I imagine HIP as well). I think the other option for implementing this is to have the vdim indexing order stuff be a completely separate class still with DerefineMatrixOpMultFunctor as a template parameter, though I think it just makes a few things slightly messier.

fem/pfespace.cpp
helloworld922: I added a tag to `VarMessageTag` for this which right now has its value defined as `291` to prevent collisions with other hard-coded tags. We should probably go through the whole code-base and update all tags to use this, then the hard-coded constants can be removed.

dylan-copeland: Sorry to be pedantic about this, but `VarMessageTag` is associated with `struct VarMessage`. Since the 291 tag is used directly in `MPI` calls, rather than `VarMessage`, it seems better to define this tag separately from `VarMessageTag`. Maybe we can start a new `enum MPI_Tag` in `ParFiniteElementSpace`, to keep its scope limited to that class? Or if we don't have enough for an `enum`, maybe something like `static constexpr int DEREFINEMENT_TAG = 291;` again limited to the class.

fem/pfespace.cpp
```diff
@@ -22,11 +22,658 @@
...
+#include "kernel_dispatch.hpp"
+#include "fes_kernels.hpp"
+
 #include <limits>
 #include <list>
 
 namespace mfem
 {
+/// \cond DO_NOT_DOCUMENT
+class ParDerefineMatrixOp : public Operator
```
dylan-copeland, line 34: We should also consider putting this in another file, to keep this file more readable. I don't have a strong opinion about whether to make the split in this PR. What do other reviewers think?

fem/pfespace.cpp
dylan-copeland: Or maybe it's better to have all the tags in `communication.hpp`, so we can easily see if a tag is duplicated. Perhaps a separate `enum ParFiniteElementSpaceTag` in `communication.hpp` would be best.

fem/pfespace.cpp
helloworld922: I think it makes sense to split it into its own separate file, I can do that. Same with the serial version.

fem/pfespace.cpp
helloworld922: I think it makes sense for them all to be in the same place so it's easy to tell if there are any tag collisions. It could be in a separate enum in `communication.hpp`.

general/communication.hpp
```diff
@@ -425,16 +425,24 @@ class GroupCommunicator
    ~GroupCommunicator();
 };
 
+/// General MPI message tags used by MFEM
+enum MessageTag
+{
+   DEREFINEMENT_MATRIX_CONSTRUCTION_DATA_VM =
```
dylan-copeland, line 431: `_VM` means `VarMessage` and could be removed here.

dylan-copeland (approved):
There are some remaining questions about whether existing classes for derefinement should be merged with the new code or eliminated. I'm not familiar enough with these classes to comment, but it seems fine to me to keep everything for now and maybe simplify later, with the priority being to introduce the new capabilities in this PR soon.

tzanio:
Please mention in `CHANGELOG`

tzanio (approved):
Thanks @helloworld922 !

tzanio:

v-dobrev:

PR #4821, opened 2025-04-22
More threads for SmemPAMassApply3D

tomstitt:
Switches `SmemPAMassApply3D_Element` to a 3D thread block loop to better use GPUs at low/medium orders


tomstitt:
> @tomstitt, is this ready for review? If it is, can you please add the `ready-for-review` label?

I need to come back to it, when I tested before it was only faster for some orders but I also saw large iteration count differences so I think something is broken

helloworld922:
superseded by #5299

PR #4826, opened 2025-04-23
Chain asan sigaction

victor-decaria-nnl:
Chain the mfem signal handlers so that previously installed signal handlers can still be called, like if asan is being used.

Includes the signal and code in the error message for illegal memory access. I didn't realize before that I was running into an access permission error specifically, which asan can't help much with, so this extra information can be helpful.

I tested this with and without asan on this MWE I posted in this #4825  

Without asan, it falls through to calling the original `MFEM_ABORT`.

With asan, there is a partial error message from mfem, and then it goes to the asan signal handler and prints the stack trace. It does not print the `MFEM_ABORT` stuff in that case, which I hope is ok.

### DIFF
--- a/general/mem_manager.cpp
+++ b/general/mem_manager.cpp
@@ -57,6 +57,9 @@
 namespace mfem
 {
 
+static struct sigaction old_segv_action;
+static struct sigaction old_bus_action;
+
 MemoryType GetMemoryType(MemoryClass mc)
 {
    switch (mc)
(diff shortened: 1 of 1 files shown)


tzanio:

camierjs (approved):
Looks good @victor-decaria-nnl, thank you for this addition!
It definitely helps to get the stack trace.

helloworld922 (approved):

tzanio:

PR #4836, opened 2025-04-29
[WIP] AD for TMOP_WorstCaseUntangleOptimizer_Metric 

kmittal2:


tzanio:

fem/tmop.hpp
```diff
@@ -69,6 +73,20 @@ class TMOP_QualityMetric : public HyperelasticModel
    virtual void EvalPW(const DenseMatrix &Jpt, DenseMatrix &PW) const
    { PW = 0.0;}
 
+   // First-derivative hook for AD-based computations:
```
vladotomov: We should stress in the comments that this function is internal, for AD connection purposes. I don't like it, but also couldn't figure a way to remove it from the `public`. Same for `EvalW_AD2`.

fem/tmop.hpp
```diff
@@ -69,6 +73,20 @@ class TMOP_QualityMetric : public HyperelasticModel
...
+                             const std::vector<AD1Type> &W) const
+   {
+      return AD1Type{0.0, 0.0};
+   }
+
+   // Second-derivative hook for AD-based computations:
+   virtual AD2Type EvalW_AD2(const std::vector<AD2Type> &T,
+                             const std::vector<AD2Type> &W) const
+   {
+      return AD2Type{{0.0, 0.0}, {0.0,0.0}};
```
vladotomov: Isn't it better to abort? Now these functions can be used with any metric, and it will just work, returning zeros silently.

fem/tmop.cpp
```diff
@@ -645,6 +748,64 @@ real_t TMOP_WorstCaseUntangleOptimizer_Metric::EvalWBarrier(
...
+
+void TMOP_WorstCaseUntangleOptimizer_Metric::EvalP(const DenseMatrix &Jpt,
+                                                   DenseMatrix &P) const
+{
+   auto mu_ad_fn = [this](
+                      std::vector<AD1Type> &T, std::vector<AD1Type> &W)
+   {
+      return EvalW_AD1(T,W);
+   };
+   if (tmop_metric.Id() == 4 || tmop_metric.Id() == 14 || tmop_metric.Id() == 66)
```
vladotomov: I don't see 66 implementing the AD. It combines 4 an 55, and 55 does not implement anything (returns zeros from the default `EvalW_AD` implementation).

vladotomov (approved):
ping @bslazarov @SchmidtMat

fem/tmop.cpp
```diff
@@ -163,6 +193,63 @@ type mu98_ad(const std::vector<type> &T, const std::vector<type> &W)
...
+// add specialization for AD2Type
+template <>
+AD2Type make_one_type<AD2Type>()
+{
+   return AD2Type{AD1Type{1.0, 0.0}, AD1Type{0.0, 0.0}};
+}
+
+using TWCUO = TMOP_WorstCaseUntangleOptimizer_Metric;
+template <typename type>
+type wcuo_ad(type mu,
```
SchmidtMat, line 216: I wonder if we should add some form of small comment explaining what the function is doing. Similar to the other ones above.

SchmidtMat (approved):

tzanio:

PR #4840, opened 2025-04-29
Add fallback for QuadratureFunction::ProjectGridFunction

pazner:
This PR adds a slower fallback for QuadratureFunction::ProjectGridFunction that will be used when QuadratureInterpolator is not supported for the finite element space.

FiniteElementSpace::GetQuadratureInterpolator and FiniteElementSpace::GetFaceQuadratureInterpolator now may return nullptr if the space isn't supported by QuadratureInterpolator.

Some discussion about the implications of this design might be useful.

### DIFF
--- a/fem/qfunction.cpp
+++ b/fem/qfunction.cpp
@@ -56,6 +56,20 @@ void QuadratureFunction::Save(std::ostream &os) const
    os.flush();
 }
 
+void QuadratureFunction::ProjectGridFunctionFallback(const GridFunction &gf)
+{
+   if (gf.VectorDim() == 1)
+   {
+      GridFunctionCoefficient coeff(&gf);
+      coeff.Coefficient::Project(*this);
+   }
+   else
+   {
+      VectorGridFunctionCoefficient coeff(&gf);
+      coeff.VectorCoefficient::Project(*this);
+   }
+}
+
 void QuadratureFunction::ProjectGridFunction(const GridFunction &gf)
 {
    SetVDim(gf.VectorDim());
@@ -76,6 +90,13 @@ void QuadratureFunction::ProjectGridFunction(const GridFunction &gf)
       // Use quadrature interpolator to go from E-vector to Q-vector
       const QuadratureInterpolator *qi =
          gf_fes.GetQuadratureInterpolator(*qs_elem);
+
+      if (!qi)
+      {
+         ProjectGridFunctionFallback(gf);
+         return;
+      }
+
       qi->SetOutputLayout(QVectorLayout::byVDIM);
       qi->DisableTensorProducts(!use_tensor_products);
       qi->Values(e_vec, *this);
@@ -99,6 +120,13 @@ void QuadratureFunction::ProjectGridFunction(const GridFunction &gf)
       // Use quadrature interpolator to go from E-vector to Q-vector
       const FaceQuadratureInterpolator *qi =
          gf_fes.GetFaceQuadratureInterpolator(qspace->GetIntRule(0), face_type);
+
+      if (!qi)
+      {
+         ProjectGridFunctionFallback(gf);
+         return;
+      }
+
       qi->SetOutputLayout(QVectorLayout::byVDIM);
       qi->DisableTensorProducts(!use_tensor_products);
       qi->Values(e_vec, *this);
--- a/fem/quadinterpolator.cpp
+++ b/fem/quadinterpolator.cpp
@@ -69,9 +69,7 @@ QuadratureInterpolator::QuadratureInterpolator(const FiniteElementSpace &fes,
 
    d_buffer.UseDevice(true);
    if (fespace->GetNE() == 0) { return; }
-   const FiniteElement *fe = fespace->GetTypicalFE();
-   MFEM_VERIFY(fe->GetMapType() == FiniteElement::MapType::VALUE ||
-               fe->GetMapType() == FiniteElement::MapType::H_DIV,
+   MFEM_VERIFY(SupportsFESpace(fes),
                "Only elements with MapType VALUE and H_DIV are supported!");
 }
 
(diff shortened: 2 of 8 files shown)


fem/qfunction.cpp
```diff
@@ -76,6 +90,13 @@ void QuadratureFunction::ProjectGridFunction(const GridFunction &gf)
       // Use quadrature interpolator to go from E-vector to Q-vector
       const QuadratureInterpolator *qi =
          gf_fes.GetQuadratureInterpolator(*qs_elem);
+
+      if (!qi)
+      {
+         ProjectGridFunctionFallback(gf);
```
najlkin, line 96: That warning could go here to let the user know at least in the debug mode that the implementation does not use a quadrature interpolator and is expected to be slower 🐢 .

tzanio:

najlkin:
I think we should move forward with this one, making it to v4.9 ideally ⭐ . It is a small code change, but systematic improving user experience 👍 .

tzanio:
> I think we should move forward with this one, making it to v4.9 ideally ⭐ . It is a small code change, but systematic improving user experience 👍 .

Can I get some approvals, please?

rcarson3:
Could a simple unit test be added just to make sure this works roughly as expected and doesn't accidentally break in the future?

pazner:
Good point @rcarson3, addressed in 59a95cc8e6a689eca49d6df5940ccf260d864c40 and 63d983ed6a365520bcf68448065aec2f453109d1.

tests/unit/fem/test_quadf_coef.cpp
```diff
@@ -300,3 +300,50 @@ TEST_CASE("Face Quadrature Function Coefficients", "[Coefficient]")
       }
    }
 }
+
+TEST_CASE("QuadratureFunction::ProjectGridFunction",
+          "[Coefficient][QuadratureFunction]")
```
najlkin, line 305: Cool 👍 , I have checked this runs both, fallback and supported cases. However, it only tests `FaceQuadratureSpace`. Would you add a section for `QuadratureSpace`?

tests/unit/fem/test_quadf_coef.cpp
```diff
@@ -300,3 +300,50 @@ TEST_CASE("Face Quadrature Function Coefficients", "[Coefficient]")
...
+{
+   auto ftype = GENERATE(FaceType::Interior, FaceType::Boundary);
+   const int order = GENERATE(1, 2);
+   const int int_order = order + 1;
+   const auto mesh_fname = GENERATE(
+                              "../../data/star.mesh",
+                              "../../data/star-mixed.mesh",
```
najlkin, line 310: I have tried `QuadratureSpace`, but I am getting ugly errors on mixed meshes. That can be expected as `QuadratureInterpolator` uses `fespace->GetFE(0)` everywhere. Maybe we could check `Mesh::GetNumGeometries()` in `SupportsFESpace()`? 🤔

fem/qfunction.cpp
najlkin: I do not insist on this one, might be useful, but also excessive depending on the application. A preprocessor declaration could be used to turn it off or something like that, but well documented `QuadratureInterpolator::SupportsFESpace()` should be enough 😉 .

tests/unit/fem/test_quadf_coef.cpp
najlkin: ... and `Mesh::IsVariableOrder()` too 😉

pazner: Good suggestions, done in 6602df4fd251ddb986c1b7c8f29c109801dd84f8.

najlkin (approved):
Thanks for the change, looks good! 👍  I hope it will make it to v4.9 ⭐

najlkin:
@v-dobrev , are you planning to take a look?

fem/quadinterpolator.cpp
```diff
@@ -86,12 +84,20 @@ QuadratureInterpolator::QuadratureInterpolator(const FiniteElementSpace &fes,
...
 }
 
+bool QuadratureInterpolator::SupportsFESpace(const FiniteElementSpace &fespace)
+{
+   const FiniteElement *fe = fespace.GetTypicalFE();
+   const Mesh &mesh = *fespace.GetMesh();
+   return (fe->GetMapType() == FiniteElement::MapType::VALUE ||
+           fe->GetMapType() == FiniteElement::MapType::H_DIV)
+          && (!fespace.IsVariableOrder())
+          && (mesh.GetNumGeometries(mesh.Dimension()) == 1);
```
v-dobrev: The condition `(mesh.GetNumGeometries(mesh.Dimension()) == 1)` should be `<= 1` instead of `== 1` to support empty ranks.

Note that the result from `mesh.GetNumGeometries(mesh.Dimension())` is generally local to the processor, so the result of this method (`SupportsFESpace`) may be different on different ranks too. This seems to be fine when `SupportsFESpace` is used to dispatch to `ProjectGridFunctionFallback` but it is probably not a good idea to have rank-dependent result in `FiniteElementSpace::GetQuadratureInterpolator`.

One option to make the result globally consistent is to use `MPI_Allreduce` when the mesh is parallel. To avoid checks for parallel build and parallel mesh, it may be convenient to use the virtual method `Mesh::ReduceInt` instead.

Another option (which is a bit more restrictive) is to use `mesh.MeshGenerator()` (which should return the same on all ranks) and check it has only one bit equal to 1, i.e. the number is a power of 2. A simple way to check if `x` is a power of 2 is `((x-1) & x) == 0`. It may be worth adding a method to `Mesh` that returns this directly, i.e. if the mesh has only one type of elements globally.

pazner: Good point. How does 4d042959e8d41d88ba7ca026b5b8082ee22e0f34 look to you?

tzanio:

PR #4882, opened 2025-06-05
Add option to run debug memory backend without issuing errors

victor-decaria-nnl:
This adds a runtime option so that the `debug` backend won't use mprotect statements if `MFEM_MMU_SILENT` is in the environment. 

The motivation for this is if you're debugging but you suspect that the `mprotect` logic is causing false postives, you can disable the `mprotect` calls and still emulate the host-device transfers.

I thought about seeing if I could make `debug-silent` an acceptable string but wasn't sure where that plugged in and this is similar to the way `MFEM_MMU_PROTECT_ERROR` is currently used.

### DIFF
--- a/general/mem_manager.cpp
+++ b/general/mem_manager.cpp
@@ -379,7 +379,9 @@ inline void MmuDealloc(void *ptr, const size_t bytes)
 inline void MmuProtect(const void *ptr, const size_t bytes)
 {
    static const bool mmu_protect_error = GetEnv("MFEM_MMU_PROTECT_ERROR");
-   if (!::mprotect(const_cast<void*>(ptr), bytes, PROT_NONE)) { return; }
+   // Don't protect if MFEM_MMU_SILENT is defined. Just emulate device memory movement.
+   static const bool mmu_silent = GetEnv("MFEM_MMU_SILENT");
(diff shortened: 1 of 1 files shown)


tzanio:

camierjs:
Thank you @victor-decaria-nnl for bringing this option: I can see the use of it.

If we want to silence the MMU and skip all the `mprotect` calls, we can also use the `StdHostMemorySpace` and `StdDeviceMemorySpace` while selecting the memory controllers in `NewHostCtrl`:
```
case MT::HOST_DEBUG:
   if (GetEnv("MFEM_MMU_STD")) { return new StdHostMemorySpace(); }
   return new MmuHostMemorySpace();
```
and
```
case MT::DEVICE_DEBUG:
   if (GetEnv("MFEM_MMU_STD")) { return new StdDeviceMemorySpace(); }
   return new MmuDeviceMemorySpace();
```
It allows you to skip all the address computations, which overall gives a good speedup.

victor-decaria-nnl:
@camierjs I implemented your suggested change and tested that this still works as intended for me.

tzanio:
@helloworld922 and @jandrej -- can you please take a look when you get a chance?

tzanio:
ping: @helloworld922 and @jandrej

tzanio:

PR #4885, opened 2025-06-10
Batched LOR & Markers filtering

camierjs:
Adds filtering for Batched LOR Assembly and a `markers` test in `AddIntegratorsAndMarkers`, which enables all `lor_solvers` tests to pass:
 - `lor_solvers -m ../../data/amr-quad.mesh -fe [h,n,r]`
 - `lor_solvers -fe l`

### DIFF
--- a/fem/lor/lor.cpp
+++ b/fem/lor/lor.cpp
@@ -48,7 +48,7 @@ void LORBase::AddIntegratorsAndMarkers(BilinearForm &a_from,
    for (int i=0; i<integrators->Size(); ++i)
    {
       BilinearFormIntegrator *integrator = (*integrators)[i];
-      if (*markers[i])
+      if (markers[i] && (*markers[i]))
       {
          (a_to.*add_integrator_marker)(integrator, *markers[i]);
       }
--- a/fem/lor/lor_batched.cpp
+++ b/fem/lor/lor_batched.cpp
@@ -58,7 +58,7 @@ bool BatchedLORAssembly::FormIsSupported(BilinearForm &a)
    // Batched LOR requires all elements to be similar
    const auto *mesh = fes->GetMesh();
    if (fes->IsVariableOrder() || mesh->Nonconforming() ||
-       mesh->GetNumGeometries(mesh->Dimension())) { return false; }
+       mesh->GetNumGeometries(mesh->Dimension()) > 1) { return false; }
 
    if (dynamic_cast<const H1_FECollection*>(fec))
    {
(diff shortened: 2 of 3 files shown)


fem/lor/lor.cpp
```diff
@@ -48,7 +48,7 @@ void LORBase::AddIntegratorsAndMarkers(BilinearForm &a_from,
    for (int i=0; i<integrators->Size(); ++i)
    {
       BilinearFormIntegrator *integrator = (*integrators)[i];
-      if (*markers[i])
+      if (markers[i] && (*markers[i]))
```
pazner, line 51: I think the correct check is just `markers[i] != nullptr` (done in #4891).

fem/lor/lor_batched.cpp
```diff
@@ -48,12 +48,18 @@ bool HasIntegrators(BilinearForm &a)
...
+   const FiniteElementCollection *fec = fes->FEColl();
    // TODO: check for maximum supported orders
 
    // Batched LOR requires all tensor elements
    if (!UsesTensorBasis(*a.FESpace())) { return false; }
 
+   // Batched LOR requires all elements to be similar
+   const auto *mesh = fes->GetMesh();
+   if (fes->IsVariableOrder() || mesh->Nonconforming() ||
+       mesh->GetNumGeometries(mesh->Dimension()) > 1) { return false; }
```
pazner, line 61: It should work with nonconforming meshes (in serial, there was a bug in this case, fixed in #4891). I haven't tested with variable orders (in principle it could be made to work using the $P$ matrix).

miniapps/solvers/CMakeLists.txt
```diff
@@ -76,5 +76,7 @@ add_mfem_miniapp(lor_solvers
   LIBRARIES mfem)
 
 if (MFEM_ENABLE_TESTING)
-  add_test(NAME lor_solvers_ser COMMAND lor_solvers -fe n -no-vis)
+  add_test(NAME lor_solvers_nd_ser COMMAND lor_solvers -fe n -no-vis)
+  add_test(NAME lor_solvers_l2_ser COMMAND lor_solvers -fe l -no-vis)
+  add_test(NAME lor_solvers_h1_amr_ser COMMAND lor_solvers -fe h -m ../../data/amr-quad.mesh -no-vis)
 endif()
```
pazner, line 82: These cases should also be covered by the miniapp sample runs, but I'm fine with adding them here too.

camierjs:
Closed in favor of #4891

PR #4894, opened 2025-06-13
Add `Ordering::Reorder`

j-signorelli:
This PR adds a function to `Ordering` that reorders any given `Vector` from its existing ordering to a different one. There are multiple different places where this kind of functionality is re-written in MFEM repetitively, like `GridFunction::ReorderByNodes` and `FindPointsGSLIB::InterpolateH1`, in addition to users having to currently manually write their own versions of this if they want to reorganize vector data. A unit test is included as well - please let me know if I should move it to an existing file or if it is OK as is in a new one.

A description was also added to Ordering::Map.

I can also go through and try to convert any existing reorderings in MFEM to use this function instead as well in this PR, just let me know!

### DIFF
--- a/fem/fespace.cpp
+++ b/fem/fespace.cpp
@@ -55,6 +55,35 @@ DofsToVDofs<Ordering::byVDIM>(int ndofs, int vdim, Array<int> &dofs)
    }
 }
 
+void Ordering::Reorder(Vector &v, int vdim, Ordering::Type in_ord,
+                       Ordering::Type out_ord)
+{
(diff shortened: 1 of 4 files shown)


tzanio:

kmittal2:
@tzanio and @v-dobrev: related to this PR, it seems that `GridFunction::ReorderByNodes()` re-orders the grid function data without changing the underlying finite element space ordering. What is the motivation for that function? Should we mark it as deprecated?

j-signorelli:
Continued in #4981

PR #4900, opened 2025-06-18
Add missing pieces for dFEM functional computation

jandrej:


tzanio:
@jandrej, should we revive this or close it?

PR #4905, opened 2025-06-24
[WIP] Reconstruction Miniapp

cjvogl:
TODO list

- [x] add quantitative evaluation of reconstruction (e.g., discrete $L^2$ and/or $L^\inf$ error) to terminal output
- [x] refinement of mesh via command line arguments
- [x] replace static implementation of 4th row of $A$ (enforcing $\frac{d^2u}{dxdy} = 0$) with integrator implementation


miniapps/tools/reconstruction.cpp
```diff
@@ -0,0 +1,178 @@
+#include "mfem.hpp"
+
+using namespace mfem;
+
+// reconstruct a field u (represented by dst) from u_hat (represented by src)
+// Note current only 2D reconstruction of L^2_1 (piecewise-linear) field from
+// L^2_0 (piecewise-constant) field is supported. The reconstruction is done
+// by enforcing
+//   1) element average: (u, psi_hat)_E = ({u_hat}, psi_hat)_E
```
cjvogl:
```suggestion
//   1) element average: (u, psi_hat)_E = (u_hat, psi_hat)_E
```

miniapps/tools/reconstruction.cpp
```diff
@@ -0,0 +1,178 @@
...
+   for (int element_ind=0; element_ind < mesh.GetNE(); element_ind++)
+   {
+      const FiniteElement& src_element = *(src_fe_space.GetFE(element_ind));
+      const FiniteElement& dst_element = *(dst_fe_space.GetFE(element_ind));
+      ElementTransformation& transform = *(src_fe_space.GetElementTransformation(element_ind));
+      DenseMatrix A(dst_element.GetDof());
+      Vector b(dst_element.GetDof());
+      DenseMatrix Arow;
+
+      // enforce (u, psi_hat)_E = ({u_hat}, psi_hat)_E
```
cjvogl:
```suggestion
      // enforce (u, psi_hat)_E = (u_hat, psi_hat)_E
```

miniapps/tools/reconstruction.cpp
```diff
@@ -152,27 +189,27 @@ int main(int argc, char* argv[])
...
+                           << "solution\n" << mesh << u_reconstruction
+                           << "window_title 'reconstruction'\n" << std::flush;
    }
    else
       MFEM_WARNING("Cannot connect to glvis server, disabling visualization.")
-   // TODO: quantitatively compare (e.g. L2 norm) original & reconstruction
+      // TODO: quantitatively compare (e.g. L2 norm) original & reconstruction
 
-   Mpi::Finalize();
+      Mpi::Finalize();
```
cjvogl: I dont think these two lines should have had the indent changed?

homeomorfismo: I will change it on the next commit! thanks 👍

homeomorfismo: I checked the git-blame and it seems it could be an `astyle` thing. I wrapped the else block so I can keep the git hook active haha

makefile
```diff
@@ -791,7 +791,7 @@ status info:
 	$(info MFEM_MPI_NP            = $(MFEM_MPI_NP))
 	@true
 
-ASTYLE_BIN = astyle
+ASTYLE_BIN ?= astyle
```
cjvogl: Is the idea here to specify `ASTYLE_BIN` as an environment variable if `astyle` is not in the environment path?

mesh/ncmesh.hpp
```diff
@@ -1025,6 +1025,7 @@ class NCMesh
                          Array<int> *neighbors, /* append */
                          Array<char> *neighbor_set = NULL);
 
+public:
```
cjvogl, line 1092: I'm wondering why `FindNeighbors` and `NeighborExpand` weren't already public.  Do they modify member variables?  If not, perhaps make them `const` member functions would alleviate any concerns with making them `public`

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+
+         // Matrix
+         DenseMatrix e_to_f;
+         ComputeFaceMatrix(e_idx, fes_dst, *face_trans, ir, e_to_f);
+         e_to_faces.SetSubMatrix(offsets[i], 0, e_to_f);
+      }
+
+      auto reg_mat = std::make_unique<DenseMatrix>();
+      auto _diff_int = std::make_unique<DiffusionIntegrator>();
+      auto _h1_int = std::make_unique<SumIntegrator>(0);
```
cjvogl, line 1031: I'm a bit of traditionalist when it comes to `auto` in that I prefer for the variable datatype to be obvious from the same line of code.  Here is an example where I think that it **is** obvious that the datatype is `std::unique_ptr<T>`, but I will ask that other uses of `auto` that don't meet this criteria be replaced with the explicit datatype.

miniapps/tools/CMakeLists.txt
```diff
@@ -53,16 +53,22 @@ add_mfem_miniapp(tmop-metric-magnitude
   ${MFEM_MINIAPPS_COMMON_HEADERS}
   LIBRARIES mfem-common)
 
+add_mfem_miniapp(least-squares
+  MAIN least-squares.cpp LIBRARIES mfem)
```
cjvogl: Will `least-squares` work if `MFEM_USE_MPI=OFF`?  I suspect there may be issues because the `Par`-prefixed classes are used, but I could be wrong

cjvogl: Also, how much work would it be to have one driver/miniapp for all the methods you've implemented?

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+{
+
+/// Enumerator for different function examples
+enum ExamplesType
+{
+   plane,
+   sine,
+   exp_sine,
+   num_examples  // last
+};
```
cjvogl, line 24: I would consider partitioning this file in general into a header file with these enum, class, and functions definitions.

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+///@{
+
+/** Class for an asymmetric (out-of-element) mass integrator $a_K(u,v) := (E(u),v)_K$,
+ *  where $E$ is an extension of $u$ (with original domain $\hat{K}$) to $K$.
+ *
+ *  @note Currently we don't need to derive from MassIntegrator, as GetRule is public.
+ *  This could be it's own standalone function.
+ *  @todo Kernel implementation: requires kernel_dispatch.hpp. MFEM_THREAD_SAFE...
+ */
+class AsymmetricMassIntegrator : public MassIntegrator
```
cjvogl, line 102: Perhaps this is a question I'll answer for myself as I read more, but does `AsymmetricMassIntegrator` need to inherit from `MassIntegrator`?  More specifically, is an AsymmetricMassIntegrator` object ever interpreted as `MassIntegrator`?  If not, I would remove the inheritance.

cjvogl: N/m, I see your comment above the class.  Yes, please avoid inheritance unless polymorphism is needed.

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+      it_solver->SetRelTol(isp.rtol);
+      it_solver->SetAbsTol(isp.atol);
+   }
+
+   // Reconstruction function
+   switch (params.rec)
+   {
+      case norm:
+         L2Reconstruction(*solver.get(), u_src, u_dst, mesh, params.newton,
+                          params.reg_type, params.reg, params.preserve_volumes);
```
cjvogl, line 1631:
```suggestion
         L2Reconstruction(*solver, u_src, u_dst, mesh, params.newton,
                          params.reg_type, params.reg, params.preserve_volumes);
```
The "deference" operator should be defined for the `std::shared_ptr` object itself.

cjvogl: Note this can also be applied in the other reconstruction methods.

miniapps/tools/least-squares.cpp
cjvogl: Another option would be to include only the function declarations before `main` and move the definitions to after `main`... either way, the goal is to have `main` appear earlier in the file.

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+      neighbor_fe.CalcPhysShape(neighbor_trans, neighbor_shape);
+      AddMult_a_VVt(e_trans.Weight()*ip.weight, neighbor_shape, elmat);
+   }
+}
+
+
+/** Integrator of the DPG form $\langle v, \lbrace w \rbrace \rangle$ where
+ *  $\lbrace \cdot \rbrace$ represents the average of the traces of the argument.
+ */
+class TraceAverageIntegrator : public BilinearFormIntegrator
```
cjvogl, line 249: Similar to a comment about the `AsymmetricMassIntegrator`, does this class need to inherit from `BilinearFormIntegrator`?  More specifically, is a `TraceAverageIntegrator` ever used as a `BilinearFormIntegrator`?  If not, please remove this inheritance.  For what it's worth, this is known as preferring composition ("has a") over inheritance ("is a").

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+      }
+   }
+}
+
+///@}
+
+/// @name Auxiliary functions
+///@{
+
+/// @brief Dense small least squares solver
```
cjvogl, line 336: Can you expand this comment to more explicitly define the least squares problem being solved?  Namely something like "solves the least squares problem $x = argmin_y \|Ay-b\|_2^2$" or "solves the least squares problem $x = argmin_y (Ay-b)^T(Ay-b)$" or "solves the least squares problem $x = argmin_y (Ay-b,Ay-b)$"

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+      }
+      AtA_reg.Reset(_AtA_reg, true);
+      solver.SetOperator(*_AtA_reg);
+   }
+   else { mfem_error("Solver not implemented for this wrapper!"); }
+
+   solver.Mult(Atb, x);
+}
+
+/// @brief Dense small least squares solver, with constrains @a C with value @a c
```
cjvogl, line 385: Can you also expand this comment to be consistent with the expanded comment above the other `LSSolver`?  Something like "Solves the constrained least squares problem $x = argmin_y \|Ay - b\|_2^2$ subject to $Cy = c$.

miniapps/tools/least-squares.cpp
cjvogl: Also, please note what the method is that is used to solved the problem... something like 
"solves the least squares problem $x = argmin_y \|Ay - b\|_2^2$ by enforcing the first-order optimality condition $A^TA x - A^T b = 0$

makefile
homeomorfismo: TLDR: Yes. 

As far as I know, the make rules of the astyle repo create the binaries in an odd directory tree and the binary is usually named `Astyle`. To avoid cluttering someone's `PATH`, the binary could be provided as an env variable. I mainly did this to run the git hooks comfortably, but the commit can be reverted too. :)

miniapps/tools/least-squares.cpp
homeomorfismo: I think a separate header file would be better!

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+/// @brief Local-averages-based reconstruction
+void AverageReconstruction(Solver& solver,
+                           ParGridFunction& src,
+                           ParGridFunction& dst,
+                           ParMesh& mesh,
+                           IterativeSolverParams& newton,
+                           RegularizationType reg_type = direct,
+                           real_t reg = 0.0,
+                           int print_level = -1,
+                           bool preserve_volumes = false)
```
cjvogl, line 689: I believe the `ParMesh` object can be obtain through one of the `ParGridFunction` arguments, so I would remove it from the function signature.

Additionally on a first pass read, I believe `print_level` is used more as a boolean flag than an integer print level... if that is the case, I would change the function signature to use a boolean flag.

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+                     std::max(fe_src_other->GetOrder(), fe_dst_other->GetOrder());
+   return IntRules.Get(face_trans.GetGeometryType(), order);
+}
+
+///@}
+
+/// @name Reconstruction functions
+///@{
+
+/// @brief Local-averages-based reconstruction
```
cjvogl, line 680: Please expand the comments here to describe the optimization problem being solved and the method for solving it (see comments on `LSSolver`)

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+   y = 0.0;
+
+   x += z.GetBlock(0);
+   y += z.GetBlock(1);
+}
+
+/// @brief Dense small least squares solver, with constrains @a C with value @a c
+/// but the constraint is one-dimensional, and passed as a vector. Does not return
+/// the multiplier.
+void LSSolver(Solver& solver,
```
cjvogl, line 473: To disambiguate the `solver` argument from the function which has Solver in the name, I suggest renaming to `linear_solver` (applies to all other use of `solver` as well)

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+      if (print_level >= 0) { CheckLSSolver(fe_dst_to_neighbors_mat, src_neighbor, dst_e); }
+
+      dst.SetSubVector(e_dst_dofs, dst_e);
+
+      e_dst_dofs.DeleteAll();
+      neighbors_e.DeleteAll();
+   }
+}
+
+/// @brief An out-of-element, L2-based reconstruction
```
cjvogl, line 833: Please expand the comments here to describe the optimization problem being solved and the method for solving it (see comments on LSSolver)

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+      }
+
+      dst.SetSubVector(e_dst_dofs, dst_e);
+
+      neighbors_e.DeleteAll();
+      e_dst_dofs.DeleteAll();
+   }
+}
+
+/// @brief A element-based, face-average reconstruction
```
cjvogl, line 966: Please expand the comments here to describe the optimization problem being solved and the method for solving it (see comments on LSSolver)

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+      fes_dst.GetElementDofs(e_idx, e_dst_dofs);
+      dst.SetSubVector(e_dst_dofs, dst_e);
+
+      e_dst_dofs.DeleteAll();
+      faces_e.DeleteAll();
+      orientation_e.DeleteAll();
+   }
+}
+
+/// @brief A element-based, face-average reconstruction using weak jumps
```
cjvogl, line 1102: Please expand the comments here to describe the optimization problem being solved and the method for solving it (see comments on LSSolver)

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+
+      dst.SetSubVector(e_dst_dofs, dst_e);
+
+      e_dst_dofs.DeleteAll();
+      faces_e.DeleteAll();
+      orientation_e.DeleteAll();
+   }
+}
+
+/// @brief A minimum-variation, face-constrained reconstruction average
```
cjvogl, line 1265: Please expand the comments here to describe the optimization problem being solved and the method for solving it (see comments on LSSolver)

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+       * VectorConstantCoefficient yhat(Vector({0.0,1.0}));
+       * Vector b_yhat(mesh.GetNE());
+       * ParBilinearForm B_yhat(src.ParFESpace());
+       * B_yhat.AddInteriorFaceIntegrator(new DGTraceIntegrator(yhat, 1.0, 0.0));
+       * B_yhat.AddBdrFaceIntegrator(new DGTraceIntegrator(yhat, 2.0,
+       *                                                   0.0)); // note the 2 enforces du/dy=0 at the y boundaries
+       * B_yhat.Assemble();
+       * B_yhat.Finalize();
+       * matrix = std::unique_ptr<HypreParMatrix>(B_yhat.ParallelAssemble());
+       * matrix->Mult(*src.GetTrueDofs(), b_yhat); */
```
cjvogl, line 1345: Delete this code if not being used

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+};
+
+function<real_t(const Vector &)> exp_sine = [](const Vector& x)
+{
+   return exp(hypot(x(0),x(1))) *
+          sin(2.0 * M_PI * k_x * x(0)) *
+          sin(2.0 * M_PI * k_y * x(1));
+};
+
+} // end namespace example
```
cjvogl, line 1378: Because these are effectively single-line lambda expressions, I would just define them where they are used in `main`.

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+
+   // Solver choice
+   // TODO(Gabriel): Support more PCs?
+   /* Notes:
+    * OperatorJacobiSmoother will call AssembleDiagonal on
+    * Operator, even if the Setup function will be called
+    * later on. DenseMatrix does not have this method
+    * implemented.
+    */
+   std::shared_ptr<Solver> solver(nullptr);
```
cjvogl, line 1581: Technically, I think the trivial constructor already sets the object to `nullptr`, so you can replace with `std::shared_ptr<Solver> solver`... although I would replace `solver` with the more descriptive `linear_solver`

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+      {
+         case 0:
+            print_level.All();
+            break;
+         case 1:
+            print_level.FirstAndLast();
+            print_level.Warnings();
+         case 2:
+            print_level.Errors();
+            break;
```
cjvogl, line 1615: Missing a break statement as part of `case 1`?

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+   // Visualization
+   {
+      char vishost[] = "localhost";
+      socketstream glvis_smooth(vishost, params.vis.port);
+      socketstream glvis_src(vishost, params.vis.port);
+      socketstream glvis_dst_avg(vishost, params.vis.port);
+      socketstream glvis_dst(vishost, params.vis.port);
+
+      if (glvis_smooth && glvis_src && glvis_dst_avg && glvis_dst &&
+          params.vis.visualization)
```
cjvogl, line 1661: I would move the check for `params.vis.visualization` to directly under the comment line `// Visualization`, i.e., to avoid the whole visualization block if visualization is disabled

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
...
+
+   if (Mpi::Root())
+   {
+      mfem::out << "\n|| Rec(u_h) - u ||_{L^2} = " << error << "\n" << std::endl;
+      mfem::out << "\n|| Avg(Rec(u_h)) - u_h ||_{L^2} = " << error_avg << "\n" <<
+                std::endl;
+   }
+
+   if (params.vis.save_to_files && Mpi::Root())
+   {
```
cjvogl, line 1709: consider combining these conditionals

miniapps/tools/least-squares.cpp
```diff
@@ -0,0 +1,1720 @@
+// Copyright (c) 2010-2025, Lawrence Livermore National Security, LLC. Produced
+// at the Lawrence Livermore National Laboratory. All Rights reserved. See files
+// LICENSE and NOTICE for details. LLNL-CODE-806117.
+//
+// This file is part of the MFEM library. For more information and source code
+// availability visit https://mfem.org.
+//
+// MFEM is free software; you can redistribute it and/or modify it under the
+// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// CONTRIBUTING.md for details.
```
cjvogl, line 10: Please add some "sample runs" that contain an example of each of the methods you've implemented.

cjvogl:
@homeomorfismo , I've made a first pass through the code.  Overall, I think this is in really good shape.  Let's focus on the corresponding write-up first, and only return to the code here if that write-up is completed before the end of the week.

cjvogl:
Closing this PR in lieu of #5362 that will include all the reconstruction work here alongside the SAMRAI-MFEM coupling work.

PR #4907, opened 2025-06-24
Boundary integration in the hyperbolic framework

najlkin:
PR adding support for boundary integration to the hyperbolic form integrator and introducing a linear form integrator for boundary flow `BoundaryHyperbolicFlowIntegrator` and `BdrHyperbolicDirichletIntegrator` for weak Dirichlet BC.

For testing, there is a derived branch `najlkin/fix-hyperbolic-bdr-test` with modified example 41. It demonstrates a BC for inflow  and free outflow on the opposite side. You may run something like: `./ex41 -p 1 -r 1 -tf 1.2` (add `-nlbc` for weak Dirichlet) to get:
GLVis_s01

❓ A simple run involving the integrator is in `darcy-hdg-dev` in `examples`: `./ex5-nguyen -nx 20 -nl -nlc -p 2` (Jacobian is used for preconditioning)

| [#4907](https://github.com/mfem/mfem/pull/4907) | @najlkin | @tzanio | @cjvogl + @helloworld922 | 7/1/25 | 9/7/25 | 9/14/25 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:

fem/hyperbolic.cpp
```diff
@@ -231,33 +239,41 @@ void HyperbolicFormIntegrator::AssembleFaceVector(
...
       }
       else
       {
          CalcOrtho(Tr.Jacobian(), nor);
       }
       // Compute F(u+, x) and F(u-, x) with maximum characteristic speed
       // Compute hat(F) using evaluated quantities
-      const real_t speed = numFlux.Eval(state1, state2, nor, Tr, fluxN);
+      const real_t speed = (dof2) ? numFlux.Eval(state1, state2, nor, Tr, fluxN):
+                           fluxFunction.ComputeFluxDotN(state1, nor, Tr, fluxN);
```
helloworld922, line 260: Does it make sense to add a method `ComputeFluxDotNBdr` to allow weak Dirichlet boundary condition enforcement/boundary flux BC?

fem/hyperbolic.cpp
```diff
@@ -335,7 +364,14 @@ void HyperbolicFormIntegrator::AssembleFaceGrad(
...
 
       // Compute hat(J) using evaluated quantities
-      numFlux.Grad(1, state1, state2, nor, Tr, JDotN);
+      if (dof2)
+      {
+         numFlux.Grad(1, state1, state2, nor, Tr, JDotN);
+      }
+      else
+      {
+         fluxFunction.ComputeFluxJacobianDotN(state1, nor, Tr, JDotN);
```
helloworld922, line 361: Same here for weak Dirichlet BCs/boundary flux BC.

fem/hyperbolic.cpp
najlkin: Two points here:

- This is an integrator, so on the level of the form you may specify a different one for the boundary if you like and provide it an arbitrary flux function, so I do not see a need to specialize things here.
- For integration of boundary conditions, you may use any `LinearFormIntegrator` and `LinearForm`.  However, it is true we do not have anything if you want to use for consistency the same flux function and just provide a value of the scalar, i.e., a `LinearFormIntegrator` taking a `FluxFunction` and a coefficient as inputs 🤔.  Should I add it?

helloworld922: hmm, either way is probably fine.

helloworld922 (approved):

fem/hyperbolic.cpp
najlkin: Ok, I added `BoundaryHyperbolicLFIntegrator` for this purpose. Could you have a look? 😉 There is also a test, see the instructions above ☝️ .

helloworld922 (approved):

najlkin:
@cjvogl , @dohyun-cse , any chance looking at this?

cjvogl:
> @cjvogl , @dohyun-cse , any chance looking at this?

Working through my to-do list: hoping next week or the one after.

fem/hyperbolic.cpp
najlkin: It goes beyond the original motivation for the PR, but I just added also `BdrHyperbolicFormIntegrator`, which does integration of weak Dirichlet if you have really a non-linear problem.

fem/hyperbolic.cpp
```diff
@@ -314,18 +338,23 @@ void HyperbolicFormIntegrator::AssembleFaceGrad(
 
       // Calculate basis functions on both elements at the face
```
cjvogl: This comment should be updated to match the change below it.

fem/hyperbolic.cpp
```diff
@@ -219,7 +219,15 @@ void HyperbolicFormIntegrator::AssembleFaceVector(
...
-      const int order = 2*std::max(el1.GetOrder(), el2.GetOrder()) + IntOrderOffset;
+      int order;
+      if (dof2)
+      {
+         order = 2*std::max(el1.GetOrder(), el2.GetOrder()) + IntOrderOffset;
+      }
+      else
+      {
+         order = 2*el1.GetOrder() + IntOrderOffset;
+      }
```
cjvogl:
```suggestion
      const int max_el_order = dof2 ? std::max(el1.GetOrder(), el2.GetOrder()) : el1.GetOrder();
      const int order = 2*max_el_order + IntOrderOffset;
```

fem/hyperbolic.cpp
```diff
@@ -302,7 +318,15 @@ void HyperbolicFormIntegrator::AssembleFaceGrad(
...
```
cjvogl:
```suggestion
      const int max_el_order = dof2 ? std::max(el1.GetOrder(), el2.GetOrder()) : el1.GetOrder();
      const int order = 2*max_el_order + IntOrderOffset;
```

fem/hyperbolic.hpp
```diff
@@ -416,6 +412,189 @@ class HyperbolicFormIntegrator : public NonlinearFormIntegrator
...
+/**
+ * @brief Abstract boundary hyperbolic form integrator, assembling
+ * <F̂(u⁻,u_b,x) n, [v]> term for scalar finite elements at the boundary.
+ *
+ * This form integrator is coupled with a NumericalFlux that implements the
+ * numerical flux F̂ at the boundary faces. The flux F is obtained from the
+ * FluxFunction assigned to the aforementioned NumericalFlux with the given
+ * boundary coefficient for the state u_b.
+ */
+class BdrHyperbolicFormIntegrator : public NonlinearFormIntegrator
```
cjvogl: Perhaps there is an obvious issue that I'm overlooking, but could the functionality of assembling `<F(u^-,u_b,x) n, [v]>` introduced in `HyperbolicFormIntegrator` (using `u^+ = u_b` in the appropriate class methods when `Tr.Elem2No < 0`) as opposed to introducing a new class?  If so, I would prefer that from a code maintenance perspective.

cjvogl (changes requested):
Sorry again for the delayed review.  I've added a few suggestions for consideration and posed a more general question about whether the functionality introduced in `BdrHyperbolicFormIntegrator` could be introduced as part of `HyperbolicFormIntegrator`.

fem/hyperbolic.hpp
```diff
@@ -416,6 +412,189 @@ class HyperbolicFormIntegrator : public NonlinearFormIntegrator
...
+ * <-F(u,x) n, v> terms for scalar finite elements.
+ *
+ * This form integrator is coupled with a FluxFunction that evaluates the
+ * flux F at the faces.
+ */
+class BoundaryHyperbolicLFIntegrator : public LinearFormIntegrator
+{
+   const FluxFunction &fluxFunction;
+   Coefficient *u_coeff;
+   VectorCoefficient *u_vcoeff;
```
dohyun-cse: Can't we use `u_vcoeff` only? When a coefficient is given, create a `VectorArrayCoefficient` with `vdim=1`. We can store `own_coeff` to mark whether we own `u_vcoeff`. Then we do not have to use branching in the main loop.

fem/hyperbolic.cpp
```diff
@@ -360,36 +396,377 @@ void HyperbolicFormIntegrator::AssembleFaceGrad(
...
+      if (nor.Size() == 1)  // if 1D, use 1 or -1.
+      {
+         nor(0) = 2*Tr.GetElement1IntPoint().x - 1.;
+      }
+      else
+      {
+         CalcOrtho(Tr.Jacobian(), nor);
+      }
+      // Compute F(u, x) with maximum characteristic speed
+      const real_t speed = fluxFunction.ComputeFluxDotN(state, nor, Tr, fluxN);
```
dohyun-cse, line 688: A quick thought on the boundary conditions: shouldn't we be checking the eigenvalues of the flux Jacobian to determine if a boundary point is inflow or outflow? My concern is that without this check, we risk creating an ill-posed problem, which could lead to non-physical solutions.

Given this, would it be better to use the original `HyperbolicFormIntegrator` and extend it with `u_vcoeff`? That approach would allow us to use flux information from the interior solution to correctly distinguish between inflow and outflow boundaries.

fem/hyperbolic.hpp
najlkin: We can, this is just for convenience when using the framework for scalar equations, which is a large use case, I think. However, you are right there are proxy classes for that. I do not see the branching as a problem from performance perspective, it is a simple condition which can be optimized out by the compiler or reliably predicted by the processor. Anyway, if you insist I may remove it.

fem/hyperbolic.cpp
najlkin: In `BdrHyperbolicFormIntegrator`, that depends on the flux function, I have implemented `ComponentwiseUpwindFlux`, which does upwinding and would split the inflow and outflow BCs. It is up to the user, so you may have different flux function inside the domain and at the boundary if you are concerned about this, but the stability argument is the same for the interior as the boundary (it is a Riemann problem of two states), so you may use Rusanov happily, for example 😉 . Or, you can of course specify the boundary attribute in the bilinear form to manually apply the BC where appropriate. It would be unsystematic to add another layer of upwinding there.

In `BoundaryHyperbolicLFIntegrator`, it is a question of the design if to include it or not. It can be like `BoundaryFlowIntegrator` with upwinding or `VectorBoundaryFluxLFIntegrator` without it. I think it makes sense to follow the first one and add it 👍 

That is a misunderstanding, the `BdrHyperbolicFormIntegrator` uses the values from the interior of the domain, so there is no need to integrate it to `HyperbolicFormIntegrator`. `BoundaryHyperbolicLFIntegrator` does not on purpose, because it allows you to form the rhs of a linear system.

fem/hyperbolic.hpp
najlkin: Well, it seems the majority is for integrating it, but let me mention a few things here why I think it is not a good idea:

- it is not a common practice in MFEM to make large monolithic integrators, but rather compound different smaller ones
- the code is relatively short and simple, so maintenance is not an issue
- the documentation and code are more clear when referring to a single case
- a separate class enables interior BCs, because the decision what is boundary or not is on the user, not a simple topology logic
- there are multiple ways how to implement weak Dirichlet BCs, like someone might want to use a discrete grid function or so, there is no particular reason to privilege one approach over other (even though the default MFEM integrators usually use coefficients)
- there are different BCs (Neumann, Newton,...), so again why to privilege one?

cjvogl: > the code is relatively short and simple, so maintenance is not an issue

How many lines must code be before maintenance is an issue?  My point is that your comment is a subjective statement that I find at best not productive and at worst demeaning.  

> the documentation and code are more clear when referring to a single case
> a separate class enables interior BCs, because the decision what is boundary or not is on the user, not a simple topology logic
> there are multiple ways how to implement weak Dirichlet BCs, like someone might want to use a discrete grid function or so, there so no particular reason to privilege one approach over other (even though the default MFEM integrators usually use coefficients)
> there are different BCs (Neumann, Newton,...), so again why to privilege one?

If this integrator has a particular purpose, then I suggest changing the comment and name to indicate the specific purpose of this additional integrator (perhaps something indicating that it's a particular way to enforce weak Dirichlet, that can be applied to interior interfaces): `BdrHyperbolicIntegrator` implies to me that it is **the** integrator you must add along to the boundary with `HyperbolicIntegrator` on in the interior.

cjvogl: > it is not a common practice in MFEM to make large monolithic integrators, but rather compound different smaller ones

I view what this PR introduces to `HyperbolicFormIntegrator` as evaluating $<F(u^-,u^+),v>$ on boundaries by assuming $u^+ = u^-$ (i.e., the "zero-th order extrapolation boundary treatment" implemented by many finite volume codes).  I suppose the implicit assumption is that $F(u^-,u^-) = f(u^-)$ but that property is a common requirement for numerical flux functions. 

From that perspective of the addition to `HyperbolicFormIntegrator`, optionally using $u^+ = u_b$ instead of $u^+ = u^-$ along the boundary is (to me at least) a natural functionality for `HyperbolicFormIntegrator`, but I can understand your large monolithic integrator concern if you anticipate more boundary treatment options being added to the framework. If that is indeed the case, then I refer to my suggestion above to rename `BdrHyperbolicFormIntegrator` (as well as update the class comments to include your intended use on interior faces).

najlkin: Yeah, that is another thing with the "free" BC. I thought about it and it is a matter of discussion what the plain integrator should do at the boundary. The reasoning here is that `F` is maybe seemingly more consistent, but might introduce unexpected boundary mass/layer when combined with BC integrators for the (physical) flux or (newly) `BdrHyperbolicFlowIntegrator`. So I think it is better to keep `f` there. If you want a Dirichlet BC with `F`, you have that `BoundaryHyperbolicDirichletIntegrator` for it (now with this hopefully more fitting name) :wink:

fem/hyperbolic.hpp
najlkin: Ok, in the light of other changes (with additional constructors), I am removing the scalar coefficient. However, it is not common to own coefficients in integrators, so I am not doing that part and rather using references.

fem/hyperbolic.cpp
najlkin: Ok, I renamed `BoundaryHyperbolicLFIntegrator` to `BoundaryHyperbolicFlowIntegrator` (following `BoundaryFlowIntegrator`) and added the component-wise upwinding logic.

fem/hyperbolic.cpp
```diff
@@ -449,7 +425,7 @@ BdrHyperbolicDirichletIntegrator::BdrHyperbolicDirichletIntegrator(
    : NonlinearFormIntegrator(),
      numFlux(numFlux),
      fluxFunction(numFlux.GetFluxFunction()),
-     u_coeff(nullptr), u_vcoeff(&bdrState),
```
cjvogl: Regardless of whether holding references is common elsewhere in MFEM, I think should be avoided here as it prevents the potential future implementation and leveraging of explicit move semantics for `VectorCoefficient`.  As an example, in the following code, the call to `reference.Do()` segfaults because the reference held by `reference` is invalidated by the move construction, whereas the call to `pointer.Do()` is valid.
```c++
#include <iostream>
#include <memory>

class Test
{
  std::unique_ptr<double> test;

  public:

    Test(double val) { test = std::make_unique<double>(val); }

    Test(Test &&other) { test = std::move(other.test); }

    Test& operator +=(const int val) { *test += val; return *this;}

    void Print() { std::cout << *test << std::endl; }
};

class Pointer
{
  Test* test;

  public:

    Pointer(Test *m_test): test(m_test) {}

    void Do() { test += 1; }
};

class Reference
{
  Test& test;

  public:

    Reference(Test &m_test): test(m_test) {}

    void Do() { test +=1; }
};


int main(int argc, char* argv[])
{

  Test test(1.0);
  Pointer pointer(&test);
  Reference reference(test);

  Test test2(std::move(test));
  pointer.Do();
  test2.Print();
  reference.Do();
  test2.Print();

  return 0;
}
```
Execution results in the following terminal output:
```bash
./test
1
Segmentation fault (core dumped)
```

cjvogl: You can ignore my comment above... the test example I put together has an issue with it.  While I still think holding a pointer over holding a reference is better practice, I'll mark this as resolved unless I can come up with an example.

najlkin: Yeah, I think the validity is the same, but the system has some segfault tolerance, not rising alarm immediately. I would say references are preferred in MFEM at such places, because they make it clear the content is not owned and also that the content is required for functionality of the class.

tzanio:

najlkin:
@tzanio , ready to be remerged...

tzanio:

PR #4918, opened 2025-07-03
Initialize `KnotVector::coarse`

tomstitt:
`KnotVector::coarse` is not initialized during construction which is causing undefined behavior [here](https://github.com/mfem/mfem/blob/master/mesh/nurbs.cpp#L97) because a non-valid value for bool is being assigned to another bool.


tomstitt:
@camierjs already did this in https://github.com/mfem/mfem/pull/4899 =]

PR #4929, opened 2025-07-11
Adding complex-valued coefficient classes

mlstowell:
This PR adds support for complex-valued coefficients to make complex FEM objects more convenient to work with. The main additions are:
- ComplexCoefficient
- ComplexVectorCoefficient
- ComplexMatrixCoefficient
As well as a handful of classes derived from these such as:
- ComplexConstantCoefficient
- ComplexVectorConstantCoefficient
- ComplexMatrixConstantCoefficient 
- ComplexFunctionCoefficient
- ComplexVectorFunctionCoefficient 
And a few related classes:
- [Real|Imag]PartCoefficient
- [Real|Imag]PartVectorCoefficient
- [Real|Imag]PartMatrixCoefficient

To create these classes we also need complex-valued vector and matrix classes so that we can implement the `Eval` methods. These are named:
- ComplexVector
- StdComplexDenseMatrix (because ComplexdenseMatrix already exists and didn't seem appropriate for this purpose).

Alternate names may be needed but we can debate that during the review process.

Another significant addition is the appearance of templated methods on the ComplexLinearForm and SesquilinearForm classes to enable adding integrators using just one integrator class and one coefficient object. The resulting API looks like this:
`a->AddDomainIntegrator<VectorFEMassIntegrator>(complexMassCoef);`
where we previously had to use:
` a->AddDomainIntegrator(new VectorFEMassIntegrator(massCoef), new VectorFEMassIntegrator(lossCoef));`

The primary motivation for these changes was the desire to eliminate the need for separate coefficients for the real and imaginary parts of various coefficients.


linalg/complex_vector.hpp
```diff
@@ -0,0 +1,480 @@
...
+   ComplexVector(const Vector &);
+
+   /// Move constructor. "Steals" data from its argument.
+   ComplexVector(ComplexVector&& v);
+
+   /// @brief Creates vector of size s.
+   /// @warning Entries are not initialized to zero!
+   explicit ComplexVector(int s);
+
+   /// Creates a vector referencing an array of complex<doubles>,
```
adam-sim-dev, line 51: complex<double>s?

adam-sim-dev:
Thanks @mlstowell  for the great addition! I use complex operations frequently.  It would be great if add and subtract functions could be implemented for ComplexVector too .
https://github.com/mfem/mfem/blob/7aa327174abea231cf3136a58d68594e0349df4f/linalg/vector.hpp#L373-L390

tzanio:
@mlstowell can you resolve conflicts

PR #4936, opened 2025-07-16
Modified  fespace initialization for NURBS computation

IdoAkkerman:
With this PR the initialization of fespaces for NURBS computations can  be done without giving an implicit NURBSextension.

If no NURBSextsion is provided but the FEcollection is a NURBSFECollection,  either the  NURBSextension from the mesh is used. Or if the order does not match a new NURBSextension is constructed on the fly.

This greatly simplifies the driver code. See ex1 for the small modification required for a NURBS computation.

| [#4936](https://github.com/mfem/mfem/pull/4936) | @IdoAkkerman | @tzanio | @justinlaughlin + @dylan-copeland | 10/2/25 | ⌛due 10/16/25 | ⌛due 10/23/25 | |

<details>
    
- [x] Code builds.
- [ ] Update `CHANGELOG`:
- [x] Update `INSTALL`:
    - [x] `.github`
    - [x] `.appveyor.yml`
- [x] Update `.gitignore`:
- [x] New examples:
- [x] New miniapps:
   - [x] If new miniapps directory (e.g.`miniapps/nurbs`), list it in `doc/CodeDocumentation.conf.in`
- [ ] New capability:
</details>


IdoAkkerman:
Please note, this PR fixes the visualization part of issue #5035.

#5050 also solves the problem, however this PR is preferred solution.

tzanio:

fem/fespace.hpp
```diff
@@ -679,8 +679,17 @@ class FiniteElementSpace
    /// Returns the mesh
    inline Mesh *GetMesh() const { return mesh; }
 
+   /// Checks if the fespace is a NURBS fespace
+   bool IsNURBS() const { return NURBSext != NULL; }
+
+   /// Returns the NURBSextension
```
dylan-copeland: `NURBSextension` -> `NURBSExtension`

examples/ex1.cpp
```diff
@@ -144,7 +144,16 @@ int main(int argc, char *argv[])
    bool delete_fec;
    if (order > 0)
    {
-      fec = new H1_FECollection(order, dim);
+      if (mesh.NURBSext)
+      {
+         fec = new NURBSFECollection(order);
```
dylan-copeland: Should this change be made also in `ex1p` and `ex11p`?

IdoAkkerman: This is the question....
The changes to obtain a NURBS space are now so minimal we might as well make all the examples (That use H1 elements) be able to run with NURBS.  So `ex1` was mainly there as a show case.
I guess this will require a discussion (with @tzanio a.o.), that might as well be postponed to a later PR, when I think of it.

The NURBS miniapps can then be used for NURBS specific stuff.

fem/fespace.hpp
IdoAkkerman: Done.

examples/ex1.cpp
IdoAkkerman: To prevent compatiblity/acceptance issues,  I have changed `ex1.cpp` back. And modified the FEspace initialization of some of the `miniapp/nurbs` to showcase the required changes.

Changes to the examples can be a separate low impact PR, with perhaps a difficult discussion ...

IdoAkkerman:
I added a Constructor to FESpace in order to pass the mater/slave periodic boundaries.

miniapps/nurbs/nurbs_ex1.cpp
```diff
@@ -236,45 +237,33 @@ int main(int argc, char *argv[])
...
-         slave.SetSize(psize);
-         master.Load(in, psize);
-         slave.Load(in, psize);
-         in.close();
-      }
-      NURBSext->ConnectBoundaries(master,slave);
+      mfem_error("Order has size 0");
    }
-   else if (order[0] == -1) // Isoparametric
+   else if (order[0] == 0) // Isoparametric
```
justinlaughlin: I think we should use `order[0] < 1` to be consistent with e.g. `ex1.cpp`

fem/fespace.cpp
```diff
@@ -101,7 +101,63 @@ FiniteElementSpace::FiniteElementSpace(const FiniteElementSpace &orig,
 FiniteElementSpace::FiniteElementSpace(Mesh *mesh,
                                        const FiniteElementCollection *fec,
                                        int vdim, int ordering)
-{ Constructor(mesh, NULL, fec, vdim, ordering); }
+{
```
justinlaughlin, line 104: This may be a personal problem, but I find the nurbs dispatch logic in the fespace constructors quite confusing. I frequently find myself going back to the source to figure out what is going on.

The dispatch logic here in `FiniteElementSpace::FiniteElementSpace(Mesh *mesh,
                                       const FiniteElementCollection *fec,
                                       int vdim, int ordering)` makes sense and is useful imo, despite making the constructor more complicated. One idea is maybe we could split up some of the dispatching of `Constructor` into specific named constructors like `NURBSConstructor`, `IsogeometricConstructor` (uses `mesh->NURBSext`), etc. for the sake of readability?

I'll need to think about this more...

@dylan-copeland  do you have any thoughts/ideas?

fem/fespace.cpp
```diff
@@ -101,7 +101,63 @@ FiniteElementSpace::FiniteElementSpace(const FiniteElementSpace &orig,
...
+                     fec, vdim, ordering);
+      }
+   }
+   else
+   {
+      MFEM_ABORT("NURBSFECollection requires a NURBS-based mesh.");
+   }
+}
+
+FiniteElementSpace::FiniteElementSpace(Mesh *mesh,
```
justinlaughlin: This is just my initial thoughts, but if this isn't too common (is it?) I think I'd prefer if this stays in the examples. It adds a new signature but is nurbs specific (which is not immediately obvious), and it does not work with variable order anyways, meaning we still need to dispatch for it in eg `nurbs_ex1`. We could add another signature that takes variable orders but then the `FiniteElementSpace` constructor is becoming very conflated with `NURBSExtension`.

Maybe we need some kind of helper dispatch/constructor external to `FiniteElementSpace` that returns a `-> (NURBSExtension, NURBSFECollection)`. Then we could place all the nurbs-specific stuff there and keep the `FiniteElementSpace` constructor signatures clean

justinlaughlin: I'm just spitballing - I'm open to any other ideas

fem/fespace.hpp
```diff
@@ -679,8 +683,17 @@ class FiniteElementSpace
...
+   bool IsNURBS() const { return NURBSext != NULL; }
+
+   /// Returns the NURBSExtension
    const NURBSExtension *GetNURBSext() const { return NURBSext; }
+
+   /// Returns the NURBSExtension
    NURBSExtension *GetNURBSext() { return NURBSext; }
+
+   /** Returns the owned NURBSExtension, and disowns NURBSExtension.
+       Throws an error if NURBSextenion is not owned.*/
```
dylan-copeland: `NURBSextenion` -> `NURBSExtension`

fem/fespace.hpp
```diff
@@ -665,6 +665,10 @@ class FiniteElementSpace
                       const FiniteElementCollection *fec,
                       int vdim = 1, int ordering = Ordering::byNODES);
 
+   FiniteElementSpace(Mesh *mesh, Array<int> masterBdr, Array<int> slaveBdr,
```
dylan-copeland: Pass the arrays by reference? Documentation for this new constructor would be helpful.

miniapps/nurbs/nurbs_ex1.cpp
dylan-copeland: Is the convention of using `-1` for the isoparametric case being changed to `0`? There are still sample runs using `-o -1`, which reach line 302 below. Should those sample runs be changed to `-o 0`?

fem/fespace.hpp
IdoAkkerman: Corrected

fem/fespace.hpp
IdoAkkerman: Passed by reference and add some documentation

miniapps/nurbs/nurbs_ex1.cpp
IdoAkkerman: Changed meaning of the order input to make `-1`  isoparametric, sticking with the convention.

fem/fespace.hpp
```diff
@@ -679,8 +693,17 @@ class FiniteElementSpace
...
+   bool IsNURBS() const { return NURBSext != NULL; }
+
+   /// Returns the NURBSExtension
    const NURBSExtension *GetNURBSext() const { return NURBSext; }
+
+   /// Returns the NURBSExtension
    NURBSExtension *GetNURBSext() { return NURBSext; }
+
+   /** Returns the owned NURBSExtension, and disowns NURBSExtension.
+       Throws an error if NURBSExtenion is not owned.*/
```
dylan-copeland: `NURBSExtenion` -> `NURBSExtension`

miniapps/nurbs/nurbs_ex1.cpp
```diff
@@ -288,15 +281,33 @@ int main(int argc, char *argv[])
...
       }
+      fespace = new FiniteElementSpace(mesh, NURBSext, fec);
+   }
+   else if (mesh->NURBSext && (order[0] > 0))
+   {
+      if (order.Size() == 1)
+      {
+         fec = new NURBSFECollection(order[0]);
+         own_fec = 1;
+         fespace = new FiniteElementSpace(mesh,master,slave,fec);
```
dylan-copeland: Spaces after commas, also below.

dylan-copeland (approved):

justinlaughlin:
I've made some proposed changes in [5071](https://github.com/mfem/mfem/pull/5071) - please let me know what you think

miniapps/nurbs/nurbs_ex1.cpp
dylan-copeland: This was resolved recently.

fem/fespace.cpp
IdoAkkerman: I think I see your point, about transparency, there is definitely some technical debt that needs to  be addressed.
In user code such as the examples and miniapps the impact is minor and thing will  be more clear.

However, I think there is also a use case for the more opaque (auto magic) way of doing things.
Having NURBS available in examples and miniapps with very minor (if any?) additional code, is perhaps not the correct motivation.

But there is when dealing with visualisers such as glvis and visit. Here a one-size-fits-all approach would be very helpful.

IdoAkkerman: My above "opaque" statement pertains to the changed (improved?) logic in the constructor with the old signature (lines 105-128). This is also the required part for the GLvis/visit bugfix.

Adding the new signature did not sit right with me either.  I agree that that is a step to far. 
I was not a direct fan of #5071 (because of the std:pair and auto stuff), but I think that would be the way to go wrt the variable order and periodic stuff.

fem/fespace.cpp
IdoAkkerman: Adding the new signature did not sit right with me either. I agree that that is a step to far.
I was not a direct fan of https://github.com/mfem/mfem/pull/5071 (because of the std:pair and auto stuff), but I think that would be the way to go wrt the variable order and periodic stuff, and all future NURBS specific things.

Would it make sense to put this function in a NURBS namespace?

fem/fespace.cpp
justinlaughlin: Yeah one of the challenges of making it auto-magical is the ownership of `fec`. Specifically because in the isoparametric case, it is owned by the mesh/gridfunction.

What about this - say we take what is in #5071 but make `fec` an argument of `NURBSConstructor`: if its null it will create an isoparametric space. Then we can avoid  std::pair (just have it return a `FiniteElementSpace`). I think I like this solution better than what is currently in #5071, and we can keep all the variable order / periodic stuff in the nurbs specific constructor. I reckon we could probably also simplify some of the dispatch logic internally too by routing other nurbs specific things to the `NURBSConstructor`

justinlaughlin: Another challenge is that the `fec` does not know whether or not it is supposed to be isoparametric. This makes the typical interface of `create fec -> use fec to create fespace` awkward because we also need to pass `order`(s) since that is how the isoparametric property is encoded.

justinlaughlin: Another kind of awkward part of `fec` and NURBS is that the orders are held by the `knotvectors` (I'm not actually sure what the order in the NURBS `fec` is used for). If you, for example, use `mesh.DegreeElevate(3)`, then create an isoparametric space, the `fec` will still be the original one loaded from file (eg `NURBS1`).

IdoAkkerman: I think order in fec is a remnant of the past. When variable order nurbs did not excist. I think the nicest way would be for the fec to hold the order(as int or Array<int>). And extension to get it from fec. 

However this looked like too much work and potentieel controversie for this small bugfix PR.

IdoAkkerman: > Another kind of awkward part of `fec` and NURBS is that the orders are held by the `knotvectors` (I'm not actually sure what the order in the NURBS `fec` is used for). If you, for example, use `mesh.DegreeElevate(3)`, then create an isoparametric space, the `fec` will still be the original one loaded from file (eg `NURBS1`).

You are right. There are some inconsistenties. Previous NURBS feature upgrades have not been done with the greatest care on myside or enough scrutiny on the review side.

justinlaughlin: >I think the nicest way would be for the fec to hold the order(as int or Array). And extension to get it from fec.

I don't quite understand how this would work, since `Order` is a property of the knotvectors. If we also held the order int/array in the nurbs fec we'd have to manage synchronization of state between the two. If a nurbs `fec` *needs* to have an associated order, for some internal reasons, I think I'd prefer a solution where it is computed lazily/on-the-fly, eg grabbing the info from `fespace -> nurbsext -> knotvectors`.

That being said...I did a quick test:

```cpp
   auto [fespace, fec_] = FiniteElementSpace::NURBSConstructor(&mesh, &order);
   NURBSFECollection* fec = fec_.get();
   fec->SetOrder(99);
   NURBSFECollection* mesh_fec = mesh.GetNodes()->OwnFEC();
   mesh_fec->SetOrder(99);
```

It looks like setting the order of the nurbs fec (for both the fespace and mesh) doesn't seem to do anything except change the fec name, at least in this case. It seems that the main purpose of a nurbs fec is when loading a nurbs mesh from file, but after that point the knotvectors hold the actual order info, and the fec is just a relic? I'm not too sure if it is necessary for that either since the knotvectors (with orders) are specified in the file 🤔. I'm starting to think maybe we don't even need order in a nurbs fec.

dylan-copeland: The NURBS mesh files have the degree in the name for the `Nodes` space, e.g. `FiniteElementCollection: NURBS2`. This means the same degree is used in all dimensions, although the code is supposed to support different degrees in each dimension. I guess this means file output and input is not supported for varying degrees. My understanding is that the common usage is with one degree, although this is not completely general. In that case, I would be in favor of getting rid of a single degree associated with the `FiniteElementCollection`, and getting it for each dimension from knotvectors, as Justin mentioned.

justinlaughlin: There is some quote along the lines of "give someone state and they'll have bugs for a day, but give them state in two separate locations that must be kept in sync and they'll have bugs for a lifetime" 😅 . I'd be okay if `Order`(s) were represented in either the fec or knotvectors, but I don't like the idea of having it in both (imo if it has to be in both, one should be referential). I personally think it makes more sense in knotvectors, and this is how it is represented now.

IdoAkkerman: We are kind of in between an old way(16 years ago) of doing things and a new way(8 years ago?) of doing things. So it is fair to say that things are sloppy to say the least, "but it works...".

What I had in mind is, generate the NURBS fec with an order array, and use that when the extension is made (in your static functions for example.) Then via the extension it ends up in the knotvector (as you correctly state).

I the fec the order is not used during the simulation itself it is only used when writing the solution. For robustness it might be good to update the order In the fec based on the extension in certain stages. 

Perhaps we can also remove the order in the solution output and solely rely in the knot vector in the extension. 

In that case  the order in the fec would only be used a way to get the order into the extension at generation stage in a "more natural"* (and implicit) way.

*Beauty is in the eye of the beholder.

I see that the storing the order in the fec can also confusing as it does not really matter for the fec itself.
Perhaps we can preempt this with a clear documentation in the fec?

Perhaps I should have read your comment beter... but yes I think you are right. Order in the fec is not necessary. I would suggest to using the fec as a carrier of the order to the extension, but this can be easily done via other means so that is more interface/api design choice, with effectually no technical impact.

IdoAkkerman: > I guess this means file output and input is not supported for varying degrees.

That is not an illogical conclusion, but I hink the incorrect order in fec actually does not influences it usage.
But this is perhaps worth checking.

Either way this is an inconsistenty worth correcting, I think dropping the order in the fec output should work(as it is not used).
And might be the easiest and best solution, IMO.

justinlaughlin: >I guess this means file output and input is not supported for varying degrees.

I was curious about this, so I tried and I think it is possible (at least, it will run) [[code](https://gist.github.com/justinlaughlin/86f711452230c129e3c9ec4bf852050c)]. When the fec is named just `NURBS` (without an integer at the end) it means it is variable order.

If we look at how a nurbs mesh is loaded (`Mesh::Loader`) the logic is kind of like (I'm simplifying the syntax):
  - `Mesh::ReadNURBSMesh(input)`
    - First, creates the nurbs extension: `ext = NURBSExtension(input)`
      - Load the patch topo
      - Load the rest of the nurbs info: `NURBSExtension::Load(input)`. This reads knotvectors, patches, spacing, etc.; it also calls `SetOrdersFromKnotVectors()`
      - At this point, we have all orders information from the knotvectors 
    - if `ext` has patches -> `fec = NURBSFECollection(NURBSext->GetOrder())` -> `fes = FiniteElementSpace(fec)`. Otherwise, set `read_gf=1`
    - if `read_gf==1` -> calls `Nodes = GridFunction(mesh, input)`
      - Initializes `fes` and `fec = fes->Load(mesh, input)`
        - This then creates a `fec` based on the text in the mesh file [[code](https://github.com/mfem/mfem/blob/master/fem/fespace.cpp#L4518)]
        - if the `fec` text order does not match the extension, a new nurbs extension is created with the `fec` orders
        - There is also some separate logic for the `v1.0` fespace format but I haven't seen any examples using it.

So the text of the `fec` in a mesh file does matter (it will override the orders of the knotvectors), but I don't really understand when you would want this. Otherwise, the way we load in a mesh already begins with defining the nurbs extension, which gets its orders from the knotvectors. I haven't found (yet) a place where nurbs `fec` order matters beyond this.

justinlaughlin: >the incorrect order in fec actually does not influences it usage

I don't think it does, but as a user I'd expect `fec.GetOrder()` to be correct when I call it. I don't think we want to update `fec.orders` everytime we update a knotvector order; we could... but it'd be a lot to maintain for almost no benefit.

justinlaughlin: >Perhaps we can also remove the order in the solution output and solely rely in the knot vector in the extension.

I am in favor of this. I need to investigate a bit more thoroughly how robust our read/write is with variable order (`fec==NURBS` )

IdoAkkerman: > it will override the orders of the knotvector

Are you sure? Where?
My guess would be that  `SetOrdersFromKnotVectors()`  gets it from the KnotVector, is this routine still necessary?

IdoAkkerman: > I don't think it does, but as a user I'd expect `fec.GetOrder()` to be correct when I call it. I don't think we want to update `fec.orders` everytime we update a knotvector order; we could... but it'd be a lot to maintain for almost no benefit.

I agree. The current state of things "work" (kind-off) but are confusing.
Orders are stored in the KnotVector, as they should, other places should probable not store it, I agree.

That being said:
When initializings a FESpace with NURBS, we need to provide a  NURBSExtension to the fespace constructor.  The Extension is generated with an order as input and will pass down the orders to the KnotVector.

What is now the  best/nicest/cleanest/clearest to get the orders in the NURBSExtension.
- My idea was to have the fec "carry" it. But this might be  a step to far in terms of automagically doing things, and creating clarity.
- Perhaps the `NURBSConstructor` function call should have an explicit order in the argument list. This is probally the clearest ( as it is the most explicit) way.

justinlaughlin: > > it will override the orders of the knotvector
> 
> Are you sure? Where? My guess would be that `SetOrdersFromKnotVectors()` gets it from the KnotVector, is this routine still necessary?

https://github.com/mfem/mfem/blob/master/fem/fespace.cpp#L4536. By "override" I mean the orders specified in the knotvectors will be ignored, not overwritten, since a separate nurbsext is created

justinlaughlin: >My idea was to have the fec "carry" it. But this might be a step to far in terms of automagically doing things, and creating clarity.

I kind of see the idea of this, and it is tempting, but I think I would like to remove responsibility from nurbs fec (as its effectively doing almost nothing) instead of giving it more, somewhat artificial, responsibility

IdoAkkerman: I agree. Explicitness and clarity trump niceness and automagically in this case.

justinlaughlin: Thank you for all of the discussion 👍 this was a very educational experience for me

fem/fespace.cpp
```diff
@@ -101,7 +101,62 @@ FiniteElementSpace::FiniteElementSpace(const FiniteElementSpace &orig,
...
+                                       Array<int> &masterBdr, Array<int> &slaveBdr,
+                                       const FiniteElementCollection *fec,
+                                       int vdim, int ordering)
+{
```
justinlaughlin: Should we remove this constructor now?

fem/fespace.hpp
```diff
@@ -661,10 +661,24 @@ class FiniteElementSpace
...
                       const FiniteElementCollection *fec,
                       int vdim = 1, int ordering = Ordering::byNODES);
 
+   /** Construct a FiniteElementSpace based on the given
+       FiniteElementCollection, @a fec.*/
+   /** @note Periodic boundary conditions can be provided via @a masterBdr and
+       @a slaveBdr, this only works for NURBS FE at this point.
+       If FiniteElementCollection, @a fec is a NURBS FE the required
+       NURBSExtension will be generated.*/
+   FiniteElementSpace(Mesh *mesh, Array<int> &masterBdr, Array<int> &slaveBdr,
```
justinlaughlin: remove?

justinlaughlin (changes requested):
Thank you Ido!

fem/fespace.hpp
IdoAkkerman: Done.

fem/fespace.cpp
IdoAkkerman: Done.

justinlaughlin (approved):
Thanks Ido

fem/fespace.hpp
```diff
@@ -1570,6 +1589,40 @@ class FiniteElementSpace
...
+    class helps to dispatch for various cases.
+ */
+struct NURBSSpace
+{
+   enum class Type { H1, Hdiv, Hcurl, };
+
+   std::unique_ptr<FiniteElementSpace> fespace;
+   std::unique_ptr<NURBSFECollection> nurbs_fec;
+   FiniteElementCollection* mesh_fec;
+   FiniteElementCollection* fec;
```
dylan-copeland: It would help to document the purposes of `nurbs_fec`, `mesh_fec`, and `fec`.

fem/fespace.hpp
```diff
@@ -1570,6 +1589,40 @@ class FiniteElementSpace
...
+
+   /** Primary constructor for NURBS FiniteElementSpace. The
+       FiniteElementCollection is related to mesh.GetNodes()->OwnFEC()
+       but with different order(s). Creates a NURBSExtension object that
+       is owned by the returned FiniteElementSpace. */
+   NURBSSpace(Mesh* mesh,
+              Array<int> orders,
```
dylan-copeland, line 1617: Pass `orders` by reference? Document the meaning of the array of orders and the boundary arrays?

fem/fespace.cpp
```diff
@@ -159,6 +183,107 @@ void FiniteElementSpace::CopyProlongationAndRestriction(
...
 
+FiniteElementSpace
+FiniteElementSpace::IsoparametricConstructor(Mesh* mesh, int vdim, int ordering)
+{
+   MFEM_VERIFY(mesh->GetNodes(),
+               "Mesh must have Nodes in order to create an isoparametric space.");
+
+   const FiniteElementCollection* fec = mesh->GetNodes()->FESpace()->FEColl();
+
+   MFEM_ASSERT(fec,
```
dylan-copeland: `MFEM_VERIFY` could be used here without a performance impact.

fem/fespace.cpp
```diff
@@ -159,6 +183,107 @@ void FiniteElementSpace::CopyProlongationAndRestriction(
...
+                     "number of master and slave boundaries do not match!");
+         connect_boundaries = true;
+      }
+   }
+   MFEM_VERIFY(orders.Size() > 0,
+               "NURBSSpace - orders cannot be empty.");
+
+   mesh_fec = mesh->GetNodes()->OwnFEC();
+   NURBSExtension *ext = nullptr;
+   // Case 1 - Isoparametric (convention: order < 1)
```
dylan-copeland, line 232: Isn't the convention `order == -1`?

fem/fespace.cpp
```diff
@@ -159,6 +183,107 @@ void FiniteElementSpace::CopyProlongationAndRestriction(
...
+   if ((orders.Size() == 1) && (orders[0] < 1))
+   {
+      mfem::out << "Generating an Isogeometric FE space" << std::endl;
+      nurbs_fec = nullptr;
+      fec = mesh_fec;
+   }
+   // Case 2/3 - Variable/Fixed order
+   else
+   {
+      int fixed_order = (orders.Size() == 1) ? orders[0] : -1;
```
dylan-copeland: `const int`

miniapps/nurbs/nurbs_ex1.cpp
```diff
@@ -209,105 +210,71 @@ int main(int argc, char *argv[])
...
+      master.SetSize(psize);
+      slave.SetSize(psize);
+      master.Load(in, psize);
+      slave.Load(in, psize);
+      in.close();
    }
 
-   FiniteElementSpace *fespace = new FiniteElementSpace(mesh, NURBSext, fec);
+   // 4. Define a finite element space on the mesh. Here we use NURBS
+   //    basis functions of the specified order(s). If order < 1, we
```
dylan-copeland, line 259: Are we keeping the convention `order == -1`? The condition `order < 1` is mentioned in the file description as well.

miniapps/nurbs/nurbs_ex3.cpp
```diff
@@ -42,7 +42,6 @@ int main(int argc, char *argv[])
    // 1. Parse command-line options.
```
dylan-copeland, line 42: For faster regression testing, could we change the 3D sample run to use `-r 3`?

miniapps/nurbs/nurbs_ex5.cpp
```diff
@@ -5,15 +5,8 @@
...
-//               nurbs_ex5 -m ../../data/beam-hex.mesh
-//               nurbs_ex5 -m ../../data/escher.mesh
-//               nurbs_ex5 -m ../../data/fichera.mesh
-//
-// Device sample runs -- do not work for NURBS:
-//               nurbs_ex5 -m ../../data/escher.mesh -pa -d cuda
-//               nurbs_ex5 -m ../../data/escher.mesh -pa -d raja-cuda
-//               nurbs_ex5 -m ../../data/escher.mesh -pa -d raja-omp
+//               nurbs_ex5 -m ../../data/beam-quad-nurbs.mesh
+//               nurbs_ex5 -m ../../data/beam-hex-nurbs.mesh
```
dylan-copeland: `NURBSExtension::GetDivExtension currently only works for single patch NURBS meshes`

dylan-copeland:
Should this PR have the "bug" label? If so, I'm not sure what the bug is.

IdoAkkerman:
@dylan-copeland The bug is quite subtle, it occurs in glvis and visit when reading in the solution. Without the fix the new FESpace will always use the mesh NURBSExtension, instead building a dedicated one with the correct order (or periodicity)

fem/fespace.hpp
IdoAkkerman: Passed by reference.

fem/fespace.cpp
IdoAkkerman: Changed to `MFEM_VERIFY`

fem/fespace.cpp
IdoAkkerman: done

miniapps/nurbs/nurbs_ex3.cpp
IdoAkkerman: I assume you mean on line 7?
This is LLNL internal regression testing right? Or am I missing something?

miniapps/nurbs/nurbs_ex5.cpp
IdoAkkerman: Removed.

fem/fespace.hpp
IdoAkkerman: Moved it back... This needs to be by value, to allow the other constructor an easy access to this function.
The overhead is pretty minimal (# of knotvectors) and only at the init stage, so I would guess this is acceptable.

fem/fespace.hpp
IdoAkkerman: Removed `mesh_fec` for clarity, and explained the others.

fem/fespace.cpp
IdoAkkerman: The current routine does not allow a fallback to H1, so there is no need to restrict it to `order == -1`.
The current logic is more `robust`, but might also be deemed confusing in relation to other cases.
I do not really mind.

@justinlaughlin what is your opinion?

miniapps/nurbs/nurbs_ex1.cpp
IdoAkkerman: This is currently consistent with the implementation of `NURBSSpace`. Lets see how that discussion pans out.

fem/fespace.hpp
dylan-copeland: I agree that it should be fine in terms of overhead. We should think about the interface, though, since it is unchangeable. `NURBSExtension` has a constructor with `const Array<int> &newOrders` and one with `int newOrder`. This is clearer, in my opinion, since passing by value makes me stop and think of why this is being done. Also, making `NURBSSpace` similar would be more consistent.

miniapps/nurbs/nurbs_ex3.cpp
dylan-copeland: Yes, I meant to make the sample runs fast for the internal regression testing. Thanks for adding `-r 3` to speed up this test.

fem/fespace.hpp
dylan-copeland: Also, a couple of typos: pecified, respectivly

fem/fespace.cpp
justinlaughlin: I thought the isoparametric convention was `order == -1` as well, until I read some of the older nurbs miniapps where it turns out it is `order < 1`. I wrote it this way to conform with that standard but I'm okay with just changing it to `order == -1` consistently throughout.

justinlaughlin: FWIW I think `order == -1` makes more sense. I'm not sure if it is ever used but I see no reason you couldn't have a DG isoparametric space?

justinlaughlin:
@IdoAkkerman the bug you are referring to is from https://github.com/mfem/mfem/issues/5035 right?

IdoAkkerman:
@justinlaughlin Yes! This issue also occurs in visit. 
It originates from initializing the gridfunction and the assumed order of the solution space.
As not ext is provided the fespace will use the ext from the mesh, which is not always correct.

PR #4974, opened 2025-08-08
Hybridization on GPU with nonconforming AMR

pazner:
This PR adds GPU support for the hybridization solver with nonconforming AMR.

It also fixes the "legacy" hybridization assembly on cases with parallel nonconforming AMR and ghost faces (there were zero rows in the matrix, causing hypre to fail - now these are replaced with identity).

There are also a few other changes along the way:
* Fix for when `ParNCMesh::GetFaceNeighbors` was called multiple times
    * (This can happen easily if, e.g., `ParMesh::GetNFbyType` is called before `EnsureNodes` and then after)
* Make sure to call `HostRead` in `SparseMatrix::BooleanMultTranspose`
* Use `unordered_map` instead of `map` in cases where we don't need ordered access
* Unify generic hashing of `std::tuple`
* Cache the `InterpolationManager` in `FiniteElementSpace`
* Clean up and fixes for some Doxygen documentation


tzanio:

general/hash_util.hpp
```diff
@@ -0,0 +1,60 @@
...
+inline size_t VariadicHash() { return 0; }
+
+template<typename TFirst, typename... TRest>
+size_t VariadicHash(const TFirst &first, const TRest&... rest)
+{
+   // The hashing formula here is taken directly from the Boost library, with
+   // the magic number 0x9e3779b9 chosen to minimize hashing collisions.
+   const size_t lhs_hash = std::hash<TFirst>()(first);
+   const size_t rhs_hash = VariadicHash(rest...);
+   return lhs_hash^(rhs_hash + 0x9e3779b9 + (lhs_hash<<6) + (lhs_hash>>2));
```
helloworld922: What do you think of using `mfem::Hasher` in `hash.hpp`? For plain old data this can efficiently concatenate an endless stream of bytes with reasonable collision resistance. For non-trivial types it probably makes the most sense to first try to call `std::hash` on the type, then append the hash bytes onto the stream.

pazner: Sounds good to me! Can you propose the diff?

helloworld922: Sure, here's a patch. It moves `Hasher` to `hash_utils.hpp`, and introduces a `ChainedHasher` type which supports expanding `std::pair`, `std::array`, and `std::tuple` recursively. I also added a specialization for `std::string` since that seems like something you might want to hash directly instead of going through `std::hash`.
```diff
diff --git a/general/hash.hpp b/general/hash.hpp
index 13fcbeb31d..26428792b2 100644
--- a/general/hash.hpp
+++ b/general/hash.hpp
@@ -15,8 +15,8 @@
 #include "../config/config.hpp"
 #include "array.hpp"
 #include "globals.hpp"
+#include "hash_util.hpp"
 
-#include <array>
 #include <cstdint>
 #include <type_traits>
 #include <utility>
@@ -455,75 +455,6 @@ protected:
    int BinSize(int idx) const;
 };
 
-///
-/// @brief streaming implementation for murmurhash3 128 (x64).
-/// Constructs the hash in 3 stages: init, append, finalize.
-///
-struct Hasher
-{
-   /// where the final hash result is stored after finalize. Use data[1] when
-   /// only 64 bits are required.
-   uint64_t data[2] = {0, 0};
-
-private:
-   uint64_t nbytes = 0;
-
-   uint64_t buf_[2] = {0, 0};
-
-public:
-
-   /// resets this hasher back to an initial seed
-   void init(uint64_t seed = 0);
-   void append(const uint8_t *vs, uint64_t bytes);
-
-   void finalize();
-
-private:
-   // add 16 bytes
-   void add_block(uint64_t k1, uint64_t k2);
-
-   // add [1-8] more bytes, then finalize
-   void finalize(uint64_t k1, int num);
-
-   // add [1-15] more bytes, then finalize
-   // 0 < num < 16
-   void finalize(uint64_t k1, uint64_t k2, int num);
-};
-
-/// Helper class for hashing std::pair. Usable in place of std::hash<std::pair<T,U>>
-struct PairHasher
-{
-   template <class T, class V>
-   size_t operator()(const std::pair<T, V> &v) const noexcept
-   {
-      Hasher hash;
-      // chosen randomly with a 2^64-sided dice
-      hash.init(0xfebd1fe69813c14full);
-      hash.append(reinterpret_cast<const uint8_t *>(&v.first), sizeof(T));
-      hash.append(reinterpret_cast<const uint8_t *>(&v.second), sizeof(V));
-      hash.finalize();
-      return hash.data[1];
-   }
-};
-
-/// Helper class for hashing std::array. Usable in place of std::hash<std::array<T,N>>
-struct ArrayHasher
-{
-   template <class T, size_t N>
-   size_t operator()(const std::array<T, N> &v) const noexcept
-   {
-      Hasher hash;
-      // chosen randomly with a 2^64-sided dice
-      hash.init(0xfebd1fe69813c14full);
-      for (size_t i = 0; i < N; ++i)
-      {
-         hash.append(reinterpret_cast<const uint8_t *>(&v[i]), sizeof(T));
-      }
-      hash.finalize();
-      return hash.data[1];
-   }
-};
-
 /// Hash function for data sequences.
 /** Depends on GnuTLS for SHA-256 hashing. */
 class HashFunction
diff --git a/general/hash_util.hpp b/general/hash_util.hpp
index 937c81e8b4..8d8e7a3413 100644
--- a/general/hash_util.hpp
+++ b/general/hash_util.hpp
@@ -12,27 +12,159 @@
 #ifndef MFEM_HASH_UTIL_HPP
 #define MFEM_HASH_UTIL_HPP
 
+#include <array>
 #include <cstddef>
 #include <tuple>
 #include <functional>
 #include <utility>
+#include <cstdint>
 
-namespace
+namespace mfem
 {
 
-inline size_t VariadicHash() { return 0; }
+///
+/// @brief streaming implementation for murmurhash3 128 (x64).
+/// Constructs the hash in 3 stages: init, append, finalize.
+///
+struct Hasher
+{
+   /// where the final hash result is stored after finalize. Use data[1] when
+   /// only 64 bits are required.
+   uint64_t data[2] = {0, 0};
+
+private:
+   uint64_t nbytes = 0;
+
+   uint64_t buf_[2] = {0, 0};
+
+public:
+
+   /// resets this hasher back to an initial seed
+   void init(uint64_t seed = 0);
+   void append(const uint8_t *vs, uint64_t bytes);
+
+   void finalize();
 
-template<typename TFirst, typename... TRest>
-size_t VariadicHash(const TFirst &first, const TRest&... rest)
+private:
+   // add 16 bytes
+   void add_block(uint64_t k1, uint64_t k2);
+
+   // add [1-8] more bytes, then finalize
+   void finalize(uint64_t k1, int num);
+
+   // add [1-15] more bytes, then finalize
+   // 0 < num < 16
+   void finalize(uint64_t k1, uint64_t k2, int num);
+};
+
+template <class T> struct ChainedHasher
 {
-   // The hashing formula here is taken directly from the Boost library, with
-   // the magic number 0x9e3779b9 chosen to minimize hashing collisions.
-   const size_t lhs_hash = std::hash<TFirst>()(first);
-   const size_t rhs_hash = VariadicHash(rest...);
-   return lhs_hash^(rhs_hash + 0x9e3779b9 + (lhs_hash<<6) + (lhs_hash>>2));
-}
+   static void Append(Hasher &hasher, const T &value)
+   {
+      if constexpr (std::is_fundamental_v<T> || std::is_pointer_v<T>)
+      {
+         hasher.append(reinterpret_cast<const uint8_t *>(&value), sizeof(T));
+      }
+      else
+      {
+         std::hash<T> h;
+         auto v = h(value);
+         hasher.append(reinterpret_cast<uint8_t *>(&v), sizeof(v));
+      }
+   }
+};
+
+template <class T, class V> struct ChainedHasher<std::pair<T, V>>
+{
+   static void Append(Hasher &hasher, const std::pair<T, V> &value)
+   {
+      ChainedHasher<T>::Append(hasher, value.first);
+      ChainedHasher<V>::Append(hasher, value.second);
+   }
+};
+
+template <class T, size_t N> struct ChainedHasher<std::array<T, N>>
+{
+   static void Append(Hasher &hasher, const std::array<T, N> &value)
+   {
+      for (size_t i = 0; i < N; ++i)
+      {
+         ChainedHasher<T>::Append(hasher, value[i]);
+      }
+   }
+};
+
+template<class... Ts> struct ChainedHasher<std::tuple<Ts...>>
+{
+private:
+   template <size_t N>
+   static void AppendImpl(Hasher &hasher, const std::tuple<Ts...> &value)
+   {
+      ChainedHasher<std::decay_t<decltype(std::get<N>(value))>>::Append(
+                                                                hasher, std::get<N>(value));
+      if constexpr (N + 1 < sizeof...(Ts))
+      {
+         AppendImpl<N + 1>(hasher, value);
+      }
+   }
+
+public:
+   static void Append(Hasher &hasher, const std::tuple<Ts...> &value)
+   {
+      if constexpr (sizeof...(Ts))
+      {
+         AppendImpl<0>(hasher, value);
+      }
+   }
+};
 
-} // anonymous namespace
+/// Helper class for hashing std::pair. Usable in place of std::hash<std::pair<T,U>>
+struct PairHasher
+{
+   template <class T, class V>
+   size_t operator()(const std::pair<T, V> &v) const noexcept
+   {
+      Hasher hash;
+      // chosen randomly with a 2^64-sided dice
+      hash.init(0xfebd1fe69813c14full);
+      ChainedHasher<std::pair<T, V>>::Append(hash, v);
+      hash.finalize();
+      return hash.data[1];
+   }
+};
+
+/// Helper class for hashing std::array. Usable in place of
+/// std::hash<std::array<T,N>>
+struct ArrayHasher
+{
+   template <class T, size_t N>
+   size_t operator()(const std::array<T, N> &v) const noexcept
+   {
+      Hasher hash;
+      // chosen randomly with a 2^64-sided dice
+      hash.init(0xfebd1fe69813c14full);
+      ChainedHasher<std::array<T, N>>::Append(hash, v);
+      hash.finalize();
+      return hash.data[1];
+   }
+};
+
+/// Helper class for hashing std::array. Usable in place of
+/// std::hash<std::Tuple<Ts...>>
+struct TupleHasher
+{
+   template <class T>
+   size_t operator()(const T &v) const noexcept
+   {
+      Hasher hash;
+      // chosen randomly with a 2^64-sided dice
+      hash.init(0xfebd1fe69813c14full);
+      ChainedHasher<T>::Append(hash, v);
+      hash.finalize();
+      return hash.data[1];
+   }
+};
+}
 
 /// Specialization of std::hash for std::tuple of hashable types.
 template<typename ...Args>
@@ -42,7 +174,8 @@ public:
    /// Returns the hash of the given @a value.
    size_t operator()(const std::tuple<Args...> &value) const
    {
-      return std::apply(VariadicHash<Args...>, value);
+      mfem::TupleHasher h;
+      return h(value);
    }
 };
 
@@ -52,7 +185,19 @@ struct std::hash<std::pair<T1, T2>>
 {
    std::size_t operator()(const std::pair<T1, T2> &value) const
    {
-      return VariadicHash(value.first, value.second);
+      mfem::PairHasher h;
+      return h(value);
+   }
+};
+
+/// Specialization of std::hash for std::array of hashable types.
+template<typename T, size_t N>
+struct std::hash<std::array<T, N>>
+{
+   std::size_t operator()(const std::array<T, N> &value) const
+   {
+      mfem::ArrayHasher h;
+      return h(value);
    }
 };
 
```

helloworld922: It probably makes sense to remove `ArrayHasher`, `TupleHasher`, and `PairHasher` since you can just use `std::hash` now, but I left them for now since at least `ArrayHasher` and `PairHasher` existed previously and I didn't want to break any code.

helloworld922: Alternatively, we could remove the `std::hash` namespace injections since these are standard types, and there's no guarantees that any other user code might also do this.

pazner: Sounds good, I made the changes you suggested. Good point about potential conflict with `std::hash` (I guess you are only supposed to specialize it for user-defined types).

fem/bilinearform.cpp
```diff
@@ -825,14 +825,46 @@ void BilinearForm::FormLinearSystem(const Array<int> &ess_tdof_list, Vector &x,
...
          FormSystemMatrix(ess_tdof_list, A);
-         ConstrainedOperator A_constrained(this, ess_tdof_list);
-         A_constrained.EliminateRHS(x, b);
-         hybridization->ReduceRHS(b, B);
+
+         std::unique_ptr<ConstrainedOperator> A_constrained([&]()
+         {
+            Operator *op;
+            Operator::FormSystemOperator(ess_tdof_list, op);
+            return dynamic_cast<ConstrainedOperator*>(op);
```
helloworld922, line 840: I think `FormSystemOperator` always returns a `ConstrainedOperator`?
```suggestion
            ConstrainedOperator *op;
            Operator::FormSystemOperator(ess_tdof_list, op);
            return op;
```

rw-anderson:
I have tested this in 2D and 3D on multiple problems using dynamic AMR and hybridization solvers and everything looks good. Thank you!

helloworld922 (approved):

fem/bilinearform.cpp
pazner: Yes, at least currently, but it's not guaranteed by the interface.

tzanio:

PR #4834, opened 2025-04-28
GPU Assembly of LOR DG Preconditioner

Toni-ko:
This pr adds a batched gpu assembly of the lor dg preconditioner. New preconditioner is used in the lor solvers miniapp. A unit test is added to compare with legacy implementation. 

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#4834](https://github.com/mfem/mfem/pull/4834) | @Toni-ko | @tzanio | @YohannDudouit + @tzanio + @pazner | 4/30/25 | 9/25/25 | 9/27/25 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
      - [ ] Add the example code to the `ALL_EXE_SRCS` variable.
      - [ ] Make sure `THIS_TEST_OPTIONS` is set correctly for the new example.
   - [ ] List the new example in `doc/CodeDocumentation.dox`.
[...]

PR #4977, opened 2025-08-12
[WIP] IMEX Implementation with DG Convection-Diffusion example [imex-conv-diff-dg]

Toni-ko:
This pull request implements three new features. 

1: We create a new class called `SplitTimeDependentOperator` which implements operators of the form F(u, du/dt, t) = G1(u,t) + G2(u,t). This class is used to implement IMEX ODE Solvers, but in principle can be used for any method which splits the ode into two parts. 

2:  Four new ODE Solvers are implemented. They are different IMEX methods which utilize the new `SplitTimeDependentOperator` class. A new `SplitODESolver` class is created which inherits from `ODESolver` but used to solve split time dependent operators. 

3: The new ODE Solvers are showcased in a new example `ex41` and `ex41p` which solves the convection-diffusion equation using a DG discretization.  `ex41p` makes use of DG LOR preconditioning, implemented in the following PR: https://github.com/mfem/mfem/pull/4834. 

Any feedback, comments, or suggestions are greatly appreciated!


tzanio:

najlkin:
Nice work, I would like to see IMEX solvers directly in MFEM 👍 However, my general concern here is that this PR reinvents the IMEX interfaces which already exist in MFEM. There is already `TimeDependentOperator::EvalMode` with a similar purpose, which  can be then combined with `Mult()` and `ImplicitSolve()`. Or there is `ExplicitMult()` and `ImplicitMult()/SUNImplicitSolve()`. Unfortunately, the interface is fractured, but adding an additional interface does not make it better and rather makes it more confusing perhaps 🤔

tzanio:
any relation to https://github.com/mfem/mfem/pull/2474?

bslazarov:
> Nice work, I would like to see IMEX solvers directly in MFEM 👍 However, my general concern here is that this PR reinvents the IMEX interfaces which already exist in MFEM. There is already `TimeDependentOperator::EvalMode` with a similar purpose, which can be then combined with `Mult()` and `ImplicitSolve()`. Or there is `ExplicitMult()` and `ImplicitMult()/SUNImplicitSolve()`. Unfortunately, the interface is fractured, but adding an additional interface does not make it better and rather makes it more confusing perhaps 🤔

The new SplitTimeOperator is removed and the now the integrators are based on the original TimeOperator with corresponding evaluation mode.

cjvogl:
This is now out of scope for this PR (given what was discussed at the technical meeting today), but if anyone is interested in future work into how refactoring `TimeDependentOperator` into modular interfaces might look (as @bslazarov had suggested at one point), I've started a WIP PR that showcases how an `ARKStepSolver`-specific interface could work: https://github.com/mfem/mfem/pull/5085

linalg/ode.hpp
```diff
@@ -931,6 +938,100 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
...
+/// Class for solving systems of split ODEs:
+/// M dx/dt = f_1(x,t) + f_2(x,t)
+/// where M^{-1}*f_1 and M^{-1}*f_2 are treated differently (e.g.,
+/// explicitly and implicitly).This is the abstract base class for
+/// such ODE solvers expecting a TimeDependentOperator with split
+/// functionality. Setting EvalMode=ADDITIVE_TERM_1 and calling
+/// Mult(...) should return k1=M^{-1}*f_1(x,t). Setting
+/// EvalMode=ADDITIVE_TERM_2 and calling ImplicitSolve(...) should
+/// solve M*k2 = f_2(x+gamma*k2,t).
+class SplitODESolver : public ODESolver
```
sohailreddy:
```suggestion
class ARKODESolver : public ODESolver // Additive Runge Kutta method
private:
const real_t *a_imp, *b_imp, *c_imp; // implicit Butcher tableau coefficients
const real_t *a_exp, *b_exp, *c_exp; // explicit Butcher tableau coefficients
```
The ARS(2,2,2) (in `class IMEXRK2`), ARS(2,3,2) (in `class IMEXRK2_3StageExplicit`) and ARS(3,4,3) (in `class IMEX_DIRK_RK3`) can be implemented/unified as a general ARK IMEX scheme (Eqs. 2.1b - 2.1e in [this paper](https://steveruuth.org/wp-content/uploads/2020/10/ars.pdf)) by using specific explicit and implicit Butcher tableaus. See [ExplicitRKSolver](https://docs.mfem.org/html/classmfem_1_1ExplicitRKSolver.html#:~:text=mfem-,ExplicitRKSolver,-Public%20Member%20Functions) to see how a tableau is used. It's better to implement the general ARK using tableaus so users can use different ARK IMEX schemes by supplying different tableaus.

linalg/ode.hpp
```diff
@@ -931,6 +938,100 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
...
+   using ODESolver::Init;
+   virtual void Init(TimeDependentOperator &f);
+
+   static MFEM_EXPORT std::unique_ptr<SplitODESolver> Select(
+      const int ode_solver_type);
+
+   virtual ~SplitODESolver() { }
+};
+
+class IMEXExpImplEuler : public SplitODESolver
```
sohailreddy: This should be implemented by defining a Butcher tableau (given in [Section 2.1 in this paper](https://steveruuth.org/wp-content/uploads/2020/10/ars.pdf)) that is used by `ARKODESolver`, and not as a separate class

linalg/ode.hpp
```diff
@@ -931,6 +938,100 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
...
+
+   void Step(Vector &x, real_t &t, real_t &dt) override;
+};
+
+
+/**IMEX RK2 Method from "On the Stability of IMEX Upwind gSBP Schemes for 1D Linear Advection‑Difusion Equations" by Sigrun Ortleb.
+ * Same as (2,2,2) from "Implicit-explicit Runge-Kutta
+    methods for time-dependent partial differential equations" by Ascher, Ruuth
+    and Spiteri, Applied Numerical Mathematics (1997).**/
+class IMEXRK2 : public SplitODESolver
```
sohailreddy: This should be implemented by defining a Butcher tableau (given in [Section 2.6 in this paper](https://steveruuth.org/wp-content/uploads/2020/10/ars.pdf)) that is used by `ARKODESolver`, and not as a separate class

linalg/ode.hpp
```diff
@@ -931,6 +938,100 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
...
+   void Init(TimeDependentOperator &f_) override;
+
+   void Step(Vector &x, real_t &t, real_t &dt) override;
+};
+
+
+/**(2,3,2) from "Implicit-explicit Runge-Kutta
+    methods for time-dependent partial differential equations" by Ascher, Ruuth
+    and Spiteri, Applied Numerical Mathematics (1997).**/
+class IMEXRK2_3StageExplicit : public SplitODESolver
```
sohailreddy: This should be implemented by defining a Butcher tableau (given in [Section 2.5 in this paper](https://steveruuth.org/wp-content/uploads/2020/10/ars.pdf)) that is used by `ARKODESolver`, and not as a separate class

linalg/ode.hpp
```diff
@@ -931,6 +938,100 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
...
+public:
+   void Init(TimeDependentOperator &f_) override;
+
+   void Step(Vector &x, real_t &t, real_t &dt) override;
+};
+
+/**(3,4,3) from "Implicit-explicit Runge-Kutta
+    methods for time-dependent partial differential equations" by Ascher, Ruuth
+    and Spiteri, Applied Numerical Mathematics (1997).**/
+class IMEX_DIRK_RK3 : public SplitODESolver
```
sohailreddy: This should be implemented by defining a Butcher tableau (given in [Section 2.7 in this paper](https://steveruuth.org/wp-content/uploads/2020/10/ars.pdf)) that is used by `ARKODESolver`, and not as a separate class

sohailreddy (changes requested):
All the IMEX methods in this PR (`class IMEXRK2`, `class IMEXRK2_3StageExplicit`, `class IMEX_DIRK_RK3`) can be implemented/unified as a general ARK IMEX scheme (Eqs. 2.1b - 2.1e in [this paper](https://steveruuth.org/wp-content/uploads/2020/10/ars.pdf)) by using specific explicit and implicit Butcher tableaus. It's better to implement the general ARK using tableaus (rather than new class for each stage/order scheme) so users can use different ARK IMEX schemes by supplying different tableaus. See [ExplicitRKSolver](https://docs.mfem.org/html/classmfem_1_1ExplicitRKSolver.html#:~:text=mfem-,ExplicitRKSolver,-Public%20Member%20Functions) to see how a tableau is used. Each IMEX scheme in this PR should be implemented as a tableaus that are then used by the general ARK ODESolver.

linalg/ode.hpp
najlkin: A general class like that is implemented in #2474 , so it could be just reused perhaps.

examples/CMakeLists.txt
```diff
@@ -151,7 +153,7 @@ if (MFEM_ENABLE_TESTING)
       ex1 ex3 ex4 ex5 ex6 ex9 ex14 ex22 ex24 ex25 ex26 ex34
```
v-dobrev, line 153: Does `ex41` support device execution? If it does, please:
* add it to this list, and
* add device sample runs in `ex41.cpp`.

examples/CMakeLists.txt
```diff
@@ -151,7 +153,7 @@ if (MFEM_ENABLE_TESTING)
...
       # parallel examples with device support:
       ex1p ex2p ex3p ex4p ex5p ex6p ex7p ex9p ex13p ex14p ex22p ex24p ex25p
-      ex26p ex34p ex35p)
+      ex26p ex34p ex35p ex41p)
```
v-dobrev: In `ex41p.cpp`, add device sample runs. Currently, there are none.

examples/ex41.cpp
```diff
@@ -0,0 +1,579 @@
+//                                MFEM Example 41
+//
+// Compile with: make ex41
+//
+// Sample runs: ex41
```
v-dobrev: For a sample run to be picked up by the script that automatically extracts the sample runs from the source, there must be **two** spaces before `ex41`.

Also, to reduce the current large indentation, the sample runs can start at a new line like this:
```
// Sample runs:
//    ex41
//    ex41 ...
//    ...
```

examples/ex41.cpp
```diff
@@ -0,0 +1,579 @@
...
+//
+// Device sample runs:
+//
+// Description:  This example code solves the time-dependent advection-diffusion
+//               equation du/dt + v.grad(u) - a div(grad(u)) = 0, where v is a
+//               given fluid velocity, a is the diffusion coefficient, and
+//               u0(x)=u(0,x) is a given initial condition.
+//
+//               The example demonstrates the use of Discontinuous Galerkin (DG)
+//               bilinear forms in MFEM (face integrators), and the use of IMEX ODE time integrators.
```
v-dobrev: Re-format to fit in 80 chars please.

examples/ex41.cpp
```diff
@@ -0,0 +1,579 @@
...
+//              ex41 -m ../data/periodic-hexagon.mesh -p 1 -r 2 -dt 0.005 -tf 9
+//              ex41 -m ../data/amr-quad.mesh -p 1 -r 2 -dt 0.002 -tf 9
+//              ex41 -m ../data/star-q3.mesh -p 1 -r 2 -dt 0.001 -tf 9
+//              ex41 -m ../data/star-mixed.mesh -p 1 -r 2 -dt 0.005 -tf 9
+//              ex41 -m ../data/disc-nurbs.mesh -p 1 -r 3 -dt 0.005 -tf 9
+//              ex41 -m ../data/disc-nurbs.mesh -p 2 -r 3 -dt 0.005 -tf 9
+//              ex41 -m ../data/periodic-square.mesh -p 3 -r 4 -dt 0.0025 -tf 9 -vs 20
+//              ex41 -m ../data/periodic-cube.mesh -p 0 -r 2 -o 2 -dt 0.01 -tf 8
+//
+// Device sample runs:
```
v-dobrev, line 20: I mentioned this above but as a reminder: please add device sample runs here that illustrate and test running on device.

examples/ex41p.cpp
```diff
@@ -0,0 +1,741 @@
...
+//              mpirun -np 4 ex41p -m ../data/periodic-hexagon.mesh -p 0 -dt 0.005 -tf 10
+//              mpirun -np 4 ex41p -m ../data/periodic-square.mesh -p 1 -dt 0.005 -tf 9
+//              mpirun -np 4 ex41p -m ../data/periodic-hexagon.mesh -p 1  -dt 0.005 -tf 9
+//              mpirun -np 4 ex41p -m ../data/star-q3.mesh -p 1 -rp 1 -dt 0.001 -tf 9
+//              mpirun -np 4 ex41p -m ../data/disc-nurbs.mesh -p 1 -rp 1 -dt 0.005 -tf 9
+//              mpirun -np 4 ex41p -m ../data/disc-nurbs.mesh -p 2 -rp 1 -dt 0.005 -tf 9
+//              mpirun -np 4 ex41p -m ../data/periodic-square.mesh -rp 2 -dt 0.0025 -tf 9 -vs 20
+//              mpirun -np 4 ex41p -m ../data/periodic-cube.mesh -p 0 -rs 2 -o 2 -dt 0.01 -tf 8
+//
+// Device sample runs:
```
v-dobrev, line 18: I mentioned this before but as a reminder: please add device sample runs here that illustrate and test running on device.

examples/ex41p.cpp
```diff
@@ -0,0 +1,741 @@
+//                                MFEM Example 41 - Parallel Version
+//
+// Compile with: make ex41p
+//
+// Sample runs: mpirun -np 4 ex41p
```
v-dobrev: Similar to the serial case, we need (at least) two spaces before `mpirun`.

Also, the indentation can be reduced by starting the sample runs on a new line like this:
```c++
// Sample runs:
//    mpirun -np 4 ex41p
//    ...
```

examples/ex41p.cpp
```diff
@@ -0,0 +1,741 @@
...
+//
+// Device sample runs:
+//
+// Description:  This example code solves the time-dependent advection-diffusion
+//               equation du/dt + v.grad(u) - a div(grad(u)) = 0, where v is a
+//               given fluid velocity, a is the diffusion coefficient, and
+//               u0(x)=u(0,x) is a given initial condition.
+//
+//               The example demonstrates the use of Discontinuous Galerkin (DG)
+//               bilinear forms in MFEM (face integrators), DG-LOR Preconditioning and the use of IMEX ODE time integrators.
```
v-dobrev: Here and in other places where adding new code or changing existing code, please re-format to fit in 80 chars, if possible. (Sample runs can go beyond that because they are expected to be on one line.)

examples/makefile
```diff
@@ -22,14 +22,14 @@ MFEM_LIB_FILE = mfem_is_not_built
...
 SEQ_EXAMPLES = ex0 ex1 ex2 ex3 ex4 ex5 ex6 ex7 ex8 ex9 ex10 ex14 ex15 ex16 \
    ex17 ex18 ex19 ex20 ex21 ex22 ex23 ex24 ex25 ex26 ex27 ex28 ex29 ex30 \
-   ex31 ex33 ex34 ex36 ex37 ex38 ex39 ex40
+   ex31 ex33 ex34 ex36 ex37 ex38 ex39 ex40 ex41 
 PAR_EXAMPLES = ex0p ex1p ex2p ex3p ex4p ex5p ex6p ex7p ex8p ex9p ex10p ex11p \
    ex12p ex13p ex14p ex15p ex16p ex17p ex18p ex19p ex20p ex21p ex22p ex24p \
    ex25p ex26p ex27p ex28p ex29p ex30p ex31p ex32p ex33p ex34p ex35p ex36p \
-   ex37p ex39p ex40p
+   ex37p ex39p ex40p ex41p
 SEQ_DEVICE_EXAMPLES = ex1 ex3 ex4 ex5 ex6 ex9 ex14 ex22 ex24 ex25 ex26 ex34
```
v-dobrev, line 30: Add `ex41` to this list, if it supports running on device.

fem/lor/lor.cpp
```diff
@@ -48,7 +48,7 @@ void LORBase::AddIntegratorsAndMarkers(BilinearForm &a_from,
    for (int i=0; i<integrators->Size(); ++i)
    {
       BilinearFormIntegrator *integrator = (*integrators)[i];
-      if (markers[i] != nullptr)
+      if (markers[i])
```
v-dobrev: Please refrain from changes like this that don't change anything. In this case, the author may have used the comparison with `nullptr` to make it clear that we are checking a pointer value which may not be automatically clear to others looking at the code.

linalg/ode.hpp
```diff
@@ -203,6 +204,12 @@ class ODESolver
    static MFEM_EXPORT std::unique_ptr<ODESolver> SelectImplicit(
       const int ode_solver_type);
 
+
+   /// Function for selecting the desired IMEX ODESolver
+   /// Returns an ODESolver pointer based on an type
+   /// Caller gets ownership of the object and is responsible for its deletion
+   //static MFEM_EXPORT std::unique_ptr<SplitODESolver> SelectIMEX(const int ode_solver_type);
```
v-dobrev: I think it is simpler to implement this method, instead of adding the class `SplitODESolver` that adds its version of `Select`. Of course, the return value will need to be changed to `std::unique_ptr<ODESolver>` as in the other `Select*` methods.

linalg/ode.hpp
v-dobrev: I'm not sure we need this class -- the `Select` method can be added to `ODESolver` as `SelectIMEX`.

Maybe one reason to keep this class is that it provides a single place where the requirements that all IMEX ODE solvers (derived from this class) have for the `TimeDependentOperator` are explicitly written down -- it needs to implement `Mult` and `ImplicitSolve` with support for setting the eval-mode with `SetEvalMode`. The minimum requirement may actually be to have `Mult` with support for `EvalMode == ADDITIVE_TERM_1` (explicit part) and `ImplicitSolve` with support for `EvalMode == ADDITIVE_TERM_2` (implicit part). This documentation is already written, so it's good to keep it in some form.

v-dobrev: One reason to have specialized implementations is that it can be possible, in special cases, to implement things more efficiently than a general implementation with two tableaus. I'm not sure if the cases implemented here have that advantage.

Speaking generally, this is internal implementation that can improved or reworked later without any impact on the interface and user code.

examples/ex41.cpp
```diff
@@ -0,0 +1,579 @@
...
+      case 3:
+      {
+         const real_t f = M_PI;
+         return sin(f*X(0))*sin(f*X(1));
+      }
+   }
+   return 0.0;
+}
+
+class Implicit_Solver : public Solver
```
v-dobrev, line 154: Please add some documentation of what this class does and how it is used.

examples/ex41.cpp
```diff
@@ -0,0 +1,579 @@
...
+
+   // 2. Read the mesh from the given mesh file. We can handle geometrically
+   //    periodic meshes in this code.
+   Mesh mesh(mesh_file);
+   const int dim = mesh.Dimension();
+
+   // 3. Define the Split ODE solver used for time integration. The IMEX solvers currently
+   // available are: 55 - Forward Backward Euler, 56 - IMEXRK2(2,2,2), 57 - IMEXRK2(2,3,2), and
+   // 58 - IMEX_DIRK_RK3.
+   unique_ptr<SplitODESolver> ode_solver = SplitODESolver::Select(ode_solver_type);
```
v-dobrev: With some simple additions this example can also work with fully implicit time integrators, not just IMEX. It may be good to have that as an option here, at least for comparison. However, this is something that can be added in a later PR.

v-dobrev: An important point is that for fully implicit, one will need to solve a non-symmetric linear system, while with IMEX both solves have a symmetric matrix (mass or mass+diffusion).

linalg/ode.cpp
```diff
@@ -106,6 +111,20 @@ std::unique_ptr<ODESolver> ODESolver::SelectImplicit(int ode_solver_type)
...
+{
+   using ode_ptr = std::unique_ptr<SplitODESolver>;
+   switch (ode_solver_type)
+   {
+      case 55: return ode_ptr(new IMEXExpImplEuler);
+      case 56: return ode_ptr(new IMEXRK2);
+      case 57: return ode_ptr(new IMEXRK2_3StageExplicit);
+      case 58: return ode_ptr(new IMEX_DIRK_RK3);
+      default: MFEM_ABORT("Unknown ODE solver type: " << ode_solver_type );
+   }
```
cjvogl:
```suggestion
   switch (ode_solver_type)
   {
      case 55: return std::make_unique<IMEXExpImplEuler>();
      case 56: return std::make_unique<IMEXRK2>();
      case 57: return std::make_unique<IMEXRK2_3StageExplicit>();
      case 58: return std::make_unique<IMEX_DIRK_RK3>();
      default: MFEM_ABORT("Unknown ODE solver type: " << ode_solver_type );
   }
```
I believe this is entirely a superficial change... so feel free to ignore this suggestion

cjvogl: Definitely feel free to ignore now that I see this is merely following what was done for the `ODESolver::Select` method

linalg/ode.hpp
cjvogl: I'm also in favor of not introducing `SplitODESolver` (and instead adding `ODESolver::SelectIMEX`).  Unless `SplitODESolver` is adding functionality over `ODESolver`, I personally prefer to not add another level of inheritance.

I'll note that "split" is ambiguous in a way that can be problematic if other splittings / partitionings are considered (e.g. nonlinear partitionings).  On the time integrator side of things, the equations this targets are specifically "additively split" or "additively partitioned"

linalg/ode.hpp
```diff
@@ -931,6 +938,100 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
...
+/// explicitly and implicitly).This is the abstract base class for
+/// such ODE solvers expecting a TimeDependentOperator with split
+/// functionality. Setting EvalMode=ADDITIVE_TERM_1 and calling
+/// Mult(...) should return k1=M^{-1}*f_1(x,t). Setting
+/// EvalMode=ADDITIVE_TERM_2 and calling ImplicitSolve(...) should
+/// solve M*k2 = f_2(x+gamma*k2,t).
+class SplitODESolver : public ODESolver
+{
+protected:
+   /// Pointer to the associated SplitTimeDependentOperator.
```
cjvogl: Noting that `SplitTimeDependentOperator` no longer exists.

linalg/ode.hpp
cjvogl: I will note that having all the IMEX solvers inherit directly from `ODESolver` instead of `SplitODESolver` will make inserting a common base class along the lines of `ARKSolver` suggested by @sohailreddy .  Having `SplitODESolver` introduced now as-is opens the door to someone else deriving `SplitODESolver` in a way that makes introducing `ARKSolver` in a future PR more difficult.

v-dobrev: @cjvogl, if `SplitODESolver` is removed, where do you think is the best place to move its documentation? I think we should keep that in some form.

examples/CMakeLists.txt
bslazarov: It does not

examples/CMakeLists.txt
bslazarov: It doesn't support.

examples/ex41.cpp
bslazarov: modified

examples/ex41.cpp
bslazarov: done

examples/ex41.cpp
bslazarov: it does bot support devices

examples/ex41p.cpp
bslazarov: does not support devices

examples/ex41p.cpp
bslazarov: corrected

examples/ex41p.cpp
bslazarov: corrected

examples/makefile
bslazarov: it doesn't

fem/lor/lor.cpp
bslazarov: corrected

linalg/ode.hpp
bslazarov: Modified

linalg/ode.hpp
bslazarov: I moved the documentation to the ODESolver - let me know if you think it should be somewhere else. 
The above is modified.

examples/ex41.cpp
bslazarov: added

examples/ex41.cpp
bslazarov: Such features will be added to upcoming Stokes/Navier-Stokes solvers in the miniapps.

linalg/ode.cpp
bslazarov: I will keep the original as it will use the same style as the other Select methods.

linalg/ode.hpp
bslazarov: Removed

linalg/ode.hpp
bslazarov: This should be implemented in following PRs in order to be tested and consistent. One options is to leave only the basic IMEX integrator in this PR - however as the interface does not change if we modify the code later, I think it is better to have the options available in MFEM now.

linalg/ode.hpp
bslazarov: This should be implemented in following PRs in order to be tested and consistent. One options is to leave only the basic IMEX integrator in this PR - however as the interface does not change if we modify the code later, I think it is better to have the options available in MFEM now.

linalg/ode.hpp
bslazarov: This should be implemented in following PRs in order to be tested and consistent. One options is to leave only the basic IMEX integrator in this PR - however as the interface does not change if we modify the code later, I think it is better to have the options available in MFEM now.

linalg/ode.hpp
bslazarov: This should be implemented in following PRs in order to be tested and consistent. One options is to leave only the basic IMEX integrator in this PR - however as the interface does not change if we modify the code later, I think it is better to have the options available in MFEM now.

linalg/ode.hpp
bslazarov: This should be implemented in following PRs in order to be tested and consistent. One options is to leave only the basic IMEX integrator in this PR - however as the interface does not change if we modify the code later, I think it is better to have the options available in MFEM now.

cjvogl: > @cjvogl, if SplitODESolver is removed, where do you think is the best place to move its documentation? I think we should keep that in some form.

Because the documentation describes how things defined in `TimeDependentOperator` are used, that is where I personally would put it ... but I think `ODESolver` works too with some advantages.

linalg/ode.cpp
```diff
@@ -1277,4 +1296,208 @@ void GeneralizedAlpha2Solver::Step(Vector &x, Vector &dxdt,
...
+   k2_imp.SetSize(n, mem_type);
+   k3_imp.SetSize(n, mem_type);
+   y.SetSize(n, mem_type);
+   z.SetSize(n, mem_type);
+   w.SetSize(n, mem_type);
+}
+
+void IMEXRK2_3StageExplicit::Step(Vector &x, real_t &t, real_t &dt)
+{
+   // WIP
```
najlkin: What is this referring to?

linalg/ode.cpp
```diff
@@ -1277,4 +1296,208 @@ void GeneralizedAlpha2Solver::Step(Vector &x, Vector &dxdt,
...
+   f->SetEvalMode(TimeDependentOperator::ADDITIVE_TERM_1);
+   f->Mult(y, k2_exp);
+
+   //K2_imp = f_2(t + gamma dt, x + dt gamma K2_imp)
+   f->SetEvalMode(TimeDependentOperator::ADDITIVE_TERM_2);
+   f->ImplicitSolve(dt*gamma, x, k2_imp);
+
+   //K3_imp = f_2(t+dt,x + dt(1-gamma)K2_imp + dt gamma K3_imp)
+   f -> SetTime(t + dt);
+   add(x, dt*(1-gamma), k2_imp, z);
```
najlkin: In these DIRK methods, `x` is formed with the same coefficients as the intermediate states, which are then unnecessary. So you may start summing up `x` already here (`x.Add(...)`) and remove `z` completely, using `x` instead. Moreover, `k2_imp` is not needed anymore, so it can be merged with `k3_imp` into one variable.

linalg/ode.cpp
```diff
@@ -1277,4 +1296,208 @@ void GeneralizedAlpha2Solver::Step(Vector &x, Vector &dxdt,
...
+   f->Mult(w, k3_exp);
+
+   //K2_imp = f_2(t + gamma dt, x + dt gamma K2_imp)
+   f->SetTime(t + gamma*dt);
+   f->SetEvalMode(TimeDependentOperator::ADDITIVE_TERM_2);
+   f->ImplicitSolve(dt*gamma, x, k2_imp);
+
+   //K3_imp = f_2(t+dt,x + dt(1-gamma)K2_imp + dt gamma K3_imp)
+   f -> SetTime(t + dt);
+   add(x, dt*(1-gamma), k2_imp, z);
```
najlkin: Also here `z` and `k3_imp` is unnecessary, just sum `x`.

linalg/ode.cpp
```diff
@@ -1277,4 +1296,208 @@ void GeneralizedAlpha2Solver::Step(Vector &x, Vector &dxdt,
...
+   //K2 exp is f_1(t + gamma dt, x + dt gamma K1)
+   f->SetTime(t + gamma*dt);
+   add(x, dt*gamma, k1_exp, y);
+   f->SetEvalMode(TimeDependentOperator::ADDITIVE_TERM_1);
+   f->Mult(y, k2_exp);
+
+   //K3 Exp is f_1(t + dt, x + dt gamma K1_exp + dt (1-gamma) K2_exp)
+   f->SetTime(t + dt);
+   add(x, dt*delta, k1_exp, y);
+   add(y, dt*(1-delta), k2_exp, w);
```
najlkin: Also `w` is unnecessary here, just use `y.Add(..)`

linalg/ode.cpp
```diff
@@ -1277,4 +1296,208 @@ void GeneralizedAlpha2Solver::Step(Vector &x, Vector &dxdt,
...
+   add(x, dt*gamma, k1_exp, y);
+   f->SetEvalMode(TimeDependentOperator::ADDITIVE_TERM_1);
+   f->Mult(y, k2_exp);
+   f->SetEvalMode(TimeDependentOperator::ADDITIVE_TERM_2);
+   f->ImplicitSolve(dt*gamma, x, k2_imp);
+
+   //K3_imp, K3_exp
+   f->SetTime(t + (1+gamma)/2*dt);
+   add(x, dt*a_31, k1_exp, y);
+   add(y, dt*a_32, k2_exp, w);
```
najlkin: `w` unnecessary -> `y.Add()`

linalg/ode.cpp
```diff
@@ -1277,4 +1296,208 @@ void GeneralizedAlpha2Solver::Step(Vector &x, Vector &dxdt,
...
+
+   //K4_imp, K4_exp
+   f->SetTime(t+dt);
+   add(x, dt*a_41, k1_exp, y);
+   add(y, dt*a_42, k2_exp, w);
+   add(w, dt*a_43, k3_exp, v);
+   f->SetEvalMode(TimeDependentOperator::ADDITIVE_TERM_1);
+   f->Mult(v, k4_exp);
+   add(x, dt*b1, k2_imp, z);
+   add(z, dt*b2, k3_imp, u);
```
najlkin: `z`, `u` and `k4_imp` unnecessary, start summing `x` already

linalg/ode.cpp
```diff
@@ -1277,4 +1296,208 @@ void GeneralizedAlpha2Solver::Step(Vector &x, Vector &dxdt,
...
+   f->Mult(w, k3_exp);
+   add(x, dt*(1-gamma)/2, k2_imp, z);
+   f->SetEvalMode(TimeDependentOperator::ADDITIVE_TERM_2);
+   f->ImplicitSolve(dt*gamma, z, k3_imp);
+
+   //K4_imp, K4_exp
+   f->SetTime(t+dt);
+   add(x, dt*a_41, k1_exp, y);
+   add(y, dt*a_42, k2_exp, w);
+   add(w, dt*a_43, k3_exp, v);
```
najlkin: `w` and `v` unnecessary -> `y.Add()`

linalg/ode.hpp
```diff
@@ -106,6 +106,15 @@ class ODEStateDataVector : public ODEStateData
...
 /// Abstract class for solving systems of ODEs: dx/dt = f(x,t)
+/// For systems of split ODEs:
+/// M dx/dt = f_1(x,t) + f_2(x,t)
+/// where M^{-1}*f_1 and M^{-1}*f_2 are treated differently (e.g.,
+/// explicitly and implicitly), the solver class
+/// expects a TimeDependentOperator with split
+/// functionality. Setting EvalMode=ADDITIVE_TERM_1 and calling
+/// Mult(...) should return k1=M^{-1}*f_1(x,t). Setting
+/// EvalMode=ADDITIVE_TERM_2 and calling ImplicitSolve(...) should
+/// solve M*k2 = f_2(x+gamma*k2,t).
```
najlkin:
```suggestion
/** For systems of split ODEs:
    $$ M dx/dt = f_1(x,t) + f_2(x,t) $$
    where $ M^{-1} f_1 $ and $ M^{-1} f_2 $ are treated differently (e.g.,
    explicitly and implicitly), the solver class expects a
    TimeDependentOperator with split functionality. Setting
    TimeDependentOperator::EvalMode = TimeDependentOperator::ADDITIVE_TERM_1
    and calling TimeDependentOperator::Mult() should return
    $ k_1=M^{-1} f_1(x,t) $. Setting TimeDependentOperator::EvalMode =
    TimeDependentOperator::ADDITIVE_TERM_2 and calling
    TimeDependentOperator::ImplicitSolve() should solve
    $ M k_2 = f_2(x+\gamma k_2,t) $. */
```

linalg/ode.hpp
```diff
@@ -931,6 +948,68 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
 
 };
 
+
+class IMEXExpImplEuler : public ODESolver
```
najlkin: A short docstring here please ☝️

linalg/ode.hpp
```diff
@@ -931,6 +948,68 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
 
 };
 
+
```
najlkin:
```suggestion
/// Forward-backward Euler method
```

linalg/ode.hpp
```diff
@@ -931,6 +948,68 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
...
+   void Step(Vector &x, real_t &t, real_t &dt) override;
+};
+
+
+/**IMEX RK2 Method from "On the Stability of IMEX Upwind gSBP Schemes for 1D
+ * Linear Advection‑Difusion Equations" by Sigrun Ortleb.
+ * Same as (2,2,2) from "Implicit-explicit Runge-Kutta
+    methods for time-dependent partial differential equations" by Ascher, Ruuth
+    and Spiteri, Applied Numerical Mathematics (1997).**/
```
najlkin:
```suggestion
/// Second order, two-stage implicit-explicit (IMEX) Runge-Kutta (RK) method
/** L-stable IMEX RK2 method adopted from "On the Stability of IMEX Upwind gSBP
    Schemes for 1D Linear Advection‑Difusion Equations" by Sigrun Ortleb. Same
    as (2,2,2) from "Implicit-explicit Runge-Kutta methods for time-dependent
    partial differential equations" by Ascher, Ruuth and Spiteri, Applied
    Numerical Mathematics (1997). */
```

linalg/ode.hpp
```diff
@@ -931,6 +948,68 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
...
+public:
+   void Init(TimeDependentOperator &f_) override;
+
+   void Step(Vector &x, real_t &t, real_t &dt) override;
+};
+
+
+/**(2,3,2) from "Implicit-explicit Runge-Kutta
+    methods for time-dependent partial differential equations" by Ascher, Ruuth
+    and Spiteri, Applied Numerical Mathematics (1997).**/
```
najlkin:
```suggestion
/// Second order, 2/3-stage implicit-explicit (IMEX) Runge-Kutta (RK) method
/** L-stable method (2,3,2) from "Implicit-explicit Runge-Kutta methods for
    time-dependent partial differential equations" by Ascher, Ruuth and
    Spiteri, Applied Numerical Mathematics (1997). */
```

linalg/ode.hpp
```diff
@@ -931,6 +948,68 @@ class WBZAlphaSolver : public GeneralizedAlpha2Solver
...
+   Vector y; Vector z; Vector w;
+public:
+   void Init(TimeDependentOperator &f_) override;
+
+   void Step(Vector &x, real_t &t, real_t &dt) override;
+};
+
+/**(3,4,3) from "Implicit-explicit Runge-Kutta
+    methods for time-dependent partial differential equations" by Ascher, Ruuth
+    and Spiteri, Applied Numerical Mathematics (1997).**/
```
najlkin:
```suggestion
/// Third order, 3/4-stage implicit-explicit (IMEX) Runge-Kutta (RK) method
/** L-stable method (3,4,3) from "Implicit-explicit Runge-Kutta methods for
    time-dependent partial differential equations" by Ascher, Ruuth and
    Spiteri, Applied Numerical Mathematics (1997). */
```

linalg/ode.cpp
```diff
@@ -28,8 +28,14 @@ std::string ODESolver::ImplicitTypes  =
    "        GA      : 40 -- 50  - Generalized-alpha,\n\t"
    "        AM      : 51 - AM1, 52 - AM2, 53 - AM3, 54 - AM4\n";
 
+std::string ODESolver::IMEXTypes =
+   "\n\tIMEX solver: \n\t"
+   "           55 - Forward Backward Euler, 56 - IMEXRK2(2,2,2), 57 - IMEXRK2(2,3,2)\n\t"
+   "           58 - IMEX_DIRK_RK3\n";
```
najlkin:
```suggestion
   "        (L-Stab): 61 - Forward Backward Euler, 62 - IMEXRK2(2,2,2),\n\t"
   "                  63 - IMEXRK2(2,3,2), 64 - IMEX_DIRK_RK3\n";
```

linalg/ode.cpp
```diff
@@ -106,6 +112,19 @@ std::unique_ptr<ODESolver> ODESolver::SelectImplicit(int ode_solver_type)
...
 
+std::unique_ptr<ODESolver> ODESolver::SelectIMEX(const int ode_solver_type)
+{
+   using ode_ptr = std::unique_ptr<ODESolver>;
+   switch (ode_solver_type)
+   {
+      case 55: return ode_ptr(new IMEXExpImplEuler);
+      case 56: return ode_ptr(new IMEXRK2);
+      case 57: return ode_ptr(new IMEXRK2_3StageExplicit);
+      case 58: return ode_ptr(new IMEX_DIRK_RK3);
```
najlkin:
```suggestion
      // L-stable IMEX methods
      case 61: return ode_ptr(new IMEXExpImplEuler);
      case 62: return ode_ptr(new IMEXRK2);
      case 63: return ode_ptr(new IMEXRK2_3StageExplicit);
      case 64: return ode_ptr(new IMEX_DIRK_RK3);
```

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+   args.PrintOptions(cout);
+
+   // 2. Read the mesh from the given mesh file. We can handle geometrically
+   //    periodic meshes in this code.
+   Mesh mesh(mesh_file);
+   const int dim = mesh.Dimension();
+
+   // 3. Define the IMEX (Split) ODE solver used for time integration. The IMEX
+   // solvers currently available are: 55 - Forward Backward Euler,
+   // 56 - IMEXRK2(2,2,2), 57 - IMEXRK2(2,3,2), and  58 - IMEX_DIRK_RK3.
```
najlkin:
```suggestion
   // 3. Define the IMEX (Split) ODE solver used for time integration. The IMEX
   // solvers currently available are: 61 - Forward Backward Euler,
   // 62 - IMEXRK2(2,2,2), 63 - IMEXRK2(2,3,2), and  64 - IMEX_DIRK_RK3.
```

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+{
+   // 1. Parse command-line options.
+   int problem = 0;
+   const char *mesh_file = "../data/periodic-square.mesh";
+   int ref_levels = 2;
+   int order = 3;
+   int ode_solver_type = 58; //55 - Forward Backward Euler
+   //56 - IMEXRK2(2,2,2)
+   //57 - IMEXRK2(2,3,2)
+   //58 - IMEXRK3(3,4,3)
```
najlkin:
```suggestion
   int ode_solver_type = 64; //IMEXRK3(3,4,3)
```

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+
+using namespace std;
+using namespace mfem;
+
+// Mesh bounding box
+Vector bb_min, bb_max;
+
+// Velocity coefficient
+template<int problem=0>
+void velocity_function(const Vector &x, Vector &v)
```
najlkin, line 43: Move the definitions to the end of the file please to clean it up here. The approach with a global `problem` variable like in other examples would be more consistent and would not require if statements in main, but I am ok with templates too.

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+   unique_ptr<Implicit_Solver> implicit_solver;
+
+   mutable Vector z;
+
+public:
+   IMEX_Evolution(BilinearForm &M_, BilinearForm &K_, BilinearForm &S_,
+                  const Vector &b_);
+
+   /// Evaluate k1=M^{-1}*G1(u,t); -> k1 = M^{-1}*(K*u + b)
+   void Mult1(const Vector &x, Vector &y) const ;
```
najlkin:
```suggestion
   void Mult1(const Vector &x, Vector &y) const;
```

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+
+   /// Evaluate k2: M*k2 = G2(u+k2*dt,t); -> (M+S*dt)*k2=-S*u
+   void ImplicitSolve2(const real_t dt, const Vector &x, Vector &k);
+
+   void Mult(const Vector &x, Vector &y) const override
+   {
+      if (TimeDependentOperator::EvalMode::ADDITIVE_TERM_1==GetEvalMode())
```
najlkin:
```suggestion
      if (GetEvalMode() == TimeDependentOperator::EvalMode::ADDITIVE_TERM_1)
```

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+      }
+      else
+      {
+         mfem_error("TimeDependentOperator::Mult() is not overridden!");
+      }
+   }
+
+   void ImplicitSolve(const real_t dt, const Vector &x, Vector &k) override
+   {
+      if (TimeDependentOperator::EvalMode::ADDITIVE_TERM_2==GetEvalMode())
```
najlkin:
```suggestion
      if (GetEvalMode() == TimeDependentOperator::EvalMode::ADDITIVE_TERM_2)
```

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+   }
+   FiniteElementSpace fes(&mesh, fec);
+
+   cout << "Number of unknowns: " << fes.GetVSize() << endl;
+
+   // 6. Set up and assemble the bilinear and linear forms corresponding to the
+   //    DG discretization. The DGTraceIntegrator involves integrals over mesh
+   //    interior faces.
+   std::unique_ptr<VectorFunctionCoefficient> velocity;
+   if (0==problem)
```
najlkin, line 362: `switch(problem)` would be better or the classical global `int problem` variable

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+   int ode_solver_type = 58; //55 - Forward Backward Euler
+   //56 - IMEXRK2(2,2,2)
+   //57 - IMEXRK2(2,3,2)
+   //58 - IMEXRK3(3,4,3)
+   real_t t_final = 10.0;
+   real_t dt = 0.001;
+   bool paraview = false;
+   bool cg = false;
+   int vis_steps = 50;
+   real_t diffusion_term = 0.01;
```
najlkin, line 277: This could have an input option.

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+   }
+
+   ConstantCoefficient diff_coeff(diffusion_term);
+
+   BilinearForm m(&fes);
+   BilinearForm k(&fes);
+   BilinearForm s(&fes);
+
+   Vector b(fes.GetTrueVSize());
+   b=0.0; //The inflow on the boundaries is set to zero.
```
najlkin:
```suggestion
   b = 0.0; //The inflow on the boundaries is set to zero.
```

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+   //57 - IMEXRK2(2,3,2)
+   //58 - IMEXRK3(3,4,3)
+   real_t t_final = 10.0;
+   real_t dt = 0.001;
+   bool paraview = false;
+   bool cg = false;
+   int vis_steps = 50;
+   real_t diffusion_term = 0.01;
+   real_t kappa = -1.0;
+   real_t sigma = -1.0;
```
najlkin, line 279: These also do not have input options, but I think it suffices to move them down to the integrators and make them `constexpr` like `alpha`.

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+   k.Assemble(skip_zeros);
+   s.Assemble(skip_zeros);
+
+   m.Finalize(skip_zeros);
+   k.Finalize(skip_zeros);
+   s.Finalize(skip_zeros);
+
+   // 7. Define the initial conditions.
+   std::unique_ptr<FunctionCoefficient> u0;
+   if (0==problem)
```
najlkin, line 416: Preferably `switch` or the global.

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+      pv->SetPrefixPath("ParaView");
+      pv->RegisterField("solution", &u);
+      pv->SetLevelsOfDetail(order);
+      pv->SetDataFormat(VTKFormat::BINARY);
+      pv->SetHighOrderOutput(true);
+      pv->SetCycle(0);
+      pv->SetTime(0.0);
+      pv->Save();
+   }
+   socketstream sout;
```
najlkin, line 475:
```suggestion

   socketstream sout;
```

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+      M_prec = make_unique<DSmoother>(M.SpMat());
+      M_solver.SetOperator(M.SpMat());
+      implicit_solver = make_unique<Implicit_Solver>(M.SpMat(), S.SpMat(),
+                                                     *M.FESpace());
+   }
+   else
+   {
+      M_prec = make_unique<OperatorJacobiSmoother>(M, ess_tdof_list);
+      M_solver.SetOperator(M);
+      implicit_solver = NULL;
```
najlkin: Rather abort here, because ` ImplicitSolver` can be extended for device support in the future, but `NULL` is not a sensible value.

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+// Description:  This example code solves the time-dependent advection-diffusion
+//               equation du/dt + v.grad(u) - a div(grad(u)) = 0, where v is a
+//               given fluid velocity, a is the diffusion coefficient, and
+//               u0(x)=u(0,x) is a given initial condition.
+//
+//               The example demonstrates the use of Discontinuous Galerkin (DG)
+//               bilinear forms in MFEM (face integrators), and the use of IMEX
+//               ODE time integrators.
+//
+//               The Option to use Continuous Finite Elements is available too.
```
najlkin:
```suggestion
//               The option to use continuous finite elements is available too.
```

examples/ex41.cpp
```diff
@@ -0,0 +1,578 @@
...
+}
+
+void IMEX_Evolution::ImplicitSolve2(const real_t dt, const Vector &x, Vector &k)
+{
+   // Perform the implicit step
+   // solve for k, k = -(M+dt S)^{-1} S x
+   MFEM_VERIFY(implicit_solver != NULL,
+               "Implicit time integration is not supported with partial assembly");
+   S.Mult(x, z);
+   z*= -1.0;
```
najlkin:
```suggestion
   z.Neg();
```

examples/ex41p.cpp
najlkin, line 1: Everything applies like in the serial version ☝️ .

examples/ex41p.cpp
```diff
@@ -0,0 +1,710 @@
...
+   ~Implicit_Solver() override
+   {
+      delete A;
+   }
+};
+
+/** A time-dependent operator for the right-hand side of the ODE. The DG weak
+    form of the advection-diffusion equation is (M + dt S) du/dt = Su - K u + b, where M and K are the mass
+    and advection matrices, and b describes the flow on the boundary. In the case of IMEX evolution, the diffusion term is treated
+    implicitly, and the advection term is treated explicitly.  */
```
najlkin: Wrap at 80 chars, please.

linalg/ode.cpp
bslazarov: fixed

linalg/ode.cpp
bslazarov: I would prefer to keep the implementation like this for better readability. As discussed earlier - these method will be reformulated in the future and in this case it is much better to have a easy to understand code.

linalg/ode.cpp
bslazarov: same as above

examples/ex41.cpp
bslazarov: I think these should be fine - in general I would like to see what is defined before being used.

examples/ex41.cpp
bslazarov: I like the old order.

examples/ex41.cpp
bslazarov: I would prefer to keep the old order.

examples/ex41.cpp
bslazarov: The above works fine

examples/ex41.cpp
bslazarov: added

examples/ex41.cpp
bslazarov: not modified - kappa is modified on line 316. Since these are valid only for DG I would prefer to keep them internal for the example.

examples/ex41.cpp
bslazarov: the if statement works fine in this case

examples/ex41.cpp
bslazarov: corrected

examples/ex41p.cpp
bslazarov: done

linalg/ode.cpp
najlkin: As @v-dobrev argued [above](https://github.com/mfem/mfem/pull/4977#discussion_r2462273896), it makes sense to keep low-order time integrators implemented separately for efficiency, but it is definitely not the case of this implementation. However, it is also not general like integrators with a Butcher table, so it is not anything. I can understand that [the partial summation of the final state in the final implicit stage](https://github.com/mfem/mfem/pull/4977#discussion_r2519860159) can be a bit confusing, even when it is obvious from the construction. But a different thing is this linear combination with an unnecessary variable for every summation. It is rather confusing, because it keeps you wondering why one needs all those temporary vectors instead of just accumulating the contributions. Considering these ODE solvers are in the context of FEM and the method of lines, where the vectors contain the whole spatial solution, so we are talking about unjustifiably wasting massive amount of memory. As other parts of algorithms typically do not need many state vectors, these ODE solvers can be the limiting factor for the size of the simulation.

fem/lor/lor.cpp
```diff
@@ -48,7 +48,7 @@ void LORBase::AddIntegratorsAndMarkers(BilinearForm &a_from,
    for (int i=0; i<integrators->Size(); ++i)
    {
       BilinearFormIntegrator *integrator = (*integrators)[i];
-      if (markers[i] != nullptr)
+      if (markers[i]!= nullptr)
```
najlkin: Accidental change? Please remove.
```suggestion
      if (markers[i] != nullptr)
```

examples/ex41.cpp
najlkin: But looks ugly 😄

examples/ex41.cpp
najlkin: I think it is more readable to start with the logically highest level, which is `main()` and then go to details. At least in most examples it is like that.

cjvogl (approved):
Looks good to me as soon as the change to LOR that @najlkin found is addressed ,i.e., https://github.com/mfem/mfem/pull/4977/files#r2561353488

examples/ex41p.cpp
najlkin: Make a diff with the serial version to see the differences.

helloworld922 (approved):

examples/ex41.cpp
bslazarov: That is a personal preference :-)

linalg/ode.cpp
bslazarov: modified

linalg/ode.cpp
bslazarov: modified

linalg/ode.cpp
bslazarov: modified

linalg/ode.cpp
bslazarov: modified

examples/ex41.cpp
bslazarov: not in my eyes

linalg/ode.cpp
```diff
@@ -1401,27 +1401,30 @@ void IMEXRK2_3StageExplicit::Step(Vector &x, real_t &t, real_t &dt)
 
    //K3 Exp is f_1(t + dt, x + dt gamma K1_exp + dt (1-gamma) K2_exp)
```
najlkin: `gamma` -> `delta`

linalg/ode.cpp
```diff
@@ -1401,27 +1401,30 @@ void IMEXRK2_3StageExplicit::Step(Vector &x, real_t &t, real_t &dt)
...
    f->SetTime(t + dt);
-   add(x, dt*delta, k1_exp, y);
```
najlkin: Forgot this line!

bslazarov: modified

linalg/ode.cpp
najlkin:
```suggestion
   //K3 Exp is f_1(t + dt, x + dt delta K1_exp + dt (1-delta) K2_exp)
```

tzanio:

linalg/ode.cpp
bslazarov: modified

v-dobrev:

tzanio:

examples/ex41p.cpp
```diff
@@ -0,0 +1,720 @@
...
+   int vis_steps = 50;
+   bool adios2 = false;
+   bool binary = false;
+   real_t diffusion_term = 0.01;
+   real_t kappa = -1.0;
+   real_t sigma = -1.0;
+   bool visualization = false;
+   bool visit = false;
+   int precision = 16;
+   cout.precision(precision);
```
v-dobrev, line 322: This extra precision causes the output to look like this:
```
...
time step: 50, time: 0.05000000000000004
...
```
which seems unnecessary and is different from the serial version and the original `ex9`/`ex9p`.

Do we want to keep it like this? @bslazarov, what do you think?

v-dobrev:
In some of the sample runs, e.g.
```console
mpirun -np 4 ex41p -m ../data/periodic-square.mesh -p 1 -dt 0.005 -tf 9 -no-vis
```
I see the output from the preconditioner construction (BoomerAMG) two times: one time in the beginning and a second time just before the last timestep. This does not seem right.

examples/ex41p.cpp
bslazarov: The above output is related to the behavior below - since the last time step is different  the the previous the solver will construct new preconditioner.  The last time step is different due to numerical accumulation of errors and the accumulation can be seen in the output of the time. 

So yes - we can dump out the first digits only - but then the users might ask why do we compute preconditioner again in the last time step.

examples/ex41.cpp
```diff
@@ -0,0 +1,598 @@
...
+#if defined(MFEM_USE_DOUBLE)
+      epsilon = std::numeric_limits<double>::epsilon();
+      epsilon*=10;
+#elif defined(MFEM_USE_SINGLE)
+      epsilon = std::numeric_limits<float>::epsilon();
+      epsilon*=10;
+#else
+#error "Only single and double precision are supported!"
+      epsilon = std::numeric_limits<float>::epsilon();
+#endif
```
v-dobrev: Why not simplify this using `std::numeric_limits<real_t>`?

bslazarov: done

v-dobrev:
Adding the tolerance for changing `dt` did not actually change the result since the actual difference in `dt` on the last step is slightly above `2.7569e-15` (this is for the sample run I posted above). However, this is probably fine.

One option is to add a message from rank 0 that the time step is changing -- that will make it more obvious why the preconditioner is being rebuilt, e.g. something like:
```c++
      if (fabs(ddt) > epsilon)
      {
         int rank;
         MPI_Comm_rank(M.GetComm(), &rank);
         if (rank == 0)
         {
            cout << "switching dt: " << dt << " -> " << dt_ << endl;
         }
         // ...
      }
```

bslazarov:
> Adding the tolerance for changing `dt` did not actually change the result since the actual difference in `dt` on the last step is slightly above `2.7569e-15` (this is for the sample run I posted above). However, this is probably fine.
> 
> One option is to add a message from rank 0 that the time step is changing -- that will make it more obvious why the preconditioner is being rebuilt, e.g. something like:
> 
> ```c++
>       if (fabs(ddt) > epsilon)
>       {
>          int rank;
>          MPI_Comm_rank(M.GetComm(), &rank);
>          if (rank == 0)
>          {
>             cout << "switching dt: " << dt << " -> " << dt_ << endl;
>          }
>          // ...
>       }
> ```

I have to sync it between the processes - for the parallel run. One of the processes might exceed eps and the others might not.   Apart from this situation the behavior is fine - I don't want to increase  the tolerance more.

najlkin:
Use `MFEM_MPI_CXX_BOOL` to fix the CI instead of the native one (but unavailable on Windows).

v-dobrev:

PR #4947, opened 2025-07-19
Trace assembly optimizations

helloworld922:
Optimizations to speed up assembly on traces. Extends #4725.

- Move face information from FaceQuadratureSpace to Mesh
- Added an interface to CUB/hipCUB for single-pass inclusive and exclusive scans
- Compress the FaceQuadratureSpace offsets in some situations
- Added a way to indicate to integrators that the mesh/fespace has been updated. Allows caching some information between assemblies.

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#4947](https://github.com/mfem/mfem/pull/4947) | @helloworld922 | @tzanio | @pazner + @camierjs | 8/5/25 | 9/8/25 | 9/17/25 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
      - [ ] Add the example code to the `ALL_EXE_SRCS` variable.
[...]

PR #4979, opened 2025-08-13
[WIP] Improvements to Vector + Array

j-signorelli:
This PR adds the following member functions:

- `Array<T>::DeleteAt` (with unit test)
- `Vector::DeleteAt` (with unit test)
- `Vector::Reserve`

### DIFF
--- a/general/array.hpp
+++ b/general/array.hpp
@@ -211,6 +211,9 @@ public:
    /// Delete the first entry with value == 'el'.
    inline void DeleteFirst(const T &el);
 
+   /// Delete entries at @a indices, and resize.
+   inline void DeleteAt(const Array<int> &indices);
+
    /// Delete the whole array.
    inline void DeleteAll();
 
@@ -933,6 +936,30 @@ inline void Array<T>::DeleteFirst(const T &el)
          return;
       }
    }
+}
+
+template <class T>
+inline void Array<T>::DeleteAt(const Array<int> &indices)
+{
+   // Make a copy of the indices, sorted.
+   Array<int> sorted_indices(indices);
+   sorted_indices.Sort();
+
+   int rm_count = 0;
+   for (int i = 0; i < size; i++)
+   {
+      if (rm_count < sorted_indices.Size() && i == sorted_indices[rm_count])
+      {
+         rm_count++;
+      }
+      else
+      {
+         data[i-rm_count] = data[i]; // shift data rm_count
+      }
+   }
+
+   // Resize to remove tail
+   size -= rm_count;
 }
 
 template <class T>
--- a/linalg/vector.hpp
+++ b/linalg/vector.hpp
@@ -171,6 +171,9 @@ public:
    /// Resize the vector to size @a s using the MemoryType of @a v.
    void SetSize(int s, const Vector &v) { SetSize(s, v.GetMemory().GetMemoryType()); }
 
+   /// Update \ref Capacity() to @a res (if less than current), keeping existing entries. 
+   void Reserve(int res);
+
    /// Delete entries at @a indices and resize vector accordingly.
    void DeleteAt(const Array<int> &indices);
 
@@ -622,6 +625,17 @@ inline void Vector::SetSize(int s, MemoryType mt)
       size = 0;
    }
    data.UseDevice(use_dev);
+}
+
+inline void Vector::Reserve(int res)
+{
+   if (res > Capacity())
+   {
+      Vector copy = *this;
+      SetSize(res);
+      SetVector(copy, 0);
+      SetSize(copy.Size());
+   }
 }
 
 inline void Vector::DeleteAt(const Array<int> &indices)
--- a/linalg/vector.hpp
+++ b/linalg/vector.hpp
@@ -171,6 +171,9 @@ public:
    /// Resize the vector to size @a s using the MemoryType of @a v.
    void SetSize(int s, const Vector &v) { SetSize(s, v.GetMemory().GetMemoryType()); }
 
+   /// Delete entries at @a indices and resize vector accordingly.
+   void DeleteAt(const Array<int> &indices);
+
    /// Set the Vector data.
    /// @warning This method should be called only when OwnsData() is false.
    void SetData(real_t *d) { data.Wrap(d, data.Capacity(), false); }
@@ -619,6 +622,29 @@ inline void Vector::SetSize(int s, MemoryType mt)
       size = 0;
    }
    data.UseDevice(use_dev);
+}
+
+inline void Vector::DeleteAt(const Array<int> &indices)
+{
+   // Make copy of the indices, sorted.
+   Array<int> sorted_indices(indices);
+   sorted_indices.Sort();
+
+   int rm_count = 0;
+   for (int i = 0; i < size; i++)
+   {
+      if (rm_count < sorted_indices.Size() && i == sorted_indices[rm_count])
+      {
+         rm_count++;
+      }
+      else
+      {
+         data[i-rm_count] = data[i]; // shift data rm_count
+      }
+   }
+
+   // Resize to remove tail
+   size -= rm_count;
 }
 
 inline void Vector::NewMemoryAndSize(const Memory<real_t> &mem, int s,
--- a/tests/unit/general/test_array.cpp
+++ b/tests/unit/general/test_array.cpp
@@ -124,3 +124,19 @@ TEST_CASE("Array stl-interactions", "[Array]")
       CHECK(x[i] == y[i]);
    }
 }
+
+TEST_CASE("Array delete at indices", "[Array]")
+{
+   Array<int>       test({0,1,2,3,4,5,6,7,8});
+   Array<int> rm_indices({0,    3,4,  6,  8});
+   Array<int>     result({  1,2,    5,  7  });
+
+   test.DeleteAt(rm_indices);
+
+   REQUIRE(test.Size() == result.Size());
+
+   for (int i = 0; i < test.Size(); i++)
+   {
+      CHECK(test[i] == result[i]);
+   }
+}
--- a/tests/unit/linalg/test_vector.cpp
+++ b/tests/unit/linalg/test_vector.cpp
@@ -247,3 +247,19 @@ TEST_CASE("Vector Sum", "[Vector],[GPU]")
 
    REQUIRE(sum_1 == MFEM_Approx(sum_2));
 }
+
+TEST_CASE("Vector delete at indices", "[Vector]")
+{
+   Vector           test({0,1,2,3,4,5,6,7,8});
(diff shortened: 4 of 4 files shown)


tests/unit/general/test_array.cpp
```diff
@@ -124,3 +124,19 @@ TEST_CASE("Array stl-interactions", "[Array]")
...
+
+   test.DeleteAt(rm_indices);
+
+   REQUIRE(test.Size() == result.Size());
+
+   for (int i = 0; i < test.Size(); i++)
+   {
+      CHECK(test[i] == result[i]);
+   }
+}
```
tzanio:
```suggestion
}

```

linalg/vector.hpp
```diff
@@ -621,6 +627,40 @@ inline void Vector::SetSize(int s, MemoryType mt)
...
 }
 
+inline void Vector::Reserve(int res)
+{
+   if (res > Capacity())
+   {
+      Vector copy = *this;
+      SetSize(res);
+      SetVector(copy, 0);
+      SetSize(copy.Size());
```
pazner:
```suggestion
      Memory<T> p(res, data.GetMemoryType());
      p.CopyFrom(data, size);
      p.UseDevice(data.UseDevice());
      data.Delete();
      data = p;
```

general/array.hpp
```diff
@@ -935,6 +938,30 @@ inline void Array<T>::DeleteFirst(const T &el)
    }
 }
 
+template <class T>
+inline void Array<T>::DeleteAt(const Array<int> &indices)
+{
+   // Make a copy of the indices, sorted.
```
pazner, line 979: If we want to do this on device, it would probably make sense to use the scan operations introduced by @helloworld922 in #4947.

For host-only, this is probably fine (with the addition of `HostReadWrite()` at the beginning).

j-signorelli:
The failing test seems to be unrelated to this PR... Could someone please assist?

pazner:
> The failing test seems to be unrelated to this PR... Could someone please assist?

It is unrelated - this will be addressed separately. Once it is fixed, you can merge master in this branch and it should resolve the issue.

v-dobrev:
Check the failures again -- some of them are related to this PR, e.g. https://github.com/mfem/mfem/actions/runs/17081226589/job/48435347687?pr=4979.

linalg/vector.hpp
```diff
@@ -621,6 +627,41 @@ inline void Vector::SetSize(int s, MemoryType mt)
    data.UseDevice(use_dev);
 }
 
+inline void Vector::Reserve(int res)
+{
+   if (res > Capacity())
+   {
+      Memory<T> p(res, data.GetMemoryType());
```
pazner:
```suggestion
      Memory<real_t> p(res, data.GetMemoryType());
```

Sorry, bad copy-paste earlier from `Array<T>`. This should work, but please check!

j-signorelli: Sorry for not double-checking!

pazner: No worries! 😄 My bad for the broken suggestion - I just didn't check myself.

general/array.hpp
j-signorelli: Ok I _think_ I understand but just to clarify, are you referring to an algorithm like this (from [here](https://www.sciencedirect.com/science/article/pii/S0304397522001232)):
Screenshot 2025-08-20 at 10 21 12 AM

If yes, I can probably implement this myself. Should I just cherry-pick the commit from that PR with the scan operations?

pazner: Yes, exactly. PR #4947 adds support for efficiently performing operations like partial sum on the GPU using the CUB library, so you don't need to implement that part yourself. Then just grabbing the data from the correct indices is pretty straightforward in an mfem::forall loop.

j-signorelli: OK I have something working but a couple of things:
- Is there a particular way I should include scan.hpp? I can't include it directly without needing to add `#include "array.hpp"` to the top of scan.hpp -- should I just do that?
- Because scan.hpp depends on `Array`, I had to move `DeleteAt` to array.cpp, so it will only be use-able for the compiled Array types. This will be slightly problematic for #4986 , as it uses `Array<unsigned int>` for particle IDs, which fails to compile for `Array<T>::Abs` (since `unsigned int` is an ambiguous argument to `std::abs`). We could use an `if constexpr` with a type check `std::is_same<T, unsigned int>`? But wasn't sure if this is getting too messy at this point

Please let me know thoughts on this, and thanks a ton!

pazner: Yeah, this is a little tricky.

Actually almost none of `Array<T>`'s member functions execute on device. So maybe the easiest solution is just to leave the CPU-only version.

For `Vector::DeleteAt` this won't be a problem.

If you're really interested in `Array<T>::DeleteAt` on device, then it can be implemented as a free function.

pazner: Here is a possible implementation for Vector:

https://github.com/mfem/mfem/blob/653a61045505db61a555b059f5e7aac72500864f/linalg/vector.cpp#L1256-L1294

helloworld922: cub also has algorithms which can directly compact a flag-based stream in `device_select.cuh`. I think `cub::DeviceSelect::Flagged` would be a suitable fit, however it doesn't support inplace compaction. I think that's generally true for all GPU compaction algorithms.

helloworld922: @pazner do you think it would be helpful to wrap a few of these other cub functions? `cub::DeviceSelect::Flagged` also has an easy CPU-based implementation using `std::remove_if`. Some other functions which might be useful to wrap are radix sort, merge sort, merge, partition, and batched versions of these.

j-signorelli: @pazner thank you for the code! I've added your commits and also updated the tests to check for both host and device, in a similar way that the test_scan.cpp file does.

As for Array, it is definitely important I think to also have it on Device... For now I have added the `HostReadWrite` at the beginning as per your suggestion. I see that that PR is still being updated, but if there is any way to hide usage of Array, that would be great. Otherwise a free function somewhere might be needed for Array.

pazner: @helloworld922, I think it could be great! I bet the CUB version is more performant than the one I provided.

@j-signorelli, I am pinging here @v-dobrev, @camierjs, @tzanio regarding Array<T> on device. There was discussion some time ago, but I don't recall what the conclusion was. For now, operations on Array are all basically host-only, but I agree that there is a lot of utility to providing device versions. Do we have a plan for introducing more device operations for Array?

v-dobrev: Generally speaking, if a device version of an `Array` method is needed in MFEM (the library sources) or in an MFEM example/miniapp, then it makes sense to add it. This can be done similar to how we handle the device dispatch in methods of class `Vector` -- based on the "use-device" flag.

pazner: Doing this in generality would require including forall.hpp in array.hpp, etc., which I think (though I could be wrong) we have intentionally avoided.

At the least, things like operator= with a constant value would be useful/convenient to have on device (when the device flag is set).

tzanio:

general/array.hpp
j-signorelli: Just following up on this - I also think long term it would definitely be best to have more Array operations on Device.

In regards to this PR, right now we could leave it as it currently is, where the `Array<T>::DeleteAt` is only on host. This is less ideal for particles but works. Alternatively, to get a `Array<T>::DeleteAt` on Device that works for unsigned integers, since we have to then implement `Array<T>::DeleteAt` in array.cpp in order to include scan.hpp, we could add an `if constexpr (...check type here...) {...}` to the beginning of `Array<T>::Abs`, wrapping all of its contents, so `std::abs` is only executed when the type is valid. This can be done with either an explicit check on just `unsigned int` or on a custom type trait we'd have to define for any general non-`std::abs` argument. 

There may also be a way to disable `Array<T>::Abs` altogether for invalid types using `std::enable_if<...check type here...>` / SFINAE, but I wasn't able to find a way to make it work that was simple.

Please let me know any thoughts! It may make more sense to leave it as is for now, and have a more general porting of array functions to device in a separate PR.

pazner: In a recent change to MFEM, we now include `forall.hpp` from `mfem.hpp`, and so there is no problem including `forall.hpp` (and hence also having GPU kernels) in `array.hpp`.

This greatly simplifies things, and I believe the way forward is relatively straightforward: we can simply remove `array.cpp` (I think the effects on compile-time are negligible), and include everything in `array.hpp`. This will take care of things like `Array<T>::Abs`, that are only well-defined for specific types `T`.

I have a proof-of-concept in the branch [array-device](https://github.com/mfem/mfem/compare/array-device).

linalg/vector.cpp
```diff
@@ -1252,4 +1253,44 @@ real_t Vector::Sum() const
...
+      {
+         // Transform inclusive scan to exclusive by shifting.
+         const int j = (i > 0) ? d_out_idx[i - 1] : 0;
+         d_out[j] = d_in[i];
+      }
+   });
+
+   // Get the new size of the vector. Copy only the last entry.
+   Memory<int> submem(out_idx.GetMemory(), out_idx.Size() - 1, 1);
+   size = submem.Read(MemoryClass::HOST, 1)[0];
```
helloworld922: Using the new scan wrappers
```suggestion
   if (indices.Size())
   {
      const bool use_dev = UseDevice();

      // extra entry for number of selected out
      Array<int> workspace(size + 1);
      const auto d_flag = workspace.Write(use_dev);
      mfem::forall_switch(use_dev, size,
                          [=] MFEM_HOST_DEVICE(int i) { d_flag[i] = true; });
      const auto d_indices = indices.Read(use_dev);
      mfem::forall_switch(use_dev, indices.Size(), [=] MFEM_HOST_DEVICE(int i)
      {
         // fine as long as indices are unique; to support non-unique indices
         // assignment to d_flag must be atomic
         d_flag[d_indices[i]] = false;
      });

      Vector copy(*this);
      auto d_in = copy.Read(use_dev);
      auto d_out = Write(use_dev);
      CopyFlagged(use_dev, d_in, d_flag, d_out, d_flag + size, size);

      // assumes indices are unique
      size -= indices.Size();
   }
```

general/array.hpp
j-signorelli: Going to resolve this conversation from our discussion + decision to have GPU-related updates for Array addressed later.

linalg/vector.cpp
j-signorelli: Should we include in the documentation that indices must be unique?

pazner: Yes, I think that that makes things clearer for the user.

helloworld922 (approved):

tzanio:
@v-dobrev and @pazner -- can we move forward with this?

pazner (approved):

tzanio:

PR #4981, opened 2025-08-13
[WIP] `MultiVector`

j-signorelli:
This PR adds a new class `MultiVector` as a generalization of `GridFunction` for representing a type of `Vector` that holds a number of entries of vector data with a given vdim and `Ordering::Type`.

- Ordering is moved to the file linalg/multivector.cpp/hpp
- A new function `Ordering::Reorder` is added with a unit test
- The new `MultiVector` class is added with unit tests

The long term goal would be to have GridFunction derived from this type.

Note that this PR relies on `Vector::DeleteAt` from #4979 for the function `MultiVector::DeleteVectorsAt`.

### DIFF
--- a/linalg/multivector.hpp
+++ b/linalg/multivector.hpp
@@ -38,6 +38,9 @@ public:
 
    template <Type Ord>
    static void DofsToVDofs(int ndofs, int vdim, Array<int> &dofs);
+
+   /// Reorder Vector \p v from its current ordering \p in_ord to \p out_ord
+   static void Reorder(Vector &v, int vdim, Type in_ord, Type out_ord);
 };
 
 template <> inline int
--- a/linalg/multivector.hpp
+++ b/linalg/multivector.hpp
@@ -17,6 +17,43 @@
 namespace mfem
 {
 
+/** @brief The ordering method used when the number of unknowns per mesh node
+    (vector dimension) is bigger than 1. */
+class Ordering
+{
+public:
+   /// %Ordering methods:
+   enum Type
+   {
+      byNODES, /**< loop first over the nodes (inner loop) then over the vector
+                    dimension (outer loop); symbolically it can be represented
+                    as: XXX...,YYY...,ZZZ... */
+      byVDIM   /**< loop first over the vector dimension (inner loop) then over
+                    the nodes (outer loop); symbolically it can be represented
+                    as: XYZ,XYZ,XYZ,... */
+   };
+
+   template <Type Ord>
+   static inline int Map(int ndofs, int vdim, int dof, int vd);
+
+   template <Type Ord>
+   static void DofsToVDofs(int ndofs, int vdim, Array<int> &dofs);
+};
+
+template <> inline int
+Ordering::Map<Ordering::byNODES>(int ndofs, int vdim, int dof, int vd)
+{
+   MFEM_ASSERT(dof < ndofs && -1-dof < ndofs && 0 <= vd && vd < vdim, "");
+   return (dof >= 0) ? dof+ndofs*vd : dof-ndofs*vd;
+}
+
+template <> inline int
+Ordering::Map<Ordering::byVDIM>(int ndofs, int vdim, int dof, int vd)
+{
+   MFEM_ASSERT(dof < ndofs && -1-dof < ndofs && 0 <= vd && vd < vdim, "");
+   return (dof >= 0) ? vd+vdim*dof : -1-(vd+vdim*(-1-dof));
+}
+
 } // namespace mfem
 
 
--- a/linalg/multivector.hpp
+++ b/linalg/multivector.hpp
@@ -1,0 +1,23 @@
+// Copyright (c) 2010-2025, Lawrence Livermore National Security, LLC. Produced
+// at the Lawrence Livermore National Laboratory. All Rights reserved. See files
+// LICENSE and NOTICE for details. LLNL-CODE-806117.
+//
+// This file is part of the MFEM library. For more information and source code
+// availability visit https://mfem.org.
+//
+// MFEM is free software; you can redistribute it and/or modify it under the
+// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// CONTRIBUTING.md for details.
+
+#ifndef MFEM_MULTIVECTOR
+#define MFEM_MULTIVECTOR
+
+#include "vector.hpp"
+
+namespace mfem
+{
+
+} // namespace mfem
+
+
+#endif // MFEM_MULTIVECTOR
--- a/linalg/multivector.hpp
+++ b/linalg/multivector.hpp
@@ -85,7 +85,7 @@ public:
    /// Initialize a MultiVector with \p num_vectors vectors each of size \p vdim_ ordered \p ordering_.
    MultiVector(int vdim_, Ordering::Type ordering_, int num_vectors);
 
-   /// Initialize a MultiVector of vdim \p vdim_ with ordering \p ordering_ , initialized with copy of data in \p vec . 
+   /// Initialize a MultiVector of vdim \p vdim_ with ordering \p ordering_ , initialized with copy of data in \p vec .
    MultiVector(int vdim_, Ordering::Type ordering_, const Vector &vec);
 
    /// Get the Vector dimension of the MultiVector.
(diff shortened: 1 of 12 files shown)


linalg/multivector.hpp
```diff
@@ -0,0 +1,145 @@
...
+   const real_t& operator()(int i, int comp) const;
+
+   /** @brief Remove Vector s at \p indices.
+    *
+    *  @details The MultiVector is resized appropriately.
+    */
+   void DeleteVectorsAt(const Array<int> &indices);
+
+   /// Set the vector dimension of the MultiVector.
+   void SetVDim(int vdim_) { vdim = vdim_; }
```
kmittal2: I think we should resize and copy the data based on ordering if necessary?

linalg/multivector.hpp
```diff
@@ -0,0 +1,145 @@
...
+   MultiVector(int vdim_, Ordering::Type ordering_, const Vector &vec);
+
+   /// Get the Vector dimension of the MultiVector.
+   int GetVDim() const { return vdim; }
+
+   /// Get the ordering of data in the MultiVector.
+   Ordering::Type GetOrdering() const { return ordering; }
+
+   /// Get the number of Vectors in the MultiVector.
+   int GetNumVectors() const { return Size()/vdim; }
```
kmittal2: It might be worth keeping track of this as `num_vectors` or something similar as a class member.

kmittal2: This should account for when the MultiVector is not initialized i.e. `vdim=0`.

linalg/multivector.hpp
kmittal2: maybe we can include a boolean that specifies if the data will be copied?

linalg/multivector.hpp
```diff
@@ -0,0 +1,145 @@
...
+    *
+    *  @details For \p ordering != \ref GetOrdering , Vector data in the MultiVector is reordered.
+    */
+   void SetOrdering(Ordering::Type ordering_);
+
+   /** @brief Set the number of vectors held by the MultiVector, keeping existing Vectors.
+    *
+    * @details If \p num_vectors * \ref GetVDim > \p Vector::Capacity , memory is re-allocated.
+    */
+   void SetNumVectors(int num_vectors);
```
kmittal2: same here as for `SetVDim`. We should consider optionally copying data if the user requests it.

linalg/multivector.hpp
```diff
@@ -0,0 +1,145 @@
...
+   int GetVDim() const { return vdim; }
+
+   /// Get the ordering of data in the MultiVector.
+   Ordering::Type GetOrdering() const { return ordering; }
+
+   /// Get the number of Vectors in the MultiVector.
+   int GetNumVectors() const { return Size()/vdim; }
+
+   /// Get a copy of Vector \p i 's data.
+   void GetVectorValues(int i, Vector &nvals) const;
```
kmittal2: Could just be `GetVector`

linalg/multivector.hpp
```diff
@@ -0,0 +1,145 @@
...
+   real_t& operator()(int i, int comp);
+
+   /// Const reference to Vector \p i component \p comp value.
+   const real_t& operator()(int i, int comp) const;
+
+   /** @brief Remove Vector s at \p indices.
+    *
+    *  @details The MultiVector is resized appropriately.
+    */
+   void DeleteVectorsAt(const Array<int> &indices);
```
kmittal2: Could just be `DeleteVectors`

linalg/multivector.hpp
j-signorelli: Only place I could see this causing issues is if a `Vector::SetSize` or `Vector::SetDataAndSize` or `Vector::NewDataAndSize` , etc. etc. is called . Maybe instead it's easier to just check if vdim = 0  before that division instead? Otherwise it could be very easy for a `num_vectors` to fall out of sync.

linalg/multivector.hpp
j-signorelli: I'm not quite sure I'm following... Are you saying if for example vdim=3, but then `SetVDim(4)` is called, that we keep the previous data, but allocate/resize/shift so that there is now space for 4 components, default-initialized?

linalg/multivector.hpp
j-signorelli: I think this already handles that - this keeps existing data and just grows size (or reduces) at the tail of the data for `Ordering::byVDIM`, and does the same but with shifting of data for `Ordering::byNODES`.

linalg/multivector.hpp
j-signorelli: Only thing I could see with this is then we should also update `SetVectorValues` --> `SetVector`, which could be confusing since there is a `Vector::SetVector`:

https://github.com/mfem/mfem/blob/f19768cc7073218bbb8abdac5fdb35cd038e2844/linalg/vector.hpp#L351-L353

Should we instead go with maybe `GetValues` and `SetValues`? (or since we will revert this to `ParticleVector`, maybe just `GetParticleValues` and `SetParticleValues`?)

linalg/multivector.hpp
kmittal2: yeah that is a good point.

j-signorelli:
Get/SetComponent and SetVDim capabilities have now been implemented with unit tests. I also added unit tests for get/set vector values.

I think we discussed having a flag to optionally copy/maintain existing data -- should I include this as well?

linalg/multivector.hpp
j-signorelli: Alright we now have `GetValues` and `SetValues`, and `GetComponents` and `SetComponents`. Please advise any thoughts on the latter two (I included the "s" at the end to indicate more clearly that it is _all_ particle `vd`-component values, but this could be confused with multiple differing components).

linalg/multivector.hpp
j-signorelli: Changed to `DeleteParticles`

kmittal2:
> I think we discussed having a flag to optionally copy/maintain existing data -- should I include this as well?

Yes, I think that will be useful.

linalg/multivector.hpp
kmittal2: I am good with the new names.

kmittal2:
@tzanio @v-dobrev @mlstowell @bslazarov @SchmidtMat As discussed yesterday during the PR meeting, we decided to keep the scope of the proposed class narrow and are calling it ParticleVector. It is good to go from @j-signorelli and my side.

tzanio:

tzanio:
thanks @j-signorelli and @kmittal2 !

linalg/ordering.cpp
najlkin, line 1: What was the reason to put it in `general`? I think it would fit better `linalg` as it works with `Vector`s. I do not see a problem with moving it, `fespace.cpp` includes from `linalg` and this file does not have any dependencies except `vector.hpp` and config.

linalg/ordering.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+   int nvdofs = v.Size();
+   int nldofs = nvdofs/vdim;
+
+   if (out_ord == Ordering::byNODES) // byVDIM -> byNODES
+   {
+      Vector temp = v;
+      for (int i = 0; i < nvdofs; i++)
+      {
+         v[i] = temp[Map<byVDIM>(nldofs,vdim,i%nldofs,i/nldofs)];
```
najlkin: For better readability, please put the index on a different line and also I would suggest replacing the division and modulo by nested `for` loop. If the stride is not a power of two and the compiler does not optimize it out, division/modulo are unnecessarily expensive operations compared to multiplication, which is usually handled by address calculation instructions.

linalg/ordering.cpp
j-signorelli: @kmittal2 had suggested it - I just tested it to verify that there are no issues moving it to `linalg` and can confirm it's fine. If there are no objections please let me know and I can push.

tzanio: no objections

tzanio:

tzanio:
Should we mention this in `CHANGELOG`, or is all particle functionality be mentioned in the next PR?

kmittal2:
> Should we mention this in `CHANGELOG`, or is all particle functionality be mentioned in the next PR?

Addressed in d45bdbc.

tzanio:

PR #4994, opened 2025-08-25
[WIP] Enhancement of AutoDiff miniapp

dohyun-cse:
## Description

This PR tries to improve the user interface of AutoDiff miniapps, located in `miniapps/autodiff/`.

This work is presented as a **Request for Comments (RFC)**. The primary goal is to gather community feedback on the proposed design and user experience.

The current implementation prioritizes a **flexible and intuitive interface** that enables rapid development and prototyping of complex physical models.

---

## 🏛️ Design Philosophy

The new implementation uses the same central concept as the previous version of this miniapp: perform auto-differentiation (AD) at the **quadrature point level**. This strategy minimizes computational overhead by differentiating small, well-defined kernel functions rather than the entire global assembly process.

The framework supports both **energy-based** and **residual-based** variational formulations through two primary user-facing classes: `ADFunction` and `ADVectorFunction`. These classes seamlessly integrate with `(Block)NonlinearForm` via `AD(Block)NonlinearFormIntegrator`.

Class integration diagram

---

### Energy-Based Formulation: `ADFunction`

For problems with a known energy functional $F(u) = \int_{\Omega} f(u, \nabla u, ...) dx$, users can derive a class from `ADFunction`. By implementing a single method to define the integrand $f$ via the `AD_IMPL` macro, the framework automatically provides:

-   **Energy:** `NonlinearForm::GetEnergy(u)` evaluates $F(u)$.
-   **Gradient:** `NonlinearForm::Mult(u, dfdu)` evaluates the first variation (the residual).
-   **Hessian:** `NonlinearForm::GetGradient(u)` returns the second variation (the tangent stiffness matrix).

The inputs to the user-defined function are the quantities of interest (e.g., value, gradient, divergence) for each field. For instance, in Darcy's law, the inputs are `[q, divq, u]`, where $q\in R^d$, $divq \in R$, and $u \in R$.

### Residual-Based Formulation: `ADVectorFunction`

For systems where an energy functional is not readily available (e.g., mixed formulations or non-variational problems), users can derive from `ADVectorFunction` and define the residual via the `AD_VEC_IMPL` macro. This class allows for the direct definition of the weak form residual. The framework then applies AD to this residual to compute the correct residual gradient matrix.
For instance, in Darcy's law, the residual is defined as $(q, w) - (u, div w) + (div q, v)$. Then the input is `[q, divq, u]` and the returned residual vector represents `[w_times, divw_times, v_times]=[q, -u, div q]`.

---

## 🚀 Key Features & Capabilities

-   **Broad Finite Element Support**: Works with standard scalar and vector `FiniteElementSpace`, including H(curl) and H(div) spaces.
-   **Mixed Problems**: Natively supports `BlockNonlinearForm` for coupled systems.
-   **Flexible Formulations**: Accommodates both energy-based (`ADFunction`) and residual-based (`ADVectorFunction`) approaches.
-   **Simple User Interface**: Requires only a simple derived class and a macro (`AD_IMPL`, `AD_VEC_IMPL`) to define a new model.
-   **Generic Parameters**: An `Evaluator` class enables easy handling of parameters like material coefficients and fixed grid functions (see `ad_ex3`).
-   **Differentiable Coefficients**: Includes a `DifferentiableCoefficient` to extract value, gradient, and Hessian as a coefficient.

---

## 💡 Examples

The flexibility of this interface is demonstrated through a series of new examples:

-   **Ex0**: Basic definition of `ADFunction` and `ADVectorFunction`.
-   **Ex1**: Poisson's equation using an energy formulation.
-   **Ex2**: Minimal surface problem with regularization.
-   **Ex3**: Linear Elasticity, showcasing vector-valued fields and parameterized coefficients.
-   **Ex4**: Obstacle problem solved with a Proximal Galerkin method (CG + DG), see also `examples/ex36.cpp`.
-   **Ex5**: Gradient obstacle problem (CG + CG).
-   **Ex6**: Darcy's law using a mixed formulation (RT + DG) and a residual-based approach.

---

## Future Work and Discussion

We welcome feedback, especially on the core design.

### Discussion Points
-   **Input Data Structures**: Extracting each quantity of interest (value, grad, div) is not user-friendly. Using a `struct` to hold these derived quantities could improve usability, but might introduce minor overhead. Feedback on this trade-off would be valuable.

### TODO
-   [ ] Face integrator support (high-priority).
-   [ ] Integration with Partial Assembly (PA) and Matrix-Free (MF) methods.
-   [ ] GPU support.


miniapps/autodiff/_ad_intg.hpp
jandrej, line 1: underscore at the beginning of filename

dohyun-cse: This is a templated implementation file that should be included in `ad_intg.hpp`, not on the user side. Is there any convention in MFEM? I can move implementations to the end of `ad_intg.hpp` if this is preferred.

jandrej:
I skimmed through the changes and felt a little confused. Are you aware of the dFEM interface? [See here for an example](https://github.com/mfem/mfem/blob/master/miniapps/dfem/dfem-minimal-surface.cpp). dFEM covers a broad spectrum of functionality of automatic differentiation capabilities and I'm not sure why there is effort in recreating this functionality in different parts of MFEM.

I see multiple strong issues with this PR, but most important

*having multiple implementation of the same/very similar things not just leaves users confused, but more important leaves the wrong impression with sponsors.*

If dFEM does not fulfill your needs, we're happy to see contributions and since it is in an early development state the interface is still flexible.

PR #5001, opened 2025-09-02
Enable CMake fetching of GSLIB (and renaming of FETCH_TPLS, HYPRE_FETCH, and METIS_FETCH)

cjvogl:
This PR enables the fetching and building of GSLIB with the use of either `MFEM_FETCH_TPLS` or `MFEM_FETCH_GSLIB` during the CMake configuration.

This PR also renames existing fetching variables:
- adding "MFEM_" prefix to existing `FETCH_TPLS`, `HYPRE_FETCH`, and `METIS_FETCH`
- renamed `MFEM_HYPRE_FETCH` and `MFEM_METIS_FETCH` to `MFEM_FETCH_HYPRE` and `MFEM_FETCH_METIS`, respectively, so all `MFEM_FETCH_*` variables appear together in alphabetically-sorted list

### DIFF
--- a/config/cmake/modules/FindGSLIB.cmake
+++ b/config/cmake/modules/FindGSLIB.cmake
@@ -9,10 +9,38 @@
 # terms of the BSD-3 license. We welcome feedback and contributions, see file
 # CONTRIBUTING.md for details.
 
-# Defines the following variables:
+# Defines the following variables if fetching of TPLs is disabled (default):
 #   - GSLIB_FOUND
 #   - GSLIB_LIBRARIES
 #   - GSLIB_INCLUDE_DIRS
+# otherwise, the following are defined:
+#   - GSLIB (imported library target)
+
+if (MFEM_FETCH_GSLIB OR MFEM_FETCH_TPLS)
+  set(GSLIB_FETCH_VERSION 1.0.9)
+  add_library(GSLIB STATIC IMPORTED)
+  # define external project and create future include directory so it is present
+  # to pass CMake checks at end of MFEM configuration step
+  message(STATUS "Will fetch GSLIB ${GSLIB_FETCH_VERSION} to be built with default options")
+  set(PREFIX ${CMAKE_BINARY_DIR}/fetch/gslib)
+  include(ExternalProject)
+  ExternalProject_Add(gslib
+    GIT_REPOSITORY https://github.com/Nek5000/gslib
+    GIT_TAG v${GSLIB_FETCH_VERSION}
+    GIT_SHALLOW TRUE
+    UPDATE_DISCONNECTED TRUE
+    PREFIX ${PREFIX}
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND cd ${PREFIX}/src/gslib && make DESTDIR=${PREFIX}
+    INSTALL_COMMAND "")
+  file(MAKE_DIRECTORY ${PREFIX}/include)
+  # set imported library target properties
+  add_dependencies(GSLIB gslib)
+  set_target_properties(GSLIB PROPERTIES
+    IMPORTED_LOCATION ${PREFIX}/lib/libgs.a
+    INTERFACE_INCLUDE_DIRECTORIES ${PREFIX}/include)
+  return()
+endif()
 
 include(MfemCmakeUtilities)
 mfem_find_package(GSLIB GSLIB GSLIB_DIR "include" gslib.h "lib" gs
--- a/INSTALL
+++ b/INSTALL
@@ -123,7 +123,7 @@ Parallel build:
 
 Parallel build with fetching of hypre and METIS:
    mkdir <mfem-buil-dir> ; cd <mfem-build-dir>
-   cmake <mfem-source-dir> -DMFEM_USE_MPI=YES -DFETCH_TPLS=YES
+   cmake <mfem-source-dir> -DMFEM_USE_MPI=YES -DMFEM_FETCH_TPLS=YES
    make -j 4
 
 CUDA build:
@@ -1081,9 +1081,10 @@ The following options are CMake specific:
 MFEM_ENABLE_TESTING  - Enable the ctest framework for testing.
 MFEM_ENABLE_EXAMPLES - Build all of the examples by default.
 MFEM_ENABLE_MINIAPPS - Build all of the miniapps by default.
-FETCH_TPLS           - Enable fetching of all supported third-party libraries.
-HYPRE_FETCH          - Enable fetching of hypre.
-METIS_FETCH          - Enable fetching of metis.
+MFEM_FETCH_TPLS      - Enable fetching of all supported third-party libraries.
+MFEM_FETCH_GSLIB     - Enable fetching of gslib.
+MFEM_FETCH_HYPRE     - Enable fetching of hypre.
+MFEM_FETCH_METIS     - Enable fetching of metis.
 
 External libraries (CMake):
 ---------------------------
@@ -1149,6 +1150,7 @@ The MFEM CMake build system also provides fetching (automated building) for the
 packages/libraries listed below. Note that when fetching is enabled, any related
 auto-detection functionality is disabled.
 
+ - GSLIB
  - HYPRE
  - METIS
 
--- a/config/cmake/modules/FindHYPRE.cmake
+++ b/config/cmake/modules/FindHYPRE.cmake
@@ -36,7 +36,7 @@ if (HYPRE_FOUND OR TARGET HYPRE)
   endif()
 endif()
 
-if (HYPRE_FETCH OR FETCH_TPLS)
+if (MFEM_FETCH_HYPRE OR MFEM_FETCH_TPLS)
   set(HYPRE_FETCH_VERSION 2.33.0)
   add_library(HYPRE STATIC IMPORTED)
   # set options and associated dependencies
--- a/config/cmake/modules/FindMETIS.cmake
+++ b/config/cmake/modules/FindMETIS.cmake
@@ -18,7 +18,7 @@
 #   - METIS (imported library target)
 #   - METIS_VERSION_5 (cache variable)
 
-if (METIS_FETCH OR FETCH_TPLS)
+if (MFEM_FETCH_METIS OR MFEM_FETCH_TPLS)
   set(METIS_FETCH_VERSION 4.0.3)
   add_library(METIS STATIC IMPORTED)
   # define external project
--- a/config/defaults.cmake
+++ b/config/defaults.cmake
@@ -91,9 +91,10 @@ option(MFEM_ENABLE_BENCHMARKS "Build all of the benchmarks" OFF)
 
 # Allow a user to specify fetching of certain third-party libraries instead of
 # searching for existing installations.
-option(FETCH_TPLS "Enable fetching of all supported third-party libraries" OFF)
-option(HYPRE_FETCH "Enable fetching of hypre" OFF)
-option(METIS_FETCH "Enable fetching of METIS" OFF)
+option(MFEM_FETCH_TPLS "Enable fetching of all supported third-party libraries" OFF)
+option(MFEM_GSLIB_FETCH "Enable fetching of GSLIB" OFF)
+option(MFEM_HYPRE_FETCH "Enable fetching of hypre" OFF)
+option(MFEM_METIS_FETCH "Enable fetching of METIS" OFF)
 
 # Setting CXX/MPICXX on the command line or in user.cmake will overwrite the
 # autodetected C++ compiler.
--- a/config/defaults.cmake
+++ b/config/defaults.cmake
@@ -92,9 +92,9 @@ option(MFEM_ENABLE_BENCHMARKS "Build all of the benchmarks" OFF)
 # Allow a user to specify fetching of certain third-party libraries instead of
 # searching for existing installations.
 option(MFEM_FETCH_TPLS "Enable fetching of all supported third-party libraries" OFF)
-option(MFEM_GSLIB_FETCH "Enable fetching of GSLIB" OFF)
-option(MFEM_HYPRE_FETCH "Enable fetching of hypre" OFF)
-option(MFEM_METIS_FETCH "Enable fetching of METIS" OFF)
+option(MFEM_FETCH_GSLIB "Enable fetching of GSLIB" OFF)
+option(MFEM_FETCH_HYPRE "Enable fetching of hypre" OFF)
+option(MFEM_FETCH_METIS "Enable fetching of METIS" OFF)
 
 # Setting CXX/MPICXX on the command line or in user.cmake will overwrite the
 # autodetected C++ compiler.


config/cmake/modules/FindGSLIB.cmake
```diff
@@ -9,10 +9,38 @@
...
+  set(PREFIX ${CMAKE_BINARY_DIR}/fetch/gslib)
+  include(ExternalProject)
+  ExternalProject_Add(gslib
+    GIT_REPOSITORY https://github.com/Nek5000/gslib
+    GIT_TAG v${GSLIB_FETCH_VERSION}
+    GIT_SHALLOW TRUE
+    UPDATE_DISCONNECTED TRUE
+    PREFIX ${PREFIX}
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND cd ${PREFIX}/src/gslib && make DESTDIR=${PREFIX}
```
nmnobre:
```suggestion
    BUILD_COMMAND cd ${PREFIX}/src/gslib && make DESTDIR=${PREFIX} MPI=$<BOOL:${MFEM_USE_MPI}>
```

cjvogl: Great suggestion, thank you!

cjvogl:
> It's awesome that we now have a system we can extend with new TPLs as needed.

@nmnobre , I agree: thank you for your part in getting a good foundation going!

tzanio:

cjvogl:
Noting a good idea from @kmittal2 in a Slack conversation:

- [x] add a “deprecated” marking by explicitly checking for use of old FETCH_TPLS flag and directing the user to new MFEM_FETCH_TPLS flag

kmittal2:
The new workflow works really well.. but an issue is that if I do a serial build followed by a parallel build, gslib is not built the second time around. I am assuming that is because we check if the folder/library already exists. Do we need to address this scenario or at-least note somewhere that the user should remove gslib folder before building again?

k-collie:
> The new workflow works really well.. but an issue is that if I do a serial build followed by a parallel build, gslib is not built the second time around. I am assuming that is because we check if the folder/library already exists. Do we need to address this scenario or at-least note somewhere that the user should remove gslib folder before building again?

I think the issue here might be that GSLIB's Makefile does not rebuild objects when the configuration is changed.

As a quick fix, adding a `make clean` to the `BUILD_COMMAND` might be acceptable since GSLIB is quick to build:
```cmake
    BUILD_COMMAND cd ${PREFIX}/src/gslib && make clean && make DESTDIR=${PREFIX} MPI=$<BOOL:${MFEM_USE_MPI}> "CFLAGS=-fPIC -O2 ${CMAKE_C_FLAGS}" CC=${CMAKE_C_COMPILER}
```
A better fix might be to replace GSLIB's Makefile with CMake, which might get accepted upstream since nekRS builds with CMake.

I think `-fPIC -O2` is a sensible default and it would be good to propagate `CMAKE_C_FLAGS` to GSLIB's CFLAGS as above - I tried to find a way to make CFLAGS correspond to `CMAKE_BUILD_TYPE` but couldn't find a simple way to do this. I needed the `-fPIC` on one machine otherwise I ran into linker errors.

config/cmake/modules/FindGSLIB.cmake
```diff
@@ -9,10 +9,38 @@
...
+  set(PREFIX ${CMAKE_BINARY_DIR}/fetch/gslib)
+  include(ExternalProject)
+  ExternalProject_Add(gslib
+    GIT_REPOSITORY https://github.com/Nek5000/gslib
+    GIT_TAG v${GSLIB_FETCH_VERSION}
+    GIT_SHALLOW TRUE
+    UPDATE_DISCONNECTED TRUE
+    PREFIX ${PREFIX}
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND cd ${PREFIX}/src/gslib && make DESTDIR=${PREFIX} MPI=$<BOOL:${MFEM_USE_MPI}>
```
nmnobre: So, having a spent a bit of time on this, here are my ~2~ 3 cents:

1) `make clean` is in fact the perfect solution @k-collie, no compromises. If the build line doesn't change, CMake won't trigger a rebuild. This means you can run cmake back to back, and make back to back, as long as the build line doesn't change, GSLIB isn't rebuilt. Now, if we rerun CMake to reconfigure with a different value of `MFEM_USE_MPI` or (with my new suggestion below) `BUILD_SHARED_LIBS`, then the build line changes and GSLIB is rebuilt. This is _exactly_ what we want. Perfect.

2) `-fPIC` is needed every time MFEM is built as a shared library. Thanks @k-collie for bringing this up! So it makes sense to include the flag when `BUILD_SHARED_LIBS` is true. Unfortunately we can't append to GSLIB's default flags, so let's also add `-O2` as suggested by @k-collie which the current GSLIB default optimisation flag. Note when MFEM is configured as a static library (which I suspect is a lot of people's preferred setup), we still use GSLIB's defaults (whatever they may become in the future), just as before.

3) I don't think we need to pass `CMAKE_C_FLAGS` or `CMAKE_C_COMPILER` onto GSLIB. MFEM is not a C project, so most users won't define these flags. GSLIB uses sensible defaults anyway and we'd have to treat these carefully which I honestly don't think is worth the extra code/effort/complexity (e.g. if `CMAKE_C_COMPILER` is empty, and we effectively pass `CC=` to GSLIB's make, it crashes).

Putting it all together:

```suggestion
    BUILD_COMMAND cd ${PREFIX}/src/gslib && make clean && make DESTDIR=${PREFIX} MPI=$<BOOL:${MFEM_USE_MPI}> "$<$<BOOL:${BUILD_SHARED_LIBS}>:CFLAGS=-O2 -fPIC>"
```

Hopefully @cjvogl agrees with this? If someone else (@cjvogl, @k-collie, @kmittal2) could test this patch, I think that would put us in a position where we'd be ready to get this over the line. We have a pressing need to get this merged :)

cjvogl: This sounds good to me: thanks to both @nmnobre and @k-collie for a concise fix!

config/cmake/modules/FindGSLIB.cmake
```diff
@@ -9,10 +9,38 @@
...
+  set(PREFIX ${CMAKE_BINARY_DIR}/fetch/gslib)
+  include(ExternalProject)
+  ExternalProject_Add(gslib
+    GIT_REPOSITORY https://github.com/Nek5000/gslib
+    GIT_TAG v${GSLIB_FETCH_VERSION}
+    GIT_SHALLOW TRUE
+    UPDATE_DISCONNECTED TRUE
+    PREFIX ${PREFIX}
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND cd ${PREFIX}/src/gslib && make clean && make DESTDIR=${PREFIX} MPI=$<BOOL:${MFEM_USE_MPI}> "$<$<BOOL:${BUILD_SHARED_LIBS}>:CFLAGS=-O2 -fPIC>"
```
k-collie: Point taken that handling `CMAKE_C_COMPILER` would be a bit messy. It looks like it should be easy to propagate `CMAKE_C_FLAGS` here though - this could be useful for quickly passing in some flags for instrumentation or debugging. I think it would be surprising behaviour for cmake to not propagate these to all C code it builds.

config/cmake/modules/FindGSLIB.cmake
```diff
@@ -9,10 +9,38 @@
...
+#   - GSLIB (imported library target)
+
+if (MFEM_FETCH_GSLIB OR MFEM_FETCH_TPLS)
+  set(GSLIB_FETCH_VERSION 1.0.9)
+  add_library(GSLIB STATIC IMPORTED)
+  # define external project and create future include directory so it is present
+  # to pass CMake checks at end of MFEM configuration step
+  message(STATUS "Will fetch GSLIB ${GSLIB_FETCH_VERSION} to be built with default options")
+  set(PREFIX ${CMAKE_BINARY_DIR}/fetch/gslib)
+  include(ExternalProject)
```
nmnobre, line 35:
```suggestion
  string(TOUPPER "${CMAKE_BUILD_TYPE}" BUILD_TYPE)
  enable_language(C)
  include(ExternalProject)
```

This gives us access to `CMAKE_C_FLAGS_<CONFIG>` so we can align GSLIB's compilation flags with the chosen MFEM build type.

config/cmake/modules/FindGSLIB.cmake
nmnobre:
```suggestion
    BUILD_COMMAND cd ${PREFIX}/src/gslib && $(MAKE) clean && $(MAKE) DESTDIR=${PREFIX} MPI=$<BOOL:${MFEM_USE_MPI}> "CFLAGS=$<$<BOOL:${BUILD_SHARED_LIBS}>:-fPIC> ${CMAKE_C_FLAGS} ${CMAKE_C_FLAGS_${BUILD_TYPE}}"
```

This passes the same C flags, and in the same order, that CMake would use if GSLIB were using CMake as its build system.

cjvogl:
@nmnobre and @k-collie , my initial reaction is that the argument could be made that if the MFEM build calls for dependency-specific flags, then those dependencies should be built external to MFEM. In that light, I'm leaning towards ensuring we have a working "default" build for compilation to avoid taking on the scope of passing dependency-specific options to various build systems of the dependencies.

config/cmake/modules/FindGSLIB.cmake
cjvogl: see https://github.com/mfem/mfem/pull/5001#issuecomment-3457820661

config/cmake/modules/FindGSLIB.cmake
cjvogl: see https://github.com/mfem/mfem/pull/5001#issuecomment-3457820661

cjvogl: see https://github.com/mfem/mfem/pull/5001#issuecomment-3457820661

cjvogl: To expand on my concern, I can see a potential issue with trying to support `CMAKE_C_FLAGS` if someone is building with both GSLIB and SUNDIALS, but wants different flags for the different dependencies. I'll add the default flags while we continue to discuss (thanks for noting how to do that!)

k-collie:
@cjvogl, thanks for the reply, I've just discussed with @nmnobre, to split out the different components:

1. `CMAKE_C_FLAGS_${BUILD_TYPE}` - if I understand correctly we all agree this should be passed through so `GSLIB`'s optimization level matches MFEM's.
2. `CMAKE_C_FLAGS` - @nmnobre and I think that this should be passed through. This is a standard flag initialized from the `CFLAGS` environment variable. My opinion is that if users put flags in this variable that breaks their compilation then it's their fault. We'd like this to be passed through but if you disagree we'd still be happy to see the rest merged without.
3. Hypothetical `MFEM_GSLIB_C_FLAGS` and `MFEM_SUNDIALS_C_FLAGS` - we think we should not support these and instead require people to build the dependency externally if they require a specific unique set of flags for each dependency. They can then set `GSLIB_DIR` etc as usual.

cjvogl:
@k-collie, I hadn't thought about the environment variable use of `CMAKE_C_FLAGS` (I was only thinking of CLI argument). That then makes sense to support.  I've pushed changes that I believe address yours and @nmnobre 's concerns in  [b9bebd0](https://github.com/mfem/mfem/pull/5001/commits/b9bebd0921415234e1dcbb93edb69a91489dea6d). Can you both mark as resolved any prior open threads that are indeed now resolved?

config/cmake/modules/FindGSLIB.cmake
cjvogl: I believe this is addressed in [b9bebd0](https://github.com/mfem/mfem/pull/5001/commits/b9bebd0921415234e1dcbb93edb69a91489dea6d)

config/cmake/modules/FindGSLIB.cmake
cjvogl: I believe this is addressed in [b9bebd0](https://github.com/mfem/mfem/pull/5001/commits/b9bebd0921415234e1dcbb93edb69a91489dea6d)

config/cmake/modules/FindGSLIB.cmake
cjvogl: I believe this is addressed in [b9bebd0](https://github.com/mfem/mfem/pull/5001/commits/b9bebd0921415234e1dcbb93edb69a91489dea6d)

k-collie: Please could someone mark this as resolved - I don't have permissions.

Thanks!

cjvogl: That's odd that you cannot resolve the conversation as the author of the opening comment.  Weird.

cjvogl: I forgot to note somewhere: thank you @nmnobre for wrangling the CMake commands to pass through the C flags!

kmittal2:
I see build errors (e.g., `make pfindpts` in `miniapps/gslib/`) if I do a parallel build and then switch to a serial build. If I build either independently from scratch, there are no errors.

cjvogl:
> I see build errors (e.g., make pfindpts in miniapps/gslib/) if I do a parallel build and then switch to a serial build. If I build either independently from scratch, there are no errors.

@k-collie and @nmnobre , did either of you see such errors?

nmnobre:
> > I see build errors (e.g., make pfindpts in miniapps/gslib/) if I do a parallel build and then switch to a serial build. If I build either independently from scratch, there are no errors.
> 
> @k-collie and @nmnobre , did either of you see such errors?

I tried reproducing but couldn't unfortunately. We don't exactly know what steps @kmittal2 followed or what errors he got, but I tried flipping `-DMFEM_USE_MPI=YES/NO` and `make (p)findpts` in different orders and it always succeeded...

nmnobre:
@cjvogl would you mind cherry-picking https://github.com/farscape-project/mfem/commit/634b8df72b54205f0a08db0be5ef502b79d361e0? I can't trust Github suggestions because the lines I want to edit are too far from the ones touched in this PR. Note that for METIS, we can still say we're building with "default options", because `COPTIONS` adds _extra_ options, but doesn't replace the default optimisation flags.

kmittal2:
@nmnobre @cjvogl The errors were due to me not pulling in the latest changes :|

cjvogl:
> @nmnobre @cjvogl The errors were due to me not pulling in the latest changes :|

@kmittal2 That's an easy fix!

> @cjvogl would you mind cherry-picking https://github.com/farscape-project/mfem/commit/634b8df72b54205f0a08db0be5ef502b79d361e0? I

@nmnobre, I can see how those changes are along the lines of supporting building shared libraries, but I'm hesitant to include them here because they increase the scope of the PR to changing how fetched METIS and HYPRE are built (this PR as-is is about to merged in for testing... but such changes would in my opinion require additional review)

nmnobre:
Not a problem. Leave it out then. :)

cjvogl:
> Not a problem. Leave it out then. :)

Thanks for the understanding: I definitely encourage including those changes in a separate PR though!

tzanio:

tzanio (approved):
Thanks for expanding this functionality, guys 🙏 

I think it will be really useful for many of our users!

PR #5019, opened 2025-09-16
Automatic memory validity flag tracking

helloworld922:
This PR experiments with the idea of how automatic memory validity flag tracking between Memory aliases might work.

# Introduction

Currently when working with Memory objects aliases, users must manually ensure that where they are valid is kept up to sync. For example:

```c++
    // full range: [0, 10)
    Memory<int> base(10);
    // alias0 points to [0, 5)
    Memory<int> alias0(base, 0, 5);
    // alias1 points to [5, 10)
    Memory<int> alias1(base, 5, 5);
    {
      auto ptr = alias1.Write(MemoryClass::HOST);
      for (int i = 0; i < alias1.Size(); ++i)
      {
        ptr[i] = i;
      }
    }
    {
      auto ptr = alias0.Write(MemoryClass::DEVICE);
      forall(alias0.Size(), [=] MFEM_HOST_DEVICE(int i) { ptr[i] = i + 10; });
    }
    // actual valid on device: [0, 5)
    // actual valid on host: [5, 10)
    // flags state:
    // - base is valid on host
    // - alias1 is valid on host
    // - alias0 is valid on device

    // example: we want base to be valid on device
    alias1.SyncAlias(base);

    // now safe to use base however we want
    {
      auto ptr = base.ReadWrite(MemoryClass::HOST);
      for (int i = 0; i < base.Size(); ++i)
      {
        ptr[i] *= 2;
      }
    }
    // need to re-sync aliases now if we want to use them again...
    base.Sync(alias0);
    base.Sync(alias1);
```

This is error prone and difficult to debug, especially if `base` or one of the aliased memories is passed into another function which will do the `Read`/`ReadWrite`/`Write` operation.

The goal of this PR is to remove the need to ever call `Sync` or `SyncAlias`, and just have everything work seamlessly for you.

```c++
    // full range: [0, 10)
    Memory<int> base(10);
    // alias0 points to [0, 5)
    Memory<int> alias0(base, 0, 5);
    // alias1 points to [5, 10)
    Memory<int> alias1(base, 5, 5);
    {
      auto ptr = alias1.Write(MemoryClass:HOST);
      for (int i = 0; i < alias1.Size(); ++i)
      {
        ptr[i] = i;
      }
    }
    {
      auto ptr = alias0.Write(MemoryClass::DEVICE);
      forall(alias0.Size(), [=] MFEM_HOST_DEVICE(int i) { ptr[i] = i + 10; });
    }
    // actual valid on device: [0, 5)
    // actual valid on host: [5, 10)

    {
      // [0, 5) is copied from device to host
      auto ptr = base.ReadWrite(MemoryClass::HOST);
      for (int i = 0; i < base.Size(); ++i)
      {
        ptr[i] *= 2;
      }
    }

    // alias0 and alias1 both are valid on host, no copies are required
    {
      auto ptr0 = alias0.ReadWrite(MemoryClass::HOST);
      auto ptr1 = alias1.Read(MemoryClass::HOST);
      for (int i = 0; i < alias0.Size(); ++i)
      {
        ptr0[i] += ptr1[i];
      }
    }
```

Existing supported features:

- External pointers

```c++
  int buffer[3];
  Memory<int> res(buffer, 3, MemoryType::HOST);
```

- Temporary buffers

```c++
  for (int i = 0; i < 100; ++i)
  {
    // re-uses existing allocation, similar to PR #4065
    Memory<int> res(3, MemoryType::TEMP_HOST);
  }
```

- Support for host-pinned and managed memory spaces

```c++
  // HOSTPINNED and MANAGED memory spaces can also be used for temporary pools
  Memory<int> res(10, MemoryType::HOST_PINNED);
  {
    auto ptr = res.HostWrite();
    for (int i = 0; i < res.Size(); ++i)
    {
      ptr[i] = i;
    }
  }
  {
    // no extra allocation/copies performed
    auto ptr = res.ReadWrite();
    forall(res.Size(), [=] MFEM_HOST_DEVICE (int i) { ptr[i] += 1; });
  }
```

- ResourceManager tracks which Device backends are enabled (or not enabled) so common code can be written which supports CPU-only runs or mixed CPU/Device runs.

See `tests/unit/general/test_resource_manager.cpp` for additional usage features/examples.

# Implementation Details

The `MemoryManager` keeps an internal list of "segments" representing a coupled region of memory. If the coupled regions point to the same memory addresses the segment operates under "zero-copy" rules, useful for avoiding double allocations on host-only runs, or host-pinned/managed buffers.

Validity tracking is performed at the byte level. Each segment uses two red-black tree to mark where where the regions changes from valid to invalid (and vice versa). There is an implicit valid marker at the start of each region. Marking a region as valid or invalid is `O(log(M) + K)`, where `M` is the total number of valid/invalid transitions in the segment and `K` is the number of `valid/invalid` transitions within the region being marked.

Core files:
- `general/internal/mmu.cpp`: Moved the MMU functions from `mem_manager.cpp` so they can be commonly shared between `mem_manager.cpp` and `resource_manager.cpp`.
- `general/internal/rb_tree_impl.hpp`: Core red-black tree implementation.
- `general/internal/reusable_storage.cpp`, - `general/internal/reusable_storage.hpp`: internal helper class used to achieve good performance in tracking red black tree nodes and memory segments
- `general/mem_manager.cpp`, `general/mem_manager.hpp`: updates to be able to easily switch between the two memory managers at compile time. Undefine `MFEM_USE_NEW_MEM_MANAGER` to switch back to the original memory manager. Note that the `MFEM_ENABLE_MEM_OP_DEBUG` and `MFEM_ENABLE_MEM_BENCH` related code are there for debugging and benchmarking purposes. These will be removed before the final merge.
- `general/rb_tree.hpp`: red black tree header file. Uses CRTP for generality (can easily implement interval trees on top of this) and performance.
- `general/resource_manager.hpp`, `general/resource_manager.cpp`: core implementation files for the new memory manager
- `general/internal/mem_op_debug.cpp`: Debugging/benchmark helpers, will be removed before the final merge.

Important changes related to the new memory manager:
- `general/array.hpp`: Added interface to create temporary workspace arrays.
- `general/device.cpp`: Updated interface to work with both the old and new memory managers
- `general/reducers.hpp`, `general/scan.hpp`: Added ability to utilize the new temporary workspaces for working buffers for gpu reduce/scan.
- `linalg/densemat.hpp`: Checks only the entry being read/written to
- `linalg/sparsemat.cpp`, `linalg/sparsemat.cpp`: Changes to try and keep pointers tracked within memory objects as much as possible, instead of extracting the pointer and constructing new memory objects from them. These changes could be reverted, but are generally more robust.
- `linalg/vector.hpp`, `linalg/vector.cpp`: Interface updates to support `CheckedWrap` and creating temporary workspace vectors.
- `tests/unit/general/test_mem.cpp`, `tests/unit/general/test_reduction.cpp`, `tests/unit/miniapps/test_debug_device.cpp`: Updated tests to work with old and new memory manager
- `tests/unit/general/test_resource_manager.cpp`: New tests specific to the new memory manager
- `umpire_unit_test_main.cpp`: Extracted umpire unit test fixes from #4065

Incidental changes (can be moved to a separate PR or reverted if needed):
- `fem/batchitrans.cpp`
- `fem/lor/lor_batched.cpp`
- `fem/lor/lor_batched.hpp`
- `general/table.hpp`
- `linalg/blockmatrix.cpp`
- `linalg/blockoperator.cpp`
- `linalg/blockvector.cpp`
- `miniapps/solvers/div_free_solver.cpp`
- `tests/unit/linalg/test_chebyshev.cpp`

| [#5019](https://github.com/mfem/mfem/pull/5019) | @helloworld922 | @tzanio | @pazner + @tomstitt + @camierjs + @v-dobrev + @tzanio | 6/6/26 | ⌛due 6/20/26 | ⌛due 6/27/26 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:
@helloworld922, can you resolve the conflicts please?

tzanio:

helloworld922:
@pazner I updated the description with a list summarizing the changes in each file.

general/resource_manager.hpp
```diff
@@ -0,0 +1,1708 @@
+#ifndef RESOURCE_MANAGER_HPP
```
camierjs: There are missing MFEM license headers in most of the new files:
- general/internal/mem_op_debug.cpp
- general/internal/mmu.cpp
- general/internal/rb_tree_impl.hpp
- general/internal/reusable_storage.cpp
- general/internal/reusable_storage.hpp
- general/rb_tree.hpp
- general/resource_manager.cpp
- general/resource_manager.hpp
- tests/unit/general/test_resource_manager.cpp

tests/unit/umpire_unit_test_main.cpp
```diff
@@ -9,7 +9,28 @@
...
+// Copyright (c) 2010-2025, Lawrence Livermore National Security, LLC. Produced
+// at the Lawrence Livermore National Laboratory. All Rights reserved. See files
+// LICENSE and NOTICE for details. LLNL-CODE-806117.
+//
+// This file is part of the MFEM library. For more information and source code
+// availability visit https://mfem.org.
+//
+// MFEM is free software; you can redistribute it and/or modify it under the
+// terms of the BSD-3 license. We welcome feedback and contributions, see file
+// CONTRIBUTING.md for details.
```
camierjs: Duplicate license/header block

general/rb_tree.hpp
```diff
@@ -0,0 +1,36 @@
+#ifndef RB_TREE_HPP
```
camierjs: Generic include guards should be MFEM-prefixed

general/internal/rb_tree_impl.hpp
```diff
@@ -0,0 +1,706 @@
+#ifndef RB_TREE_IMPL_HPP
```
camierjs: Generic include guards should be MFEM-prefixed

general/resource_manager.hpp
camierjs: Generic include guards should be MFEM-prefixed

general/resource_manager.hpp
```diff
@@ -0,0 +1,1708 @@
...
+#include "internal/reusable_storage.hpp"
+
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <stdexcept>
+#include <tuple>
+#include <type_traits>
+#include <vector>
```
camierjs, line 27: There are some missing includes: the file uses `std::map`, `std::ostream`, `std::copy` but relies on transitive includes.
There also appears to be some unused include `<stdexcept>`, `<tuple>`, `< ype_traits>`.

general/internal/mmu.cpp
```diff
@@ -0,0 +1,190 @@
+// this file should only be included in mem_manager.cpp or resource_manager.cpp!
+// MFEM_MMU_CPP is defined in the associated cpp files
+#ifdef MFEM_MMU_CPP
```
camierjs, line 14: If all code paths now define `MFEM_MMU_CPP`, is this still necessary?

general/internal/mmu.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+#ifndef _WIN32
+#include <unistd.h>
+#include <signal.h>
+#include <sys/mman.h>
+#endif
+
+#include <cstddef>
+#include <cstdint>
+#include <cstdlib>
```
camierjs, line 25: Some includes might be needed: `cstdio`, `new` ?

tests/unit/linalg/test_chebyshev.cpp
```diff
@@ -21,8 +21,8 @@ TEST_CASE("OperatorChebyshevSmoother", "[Chebyshev symmetry]")
       const int cheb_order = 2;
 
       Mesh mesh = Mesh::MakeCartesian3D(4, 4, 4, Element::HEXAHEDRON);
-      FiniteElementCollection *fec = new H1_FECollection(order, 3);
-      FiniteElementSpace fespace(&mesh, fec);
+      std::unique_ptr<FiniteElementCollection> fec(new H1_FECollection(order, 3));
```
camierjs, line 26: If using `std::unique_ptr`, we should add `<memory>`.

general/internal/mem_op_debug.cpp
```diff
@@ -0,0 +1,513 @@
...
+   {
+      idcs.push_back(0);
+   }
+   return idcs.at(idx)++;
+}
+
+std::ostream &mem_op_debug(size_t idx, int)
+{
+   auto pidx = mem_op_debug(idx);
+   // cannot use mfem::out because of potential static variiable initialization
```
camierjs:
```suggestion
   // cannot use mfem::out because of potential static variable initialization
```

general/internal/reusable_storage.hpp
```diff
@@ -0,0 +1,81 @@
...
+   void EraseInternal(size_t idx);
+
+   /// next entry has stable index next after calling this function
+   void CreateNextInternal();
+};
+
+/// Flat storage of entries of type T such that their index is stable (but
+/// address is not). Can quickly "delete" (mark unused) and create values at any
+/// index.
+/// T must be default constructable and either copy or move constructable.
```
camierjs:
```suggestion
/// T must be default constructible and either copy or move constructible.
```

general/resource_manager.hpp
```diff
@@ -0,0 +1,1708 @@
...
+            // de-allocated by inst.Destroy()
+            h_ptr = reinterpret_cast<T *>(seg.lowers[0]);
+         }
+         MFEM_ASSERT(seg.mtypes[0] == h_mt, "host memory type mismatch");
+         MFEM_ASSERT((void *)h_ptr == (void *)seg.lowers[0],
+                     "host memory pointer mismatch");
+         // MFEM_ASSERT(flags & OWNS_DEVICE, "expected to also own device");
+         MFEM_ASSERT(offset_ == 0, "should not have any offset");
+         MFEM_ASSERT(size_ * sizeof(T) == size_t(seg.nbytes),
+                     "should hot refer to a subsection");
```
camierjs:
```suggestion
                     "should not refer to a subsection");
```

general/resource_manager.hpp
```diff
@@ -0,0 +1,1708 @@
...
+{
+   if (segment)
+   {
+      auto &inst = MemoryManager::Instance();
+      if ((flags & OWNS_DEVICE))
+      {
+         auto &seg = inst.storage.GetSegment(segment);
+         MFEM_ASSERT(offset_ == 0, "should not have any offset");
+         MFEM_ASSERT(size_ * sizeof(T) == size_t(seg.nbytes),
+                     "should hot refer to a subsection");
```
camierjs:
```suggestion
                     "should not refer to a subsection");
```

general/resource_manager.cpp
```diff
@@ -0,0 +1,2527 @@
...
+      Protect(seg.lowers[1], seg.mtypes[1], offset, nbytes, seg.IsTemporary());
+   }
+}
+
+#ifdef MFEM_USE_UMPIRE
+void MemoryManager::SetUmpireHostAllocatorName_(const char *h_name)
+{
+   MFEM_ASSERT(
+      !allocs_storage[10],
+      "Umpire Host Allocator has already been created and cannot be change");
```
camierjs:
```suggestion
      "Umpire Host Allocator has already been created and cannot be changed");
```

general/resource_manager.cpp
```diff
@@ -0,0 +1,2527 @@
...
+   h_umpire_name = h_name;
+}
+
+void MemoryManager::SetUmpireDeviceAllocatorName_(const char *d_name)
+{
+   MFEM_ASSERT(
+      !allocs_storage[11],
+      "Umpire Device Allocator has already been created and cannot be change");
```
camierjs:
```suggestion
      "Umpire Device Allocator has already been created and cannot be changed");
```

general/resource_manager.cpp
```diff
@@ -0,0 +1,2527 @@
...
+   d_umpire_name = d_name;
+}
+
+void MemoryManager::SetUmpireDevice2AllocatorName_(const char *d_name)
+{
+   MFEM_ASSERT(!allocs_storage[12], "Umpire Device 2 Allocator has already "
+               "been created and cannot be change");
```
camierjs:
```suggestion
               "been created and cannot be changed");
```

general/resource_manager.cpp
```diff
@@ -0,0 +1,2527 @@
...
+         {
+            curr_end = std::get<0>(blocks.back());
+            --std::get<2>(blocks.back());
+            --total_allocs;
+         }
+      }
+   }
+   else if (blocks.end() - mid == 1)
+   {
+      // last block completely free'd
```
camierjs:
```suggestion
      // last block completely freed
```

general/resource_manager.cpp
```diff
@@ -0,0 +1,2527 @@
...
+      for (size_t idx = blocks.size(); idx; --idx)
+      {
+         size_t i = idx - 1;
+         if (std::get<0>(blocks[i]) <= ptr && ptr < std::get<1>(blocks[i]))
+         {
+            --std::get<2>(blocks[i]);
+            --total_allocs;
+            if (std::get<2>(blocks[i]) == 0 && i + 1 == blocks.size())
+            {
+               // last block completely free'd
```
camierjs:
```suggestion
               // last block completely freed
```

tests/unit/CMakeLists.txt
```diff
@@ -40,14 +40,15 @@ set(UNIT_TESTS_SRCS
   dfem/test_lvector_interface.cpp
   dfem/test_mass.cpp
   general/test_array.cpp
+  general/test_reduction.cpp
```
camierjs: duplicate `test_reduction.cpp`

CHANGELOG
```diff
@@ -70,6 +70,10 @@ GPU computing
   added to ex1 and ex1p. See https://developer.nvidia.com/cudss for more
   details. Supported versions >= 0.6.0.
 
+- Rework of the memory manager so that all calls to Sync and SyncAlias are now
+  no-op and can be removed from user code. Note that the copy sometimes performed
+  by SyncAlias is deferred to the next call to Read/ReadWrite.
+
```
camierjs: Trailing whitespace to remove

general/internal/rb_tree_impl.hpp
```diff
@@ -0,0 +1,706 @@
...
+#define RB_TREE_IMPL_HPP
+
+namespace mfem
+{
+
+template <class ChildType> size_t RBTree<ChildType>::Successor(size_t idx)
```
camierjs: Is there a way to avoid this implementation to be include from the the top `general/rb_tree.hpp`?
It would help to have this source file know about the `RBTree` and some other included headers (`size_t`).

general/internal/rb_tree_impl.hpp
```diff
@@ -0,0 +1,706 @@
...
+{
+   if (static_cast<ChildType &>(*this).GetNode(idx).child[1])
```
camierjs: Can we find an elegant way to avoid the repetitions of `static_cast<ChildType &>(*this)` throughout the file?

tests/unit/makefile
```diff
@@ -82,7 +84,7 @@ PAR_TMOP_GPU_OBJ_FILES = $(if $(USE_GPU),$(TMOP_FILES:$(SRC)%.cpp=%.pgpu.o))
 PAR_TMOP_GPU_UVM_OBJ_FILES = $(if $(USE_GPU),$(TMOP_FILES:$(SRC)%.cpp=%.pgpu_uvm.o))
 
 # seq/par files and tests
-SEQ_UNIT_TESTS = unit_tests $(if $(USE_GPU),gpu_unit_tests)
+SEQ_UNIT_TESTS = unit_tests $(if $(USE_GPU),gpu_unit_tests) $(if $(USE_UMPIRE),umpire_unit_tests)
```
camierjs: This will build whenever `MFEM_USE_UMPIRE` is set, but does `umpire_unit_test_main.cpp` support CPU-only Umpire builds?

general/resource_manager.cpp
```diff
@@ -0,0 +1,2527 @@
...
+{
+   size_t alignment = 64;
+   StdAlignedAllocator() = default;
+   StdAlignedAllocator(size_t align) : alignment(align) {}
+   void Alloc(void **ptr, size_t nbytes) override
+   {
+#ifdef _WIN32
+      *ptr = _aligned_malloc(nbytes, alignment);
+#else
+      *ptr = aligned_alloc(alignment, nbytes);
```
camierjs: Doesn't `nbytes` required to be a multiple of alignment?
The return value could also be checked.

general/resource_manager.hpp
```diff
@@ -0,0 +1,1708 @@
...
+      auto &inst = MemoryManager::Instance();
+      void *res;
+      inst.allocs[idx]->Alloc(&res, n * sizeof(T));
+      return static_cast<T *>(res);
+   }
+
+   void deallocate(T *ptr, size_t n)
+   {
+      auto &inst = MemoryManager::Instance();
+      inst.allocs[idx]->Dealloc(ptr, n);
```
camierjs: Should it be `n` or `n * sizeof(T)` like in `allocate` ?

general/scan.hpp
```diff
@@ -26,12 +26,50 @@
...
+         return;
+      }
+#endif
+   }
+   bytes = 0;
+#endif
+   // determine buffer size
+   MFEM_GPU_CHECK(op(nullptr, bytes));
+#ifdef MFEM_USE_TEMPORARY_WORK_BUFFERS
+   Array<std::byte> workspace(
```
camierjs, line 66: If this local temporary `workspace` is passed down to `CUB/hipCUB` calls, which can be potentially asynchronous, isn't there a risk of them been released before usage?

general/internal/rb_tree_impl.hpp
```diff
@@ -0,0 +1,706 @@
...
+      if (x)
+      {
+         auto &xn = static_cast<ChildType &>(*this).GetNode(x);
+         if (xn.IsRed())
+         {
+            if (x_parent)
+            {
+               auto &yn = static_cast<ChildType &>(*this).GetNode(y);
+               auto &yp = static_cast<ChildType &>(*this).GetNode(yn.parent);
+               EraseFixup(root, x, x_parent, yp.child[0] == y);
```
camierjs: Does `EraseFixup` only needed for red nodes? Shouldn't it be symetric?

general/internal/rb_tree_impl.hpp
```diff
@@ -0,0 +1,706 @@
...
+         idx = y;
+         y = static_cast<ChildType &>(*this).GetNode(y).parent;
+      }
+      return y;
+   }
+}
+
+template <class ChildType>
+void RBTree<ChildType>::EraseFixup(size_t &root, size_t x, size_t x_parent,
+                                   bool y_is_left)
```
camierjs: Most of the functions in this file require some documentation, especially for the deep nested `if` cases.

general/rb_tree.hpp
camierjs: This RB tree would benefit from dedicated unit tests.

general/scan.hpp
helloworld922: I don't think so, cub enqueues any kernels used onto the specified (in this case default 0) stream, where we also perform all alloc/dealloc operations. There are also checks inside of the temporary allocators for host-pinned/UVM spaces which will explicitly call a device sync to prevent aliasing issues if a chunk of memory is going to be reused.

camierjs: I see. A short comment near the workspace allocation documenting why it's safe (stream ordering + pooled temp allocator) and that it is dependent on all GPU alloc/use happening on stream 0 would be useful.

camierjs: We might want to double check that the RAJA+CAMP for all backends use the same stream ordering.

general/resource_manager.hpp
helloworld922: Added

tests/unit/umpire_unit_test_main.cpp
helloworld922: Removed duplicate code

general/rb_tree.hpp
helloworld922: fixed

general/internal/rb_tree_impl.hpp
helloworld922: fixed

general/resource_manager.hpp
helloworld922: fixed

general/resource_manager.hpp
helloworld922: fixed headers

general/internal/mmu.cpp
helloworld922: added

tests/unit/linalg/test_chebyshev.cpp
helloworld922: added

general/internal/mem_op_debug.cpp
helloworld922: fixed

general/internal/reusable_storage.hpp
helloworld922: fixed

general/resource_manager.hpp
helloworld922: fixed

general/resource_manager.hpp
helloworld922: fixed

general/resource_manager.cpp
helloworld922: fixed

general/resource_manager.cpp
helloworld922: fixed

general/resource_manager.cpp
helloworld922: fixed

general/resource_manager.cpp
helloworld922: fixed

general/resource_manager.cpp
helloworld922: fixed

tests/unit/CMakeLists.txt
helloworld922: removed

CHANGELOG
helloworld922: removed

general/internal/rb_tree_impl.hpp
helloworld922: Changed to a local variable static cast so there should only be one per function now

tests/unit/makefile
helloworld922: It does not, changed to only build with umpire+GPU.

general/resource_manager.cpp
helloworld922: Added check

general/resource_manager.hpp
helloworld922: Yes, fixed

general/scan.hpp
helloworld922: Added comment, https://github.com/mfem/mfem/pull/5383 should make sure at least mfem's internal usage of RAJA/CAMP always use stream 0.

general/internal/rb_tree_impl.hpp
helloworld922: refactored code so this function is no longer needed

general/internal/rb_tree_impl.hpp
helloworld922: added documentation for Red-Black tree implementation

general/rb_tree.hpp
helloworld922: Added a few unit tests

general/internal/mmu.cpp
helloworld922: edit: I remember why I left this in there. It's for compatibility with the makefile build system which globs all cpp files for individual compilation, but this file cannot/should not be individually compiled.

I also avoided naming the file `mmu.hpp` since it shouldn't be installed.

general/internal/rb_tree_impl.hpp
helloworld922: Do you mean like include the template function implementations directly in `rb_tree.hpp`? This header isn't meant to be directly included and only holds the implementation since the implementation of the red black tree functions are quite long.

Maybe we could add `#include "../rb_tree.hpp" to the top of `rb_tree_impl.hpp` in case anyone accidentally directly includes this?

general/rb_tree.hpp
```diff
@@ -0,0 +1,66 @@
...
+namespace mfem
+{
+
+/// Red-black tree
+template <class ChildType> struct RBTree
+{
+   // @return the node which has value immediately after node @a idx, or 0 if
+   // idx has the largest value.
+   size_t Next(size_t idx);
+   /// internal use only
```
pazner, line 24: Should this be a class, and these private members?

general/internal/reusable_storage.hpp
```diff
@@ -0,0 +1,92 @@
...
+#include "../error.hpp"
+
+namespace mfem
+{
+namespace internal
+{
+
+/// Flat storage of entries of type T such that their index is stable (but
+/// address is not). Can quickly delete and create values at any index.
+struct ReusableStorageBase
```
pazner, line 27: Why do we need this base class (and not just `ReusableStorage`)?

general/array.hpp
```diff
@@ -107,6 +107,16 @@ class Array
...
 
+#ifdef MFEM_USE_NEW_MEM_MANAGER
+   Array(int size_, bool temporary) : data(size_, temporary), size(size_) {}
+   Array(int size_, MemoryType mt, bool temporary)
+      : data(size_, mt, temporary), size(size_)
+   {}
+   Array(int size_, MemoryType h_mt, MemoryType d_mt, bool temporary)
+      : data(size_, h_mt, d_mt, temporary), size(size_)
+   {}
+#endif
```
pazner: I personally don't like the `bool temporary` API - hard to understand from the call-site what `true` or `false` signify.

tests/scripts/branch-history
```diff
@@ -59,7 +59,7 @@ if (!chdir $mfem_dir) {
 }
 
 # Maximum number of acceptable commits in the branch
-my $branch_max_commits = 200;
```
pazner, line 62: This should probably be reverted.

I also just opened PR #5407 that adds a way to override the branch-history checks.

linalg/vector.hpp
```diff
@@ -103,9 +103,20 @@ class Vector
 
    /// Creates a vector referencing an array of doubles, owned by someone else.
    /** The pointer @a data_ can be NULL. The data array can be replaced later
-       with SetData(). */
-   Vector(real_t *data_, int size_)
-   { data.Wrap(data_, size_, false); size = size_; }
+       with SetData(). @a check = true if @a data_ could have been previously
+      registered with the memory manager. */
+   Vector(real_t *data_, int size_, bool check = false)
```
pazner, line 108: Not sure about this `check` flag... when should it be set?

linalg/vector.hpp
```diff
@@ -133,6 +144,15 @@ class Vector
...
    { std::copy(values.begin(), values.end(), begin()); }
 
+#ifdef MFEM_USE_NEW_MEM_MANAGER
+   Vector(int size_, MemoryType mt, bool temporary)
+      : data(size_, mt, temporary), size(size_)
+   {}
+   Vector(int size_, MemoryType h_mt, MemoryType d_mt, bool temporary)
+      : data(size_, h_mt, d_mt, temporary), size(size_)
+   {}
+#endif
```
pazner: As for `Array<T>`, I don't love this API for creating temporary vectors.

general/reducers.hpp
```diff
@@ -592,6 +592,78 @@ void reduce(int N, T &res, B &&body, const R &reducer, bool use_dev,
...
+/**
+ @brief Performs a 1D reduction on the range [0,N).
+ @a res initial value and where the result will be written.
+ @a body reduction function body.
+ @a reducer helper for joining two reduced values.
+ @a use_dev true to perform the reduction on the device, if possible.
+ @tparam T value_type to operate on
+ */
+template <class T, class B, class R>
+void reduce(int N, T &res, B &&body, const R &reducer, bool use_dev)
```
pazner, line 609: Should we just get rid of the explicitly passed-in workspace version?

general/resource_manager.hpp
```diff
@@ -0,0 +1,1717 @@
...
+            inst.Read(segment, 0, seg.nbytes, false);
+         }
+         if (seg.lowers[1] != seg.lowers[0] && seg.lowers[1])
+         {
+            if (OwnsDevicePtr())
+            {
+               MFEM_MEM_OP_DEBUG_REMOVE2(
+                  1, seg.lowers[1], seg.lowers[1] + seg.nbytes,
+                  "dealloc " << (int)seg.mtypes[1] << ", "
+                  << seg.ITemporary());
```
pazner:
```suggestion
                  << seg.IsTemporary());
```

general/resource_manager.hpp
```diff
@@ -0,0 +1,1717 @@
...
+   {
+      MFEM_ASSERT(!segment, "unexpected valid segment");
+      if (dloc == MemoryType::DEFAULT)
+      {
+         MFEM_ASSERT(int(hloc) < MemoryTypeSize, "invalid hloc");
+         dloc = inst.GetDualMemoryType(hloc);
+      }
+      MFEM_ASSERT(h_ptr != nullptr,
+                  "Cannot wrap d_ptr != nullptr with h_ptr == nullptr");
+      segment = inst.Insert(reinterpret_cast<char *>(h_ptr),
```
pazner, line 1206: Does `offset_` need to be set here?

general/resource_manager.hpp
```diff
@@ -0,0 +1,1717 @@
...
+                            reinterpret_cast<char *>(d_ptr), size * sizeof(T),
+                            hloc, dloc, valid_host, valid_device, false);
+   }
+   if (own_host)
+   {
+      flags = static_cast<Flags>(flags | OWNS_HOST);
```
pazner, line 1212: Do the flags need to be cleared before this?

general/resource_manager.cpp
pazner: Is this intended to work for arbitrary allocation sizes, or only when `nbytes` is a multiple of `alignment` (size could be rounded up)? Or use `mfem_memalign`

general/internal/reusable_storage.hpp
helloworld922: I separated out the parts of `ReusableStorage` which are independent of template parameters and put them in the base class, so the details of status flag management could be in `reusable_storage.cpp`.

general/array.hpp
helloworld922: I think I agree as well. I'm not sure what a good API for this should be, I think ideally these would be separate memory spaces, but then there is the question of how to handle comparing spaces for validity. Maybe this could be a separate enum class with two values `TEMPORARY` and `PERSISTENT`, defaulting to `PERSISTENT`?

linalg/vector.hpp
helloworld922: This is specifically to handle the case of wrapping a pointer we want the memory manager to check if it's been previously registered or not to support Sundials. It is a bit expensive to do this on every wrap attempt, so this is a compromise.

Maybe it needs a better name?

general/reducers.hpp
helloworld922: I left it for backwards compatibility.

general/resource_manager.hpp
helloworld922: Yes, added to the early check for erasing an old segment which should be the only time `offset_ != 0`.

general/resource_manager.hpp
helloworld922: Also yes, nice catch

helloworld922:
> Is this intended to work for arbitrary allocation sizes, or only when nbytes is a multiple of alignment (size could be rounded up)? Or use mfem_memalign

@pazner That's a good question. I can see allocating something that isn't a perfect multiple of alignment being useful.

The [posix_memalign](https://man7.org/linux/man-pages/man3/posix_memalign.3.html) documentation recommends using `aligned_alloc` instead, but this was changed in a C11 defect report to return null if size isn't a multiple of `nbytes` so it's not clear to me if `posix_memalign` will also do this or not. The MSVC documentation for `_aligned_malloc` is similarly silent about this.

We could roll our own version which is guaranteed to work across platforms

general/reducers.hpp
pazner: Would be nice to deprecate/remove it if we can.

helloworld922: I added a note in the changelog/documentation to say that this version is deprecated.

general/array.hpp
tomstitt: I liked what you have in https://github.com/mfem/mfem/pull/4065 Will

general/resource_manager.hpp
```diff
@@ -0,0 +1,1719 @@
...
+   /// only non-zero if registered. If segment == 0 and h_ptr != nullptr, then
+   /// this is an un-registered Memory
+   mutable size_t segment = 0;
+   enum Flags : unsigned char
+   {
+      NONE = 0,
+      OWNS_HOST = 1 << 1,
+      OWNS_DEVICE = 1 << 2,
+      USE_DEVICE = 1 << 3,
+      TEMPORARY = 1 << 4,
```
tomstitt: I think I prefer the https://github.com/mfem/mfem/pull/4065 approach where temporary is just another allocator and the user can pick the one they want. Are there advantages of having it as an attribute of the memory itself?

general/rb_tree.hpp
helloworld922: Yes, made private

general/array.hpp
helloworld922: I changed this to have separate temporary memory types instead of a boolean flag. This only works with the new memory manager. By convention, even memory types are permanent/persistent, while odd ones are temporary. I can also add a "NewTemporary" helpers similar to #4065 if you think these would be easier to work with at the cost of being less flexible for creating for example host pinned/managed temporary workspaces.

general/resource_manager.hpp
helloworld922: I switched to the approach in #4065, and added the `NewTemporary` helpers for Array and Vector. These don't behave exactly the same as they did in #4065, notably the user is responsible for calling `UseDevice` on the resulting vector since they could request any memory space.

pazner: Could `NewTemporary` call `UseDevice` depending on which memory space is being requested?

helloworld922: I can do that for `IsDeviceMemory(mt)`, though sometimes for host pinned/managed memory users might also want this as well.

pazner: I think it makes sense to do for `IsDeviceMemory(mt)`.

helloworld922: Do we also want to do this for Array temporaries? It looks like there isn't any public member for setting the UseDevice flag in Array, though nothing is stopping users from getting the internal memory object and calling UseDevice on that.

tests/scripts/branch-history
pazner: #5407 has been merged; can revert this now.

helloworld922: reverted

pazner: Thanks! CI logs look good now: https://github.com/mfem/mfem/actions/runs/30849397037/job/91805471454?pr=5019#check-step-5

PR #5030, opened 2025-09-23
MTOP solver GPU support through PA or dFEM

camierjs:


camierjs:
Not sure we want to bring these meshes here - closing this PR, switching directly to `uqsolvers` branch for the PR.

PR #5049, opened 2025-10-01
Fix parallel nurbs miniapps

justinlaughlin:
# Main change
A recent issue pointed out some weird looking answers when running `nurbs_ex1p`. I was able to reproduce this even without the `-no-ibp` flag, simply by running `mpirun -np 4 nurbs_ex1p -m ../../data/square-nurbs.mesh -o 3` we get an erroneous looking answer.

image

Similarly, running `mpirun -np 4 nurbs_ex11p -m ../../data/square-nurbs.mesh -o 3` will give incorrect answers.

After some investigation this seems to be related to changes in how nurbs `FiniteElementCollection` and `FiniteElementSpace` are constructed. This PR makes appropriate changes which results in correct outputs.

# Other changes
~~Another minor change in this PR - in these examples, `order` is an array but I don't see any way this would be set? I've changed these to scalar values and made changes accordingly.~~ I guess arrays can be set via cli args. I'd still like to get a second opinion on how we are setting up the problem. e.g. this line: `NURBSext = new NURBSExtension(pmesh->NURBSext, order)` - should we be creating a whole new (serial) `NURBSExtension`? The approach of `nurbs_patch_ex1` using `incdeg` instead of `order` (iga is assumed) to change the order of the mesh makes more sense to me. With the current approach, we now have the `NURBSExtension` associated with the mesh, and an independent one associated with the finite element space.

Finally, one last minor change is that the initialization of the default `kappa` value is moved down, because it should rely on the order of the `FiniteElementCollection` (if using IGA where `order=-1` this would result in an incorrect value).

### DIFF
--- a/miniapps/nurbs/nurbs_ex1.cpp
+++ b/miniapps/nurbs/nurbs_ex1.cpp
@@ -311,14 +311,14 @@ int main(int argc, char *argv[])
       {
          cout << "No integration by parts requires a NURBS mesh, with only 1 patch."<<
               endl;
-         cout << "A C_1 discretisation is required."<< endl;
+         cout << "A C_1 discretization is required."<< endl;
          cout << "Currently only C_0 multipatch coupling implemented."<< endl;
          return 3;
       }
       if (order[0]<2)
       {
          cout << "No integration by parts requires at least quadratic NURBS."<< endl;
-         cout << "A C_1 discretisation is required."<< endl;
+         cout << "A C_1 discretization is required."<< endl;
          return 4;
       }
    }
--- a/miniapps/nurbs/nurbs_ex11p.cpp
+++ b/miniapps/nurbs/nurbs_ex11p.cpp
@@ -57,8 +57,7 @@ int main(int argc, char *argv[])
    const char *mesh_file = "../../data/star.mesh";
    int ser_ref_levels = 2;
    int par_ref_levels = 1;
-   Array<int> order(1);
-   order[0] = 0;
+   int order = 0;
    int nev = 5;
    int seed = 75;
    bool slu_solver  = false;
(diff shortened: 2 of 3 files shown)


IdoAkkerman:
@justinlaughlin There is indeed an issue. I was able to exactly replicate your image. In my case there is a jump between processors and the Dirichlet BC is not satisfied on all ranks.

However your fix, is not really a fix though.... the NURBS elements are not selected. This is a standard higher-order polynomial solution on  NURBS mesh. See the header of the solution.00000x for confirmation.

I will have a closer look.

Please not PR  #4936 , that is tangentially related.

IdoAkkerman:
Moreover, the idea of using an array for the order, is that it allows you to specify a order per knotvector.

IdoAkkerman:
I think the problem occurs when the mesh and the solution use a NURBSext with different order.
The problem is not directly the simulation itself, but the subsequent reading of the solution by either glvis or visit in the GridFunction constructor.

This can be verified running for instance (in the master branch)

`mpirun -np 1 nurbs_ex1p -m ../../data/square-nurbs.mesh -o 5 -r 1`

The visualized solution is clearly bilinear (which is the order of the mesh).

There is a finicky thing in the GridFunction constructor(that might have been broken), and that is that when a solution is read form file we need both a Nodes Gridfunction for the Nodes of the mesh, and the actual solution GridFunction.
However these gridfunction are stored very differently in the files and the reconstruction is different. Resulting in a difficult to enforce branching in  the GridFunction (I think this branching might have been removed).

IdoAkkerman:
I think I found the issue. It the GridFunction constructor that reads separate parallel pieces, the fespace is not build with its own NURBSextension.

IdoAkkerman:
Please see #5050.
Does this fix none/some/all of the problems?

justinlaughlin:
Thanks for taking a look @IdoAkkerman! I agree this isn't really a solution - I realized later this was supposed to be nurbs bases (independent from the mesh nurbs bases). I'll take a look at 4936 and 5050

justinlaughlin:
I'm not sure if there is any code worth keeping here, so I'll close this PR and move the conversation over to the other PRs 👍

PR #5050, opened 2025-10-01
Add NURBSextension to fespace in gridfunction constructor

IdoAkkerman:
This PR fixes a bug when postprocessing parallel NURBS solutions.

Instead of using the NURBSextension of the mesh, a dedicated NURBSextension is made for the solution.
This is essential as it is know a priori that the solution has the same order as the mesh.

Potential speed up: check if the dedicated extension is required. This might not be worth the risk...

### DIFF
--- a/fem/gridfunc.cpp
+++ b/fem/gridfunc.cpp
@@ -89,7 +89,8 @@ GridFunction::GridFunction(Mesh *m, GridFunction *gf_array[], int num_pieces)
    vdim = fes->GetVDim();
    ordering = fes->GetOrdering();
    NURBSExtension *NURBSext = NULL;
-   if (dynamic_cast<NURBSFECollection *>(fec_owned))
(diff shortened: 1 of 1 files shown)


IdoAkkerman:
The `MergeGridFunction` has been made `static`, but I think it should be moved to `GridFunction`.
Perhaps the code can just copied in-place, no need for a separate function.

There is no need for this, but I think it will make things just a little bit clearer.

IdoAkkerman:
Perhaps #4936 already fixed the related issue without realizing.
I'll check tomorrow(CET).
This might be a good argument for the changes in proposed in #4936, it is not just for convenience, but also for robustness.

IdoAkkerman:
I just checked #4936 also solves the underlying issue.

The solution in #4936  is the preferred fix, IMHO.

PR #5071, opened 2025-10-14
NURBS Specific Constructors for FiniteElementSpace 

justinlaughlin:
This PR moves a lot of the NURBS specific construction logic into `FiniteElementSpace::IsogeometricConstructor()` and `FiniteElementSpace::NURBSConstructor()`. The main reasons are 1) the dispatch logic was getting quite complex, and 2) to avoid signature pollution in the main `FiniteElementSpace` constructor. I'd also like to propose that all of the miniapps under `miniapps/nurbs/` assume a NURBS mesh, which will also help make dispatch much simpler. For simple NURBS usage like in `examples/` I don't think we need to change anything. This is more-so to handle the dispatch of NURBS specific things (like assigning an array of orders equal to the number of knotvectors).

* I've modified `nurbs_ex1` to use this new constructor. I've also added a check that enforces the problem be run with NURBS meshes (default is changed to `beam-hex-nurbs.mesh`). Finally, I stack allocated variables (and `NURBSConstructor` returns a `unique_ptr`) so we don't have to figure out what memory needs to be freed at the end. If these changes seem reasonable, I will edit the other NURBS miniapps in a similar way. Please let me know what you think.
* I added `NURBSExtension::DegreeElevate(const Array<int> &degrees)` which was intended to be used for a deep-copy constructor; I couldn't quite figure out how to get the deep-copy constructor to be robust though because of potential implicit relationships with `Mesh` (compared to say `DegreeElevate` in `Mesh` which will update the nodes). I am open to removing this method.

TODO:
 - [ ] Modify other examples (pending comments from Dylan/Ido)
 - [ ] Possibly, replace other constructors?

### DIFF
--- a/fem/fespace.cpp
+++ b/fem/fespace.cpp
@@ -219,11 +219,13 @@ FiniteElementSpace::IsoparametricConstructor(Mesh* mesh, int vdim, int ordering)
 {
    MFEM_VERIFY(mesh->GetNodes(),
                "Mesh must have Nodes in order to create an isoparametric space.");
-   MFEM_VERIFY(mesh->GetNodes()->OwnFEC(),
-               "Mesh Nodes should own their FiniteElementCollection and "
+
+   const FiniteElementCollection* fec = mesh->GetNodes()->FESpace()->FEColl();
+
+   MFEM_ASSERT(fec,
+               "No FiniteElementCollection extracted from GridFunction"
                "FiniteElementSpace in order to create an isoparametric space.");
 
-   FiniteElementCollection* fec = mesh->GetNodes()->OwnFEC();
    // A NULL NURBSext argument means mesh.NURBSext will be used
    // (if this is a nurbs mesh)
    return FiniteElementSpace(mesh, nullptr, fec, vdim, ordering);
@@ -262,7 +264,7 @@ NURBSSpace::NURBSSpace(Mesh* mesh,
    // Case 1 - Isoparametric (convention: order < 1)
    if ((orders.Size() == 1) && (orders[0] < 1))
    {
-      mfem::out << "ISOPARAMETRIC" << std::endl;
+      mfem::out << "Generating an Isogeometric FE space" << std::endl;
       nurbs_fec = nullptr;
       fec = mesh_fec;
    }


IdoAkkerman:
I think `FiniteElementSpace::IsogeometricConstructor()` should be named `FiniteElementSpace::IsoparametiricConstructor()` , as it would also work for other element types.

tzanio:
What is `Isoparametiric`?

IdoAkkerman:
Sorry, Isoparametric

IdoAkkerman:
Isoparametric is when the mesh geometry and the solution use the same parametrization.
Isogeometric is a spin on this where the  analysis (mesh geometry and solution)  and CAD geometry use the the same parametrization.

IdoAkkerman:
Would it make sense to put the `NURBSConstructors`  in a NURBS namespace?
Otherwise it would still clutter `FESpace`?


`auto[space, fec] = NURBS::ConstructFESpace(bla)`

IdoAkkerman:
`fec_.get()` is this `get` a member of the `std::move` return type?
I still a bit puzzled how this works for the `nullptr` return type? 

I guess it is part of unique_ptr?
Rubber ducking here :)

justinlaughlin:
> Isoparametric is when the mesh geometry and the solution use the same parametrization.
Isogeometric is a spin on this where the analysis (mesh geometry and solution) and CAD geometry use the the same parametrization.

Since we are using NURBS, IMO its fair to call it isogeometric since that is the typical connotation as far as I understand (isoparametric+NURBS -> isogeometric). We don't know what basis functions were used in CAD but its almost always NURBS so that is the typical implication. In many areas we call "isoparametric", I believe we are only supporting NURBS bases anyways.

>Would it make sense to put the NURBSConstructors in a NURBS namespace?
Otherwise it would still clutter FESpace?

The main intention was to not clutter the signatures of the `FiniteElementSpace` constructor; it is getting too complicated. IMO these methods are appropriate in this namespace since they are related to `FiniteElementSpace`. I wouldn't be opposed to a NURBS namespace, but I think that change should be made more broadly.

>fec_.get() is this get a member of the std::move return type?

Yeah, I'm not totally satisfied with this solution but I haven't thought of a better way to do this yet, I'm open to any suggestions. `FiniteElementSpace::NURBSConstructor` returns a unique pointer to the `fec`, which is null when its IGA, since that `fec` is owned by the `Mesh`. `.get()` is a `unique_ptr` method that returns the associated raw pointer. So this line 
```cpp
   FiniteElementCollection* fec = !fec_.get() ? mesh.GetNodes()->OwnFEC() : fec_.get();
```
is just returning the `Mesh`s raw `fec` pointer if the returned `fec` is null. I don't know of a better way to do this because the returned pointer's ownership is dynamic.

The only thing its used for is to print the name, so maybe we don't even need the line? Another idea I had was to wrap the unpacking logic inside of `NURBSSpace` (right now it is just `using NURBSSpace = std::pair<FiniteElementSpace, std::unique_ptr<NURBSFECollection>`); I'm not a huge fan of this solution though :/. A shared pointer would probably be ideal, but would require changing `GridFucntion.fec_owned` to be a shared pointer.

justinlaughlin:
Another option is to add a `own_fec` bool to `FiniteElementSpace` (or a `nurbs_fec` unique pointer), i.e. change `fespace` so that it sometimes owns `fec`. There is some precedent for this (e.g. `own_ext`), but I'm not a huge fan of this solution since it complicates the ownership semantics.

IdoAkkerman:
@justinlaughlin I agree with the namespace argument, just making sure this is the best.

The pointer stuff is fine. Its is mostly my lack of experience with the c++11 constructs. Their might still be some massaging to be done, such as the `own` flag. But we can postponed that to later i would say.

I think we had a misunderstanding wrt the isoparametric vs isogeometric item. My point is, the current function `IsogeometricConstructor` will also work for meshes that have a Nodes gridfunction that is not defined using NURBS.
Therefor the more generic name would be more appropriate.

BTW:  there I think there needs to be a check as the `mesh` might not have `Nodes` (GetNodes() might return nullptr).

justinlaughlin:
Re pointers: Another possible solution is to return a shared pointer, but with a no-op destructor when it points to the mesh fec; this eliminates this one awkward line, but means we return a shared pointer that *sometimes* has ownership (and this isn't obvious until examining the source). There is a larger discussion about this [here](https://github.com/orgs/mfem/discussions/4614); although I am partial to a custom solution similar to what @pazner implemented [here](https://github.com/pazner/mfem/blob/handle-release-ownership/general/handle.hpp). 

Ah, I misunderstood the `FiniteElementSpace` constructor and assumed if `NURBSext` is null it just means it will use the `NURBSext` of the mesh. But I guess what you're saying is `NURBSext==nullptr` + non-nurbs mesh means isoparametric? I agree the more generic name is more appropriate.

justinlaughlin:
If this seems okay with you, I propose the following
* I will make any more changes per suggestions from you/Dylan/Tzanio (let me know what you think of the idea in this [comment](https://github.com/mfem/mfem/pull/4936/files#r2471466151))
* I will modify the other nurbs miniapps to use a similar pattern (assume a nurbs mesh, simplify dispatch logic, eliminate unnecessary heap allocations/deletes)
* If all seems okay at that point, we can merge this into !4936

IdoAkkerman:
Sounds good.

As I said there is some use to be able to do non-NURBS simulations, but there is also something to say for clean code without clutter. So I am okay with the NURBS only assumption.

fem/fespace.cpp
```diff
@@ -215,6 +214,87 @@ void FiniteElementSpace::CopyProlongationAndRestriction(
    delete perm_mat_tr;
 }
 
+FiniteElementSpace
+FiniteElementSpace::IsogeometricConstructor(Mesh* mesh, int vdim, int ordering)
```
IdoAkkerman: This should be called `IsoparametricConstructor`, as it will also work for non NURBS `fecs`.

fem/fespace.cpp
```diff
@@ -215,6 +214,87 @@ void FiniteElementSpace::CopyProlongationAndRestriction(
...
+{
+   auto fec = mesh->GetNodes()->OwnFEC();
```
IdoAkkerman: Need to check first if `GetNodes()` gives a valid pointer.

IdoAkkerman: This is just a personal style thing... but I think in this case there is no need for an `auto`.  Wouldn't it help readability and maintainability if we are explicit?
But perhaps I am wrong on this one.

justinlaughlin:
Thanks for reviewing @IdoAkkerman - I've been a bit behind but I'm working on changes to this today.

fem/fespace.cpp
justinlaughlin: Agreed - changed in [978d62b](https://github.com/mfem/mfem/pull/5071/commits/978d62b13901e5e5f03065567fb26c6351b7cf97) among other things

fem/fespace.cpp
justinlaughlin: changed in 978d62b

fem/fespace.cpp
justinlaughlin: added nodes pointer check in e144872

justinlaughlin:
I ended up wrapping these pointers into a new lightweight struct `NURBSSpace`; the isoparametric case felt awkward and implicit in the initial implementation. I updated at least the serial nurbs miniapps. Let me know what you think

PR #5085, opened 2025-10-24
ARKStepODE interface for ARKStepSolver specific TimeDependentOperator functions

cjvogl:
This PR serves as an example of how the `TimeDependentOperator` might be partitioned into modular interfaces for easier understanding what needs to be implemented for a given time integrator (i.e., `ODESolver` derived class). 

Here, the focus is on the `ARKStepSolver`, which is now initialized with an `ARKStepODE` object instead of a `TimeDependentOperator`.  While the `ARKStepODE` interface still has default implementations that throw an error, those errors are (in my opinion) much more informative than the current error thrown in `TimeDependentOperator`. Additionally, the modular nature of this interface means the derive classes can interpret functionality (e.g. like using "mfem form" and "mass form" instead of "explicit" and "implicit", the latter of which confuses with explicit vs implicit time integration) without conflict with how other ODE solvers might interpret things (e.g., PETSc)

I'll also note that because a class can derive both `TimeDependentOperator` and `ARKStepODE`, the user can still write one class and use both MFEM native and ARKODE time integrators (see the sundials examples 9 and 16). Furthermore, classes that only derive `TimeDependentOperator` can still use ARKODE time integrators (see sundials example 10). The backwards compatibility is supported by introducing a wrapper in the `ARKStepSolver` class. 

This is very much a work in progress but I'd like input before fleshing the rest of this out (namely, including something similar for `KINSovler` and `CVODESolver` so the `TimeDependentOperator::SUN`* methods are fully in their own modular interface(s).

### DIFF
--- a/examples/sundials/ex16.cpp
+++ b/examples/sundials/ex16.cpp
@@ -59,7 +59,7 @@ using namespace mfem;
  *    2. F(u, du/dt, t) = M du/dt          (ODE is expressed in IMPLICIT form)
  *       G(u, t) = - K(u) u
  */
-class ConductionOperator : public TimeDependentOperator
+class ConductionOperator : public TimeDependentOperator, public ARKStepODE
 {
    FiniteElementSpace &fespace;
    Array<int> ess_tdof_list; // this list remains empty for pure Neumann b.c.
@@ -81,11 +81,17 @@ class ConductionOperator : public TimeDependentOperator
 
    mutable Vector z; // auxiliary vector
 
+   const bool use_explicit_form;
+
 public:
 
    ConductionOperator(FiniteElementSpace &f, const real_t alpha,
                       const real_t kappa, const Vector &u,
-                      const Type &ode_expression_type);
+                      const bool use_explicit_form);
+
+   int Size() const override;
+
+   bool inExplicitForm() const override;
 
    // Compute K(u_n) for use as an approximation in - K(u) u
    void SetConductionTensor(const Vector &u);
@@ -245,16 +251,8 @@ int main(int argc, char *argv[])
    u_gf.GetTrueDofs(u);
 
    // 6. Initialize the conduction ODE operator and the visualization.
-   ConductionOperator::Type ode_expression_type;
-   if (use_mass_solver)
-   {
-      ode_expression_type = ConductionOperator::Type::IMPLICIT;
-   }
-   else
-   {
-      ode_expression_type = ConductionOperator::Type::EXPLICIT;
-   }
-   ConductionOperator oper(fespace, alpha, kappa, u, ode_expression_type);
+   const bool use_explicit_form = use_mass_solver ? false : true;
+   ConductionOperator oper(fespace, alpha, kappa, u, use_explicit_form);
 
    u_gf.SetFromTrueDofs(u);
    {
@@ -445,9 +443,10 @@ int main(int argc, char *argv[])
 ConductionOperator::ConductionOperator(FiniteElementSpace &fes,
                                        const real_t alpha, const real_t kappa,
                                        const Vector &u,
-                                       const Type &ode_expression_type)
-   : TimeDependentOperator(fes.GetTrueVSize(), 0.0, ode_expression_type),
-     fespace(fes), M(&fespace), alpha(alpha), kappa(kappa), z(height)
+                                       const bool use_explicit_form)
+   : TimeDependentOperator(fes.GetTrueVSize(), 0.0),
+     fespace(fes), M(&fespace), alpha(alpha), kappa(kappa), z(height),
+     use_explicit_form(use_explicit_form)
 {
    // specify a relative tolerance for all solves with MFEM integrators
    const real_t rel_tol = 1e-8;
@@ -472,6 +471,16 @@ ConductionOperator::ConductionOperator(FiniteElementSpace &fes,
    T_solver.SetPreconditioner(T_prec);
 
    SetConductionTensor(u);
+}
+
+int ConductionOperator::Size() const
+{
+   return z.Size();
+}
+
+bool ConductionOperator::inExplicitForm() const
+{
+   return use_explicit_form;
 }
 
 void ConductionOperator::SetConductionTensor(const Vector &u)
@@ -533,7 +542,7 @@ int ConductionOperator::SUNImplicitSolve(const Vector &r, Vector &dk,
    //   EXPLICIT form: r = -inv(M) K(u_n) u - k
    //   IMPLICIT form: r = -K(u_n) u - M k
    T_solver.SetRelTol(tol);
-   if (isExplicit())
+   if (use_explicit_form)
    {
       Mmat.Mult(r, z);
       T_solver.Mult(z, dk);
--- a/examples/sundials/ex16.cpp
+++ b/examples/sundials/ex16.cpp
@@ -28,10 +28,8 @@
 //
 //               The example demonstrates the use of nonlinear operators (the
 //               class ConductionOperator defining C(u)), as well as their
-//               implicit time integration. Note that implementing the method
-//               ConductionOperator::ImplicitSolve is the only requirement for
-//               high-order implicit (SDIRK) time integration. By default, this
-//               example uses the SUNDIALS ODE solvers from CVODE and ARKODE.
+//               implicit time integration. By default, this example uses the
+//               SUNDIALS ODE solvers from CVODE and ARKODE.
 //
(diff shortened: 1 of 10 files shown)


v-dobrev:
Some thoughts: some of the methods in `ARKStepODE` will be common with other similar classes. How do you envision avoiding repeated definitions? Also, when you inherit from both `TimeDependentOperator` and `ARKStepODE`, you get `Mult` from both and you have `Width`/`Height` (from the former) and `Size` from the latter which is not ideal.

linalg/sundials.cpp
```diff
@@ -1518,13 +1523,14 @@ ARKStepSolver::ARKStepSolver(MPI_Comm comm, Type type)
...
 #endif
 
-void ARKStepSolver::Init(TimeDependentOperator &f_)
+void ARKStepSolver::Init(ARKStepODE &f_ark_)
 {
-   // Initialize the base class
-   ODESolver::Init(f_);
+   this->f_arkstep = &f_ark_;
+
+   this->f = dynamic_cast<TimeDependentOperator*>(&f_ark_);
```
najlkin: I think this is the most painful part of such design :face_with_diagonal_mouth: . It is not allowed to dynamic cast to an independent class. The compiler might do that, but it is illegal by the standard.

linalg/operator.hpp
```diff
@@ -571,40 +570,6 @@ class TimeDependentOperator : public Operator
...
-
-       @param[in]      b   The linear system right-hand side.
-       @param[in,out]  x   On input, the initial guess. On output, the solution.
-       @param[in]      tol Linear solve tolerance.
-
-       If not re-implemented, this method simply generates an error.
-
-       Presently, this method is used by SUNDIALS ARKStep integrator, for more
-       details, see the ARKode User Guide. */
-   virtual int SUNMassSolve(const Vector &b, Vector &x, real_t tol);
```
najlkin, line 594: This is not possible because of backward compatibility. We would need a dynamic cast to call `ARKStepODE`, but that is not possible as said [below](https://github.com/mfem/mfem/pull/5085/files#r2462969069).

cjvogl: It's not clear to me what is not possible, can you elaborate?

linalg/sundials.cpp
cjvogl: I agree this is painful.  Given the only functionality used in this class is `TimeDependentOperator`, I'll remove this pain point by having `SetTime` in `ARKStepODE` (and then removing this cast)... unless there are objections?

cjvogl:
Thanks for the thoughts @v-dobrev!

> Some thoughts: some of the methods in `ARKStepODE` will be common with other similar classes. How do you envision avoiding repeated definitions?

That is a good question.  For now, my thought is to embrace repeated definition by having the method in the derived class override both methods in the two base classes (e.g., `TimeDependentOperator::Mult` and `ARKStepODE::Mult`.  I suppose there's also the issue of repeated definitions with different interpretations... that seems a bit trickier, but as a first pass, I suppose there could be use of prefixes (e.g., `SUNImplicitSolve`)  

> Also, when you inherit from both `TimeDependentOperator` and `ARKStepODE`, you get `Mult` from both and you have `Width`/`Height` (from the former) and `Size` from the latter which is not ideal.

I see your point.  I suppose I can remove `ARKStepODE::Size` and introduce at least `ARKStepODE::Height` (as it was used in `ARKStepSolver`)?

Updated response on 10/28: @najlkin shares the concern about inheriting functions with the same signature from two different base classes.  I’ll try to address this by adding ARKStepSolver-specific methods while keeping Mult and ExplicitMult for backwards compatibility (and deprecating them).

In that perspective though, I think Size is the appropriate method for ARKStepODE because ARKStepSolver doesnt distinguish between Height and Width (ie it only uses the size of the dof vector aka Height).

linalg/sundials.cpp
najlkin: I think it is also not possible, because it will lead to ambiguity when you inherit from both, `TimeDependentOperator` and `ARKStepODE`. It can be resolved manually, but you do not want to leave that on the users. I simply do not know a perfect solution to this, which is maybe the reason why there are so many monolithic interfaces in MFEM 🤷 .

cjvogl: I see the concern about ambiguity, especially given how I catered the documentation of the ARKStepODE methods to how they are used by ARKStepSolver.

I believe the ambiguity could be addressed by changing the documentation to indicate the various methods have the same definitions as TimeDependentOperator with a given definition of F and G.

Maybe a better option is to introduce new methods that are better suited to ARKStepODE, namely ones that have time as an argument (ARKStepSolver never calls Mult or ExplicitMult without first calling SetTime).  I will give that a try, while also updating the comments (and deprecating) Mult and ExplicitMult to indicate they should be overridden to match TimeDependentOperator.

linalg/operator.hpp
cjvogl: On a second pass, is the concern here the removal from TimeDependentOperator causing backwards compatibility problems? FWIW, the only SUNDIALS solver that uses SUNMassSolve (and Setup) is ARKStepSolver, which would now call ARKStepODE::SUNMassSolve

linalg/sundials.cpp
cjvogl: The refactor in [cff8552](https://github.com/mfem/mfem/pull/5085/commits/cff8552e1a90b97a82a53a1ba9c0d4fa85841d9c) now avoids dynamic casting against the standard and having classes inherit functions with the same name and signature.  Backwards compatibility (for classes that only derive `TimeDependentOperator`) is supported by introducing a wrapper in the `ARKStepSolver` class.

linalg/operator.hpp
cjvogl: I reverted the function removal in [5da21c8](https://github.com/mfem/mfem/pull/5085/commits/5da21c841b3e65f8a51df90cc6722fead133e221) and instead introduced deprecated markings in [ce6b9bd](https://github.com/mfem/mfem/pull/5085/commits/ce6b9bd4d87f1c95c4f29ffb9b9a9829a108a833)

cjvogl:
I've updated the PR description to reflect recent changes.  I'd like to find a way to avoid deprecation warnings if the user code is deriving `ARKStepODE` (b/c the `ARKStepSolver::TimeDependentODEWrapper` is still compiled); however, I want to leave that as a to-do item for now:

- [ ] avoid deprecation warnings if code derives `ARKStepODE`

linalg/sundials.hpp
```diff
@@ -706,9 +706,130 @@ class CVODESSolver : public CVODESolver
 // Interface to ARKode's ARKStep module -- Additive Runge-Kutta methods
 // ---------------------------------------------------------------------------
 
+// Interface for defining ODE systems to be evolved using ARKStepSolver:
+//
+//  1)   du/dt = inv(M) f(u,t) ("explicit" form)
+//  2) M dy/dt = f(u,t)        ("implicit" form)
```
cjvogl:
```suggestion
//  1)   du/dt = inv(M) f(u,t) ("MFEM" form)
//  2) M dy/dt = f(u,t)        ("mass" form)
```

linalg/sundials.hpp
```diff
@@ -706,9 +706,130 @@ class CVODESSolver : public CVODESolver
...
+   virtual void ARKEvaluateRHS(const Vector &u, const real_t t, Vector &result) const
+   {
+      mfem_error("This function must be specified for ERK or IMEX methods.");
+   }
+
+  /** setup linear system for solving [M(t) - gamma Jf(u)] dk = f(u) - M(t) k,
+      which is necessary for solving ODEs with DIRK or IMEX methods
+      @param[in]  u     The state at which A(@a u,t) should be evaluated.
+      @param[in]  t     The time at which A(u,@a t) should be evaluated.
+      @param[in]  v     The value of inv(M) f(u,t) for implicit or f(u,t) for explicit
```
cjvogl:
```suggestion
      @param[in]  v     The value of inv(M) f(u,t) or f(u,t) for depending on form.
```

examples/sundials/ex16.cpp
```diff
@@ -81,50 +80,90 @@ class ConductionOperator : public TimeDependentOperator
...
    mutable Vector z; // auxiliary vector
 
+   const bool use_mass_form;
+
 public:
 
    ConductionOperator(FiniteElementSpace &f, const real_t alpha,
                       const real_t kappa, const Vector &u,
-                      const Type &ode_expression_type);
+                      const bool use_explicit_form);
```
cjvogl:
```suggestion
                      const bool use_mass_form);
```

PR #5084, opened 2025-10-24
Adding [Par]ComplexGridFuction::Save

mlstowell:
This PR adds `Save` methods to the serial and parallel `ComplexGridFunction` classes. To test this functionality I write the solutions from examples `ex22` and `ex22p`. I also read the files back in and computed the difference with the original solution. I removed this difference check but left the writing.

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#5084](https://github.com/mfem/mfem/pull/5084) | @mlstowell | @tzanio | @psocratis + @najlkin | 10/24/25 | 11/4/25 | 11/8/25 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
      - [ ] Add the example code to the `ALL_EXE_SRCS` variable.
      - [ ] Make sure `THIS_TEST_OPTIONS` is set correctly for the new example.
[...]

PR #5094, opened 2025-10-29
Adding [Par]GridFunction::Save unit test (and bug fix)

mlstowell:
The `ParGridFunction::Save` function flips the sign of certain DoFs so that the single processor files can be properly read as serial `GridFunction` objects. This PR flips the signs when reading these saved files as a new `ParGridFunction` object so that the new object matches the original object.

This bug was noticed when preparing the unit tests for PR #5084. 

| [#5094](https://github.com/mfem/mfem/pull/5094) | @mlstowell | @tzanio | @acfisher + @cjvogl | 1/24/26 | 7/28/26 | 7/30/26 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


fem/pgridfunc.cpp
```diff
@@ -80,6 +80,12 @@ ParGridFunction::ParGridFunction(ParMesh *pmesh, std::istream &input)
                                     fes->GetOrdering());
    delete fes;
    fes = pfes;
+
+   real_t *data_  = const_cast<real_t*>(HostRead());
+   for (int i = 0; i < size; i++)
+   {
+      if (pfes->GetDofSign(i) < 0) { data_[i] = -data_[i]; }
+   }
```
mlstowell: To see this bug in action comment out these lines and run the unit tests in parallel:
```
mpirun -np 4 ./punit_tests "ParGridFunction Save"
```
This should produce something like the following:
```
ParGridFunction Save
tests/unit/fem/test_gridfunc_save.cpp:97

tests/unit/fem/test_gridfunc_save.cpp:186: FAILED:
  REQUIRE( diff_nd < 1e-8 * norm_nd )
with expansion:
  0.621622994795256 < 5.138964563872391e-08

test cases: 1 | 1 failed
assertions: 2 | 1 passed | 1 failed
```

najlkin:
Hmm, if the sign flipping is needed, is not it missing also in other constructors from local `GridFunction`s? 🤔 There, it would be more complicated, because we should not modify the data to keep the dual access 💣 .

mlstowell:
I don't know. It might be worth investigating.

mlstowell:
> Hmm, if the sign flipping is needed, is not it missing also in other constructors from local `GridFunction`s? 🤔 There, it would be more complicated, because we should not modify the data to keep the dual access 💣 .

The only constructor I see that might have a problem is `ParGridFunction(ParFiniteElementSpace *pf, GridFunction *gf)` because it does not take ownership of the data vector and therefore cannot change it. The comment for this c'tor states that the `ParFiniteElementSpace` must match the `FiniteElementSpace` in the `GridFunction`. This is something we could check. Clearly if they point to the same object everything should be fine but we could also compare the signs of the element DoFs in every element (h-refinement, p-refinement, or mesh repartitioning could all lead to problems which might not be apparent during construction). A more robust check would involve checking the `FiniteElementCollection` for ND or RT. The best solution would probably be to warn against using this c'tor with ND or RT and providing an alternative which copies the data.

mlstowell:
Of course the vector c'tors are also vulnerable to this issue but they seem risky in other contexts as well. We should probably put detailed warnings in their documentation blocks and place the burden to use them properly on the user.

najlkin:
With vectors, we do not know where they come from, so the documentation is the only hope, but with a `GridFunction` we can do better. I am a bit hesitant to put a check tied to some specific spaces. Maybe in debug, we could check something like `ldof_sign.Min()`. Or go through vdofs one by one and check `GetDofSign()`, which is also fine once in the constructor, I guess 🤔 .

tzanio:

cjvogl:
Sorry if what follows is naive.  My understanding of the issue is that, in order to support creating a `GridFunction` object from the output of `ParGridFunction::Save`, the dof signs and data are combined before calling `GridFunction::Save` within `ParGridFunction::Save`... and this results in an issue when a `ParGridFunction` object is created from the output of `ParGridFunction` (because the data in the output is assumed to not be combined with the dof signs).

If that understanding is correct, what are your thoughts on instead introducing `ParGridFunction::SaveAsSerial` (or similar name) that combines the dof signs and data for later creation of a `GridFunction`, so that `ParGridFunction::Save` can save the unaltered data (which can then be used to create a `ParGridFunction`)? My initial reaction is such an approach would be easier to maintain than requiring the `ParGridFunction` and `Vector` constructors to determine whether the input already has dof sign information in it... but I admit that I likely am not seeing the full picture here.

najlkin:
I think you have a good understanding of the essence of the problem, but we are dealing here with the "historical heritage" of MFEM as usual 😬 . Originally, everything was simple and easy with local grid functions matching the parallel. Then, the signs were introduced for certain spaces and elements, but without breaking changes even when the system was not able to accommodate it fully, which was not clear at that point perhaps. Now, we are in a half-broken state of things without a clear way out 🫣 .
In particular, methods like `SaveAsSerial()` make sense for a new project and would result in a clean design, but we have all MFEM code dependent on `Save()` already, which always saved the local grid function as standalone. There is no header distinguishing a `ParGridFunction` from `GridFunction` or `ParFiniteElementSpace` from `FiniteElementSpace` (Just recently, we introduced that for `(Par)ComplexGridFunction`s in #5084 to not run into issues of a similar kind). Parallel grid functions are always loaded as serial and then the `FiniteElementSpace` is converted to `ParFiniteElementSpace` hoping everything will match 🫠 . In my eyes, we need to fix both sides to keep the paradigm of serial functions still working as the sign flips are not frequent and we do not want to "rewrite everything". That might come at the prize of getting some abort messages in certain cases, because the classical construction cannot work with the flipping maybe, but the user could get an alternative and be pointed there 😉 .

cjvogl:
@najlkin , thanks for getting me up to speed! I have two follow-up questions (if you have time to answer them):

> Now, we are in a half-broken state of things

Is the broken state that calling the `ParGridFunction` constructor with `istream` input from the result of `ParGridFunction::Save` fails if certain spaces (that involve dof signs) are used? Or does the broken state go beyond just that workflow.

> In particular, methods like SaveAsSerial() make sense for a new project and would result in a clean design, but we have all MFEM code dependent on Save() already, which always saved the local grid function as standalone. 

What is the workflow that would break if `ParGridFunction::Save` did not flip the dof signs before saving? Another way to ask is what workflow would modifying `ParGridFunction::Save` in such a manner not be backwards compatible for?

The context for the questions, which could be asked of @mlstowell and/or @acfisher as well, is to understand the implications of one fix over another. I understand the concern that there exists code that assumes `ParGridFunction::Save` flips the dof signs before saving. Do you have a feel for if that existing code is more or less code than any code that assumes the `ParGridFunction` constructor (modified in this PR) does not flip the signs? Perhaps there is no code making that assumption about the `ParGridFunction`, in which case I agree this PR that maintains backwards compatibility with `ParGridFunction::Save` seems like the best fix. If there is "about as much" or "more" code that makes the assumption about the `ParGridFunction` constructor (i.e., code that this PR would **not** be backwards compatible with), then I would suggest weighing future maintainability given there will be some breaking of backwards compatibility.

cjvogl:
Another option to keep backwards compatibility with ‘ParGridFunction::Save’ (and avoiding introducing more dof flipping elsewhere in ParGridFunction) is to add an optional argument to its signature that flags whether the signs should be flipped, with a default value of true.  The user would then call myParGridFunc.Save(ofstream(“file.gf”)) to have the dof signs flipped to be used to construct a GridFunction… or myParGridFunc.Save(ofstream(“file.gf”), false) to avoid flips so the output can be used to construct a ParGridFunction using the existing constructor implementation.

najlkin:
> > Now, we are in a half-broken state of things
> 
> Is the broken state that calling the `ParGridFunction` constructor with `istream` input from the result of `ParGridFunction::Save` fails if certain spaces (that involve dof signs) are used? Or does the broken state go beyond just that workflow.

It is this workflow (Save/Load), which is (semi-)broken, but also other constructors from serial grid functions/vectors, where it was assumed the parallel grid function just views the local grid functions. This is not possible with the paradigm of local vs. global signs. Frankly, I would not go this pathway and rather would absorb the signs in the prolongation/restriction operators to keep local functions without a change, but there were some reasons for it, I guess, and it was done a long time ago...

> > In particular, methods like SaveAsSerial() make sense for a new project and would result in a clean design, but we have all MFEM code dependent on Save() already, which always saved the local grid function as standalone.
> 
> What is the workflow that would break if `ParGridFunction::Save` did not flip the dof signs before saving? Another way to ask is what workflow would modifying `ParGridFunction::Save` in such a manner not be backwards compatible for?

The files should be really serial, so you should be able to load each local part separately and process it or visualize etc. 

> The context for the questions, which could be asked of @mlstowell and/or @acfisher as well, is to understand the implications of one fix over another. I understand the concern that there exists code that assumes `ParGridFunction::Save` flips the dof signs before saving. Do you have a feel for if that existing code is more or less code than any code that assumes the `ParGridFunction` constructor (modified in this PR) does not flip the signs? Perhaps there is no code making that assumption about the `ParGridFunction`, in which case I agree this PR that maintains backwards compatibility with `ParGridFunction::Save` seems like the best fix. If there is "about as much" or "more" code that makes the assumption about the `ParGridFunction` constructor (i.e., code that this PR would **not** be backwards compatible with), then I would suggest weighing future maintainability given there will be some breaking of backwards compatibility.

A problem is that saving functions without flipping should introduce a new file format, because we do not make that distinction between serial and parallel grid functions. But then you have to support also the case someone tries to load an older file without the header, so you actually have to do something like flipping back, which makes that pathway of not flipping somewhat pointless. Overall, it seems to me an overkill 🪓 to introduce a special format here for some corner cases and definitely we should not break compatibility here of loading serial functions. So I am hesitant to add parameters for switching optionally. The only place where this is necessary is the `Vector` wrapper, where we cannot know how it was produced, so we need a switch there perhaps.

A question is what to do with the view of serial grid functions. This is broken by design as said. I thought it could give an error when the flipping is necessary. There is the more general `Vector` version (`GridFunction` is derived from `Vector`)  letting the user load the serial functions that way, but it would be clear we treat that input only as raw data without a guarantee keeping them valid as local grid functions. An alternative could be to auto-magically do a deep copy instead of a shallow (i.e., pointing to the data) 🪄 , but we have no guarantee the user will not use the original grid functions to further access the solution, so it can cause some hard-to-detect bugs for certain configurations of meshes and elements, so not really a good idea in my opinion 🫤 .

tests/unit/fem/test_gridfunc_save.cpp
```diff
@@ -0,0 +1,198 @@
...
+TEST_CASE("ParGridFunction Save", "[ParGridFunction][Parallel]")
+{
+   const int num_procs = Mpi::WorldSize();
+   const int my_rank = Mpi::WorldRank();
+   const int order = 3;
+
+   const int nx = (int)rint(cbrt(real_t(num_procs)));
+   const int ny = (int)rint(2.0 * cbrt(real_t(num_procs)));
+   const int nz = (int)rint(3.0 * cbrt(real_t(num_procs)));
+   Mesh mesh = Mesh::MakeCartesian3D(nx, ny, nz, Element::PYRAMID,
```
acfisher, line 106: Are these tests specific to Pyramids?  If so should we tag them with [Pyramids]?

fem/pgridfunc.cpp
acfisher: Good catch!

acfisher (approved):

cjvogl (approved):
> Another option to keep backwards compatibility with ‘ParGridFunction::Save’ (and avoiding introducing more dof flipping elsewhere in ParGridFunction) is to add an optional argument to its signature that flags whether the signs should be flipped, with a default value of true. The user would then call myParGridFunc.Save(ofstream(“file.gf”)) to have the dof signs flipped to be used to construct a GridFunction… or myParGridFunc.Save(ofstream(“file.gf”), false) to avoid flips so the output can be used to construct a ParGridFunction using the existing constructor implementation.

Given that some time has passed since I threw this option out there, I assume it is not gaining any popularity (and I can see an issue if the goal is to be able to create `ParGridFunction` objects from existing `ParGridFunction::Save` output).  Good catch on the bug in any case.

najlkin:
@mlstowell , see f1174bfbf5fb7aa26f395793025b0bc409aef89a commit from @v-dobrev , that can be used here with benefit.

fem/pgridfunc.cpp
najlkin: Newly, you may use `ApplyDofSigns()`

najlkin:
> With vectors, we do not know where they come from, so the documentation is the only hope, but with a `GridFunction` we can do better. I am a bit hesitant to put a check tied to some specific spaces. Maybe in debug, we could check something like `ldof_sign.Min()`. Or go through vdofs one by one and check `GetDofSign()`, which is also fine once in the constructor, I guess 🤔 .

Regarding this, @v-dobrev added `HaveDofSigns()` with optimization of having the signs only when really needed, so this check can be easily implemented.

tests/unit/fem/test_gridfunc_save.cpp
mlstowell: That's a good question. The tests are certainly written explicitly for meshes of pyramids, however, the problem could arise in an arbitrary mesh. I chose pyramids because I observed the problem while working on pyramid code but also because I knew this mesh constructor would reproduce the issue on a mesh of any size. I wasn't so confident about meshes of other types of elements.

I can certainly add the tag.

mlstowell:
> > With vectors, we do not know where they come from, so the documentation is the only hope, but with a `GridFunction` we can do better. I am a bit hesitant to put a check tied to some specific spaces. Maybe in debug, we could check something like `ldof_sign.Min()`. Or go through vdofs one by one and check `GetDofSign()`, which is also fine once in the constructor, I guess 🤔 .
> 
> Regarding this, @v-dobrev added `HaveDofSigns()` with optimization of having the signs only when really needed, so this check can be easily implemented.

@najlkin, thank you for pointing out the relevant additions by @v-dobrev. I agree that these should be used here as well. As was discussed in the PR meeting today we should wait for PR #5267 to be pushed to `master` then we can modify this PR to make use of @v-dobrev 's new functions.

najlkin:
Let us see how @tzanio decides, it is his PR. I thought this one could be merged first (with the additional fixes), which is cleaner, but I understand it is easier for testing to have one branch/PR for BLAST folks.

tzanio:
Yes, lets do PR https://github.com/mfem/mfem/pull/5267 first, sorry for the delay

tzanio:
@mlstowell, now that https://github.com/mfem/mfem/pull/5267 is merged, can you update this?

mlstowell:
After PR #5267 this PR simply contains a new unit test to verify the `[Par]GridFunction::Save` in both serial and parallel.

najlkin:
We were considering it for a followup, but now this PR is a/the followup by itself 🙂 , so I would consider protecting `ParGridFunction(ParFiniteElementSpace *pf, GridFunction *gf)` constructor as I mentioned [above](https://github.com/mfem/mfem/pull/5094#issuecomment-4298347352).

mlstowell:
> We were considering it for a followup, but now this PR is a/the followup by itself 🙂 , so I would consider protecting `ParGridFunction(ParFiniteElementSpace *pf, GridFunction *gf)` constructor as I mentioned [above](https://github.com/mfem/mfem/pull/5094#issuecomment-4298347352).

@najlkin, I don't think I understand what you're suggesting. I see at least three options but maybe you've got a better idea. First, we could check for the existence of DoF signs and flips them if necessary to create a valid `ParGridFunction` (this would seem to violate an implicit promise made by the comment which states that it doesn't take ownership.) Second and third, we could issue a warning or failed assertion, respectively, if `HaveDofSigns` is true.

I'd be happy to implement either of these three options or consider alternatives.

najlkin:
All make sense to me 😉 . Let me comment on them:

1. Changing the signs of the grid functions breaks the implicit assumption that the constructor is a "view" constructor, but the documentation only says "The data from gf is used as the local data..." This can be interpreted different ways 🤔 . Of course, there should be a note in the documentation and a runtime warning would be also appropriate. I do not see a problem with ownership, the constructor does not take ownership, but it indicates write access to the grid function by the non-const pointer, so that is fine in my opinion. This probably follows the intention of most users to create the parallel grid function and not use the serial ones anymore.
2. Giving a warning and not fixing the signs is perfectly backward compatible, but leaves users with a buggy code and requires their intervention to fix it (which is to apply the signs on (a copy of) the original data and use the vector constructor to avoid the warning).
3. Aborting the code forces the users to fix the code and leads to the best code potentially (which will not rely on the invalid concept of a view on serial data). But it might not be well received by users, as this is a relatively rare case.

mlstowell:
@najlkin, I like your analysis of these options. I suggest we follow option 1 with a small modification. Let's add an optional  argument to the constructor, a boolean offering the option to preserve the serial grid functions or not. If no sign flip is necessary we can just continue to share data pointers and everything will be fine. If a sign flip is needed and the caller wants to preserve the serial grid functions we issue an error, if they don't want to preserve we flip the signs without error or warning.

I also suggest we default to preserve the serial grid functions and issuing an error. This way callers will be notified of the potential problem and they can make an informed decision.

najlkin:
I think that is a good compromise between too forgiving, but potentially dangerous, 1 and too strict 3 👍 . Also, there is the advantage that we do not need to show any warnings, which might be sometimes too soft and other times excessive when users know what they are doing, but just want to do it the easy way ☺️ .

mlstowell:
Option 1, as described above, has been implemented. I also added a unit test to verify that creation of a `ParGridFunction` object from the serial `GridFunction` objects on each processor works as expected.

@cjvogl and @acfisher, you may want to take another quick look.

fem/pgridfunc.hpp
```diff
@@ -100,8 +100,12 @@ class ParGridFunction : public GridFunction
    /// Construct a ParGridFunction using a GridFunction as external data.
    /** The parallel space @a *pf and the space used by @a *gf should match. The
        data from @a *gf is used as the local data of the ParGridFunction on each
-       processor. The ParGridFunction does not assume ownership of the data. */
-   ParGridFunction(ParFiniteElementSpace *pf, GridFunction *gf);
+       processor. The ParGridFunction does not assume ownership of the data.
+       The boolean, @a preserve, indicates that the data stored in @a *gf should
+       remain unchanged. An error will occur if @a preserve is true and this
+       preservation would produce an invalid ParGridFunction. */
```
najlkin: It is a theoretical construction, I would go the other way around and rather say something like: "..., but construction of ParGridFunction requires to change the data." 🤔

fem/pgridfunc.cpp
```diff
@@ -22,10 +22,20 @@ using namespace std;
...
-ParGridFunction::ParGridFunction(ParFiniteElementSpace *pf, GridFunction *gf)
+ParGridFunction::ParGridFunction(ParFiniteElementSpace *pf, GridFunction *gf,
+                                 bool preserve)
 {
    fes = pfes = pf;
    SetDataAndSize(gf->GetData(), gf->Size());
+
+   if (pfes->HaveDofSigns())
+   {
+      MFEM_ASSERT(!preserve, "Differing sign conventions for the serial and "
```
najlkin: We may do even `MFEM_VERIFY` in this case, I think.

mlstowell: My mistake. I confused the behavior of these two. Thanks!

fem/pgridfunc.hpp
mlstowell: I changed the wording but if I misunderstood your point just let me know.

najlkin (approved):
Looks good 👍 Thanks! We will see what happens in nigthly tests 😄

mlstowell:
> Looks good 👍 Thanks! We will see what happens in nigthly tests 😄

Thanks, @najlkin, I appreciate the feedback. It was certainly a good choice to extend this PR to cover another save/read scheme.

tzanio:
any idea why the static analysis check is failing?

najlkin:
It is not finished. The annotation says: `The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.` 🤔

najlkin:
Some fluke, I guess 🤷 . Finished fine now.

tzanio:

PR #4441, opened 2024-08-10
Parallel Hcurl NURBS space and an example named nurbs_25p

kaipengliuHIT:
After some minor changes, parallel Hcurl NURBS space IGA supported. And add an example nurbs_25p, it is the parallel nurbs version of the example 25. the resualt can be shown by paraview:
paraview

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#4441](https://github.com/mfem/mfem/pull/4441) | @kaipengliuHIT | @tzanio | @justinlaughlin + @dylan-copeland | 8/13/24 | ⌛due 8/27/24 | ⌛due 9/3/24 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
      - [ ] Add the example code to the `ALL_EXE_SRCS` variable.
      - [ ] Make sure `THIS_TEST_OPTIONS` is set correctly for the new example.
[...]

PR #5098, opened 2025-11-05
Add parallel capability to NURBS based VectorFE

IdoAkkerman:
To achieve this most of the partitioning logic has moved from `ParNURBSExtension` to `ParFiniteElementSpace`.

As a consequence `ParNURBSExtension` loses some data members and  member functions.
Moreover   `ParFiniteElementSpace` has a pointer to a `GroupTopology`.
In the case of NURBS this will be generated and destroyed in the  `ParFiniteElementSpace`, this is proteced with an appropriate `own` flag.

Crucial parts to check:
1.  On line 47&48 of pfespace.hhp
```
   GroupTopology *gtopo;
   bool own_gtopo = false;
```
Is it okay to store this data here?

2.  On line 134 of pfespace.hhp
```
 GroupTopology &GetGroupTopo() const { return *gtopo; }
```
Will this work for all the other cases??? All test cases pass but perhaps this function should have a check if the pointer exists: If not,  run  the old code.

3. Should this function  be deleted or deprecated? It is not used in MFEM:
On line 1112&1123 of pfespace.hhp
```
GroupCommunicator *ParFiniteElementSpace::ScalarGroupComm()
{
   GroupCommunicator *gc = new GroupCommunicator(GetGroupTopo());
   if (NURBSext)
   {
      //gc->Create(pNURBSext()->ldof_group); // TODO
   }
   else
   {
      GetGroupComm(*gc, 0);
   }
   return gc;
}
```

### DIFF
--- a/fem/pfespace.cpp
+++ b/fem/pfespace.cpp
@@ -1872,7 +1872,6 @@ void ParFiniteElementSpace::ConstructTrueDofs()
 void ParFiniteElementSpace::ConstructTrueNURBSDofs()
 {
    int n = GetVSize();
-   int nt;
 
    IntegerSet         group;
    ListOfIntegerSets  groups;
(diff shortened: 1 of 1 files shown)


kaipengliuHIT:
Regarding this feature, it can be seen in the continuation of #4441 and #5143.

PR #5104, opened 2025-11-11
Remove order from NURBS FE Collection

IdoAkkerman:
Remove largely unused order memberfunction from  NURBS FE Collections.

Problems will arise in reading a mesh in format 90. Perhaps this format should be removed.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,11 @@
 
                                https://mfem.org
 
+Refactor of NURBS
+
+Remove format 90??
+
+
 
 Version 4.8.1 (development)
 ===========================
--- a/fem/fe/fe_nurbs.cpp
+++ b/fem/fe/fe_nurbs.cpp
@@ -26,6 +26,9 @@ void NURBS1DFiniteElement::SetOrder() const
 
    weights.SetSize(dof);
    shape_x.SetSize(dof);
+#ifndef MFEM_THREAD_SAFE
+   vshape.SetSize(dof, dim);
+#endif
 }
 
 void NURBS1DFiniteElement::CalcShape(const IntegrationPoint &ip,
@@ -101,6 +104,9 @@ void NURBS2DFiniteElement::SetOrder() const
    u.SetSize(dof);
    du.SetSize(dof);
    weights.SetSize(dof);
+#ifndef MFEM_THREAD_SAFE
+   vshape.SetSize(dof, dim);
+#endif
 }
 
 void NURBS2DFiniteElement::CalcShape(const IntegrationPoint &ip,
@@ -238,6 +244,9 @@ void NURBS3DFiniteElement::SetOrder() const
    u.SetSize(dof);
    du.SetSize(dof);
    weights.SetSize(dof);
+#ifndef MFEM_THREAD_SAFE
+   vshape.SetSize(dof, dim);
+#endif
 }
 
 void NURBS3DFiniteElement::CalcShape(const IntegrationPoint &ip,
@@ -437,6 +446,9 @@ void NURBS_HDiv2DFiniteElement::SetOrder() const
    u.SetSize(dof);
    du.SetSize(dof);
    weights.SetSize(dof);
+#ifndef MFEM_THREAD_SAFE
+   vshape.SetSize(dof, dim);
+#endif
 }
 
 void NURBS_HDiv2DFiniteElement::CalcVShape(const IntegrationPoint &ip,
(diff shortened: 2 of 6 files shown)


IdoAkkerman:
The problem is when reading the mesh a `GridFunction` is created on L5130 in mesh.cpp. The GridFunction reads the associated FiniteElementSpace from file using `fes->Load(mesh, stream)`.
In this `Load` function there is some logic to decide whether a new NURBSExtension needs to be build or not. The problem is that this is not a "normal" fespace/gridfunction this is the fespace/gridfunction that is essential to define  the mesh.
This distinction is not very clearly conveyed. I remember from the past that this was always tricky (=fragile!) . This is also the case when reading in Visit/glvis solutions.

What would be a good way to proceed?

Perhaps we should remove fes_format 90. As I think fes_format 100 forces the NURBSExtension reuse if order info is not explicitly provided.

IdoAkkerman:
This is part of #5103

justinlaughlin:
>The problem is that this is not a "normal" fespace/gridfunction this is the fespace/gridfunction that is essential to define the mesh.

I *think* these special gridfunctions will own the fec/fespace (denoted by `fec_owned`). I agree it is confusing.

>What would be a good way to proceed?

I've been trying to figure out a satisfactory answer to this for the past few weeks 😅 Thank you for the discussions and helping to figure something out together.

I think I like the idea of removing order from NURBS FE Collection completely, but it does serve one purpose (that I've discovered), which is upon reading a nurbs mesh file you could have a nurbs space with different order than that of the knot vectors. 

There is also the ownership issue; one possibility I thought of was fespace could own the nurbs_fec but I don't want to introduce another custom "maybe ownership" rule.

justinlaughlin:
This is my understanding of the basic nurbs classes and ownership. Thin arrows are just a raw pointer, thick arrows represent ownership (inferred from the destructor), thick arrows with text indicate maybe-ownership (depends on the listed variable). I have this written on paper but figured it'd be helpful to share it here 😅 

```mermaid
flowchart TD
    fec[fec]
    gf[gf] ==>|fec_owned| fec
    gf[gf] ==>|fec_owned| fespace
    mesh[mesh] ==>|own_nodes| gf
    fespace[fespace] --> fec
    mesh ==> nurbsext
    fespace ==>|own_ext| nurbsext
    fespace --> mesh
    nurbsext ==> knotvectors
    fec ==> nurbsfe
    nurbsfe --> knotvectors
```

IdoAkkerman:
Thanks for the figure. I will study more closely. Worth mentioning I am working on a branch where I remove `patches` and `KnotVectorCompr` as data members. Removing the patches has a bit more consequences then I intended, with the benefit that it forces a rethink on how to approach things.

I think I found a solutions, the NURBSExt constructor that reads data from file should result in the same state independent whether the data is in "original" or in patch format.  Proposed solution let the constructor read or construct the coefficient vector needed for the Nodes grid function.

Also the combo `SetCoordFormPatches`  and `ConvertToPatches` loads or removes the patches, which results in the ephemeral problems. The underlying issue is the different states of the object, aka side effects. I am reading up on functional programming... or perhaps another way of phrasing, function calls should  be atomic...

I want to have `patches` be a local data structure in degree-elevate/refine/coarsen etc. This is perhaps not 100% efficient but should result in a big leap in clarity, which I think is preferable. As the efficiency only pertains to the preprocessing stage, not the computations them selves.

IdoAkkerman:
This is just a small graphical trick, but I would say this figure explains the FESpace used for the solution, such as defined in `nurbs_ex1` for instance. However, there  is also a Nodes `gridfunction` in the `Mesh` for this one I would move he `fespace` below `Mesh`. The connectivity remains the same, but visually it communicates "how is first"/"how is the bose"/"who owns whom".

IdoAkkerman:
Perhaps `degreeElevate`/`refine`/etc should reside in `Gridfunction`?

IdoAkkerman:
Perhaps the simplest is that the `mesh->Refine` calls `NURBSExt->Refine`, with `Nodes` as argument.

justinlaughlin:
>Also the combo SetCoordFormPatches and ConvertToPatches loads or removes the patches, which results in the ephemeral problems. The underlying issue is the different states of the object, aka side effects. I am reading up on functional programming... or perhaps another way of phrasing, function calls should be atomic...

Yes exactly! I believe there are a few other commonly called methods (`UpdateNURBS`?) that will just delete the patches; although they are a very helpful data representation, I have found this confusing to work with. Imo one of our main goals should be to manage state better (or, where reasonable, not have it at all). Although there is something to be said about state and performance (eg PA and caching precomputed values), imo its always a trade-off with code complexity. For classes that are light on data such as knotvectors, imo we could very well make them stateless, which would make inference of higher-level classes easier (eg we don't have to ask, have the knotvectors been changed?).

justinlaughlin:
I think size of data and how often it will need to be moved are important factors to consider regarding the design choice of state vs state-less

PR #5105, opened 2025-11-11
Remove tables from NURBSext

IdoAkkerman:
As the some of the `active` logic is performed in GenerateTable, but this logic is trivial in the serial case, I had to move routines to `ParNURBSExtension` and make them virtual.

In the future all `active` data members should perhaps be stored in the `ParNURBSExtension`

### DIFF
--- a/mesh/nurbs.cpp
+++ b/mesh/nurbs.cpp
@@ -4528,7 +4528,8 @@ void NURBSExtension::GetElementLocalToGlobal(Array<int> &lelem_elem)
       }
 }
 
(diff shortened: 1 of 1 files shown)


IdoAkkerman:
This is related to #5103

justinlaughlin:
I think I like the idea in principle. Do you know if it affects performance?

IdoAkkerman:
I think the performance cost is very little, extra tables are only generated during (some) initializations.


Bigger problem is the use of the table in the `LoadFE` and `LoadBE` routines. I do not think there is an easy way around this.
I tried to fix things in the calling FESpace, but I don't think it there is an easy fix.

PR #5106, opened 2025-11-11
Remove format 90

IdoAkkerman:
Bootstrapping of the FiniteElement space for the mesh Nodes is difficult/fragile in format 90.

### DIFF
--- a/data/nc-nurbs3d.mesh
+++ b/data/nc-nurbs3d.mesh
@@ -275,11 +275,11 @@ weights
 1.0
 1.0
 
-FiniteElementSpace
+MFEM FiniteElementSpace v1.0
 FiniteElementCollection: NURBS1
 VDim: 3
 Ordering: 1
-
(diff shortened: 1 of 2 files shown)


IdoAkkerman:
Related to #5103

justinlaughlin:
All of the nurbs examples are v90 currently right? Maybe we should deprecate for a period before removing it (I'd imagine this could impact downstream codes).

One other thought is that since v100 doesn't seem to be in-use (I think?), if we're going to upgrade the versions we could definitely make a "v110", if there are other features that would be useful.

IdoAkkerman:
I think the fix to the files are very easy. A simple change on the first and last line of the header of the FiniteElementSpace will fix things.

I am not sure the proposed changes in other PRs, like removing order form FEC, can be achieved without this change.
Perhaps we should make it DEPRECATED now(or ASAP) to be in a position to remove to for v5.0.

PR #4835, opened 2025-04-28
NURBSExtension Constructor and edge_to_knot mapping

justinlaughlin:
## Summary
This PR 
- [x] Adds a new algorithm, `Mesh::GetEdgeToUniqueKnotvector`, to calculate the `edge_to_knot` map (now renamed `edge_to_ukv`, where `ukv` is the "unique knot vector" index) used by NURBS meshes
- [x] `Mesh::GetEdgeToUniqueKnotvector` also produces the intermediary index maps. For now the only one that is needed/returned is `ukv_to_rpkv`, used by the patch-based `NURBSExtension` constructor.
- [x] Updates the `NURBSExtension` constructor using `NURBSPatch`(es) and `patchTopo` to use `Mesh::GetEdgeToUniqueKnotvector`
- [x] Adds a new test for the `NURBSExtension` constructor
- [x] Adds a few more NURBS test meshes; `miniapps/nurbs/plus-nurbs-permuted.mesh` is an example mesh that would not work previously but now does.
- [x] Modifies the `miniapps/nurbs` CMakeLists so that `miniapps/nurbs/meshes` is copied over on builds

## Background

This work began while implementing code to automatically create a low-order version of a NURBS mesh — keeping the number of DOFS and placing the new knots at e.g. Greville abscissa of the higher order mesh. The cleanest approach in my opinion is to grab the `NURBSPatch`(es) and `patchTopo` of a `Mesh`, make adjustments, then feed that into the new `NURBSExtension` constructor introduced in #4566.

Doing so, I found that we'd often get a `Inconsistent edge-to-knot mapping` error. The setting of `edge_to_knot` in the constructor was 1) not accounting for the sign-convention `if (v[0] > v[1]) { edge_to_knot[j] = -1 - edge_to_knot[j]; }` and 2) using the "patch knot vector index", `pkv = p*dim + d` (`p` is the patch index, `dim` is the total number of dimensions, and `d` is the local dimension index) rather than the "unique knot vector index". E.g. 3 patches in 3d joined on the same axis would have 5 "unique knot vectors", but 9 "patch knot vectors".

[...]

PR #5123, opened 2025-11-23
Fix nurbs orientations

justinlaughlin:
Adds `Mesh::CorrectPatchTopoOrientations()` which will attempt to flip signs in `edge_to_ukv` such that the condition in `NURBSExtension::CheckPatches()` is satisfied.

~~Flipping strategy is to loop through patches, keep track of number of flips per edge, and flip edges that have been flipped the least amount of times.~~
Flipping strategy uses neighboring patches to determine next flips. Thank you @dylan-copeland.

Related to PR #4835 and Issue https://github.com/mfem/mfem/issues/4915

| [#5123](https://github.com/mfem/mfem/pull/5123) | @justinlaughlin | @tzanio | @dylan-copeland + @tzanio | 12/16/25 | 12/25/25 | 12/29/25 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>

### DIFF
--- a/mesh/mesh.cpp
+++ b/mesh/mesh.cpp
@@ -6550,8 +6550,8 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
    const int NPKV = NP * dim;    // number of patch knotvectors
    constexpr int notset = -9999999;
    // Sign convention
-   auto sign = [](int i) { return -1 - i; };
-   auto unsign = [](int i) { return (i < 0) ? -1 - i : i; };
+   auto flipSign = [](int i) { return -1 - i; };
+   auto unSign = [](int i) { return (i < 0) ? -1 - i : i; };
    // Local edge index -> dimension convention
    auto edge_to_dim = [](int i) { return (i < 8) ? ((i & 1) ? 1 : 0) : 2; };
 
@@ -6567,7 +6567,7 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
       {
          GetElementVertices(i, v);
          // Sign is based on the edge's vertex indices
-         edge_to_ukv[i] = (v[1] > v[0]) ? i : sign(i);
+         edge_to_ukv[i] = (v[1] > v[0]) ? i : flipSign(i);
          ukv_to_rpkv[i] = i;
       }
       return;
@@ -6617,14 +6617,14 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
          // We've set this edge already - link this index to it
          if (edge_to_pkv[edge] != notset)
          {
-            const int pkv_other = unsign(edge_to_pkv[edge]);
+            const int pkv_other = unSign(edge_to_pkv[edge]);
             unite(pkv, pkv_other);
          }
          else
          {
             GetEdgeVertices(edge, v);
             // Sign is based on the edge's vertex indices
-            edge_to_pkv[edge] = (v[1] > v[0]) ? pkv : sign(pkv);
+            edge_to_pkv[edge] = (v[1] > v[0]) ? pkv : flipSign(pkv);
          }
       }
    }
@@ -6651,10 +6651,10 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
    edge_to_ukv.SetSize(NumOfEdges);
    for (int i = 0; i < NumOfEdges; i++)
    {
-      const int pkv = unsign(edge_to_pkv[i]);
+      const int pkv = unSign(edge_to_pkv[i]);
       const int rpkv = pkv_to_rpkv[pkv];
       const int ukv = rpkv_to_ukv[rpkv];
-      edge_to_ukv[i] = (edge_to_pkv[i] < 0) ? sign(ukv) : ukv;
+      edge_to_ukv[i] = (edge_to_pkv[i] < 0) ? flipSign(ukv) : ukv;
    }
 
    CorrectPatchTopoOrientations(edge_to_ukv);
@@ -6666,7 +6666,7 @@ void Mesh::CorrectPatchTopoOrientations(Array<int> &edge_to_ukv) const
    if (dim == 1) { return; }
 
    // Sign convention
-   auto sign = [](int i) { return -1 - i; };
+   auto flipSign = [](int i) { return -1 - i; };
 
    const Table *face2elem = GetFaceToElementTable();
    Array<int> pfaces, orient;
@@ -6686,7 +6686,7 @@ void Mesh::CorrectPatchTopoOrientations(Array<int> &edge_to_ukv) const
          for (auto e : fe)
          {
             const int skv = edge_to_ukv[e];
-            if (skv == kv || sign(skv) == kv) { hasKV = true; }
+            if (skv == kv || flipSign(skv) == kv) { hasKV = true; }
          }
          if (hasKV)
          {
@@ -6730,7 +6730,7 @@ void Mesh::CorrectPatchTopoOrientations(Array<int> &edge_to_ukv) const
       for (int i = 0; i < pe.Size(); i++)
       {
          ukvs[i] = edge_to_ukv[pe[i]];
-         ukvs[i] = (oe[i] < 0) ? sign(ukvs[i]) : ukvs[i];
+         ukvs[i] = (oe[i] < 0) ? flipSign(ukvs[i]) : ukvs[i];
       }
 
       // Find the direction with this kv.
@@ -6738,7 +6738,7 @@ void Mesh::CorrectPatchTopoOrientations(Array<int> &edge_to_ukv) const
       for (int d=0; d<dim; ++d) // Loop over directions.
       {
          const int skv = edge_to_ukv[pe[dir_edges[d][0]]];
-         if (skv == kv || sign(skv) == kv)
+         if (skv == kv || flipSign(skv) == kv)
          {
             thisDir = d;
          }
@@ -6775,12 +6775,12 @@ void Mesh::CorrectPatchTopoOrientations(Array<int> &edge_to_ukv) const
          }
 
          const int edge = pe[i];
-         if ((dim == 2 && ukvs[i] != sign(ukvs[ref_edge0])) ||
-             (dim == 3 && ukvs[i] == sign(ukvs[ref_edge0])))
+         if ((dim == 2 && ukvs[i] != flipSign(ukvs[ref_edge0])) ||
+             (dim == 3 && ukvs[i] == flipSign(ukvs[ref_edge0])))
          {
             // Flip the sign of this edge
             MFEM_VERIFY(!edgeSet[edge], "");
-            edge_to_ukv[edge] = sign(edge_to_ukv[edge]);
+            edge_to_ukv[edge] = flipSign(edge_to_ukv[edge]);
          }
 
          edgeSet[edge] = true;
@@ -6838,7 +6838,7 @@ void Mesh::CorrectPatchTopoOrientations(Array<int> &edge_to_ukv) const
       }
 
       const int kv_signed = edge_to_ukv[pe[dir_edges[unsetDim][0]]];
-      const int kv = kv_signed < 0 ? sign(kv_signed) : kv_signed;
+      const int kv = kv_signed < 0 ? flipSign(kv_signed) : kv_signed;
       MFEM_VERIFY(!edgeSet[pe[dir_edges[unsetDim][0]]], "");
 
       initKV = false;


IdoAkkerman:
@justinlaughlin As part of the NurB refactoring, I did some preliminary work on deleting data structures, you have already commented on somethings...

Perhaps also have a look at:

https://github.com/mfem/mfem/tree/rn_rm_compreKV_Patches

This PR removes the comprehensiveKVs and the Patches.
But still needs fixing wrt the orientations.
I think this PR likely touches some of my proposed changes.

IdoAkkerman:
I think the orientation stuff will be one of the main things to address in the refactoring.


We need to document and explain clearly what is going on where, why and how.
Think there are now 3 or 4 points in the code where "stuff happens", and they all kind of fix a problem somebody was having at some point. But a comprehensive view (and solution) seems missing.

Removing the comprehensive KV was my first attempt at this.

There seems to be a choice between two "evils"

1) Provide your own edge_to_ukv map. This can be annoying, and difficult for complex situations.

2) Let MFEM sort it out. But in this case what assumptions will the algorithm make? 


Perhaps the fix should be to enforce the patch format? This does seem to be quite natural, but we would need a policy if thinks do not match
1 Give an error 
2 Provide the common refinement
3 Allow a NC type solution

dylan-copeland (approved):

justinlaughlin:
@IdoAkkerman I think the case where there are no edges provided for the patch-topology will always be tricky - and this is something that would be added to `Mesh` as it is not NURBS specific (see eg `Mesh::CheckElementOrientation` -> `case Element::HEXAHEDRON`). Assuming that the user provides edges that are correct (or we happen to generate correct ones), I believe the algorithm Dylan implemented should consistently resolve the correct knotvector orientations.

There is a separate problem of how we use these orientations downstream; I agree that there are multiple different ways it is being used. Nevertheless, this PR ensures they are correct upon initialization - we can refactor internals in a separate PR.

tzanio:

mesh/mesh.cpp
```diff
@@ -9563,6 +9802,8 @@ void Mesh::GetVertices(Vector &vert_coord) const
 
 void Mesh::SetVertices(const Vector &vert_coord)
 {
+   MFEM_VERIFY(vert_coord.Size() == spaceDim * NumOfVertices, "");
+   vertices.SetSize(NumOfVertices);
```
IdoAkkerman, line 9814: Is this the correct location for the `SetSize` command. I think the routine was designed to have the vertices already  set. This is an assumption...
Either way this change seems to have "wider implications", although very limited in terms of actual impact. 

If this resize is deemed appropriate: Should  `vertices.SetSize` outside this routine be removed?

mesh/mesh.hpp
```diff
@@ -527,6 +527,9 @@ class Mesh
    void PrintTopoEdges(std::ostream &out, const Array<int> &e_to_k,
                        bool vmap = false) const;
 
+   /// Set signs to ensure knotvectors are pointed in the same direction.
+   void CorrectPatchTopoOrientations(Array<int> &edge_to_ukv) const;
+
```
IdoAkkerman, line 532: Should this be a member function of `Mesh`? Could it also be a memberfunction in NURBSExtension.

PS: Due to this PR I checked the `Mesh` class, and I personally think there are a bit to many NURBS specific references. This would be for the NURBS rewrite.
But perhaps we should start with not "polluting" `Mesh` even more in the mean time.

tests/unit/CMakeLists.txt
```diff
@@ -68,8 +68,6 @@ set(UNIT_TESTS_SRCS
   mesh/test_psubmesh.cpp
   mesh/test_submesh.cpp
   mesh/test_vtu.cpp
-  mesh/test_nurbs.cpp
-  mesh/test_exodus_writer.cpp
```
IdoAkkerman, line 72: Typo??

mesh/mesh.cpp
```diff
@@ -6653,6 +6656,242 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
...
+   CorrectPatchTopoOrientations(edge_to_ukv);
+}
+
+void Mesh::CorrectPatchTopoOrientations(Array<int> &edge_to_ukv) const
+{
+   const int dim = Dimension(); // Topological (not physical) dimension
+   if (dim == 1) { return; }
+
+   // Sign convention
+   auto sign = [](int i) { return -1 - i; };
```
IdoAkkerman: name of this lambda is slightly confusing. Perhaps `flip` or `flipOrientation` is more appropriate? Or `reverse`.

mesh/mesh.hpp
IdoAkkerman: Is the name of the routine appropriate?
It does not correct the mesh right? But only communicates the directions, of edges aka kvs

IdoAkkerman: Perhaps this correction of edges can be merged with `NURBSExtension::CheckPatches()` with a flag as parameter 
`NURBSExtension::CheckPatches(bool fixit = true)`  similar to `Mesh::CheckElementOrientation`

mesh/mesh.cpp
```diff
@@ -6653,6 +6656,242 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
...
+      {
+         // Check whether this face contains kv.
+         GetFaceEdges(face, fe, feo);
+         bool hasKV = false;
+         for (auto e : fe)
+         {
+            const int skv = edge_to_ukv[e];
+            if (skv == kv || sign(skv) == kv) { hasKV = true; }
+         }
+         if (!hasKV) { continue; }
```
IdoAkkerman: I think this is more readable when written as:

```
if (hasKV)
{
   bla
}
```

mesh/mesh.cpp
```diff
@@ -6653,6 +6656,242 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
...
+   if (dim == 1) { return; }
+
+   // Sign convention
+   auto sign = [](int i) { return -1 - i; };
+
+   const Table *face2elem = GetFaceToElementTable();
+   Array<int> pfaces, orient;
+   Array<int> fe, feo;
+
+   auto faceNeighbors = [&](int p, int kv, std::unordered_set<int> &nghb)
```
IdoAkkerman, line 6676: Please comment what this lamda does.
Does it find all elements that  share the edge on which the knotvector is defined?  
If so you could also use
```
  Table edge_face = Transpose(face_edge);
```
to make the appropriate `Table` and use that?

Or does the lambda do something slightly different?


PS: The table names (and presence) in `Mesh`  should/could be sanitized.

mesh/mesh.cpp
```diff
@@ -6653,6 +6656,242 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
...
+         }
+         if (!hasKV) { continue; }
+
+         Array<int> row;
+         face2elem->GetRow(face, row);
+         for (auto elem : row) { nghb.insert(elem); }
+      }
+   };
+
+   std::vector<std::vector<int>> dir_edges;
```
IdoAkkerman, line 6700: This info is not available in `fem/geom.hpp` is it?

Should it? If yes, then beyond this PR I guess...

mesh/mesh.cpp
```diff
@@ -6653,6 +6656,242 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
...
+      GetElementEdges(p, pe, oe);
+
+      // Get the signed unique knot vector indices
+      for (int i = 0; i < pe.Size(); i++)
+      {
+         ukvs[i] = edge_to_ukv[pe[i]];
+         ukvs[i] = (oe[i] < 0) ? sign(ukvs[i]) : ukvs[i];
+      }
+
+      // Find the dimension with this kv.
```
IdoAkkerman: dimension --> direction
thisDim -> thisDir

mesh/mesh.cpp
```diff
@@ -6653,6 +6656,242 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
...
+      {
+         const int skv = edge_to_ukv[pe[dir_edges[d][0]]];
+         if (skv == kv || sign(skv) == kv)
+         {
+            thisDim = d;
+         }
+      }
+      if (thisDim == -1)
+      {
+         return false;
```
IdoAkkerman: Or should this give an error?

mesh/mesh.cpp
```diff
@@ -6653,6 +6656,242 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
...
+         if (edgeSet[edge])
+         {
+            ref_edge0 = ref_edge;
+         }
+      }
+
+      if (initKV && !edgeSet[pe[ref_edge0]])
+      {
+         visited.erase(p);
+         return false; // There is no set edge in this direction on this patch.
```
IdoAkkerman, line 6763: Should this give and error? 
Or should we set `int ref_edge0 = -1` as start value?
Why exit the routine instead of assigning all kvs in patch?

mesh/mesh.cpp
```diff
@@ -6653,6 +6656,242 @@ void Mesh::GetEdgeToUniqueKnotvector(Array<int> &edge_to_ukv,
...
+   };
+
+   Array<bool> edgeSet(NumOfEdges);
+   edgeSet = false;
+
+   std::unordered_set<int> unset;
+   for (int i=0; i<NumOfElements; ++i) { unset.insert(i); }
+
+   const int max_iter = 3 * NumOfElements;
+   for (int iter=0; iter<max_iter; ++iter)
```
IdoAkkerman, line 6799: I think I roughly understand what this loop will do.
But I think the  this entire routine, and this loop in particular could benefit from extra explanation. How do the seet, unset, visited atc flags work? What are the assumptions. My guess is that Mobius strip like meshes will not work  for this routine as they create an error, infinite loop, or an end result that does not pass the test. This is worth checking..
This is one of the reasons I have not started on multipatch VectorFE.

So perhaps this reasonable limitation  (if it is indeed the case) should be documented.

miniapps/meshing/reflector.cpp
```diff
@@ -762,7 +761,10 @@ bool GetMeshElementOrder(Mesh const& mesh, Vector const& origin,
    return true;
 }
 
-Mesh* ReflectHighOrderMesh(Mesh & mesh, Vector origin, Vector normal)
+Mesh* ReflectHighOrderMesh(Mesh &mesh,
```
IdoAkkerman, line 764: Only the signature changed. The new variables are not used- apparently- is this correct?

miniapps/meshing/reflector.cpp
```diff
@@ -1006,6 +1007,125 @@ Mesh* ReflectHighOrderMesh(Mesh & mesh, Vector origin, Vector normal)
...
+
+   // Reflect the control points in this reflected patch `rpatch`.
+   Vector vr(3);
+   for (int i=0; i<dims[0]; ++i)
+   {
+      for (int j=0; j<dims[1]; ++j)
+      {
+         for (int k=0; k<dims[2]; ++k)
+         {
+            const int old = permArray(i,j,k);
```
IdoAkkerman, line 1050: Why the need in these 4 lines to shuffle the i,j,k indices?
Why not simply flipping the coordinates? Or is this to prevent negative jacobians (incorrect orientation)?


Perhaps explain the shuffle --> or is this to complicated? Which is kind-of why I ask :)?

IdoAkkerman:
@justinlaughlin I also gave some more generic comments that are more notes-to-self wrt the NURBS rewrite.

tests/unit/CMakeLists.txt
dylan-copeland: In the `master` branch, these files are listed twice.

mesh/mesh.cpp
dylan-copeland: I think `flipSign` would be clearer and still brief.

mesh/mesh.cpp
dylan-copeland: Vertices is not always set. This issue arises for example at `reflector.cpp` line 1100: `patchTopo.SetVertices(vert_coord);`. In this case, `vertices` has size 0 initially. If it is already set, then its size should already be `NumOfVertices`, and the fact that `SetVertices` is being called means it is fine to change its size anyway. I don't see a problem with setting the size, if we are setting the coordinates.

mesh/mesh.hpp
dylan-copeland: I agree that it would be best to separate NURBS-related functionality from the `Mesh` class. In this case, this function is called from the existing function `Mesh::LoadPatchTopo` or `Mesh::GetEdgeToUniqueKnotvector`. In a major refactoring, we could try to move these functions out of `Mesh`, but I see this PR as just correcting NURBS functionality already in the `Mesh` class. If there is an easy to move some of this code into `NURBSExtension`, that may make sense in this PR, but I'm not sure it can be done without large changes.

mesh/mesh.cpp
dylan-copeland: Here is my proposed comment, which @justinlaughlin can add if he agrees: `// Find elements sharing a face containing knotvector kv`.

mesh/mesh.cpp
dylan-copeland: I don't think it is available there, just `Constants<Geometry::CUBE>::Edges[12][2]` which lists the edge vertices but does not give a clear way to get the edges in the X, Y, Z reference directions. If these arrays end up being useful in other places, we could move them somewhere for reuse. I actually defined these already as `constexpr char` arrays in `NCNURBSExtension::SetPatchFactors` (which was a starting point logically for this PR), so maybe we should go ahead and reuse them instead of repeating their definition. @justinlaughlin do you think we can make a function for these directional edge indices, maybe static in `NURBSExtension` so that in `Mesh` we can call something like `NURBSext->ReferenceDirectionEdge(dir, index)`?

mesh/mesh.cpp
dylan-copeland: I agree that it appears wrong for this case to be reached, and I think I needed this at some point in the earlier development to make something work. Rerunning my tests shows that this case is not reached. @justinlaughlin it seems we can replace this with `MFEM_VERIFY(thisDim >= 0, "")`.

mesh/mesh.cpp
dylan-copeland: Only in the case `!initKV` should we initialize the orientation for an edge. In the case `initKV`, an orientation has already been chosen, and consistency can be ensured only by using an edge in this patch already set. The `list<int> nextPatches` should have patches ordered so that a patch is not visited until a face-neighbor has already set edges for this `kv`. This should prevent this case from being reached, where no edges are set yet. In theory, I think we should be able to remove this case, or replace it with an assertion that `edgeSet[pe[ref_edge0]]` is set. The problem is that I have some meshes where the same `kv` is used in 2 directions in the same patch, for small patches with a single element where the `kv` has a single knot-span. It would be better to define different `kv` for each direction, but these meshes simply used the same trivial knotvector for all the single-span patches, to keep the mesh file minimal. For example, suppose there are 3 nontrivial knotvectors for 1 patch and many small patches with a single element, in which case the mesh file can simply list 3 + 1 knotvectors instead of many for all the patch edges. This might be considered an abuse of the NURBS mesh file format and data structures, but it avoids many repeated trivial knotvectors. This `return false` from the lambda function makes such meshes work. This is an unfortunate complication, where the code is not ideal to make an imperfect mesh work. Another way to view this is that the logic is general enough to handle such cases. Maybe we can work on a better way to allow for a generic single-span knotvector to be used repeatedly for many patches and different directions. For now, I propose keeping it this way to handle this issue.

mesh/mesh.cpp
dylan-copeland: I will add some documentation. The basic idea is to pick one knotvector on one starting patch, and then visit all the face-neighbor patches in order, setting orientations. Each of these visiting sweeps should set all the orientation for all edges associated with that knotvector. Then another knotvector is chosen, until all edges on all patches have been set. I would not be surprised if some complicated mesh breaks this code and requires some bug fixes or generalization, but I don't see any obvious limitations or cases for infinite loops (termination will happen when all patches have all edges set).

miniapps/meshing/reflector.cpp
dylan-copeland: The new variables are used only in `ReflectNURBSMesh`. There are 2 calls to `ReflectNURBSMesh`, one which uses the last 2 arguments and one which does not. I was not happy with this design but thought it's not too bad for a small miniapp with only 2 calls of this function. A better way to get access to the arrays `hexPerm` and `elOrder` might be to make a struct or class for `ReflectNURBSMesh`, but I thought that would be overkill. I'm fine either way. If you think a struct/class would be better, or have a better idea, I am open to that.

mesh/mesh.cpp
dylan-copeland: However, `sign` is consistent with `GetEdgeToUniqueKnotvector` which has `sign` and `unsign`. What do you think @justinlaughlin? Maybe we can put these conventions in the `Mesh` class or somewhere even more general? We use these conventions throughout MFEM.

miniapps/meshing/reflector.cpp
dylan-copeland: Documented in c0d5cd54c58fdef3c3b30a1b276e89521f2b8c20

mesh/mesh.cpp
dylan-copeland: Documented in c0d5cd54c58fdef3c3b30a1b276e89521f2b8c20

mesh/mesh.cpp
dylan-copeland: Done in c0d5cd54c58fdef3c3b30a1b276e89521f2b8c20

mesh/mesh.cpp
dylan-copeland: Done in c0d5cd54c58fdef3c3b30a1b276e89521f2b8c20

mesh/mesh.cpp
dylan-copeland: Done in c0d5cd54c58fdef3c3b30a1b276e89521f2b8c20

mesh/mesh.cpp
justinlaughlin: I don't think we can keep this info in `NURBSExtension` because this method is called by the patch topo `Mesh`, not the nurbs `Mesh`, so it does not have a `NURBSExtension`

mesh/mesh.cpp
dylan-copeland: Done in 3446841376247a922c17f7574a11130a44ccaad3

mesh/mesh.cpp
justinlaughlin: Changed in [5db0bff](https://github.com/mfem/mfem/pull/5123/commits/5db0bffc2f1694244ee5c21fd9177a674f272c91)

justinlaughlin: I like the idea of consolidating conventions but I'd rather keep it in a separate PR for the sake of getting this in asap. I suspect it is something people may have opinions about.

mesh/mesh.hpp
justinlaughlin: I agree this might make more sense in `NURBSExtension` but I'd like to save that for the refactoring; imo it would be cleaner to move things in one go with a systematic approach. Some methods may make sense in `nurbs.cpp` but should be decoupled from the class.

On a related note - I think there needs to be a clearer distinction between "NURBS Mesh" methods and "Patch Topology Mesh" methods. But that is another thing I'd like to save for the refactor.

justinlaughlin:
> @justinlaughlin I also gave some more generic comments that are more notes-to-self wrt the NURBS rewrite.

Thank you! I appreciate it and agree with a lot

mesh/mesh.cpp
dylan-copeland: This raises a general question for the MFEM NURBS format: should the same knotvector be allowed in different directions? So far, MFEM has allowed this, and it works fine for many meshes. This algorithm associates knotvector indices with directions, which gets more complicated when the same knotvector can be used repeatedly for many edges in different directions. This is why there is logic to allow for the possibility of finding the wrong direction here, for a given knotvector index, in which case no edges in that direction may have been set. By simply returning from the lambda function, the patch will be visited again later when more edges have been set. Maybe we should consider this issue in NURBS refactoring and try to find the best policies for handling simple/trivial knotvectors that are used in multiple directions.

mesh/mesh.cpp
IdoAkkerman: Perhaps I am overly zealous, but is there a performance impact?
I guess it is minimal in the case when resizing is not necessary.

I am fine with this, just double checking.

mesh/mesh.hpp
IdoAkkerman: Okay, agree this is for a refactor. But nice to know we seem to be on the same page.

tests/unit/CMakeLists.txt
IdoAkkerman: Okay. I only checked the diff :)

mesh/mesh.cpp
IdoAkkerman: Perhaps these should be small universal utilities in fespace, named `flipDOF` and `trueDOf`. 
But that would be for another PR, I guess.

Maybe we need to make a list of these :)

mesh/mesh.cpp
IdoAkkerman: Agree. Mobius strip will then not result in infinite loop (impossible with a finite loop), but the result will not pass the check.
On patch will have opposing edges with incompatible knotvectors.

Perhaps this is not a problematic restriction.
It is also something we could relatively easily fix, by forcing the knotvectors to me symmetric (like a palindrome).
But thus would be for another PR :)

miniapps/meshing/reflector.cpp
IdoAkkerman: I have no issues with this, was just wondering and double checking.

mesh/mesh.cpp
IdoAkkerman: I am okay with the (slightly) more complicated logic to ensure more generic use cases.

But perhaps when reading/initializing we should make a distinction between "the same kv"  and "another kv with the same (initial) specification". But this is for an other PR :)

mesh/mesh.cpp
IdoAkkerman: Not quiet the same as the Table I suppose.

mesh/mesh.cpp
IdoAkkerman: I can also remembering defining and using similar data. I guess including it in geom would be the right thing... 
eventually.

IdoAkkerman (approved):

IdoAkkerman:
@justinlaughlin  & @dylan-copeland I agreed with all the changes made, but did have some additional comments/thoughts, which now might have disappeared do to marking them as resolved.

mesh/mesh.cpp
dylan-copeland: If `Array<Vertex> vertices` is already set, then `Array<T>::SetSize(int nsize)` is a no-op, so I don't think there is any performance impact except when `vertices` is not set, in which case the cost of this function is necessary.

dylan-copeland:
Thank you @IdoAkkerman for your careful review. There were many important issues raised in this review, which are hard to find in the resolved comments. Here is a list to summarize some issues to revisit later.

1. NURBS and patch topology functions should be moved from the `Mesh` class to NURBS classes, as much as possible.
2. The sign conventions and flips `-1 - i` are used throughout MFEM, and we have multiple local functions to do the same thing. MFEM should have some universal utilities for these. This could impact the entire MFEM library, so it may be best to make a PR just for this.
3. Reference quad and hex element edge info should be put in a place accessible by NURBS and Mesh functions (currently we hard-code arrays for this info inside multiple functions). This includes reference vertices and directions of reference edges.
4. We should decide on a general policy for handling knotvectors used repeatedly in different directions or locations.
5. We may need to generalize the knotvector orientation algorithm for cases like the Mobius strip.

tzanio:

PR #4326, opened 2024-05-23
Nodal interpolation for NURBS

IdoAkkerman:
Changed PR header (27 June 2024)
This PR adds a Projector to the NURBS-based elements.
Furthermore, initialization of the a NURBS-based` FESpace` made to look similar to the initialization of the `FESpace` for other FEs.
By performing the selection of the `FE_Collection` in  small utility functions, we can run NURBS-based discretisations in for some of the examples with minimal changes.
Some small changes were made to some NURBS classes, `KnotVector`  most notably. Some examples where slightly modified to accommodate this.

Notes on the  `Project` member function of the different NURBS-based FE:
The `coefficient` is evaluated at the Bottella points,  this value is directly injected in the solution vector for the shape function associated with the Botella point. 
The Botella point is the point where the shapefunction attains its maximum.
NOTE: this is not a true projector in the mathematical sense, applying the 
operation a second time might change the solution.

ORIGINAL MESSAGE (23 May 2024)
By inheriting  from `NodalFiniteElement` instead of `ScalarFiniteElement` (the base class of `NodalFiniteElement` ), we get a free _nodal_ interpolator.

This still  needs some fundamental and practical checking.

Please note:

That the terminology of _projection_ is not correct in this context instead it is a interpolation(I think??).
The idempotent property is not satisfied, a "projection of a projection" should be the same as the original projection.

I do not expect this to be a fundamental problem, I guess  an interpolation would also work in most (all?) situations, or am I wrong on this?

Changing **all** the projection calls in interpolation calls would be way to excessive (but technically correct IMO).

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#4326](https://github.com/mfem/mfem/pull/4326) | @IdoAkkerman | @tzanio | @v-dobrev + @justinlaughlin + @tzanio + @dylan-copeland | 7/14/24 | 1/21/26 | 1/21/26 | |

<details>
<summary>PR Checklist</summary>
    
- [x] Code builds.
- [x] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [x] Update `INSTALL`:
    - [x] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
[...]

PR #5143, opened 2025-12-05
Add parallel H(curl) NURBS support for IGA electromagnetic problem and nurbs_ex25p

kaipengliuHIT:
**Note: This PR continues the work from #4441, which was closed due to branch force-push. Please refer to #4441 for previous discussion history.**

---

## Summary

This PR adds support for parallel H(curl) NURBS finite element spaces for isogeometric analysis (IGA) electromagnetic optimization applications, and add AMS Preconditioner Support for NURBS H(curl) Elements.

## Changes

- **Added VNURBSext support** in `ParFiniteElementSpace` to handle vector NURBS extensions for H(curl) spaces
- **Extended ParNURBSExtension constructor** to support vector NURBS spaces with proper parallel partitioning  
- **Added global element DOF table methods** in `NURBSExtension` for improved DOF management
- **Added nurbs_ex25p example** demonstrating parallel H(curl) NURBS usage

## Modified Files

7 files changed across 2 commits:

- [fem/fespace.cpp](file:///home/kaipeng/code/mfem/fem/fespace.cpp), [fem/fespace.hpp](file:///home/kaipeng/code/mfem/fem/fespace.hpp) - Core finite element space modifications
- [fem/pfespace.cpp](file:///home/kaipeng/code/mfem/fem/pfespace.cpp) - Parallel finite element space with VNURBSext support
- [mesh/nurbs.cpp](file:///home/kaipeng/code/mfem/mesh/nurbs.cpp), [mesh/nurbs.hpp](file:///home/kaipeng/code/mfem/mesh/nurbs.hpp) - NURBS extension enhancements  
- [miniapps/nurbs/makefile](file:///home/kaipeng/code/mfem/miniapps/nurbs/makefile) - Build configuration
- [miniapps/nurbs/nurbs_ex25p.cpp](file:///home/kaipeng/code/mfem/miniapps/nurbs/nurbs_ex25p.cpp) - New parallel example (752 lines)

## Related Issues

Continues #4441

| [#5143](https://github.com/mfem/mfem/pull/5143) | @kaipengliuHIT | @tzanio | @mlstowell + @dylan-copeland + @psocratis | 12/16/25 | ⌛due 12/30/25 | ⌛due 1/6/26 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


kaipengliuHIT:
This is a follow-up to #4441 . The calculation results of the case can refer to those presented in #4441.
ex_nurbs_25p

- Added VNURBSext support for parallel H(curl) NURBS spaces
- Implemented GetGlobalElementDofTable methods in NURBSExtension
- Added new ParNURBSExtension constructor for VNURBSext arrays
- Included nurbs_ex25p example demonstrating the new functionality
- All tests pass, code style checked

Ready for review.

IdoAkkerman:
In PR #5098 I tried implementing the same feature. I followed a slightly different route to solve the issue.

kaipengliuHIT:
> In PR #5098 I tried implementing the same feature. I followed a slightly different route to solve the issue.

This PR is a resubmission of #4441  with a refactored code style. I believe my modifications are more concise and effective, offering greater scalability and compatibility in the future. They only introduce a minimal amount of code to the original base without altering any major logical structures. Moreover, it will remain compatible with parallelization on the GPU in the future.

kaipengliuHIT:
Dear MFEM team, this PR is ready for review.@tzanio : )

tzanio:

miniapps/nurbs/makefile
```diff
@@ -22,7 +22,7 @@ MFEM_LIB_FILE = mfem_is_not_built
 
 SEQ_MINIAPPS = nurbs_ex1 nurbs_patch_ex1 nurbs_ex3 nurbs_ex5 nurbs_ex24 \
    nurbs_curveint nurbs_printfunc nurbs_solenoidal nurbs_naca_cmesh nurbs_surface
-PAR_MINIAPPS = nurbs_ex1p nurbs_ex11p
+PAR_MINIAPPS = nurbs_ex1p nurbs_ex11p nurbs_ex25p
```
dylan-copeland: In this directory, `CMakeLists.txt` should also be updated with `nurbex_ex25p`.

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
+//                                MFEM Example nurbs_25p -- modified for NURBS FE
```
dylan-copeland: `nurbs_ex25p`

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+//
+// Compile with: make nurbs_ex25p
+//
+// Sample runs:  mpirun -np 4 nurbs_ex25p -m ../../data/square-nurbs.mesh
+//               mpirun -np 4 nurbs_ex25p -m ../../data/square-nurbs.mesh -o 2
+//               mpirun -np 4 nurbs_ex25p -m ../../data/cube-nurbs.mesh
+
+// Description:  This example code solves a simple electromagnetic wave parallel IGA(NURBS)
```
dylan-copeland: MFEM has in informal (not enforced by CI) maximum of 80 columns per line.

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+   // 12. Recover the solution as a finite element grid function.
+   a.RecoverFEMSolution(X, b, x);
+
+   if (myid == 0)
+   {
+      cout << "\n Iterations: " << gmres.GetNumIterations() << "\n";
+      cout << " Converged: " << gmres.GetConverged() << "\n";
+      if (use_ams)
+      {
+         cout << "\nAMS preconditioner successfully used for NURBS HCurl space!" << endl;
```
dylan-copeland: Long line.

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+      }
+      else
+      {
+         cout << "\nBoomerAMG preconditioner successfully used!" << endl;
+      }
+      cout << string(80, '-') << endl;
+   }
+
+   // 13. Save the refined mesh and the solution in parallel. This output can be
+   //     viewed later using GLVis: "glvis -np <np> -m mesh -g sol".
```
dylan-copeland: The ParaView visualization worked for me, but not glvis in parallel. It would help to simplify the file names, as in `examples/ex25p.cpp`. Aside from the naming of the files, glvis fails when loading the parallel files with `-np`, in `NURBSExtension::LoadFE()`, although it works for the files associated with a single rank.

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+int main(int argc, char *argv[])
+{
+   // 1. Initialize MPI and HYPRE.
+   Mpi::Init(argc, argv);
+   int myid = Mpi::WorldRank();
+   Hypre::Init();
+   // Parse command-line options.
+   const char *mesh_file = "../../data/square-nurbs.mesh";
+   int ref_levels = -1;
+   bool NURBS = true;
```
dylan-copeland: These are not used and can be removed: `NURBS`, `static_cond`. It is possible to remove `pa` and `device_config` as well, but it may make sense to keep these for possible future use. Also, `ref_levels` is not used but should be, instead of the hard-coded 5. The value 5 could be the initial value used by default.

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+void detJ_inv_JT_J_abs(const Vector &x, PML * pml, Vector & D);
+
+Array2D<real_t> comp_domain_bdr;
+Array2D<real_t> domain_bdr;
+
+real_t mu = 1.0;
+real_t epsilon = 1.0;
+real_t omega;
+int dim;
+bool exact_known = false;
```
dylan-copeland: Unused.

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+   {
+      args.PrintUsage(cout);
+      return 1;
+   }
+   if (myid == 0 )
+   {
+      args.PrintOptions(cout);
+   }
+
+
```
dylan-copeland: Extra empty line.

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+      mesh->UniformRefinement();
+   }
+
+   ParMesh *pmesh = new ParMesh(MPI_COMM_WORLD, *mesh);
+   delete mesh;
+   pml->SetAttributes(pmesh);
+
+
+   // 5. Define a finite element space on the mesh. Here we use the
+   //    Raviart-Thomas finite elements of the specified order.
```
dylan-copeland: Was `Raviart-Thomas` mistakenly copied from another example?

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+   FiniteElementCollection *fec = nullptr;
+   NURBSExtension *NURBSext = nullptr;
+
+   fec = new NURBS_HCurlFECollection(order,dim);
+   NURBSext  = new NURBSExtension(pmesh->NURBSext, order);
+   mfem::out<<" Create NURBS fec and ext"<<std::endl;
```
dylan-copeland: Instead of `fec and ext`, it would be more readable to say something like `finite element space`. This output is also unnecessary and could be removed.

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+   ConstantCoefficient omeg(-pow2(omega) * epsilon);
+   RestrictedCoefficient restr_muinv(muinv,attr);
+   RestrictedCoefficient restr_omeg(omeg,attr);
+
+   // Integrators inside the computational domain (excluding the PML region)
+   ParSesquilinearForm a(fespace, conv);
+   a.AddDomainIntegrator(new CurlCurlIntegrator(restr_muinv),NULL);
+   a.AddDomainIntegrator(new VectorFEMassIntegrator(restr_omeg),NULL);
+
+   int cdim = (dim == 2) ? 1 : dim;
```
dylan-copeland: `const int cdim`

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+   OperatorPtr A;
+   Vector B, X;
+   a.FormLinearSystem(ess_tdof_list, x, b, A, X, B);
+
+   cout << "Size of linear system: " << A->Height() << endl;
+
+   ConstantCoefficient absomeg(pow2(omega) * epsilon);
+   RestrictedCoefficient restr_absomeg(absomeg,attr);
+
+   ParBilinearForm prec(fespace);
```
dylan-copeland, line 329: It would be helpful to add some comments about the preconditioner assembled here.

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,804 @@
...
+   //     preconditioner. Choose between AMS (for HCurl) or AMG (general purpose).
+   Array<int> offsets(3);
+   offsets[0] = 0;
+   offsets[1] = fespace->GetTrueVSize();
+   offsets[2] = fespace->GetTrueVSize();
+   offsets.PartialSum();
+
+   std::unique_ptr<Operator> pc_r;
+   std::unique_ptr<Operator> pc_i;
+   int s = (conv == ComplexOperator::HERMITIAN) ? -1 : 1;
```
dylan-copeland: const

mesh/nurbs.hpp
```diff
@@ -1027,6 +1027,13 @@ class NURBSExtension
 
    /// Print control points for coarse patches.
    virtual void PrintCoarsePatches(std::ostream &os);
+
+protected:
+   /// Construct and return a table of DOFs for each global element.
+   Table *GetGlobalElementDofTable();
```
dylan-copeland: These functions are already defined in `ParNURBSExtension`. Should they be `virtual`? Or should these be in `ParNURBSExtension` with different names? How should these functions differ from the existing ones?

fem/fe/fe_nurbs.cpp
```diff
@@ -817,6 +817,99 @@ void NURBS_HCurl2DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
    }
 }
 
+// ProjectGrad implementation for NURBS_HCurl2DFiniteElement
+void NURBS_HCurl2DFiniteElement::ProjectGrad(const FiniteElement &fe,
```
dylan-copeland, line 1193: It would be nice to add some small unit tests `mfem/tests/unit/fem` for the 2D and 3D `ProjectGrad` functions. There are some tests already for `GradientInterpolator`, so maybe it would be easy to add another case for NURBS? Also for `Project`.

fem/fe/fe_nurbs.cpp
```diff
@@ -817,6 +817,99 @@ void NURBS_HCurl2DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
    }
 }
 
+// ProjectGrad implementation for NURBS_HCurl2DFiniteElement
```
dylan-copeland: This comment is not informative.

fem/fe/fe_nurbs.cpp
```diff
@@ -817,6 +817,99 @@ void NURBS_HCurl2DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
...
+void NURBS_HCurl2DFiniteElement::Project(const FiniteElement &fe,
+                                         ElementTransformation &Trans,
+                                         DenseMatrix &I) const
+{
+   if (fe.GetRangeType() == SCALAR)
+   {
+      int sdim = Trans.GetSpaceDim();
+      real_t vk[Geometry::MaxDim];
+      Vector shape(fe.GetDof());
+      real_t tk_2d[4] = {1.0, 0.0, 0.0, 1.0};  // Unit tangents in x and y
```
dylan-copeland: `constexpr real_t`

fem/fe/fe_nurbs.cpp
```diff
@@ -1008,6 +1101,110 @@ void NURBS_HCurl3DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
...
+void NURBS_HCurl3DFiniteElement::Project(const FiniteElement &fe,
+                                         ElementTransformation &Trans,
+                                         DenseMatrix &I) const
+{
+   if (fe.GetRangeType() == SCALAR)
+   {
+      int sdim = Trans.GetSpaceDim();
+      real_t vk[Geometry::MaxDim];
+      Vector shape(fe.GetDof());
+      real_t tk_3d[9] = {1.0, 0.0, 0.0,  0.0, 1.0, 0.0,  0.0, 0.0, 1.0};  // Unit tangents
```
dylan-copeland: `constexpr real_t`

kaipengliuHIT:
Thank you for the feedback! I've fixed the glvis parallel loading issue.

Changes made:
1. Simplified output file naming in nurbs_ex25p.cpp to match ex25p.cpp
2. Initialized sFE/qFE/hFE pointers to nullptr in NURBS_HCurlFECollection and NURBS_HDivFECollection constructors
3. Added null pointer checks in NURBSExtension::LoadFE() and LoadBE()
4. Fixed parallel loading by calling SetDim() in GridFunction merge constructor

Root cause:
The GridFunction merge constructor (used during parallel loading) was creating a new FiniteElementCollection but not calling SetDim(), leaving sFE/qFE/hFE pointers uninitialized for NURBS H(curl)/H(div) collections.

Testing:
Serial loading: `glvis -m mesh.000000 -g ex25p-sol_r.000000`
Parallel loading: `glvis -np 2 -m mesh -g ex25p-sol_r`

I'm now working on addressing the other feedback items. Will update soon.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: Thank you for the feedback! I've fixed the glvis parallel loading issue.

Changes made:

Simplified output file naming in nurbs_ex25p.cpp to match ex25p.cpp
Initialized sFE/qFE/hFE pointers to nullptr in NURBS_HCurlFECollection and NURBS_HDivFECollection constructors
Added null pointer checks in NURBSExtension::LoadFE() and LoadBE()
Fixed parallel loading by calling SetDim() in GridFunction merge constructor
Root cause:
The GridFunction merge constructor (used during parallel loading) was creating a new FiniteElementCollection but not calling SetDim(), leaving sFE/qFE/hFE pointers uninitialized for NURBS H(curl)/H(div) collections.

Testing:
Serial loading: glvis -m mesh.000000 -g ex25p-sol_r.000000
Parallel loading: glvis -np 2 -m mesh -g ex25p-sol_r

I'm now working on addressing the other feedback items. Will update soon.

mesh/nurbs.hpp
kaipengliuHIT: I've made the GetGlobalElementDofTable methods virtual in the NURBSExtension base class and added override in ParNURBSExtension. This is necessary because these methods are called through NURBSExtension* pointers in VNURBSExt arrays (nurbs.cpp:5859-5872), so we need proper polymorphic behavior. Tested and verified - nurbs_ex25p runs correctly with the changes.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: I have revised.

miniapps/nurbs/nurbs_ex25p.cpp
dylan-copeland: Are these changes committed? In the current version, the MPI rank is not in the mesh filename, so glvis only shows the subdomain for rank 0.

dylan-copeland: Also, the glvis output should be done `if (visualization)`.

mesh/nurbs.hpp
dylan-copeland: The code for these functions is identical, for the base and derived class. That should not be necessary. The functions can remain only in the base class.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: Actually, the submitted code doesn't use the method described above to enable glvis. I will write a new reply later.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Thank you for the suggestion! I have added unit tests for the NURBS ProjectGrad and Project functions.Specifically, I created (mfem/tests/unit/fem/test_nurbs_projectgrad.cpp) which includes the following test cases :
1) NURBS ProjectGrad 2D/3D: Tests for ProjectGrad functionality for 2D and 3D NURBS_HCurl finite elements
2) NURBS ProjectGrad Detailed 2D/3D: More detailed verification tests for the ProjectGrad functionality
3) NURBS Project 2D/3D: Tests for Project functionality for 2D and 3D NURBS_HCurl finite elements
These tests cover the ProjectGrad and Project methods in NURBS_HCurl2DFiniteElement and NURBS_HCurl3DFiniteElement classes, using data/square-nurbs.mesh and data/cube-nurbs.mesh as test mesh files. The tests verify matrix dimension correctness and basic functionality implementation.All tests have passed and the CI checks have successfully passed as well. These tests follow the existing test framework patterns and maintain consistent style and quality standards with the existing GradientInterpolator tests.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: I have modified the code as you suggested.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: I have modified the code as you suggested.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: I removed this unnecessary comment.

mesh/nurbs.hpp
kaipengliuHIT: I removed the duplicated code in the derived classes.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: I have modified the code as you suggested.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: Yes， I have added the comments.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: I have modified the code as you suggested.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: It has been removed.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: Yes, it has been removed

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: It has been removed

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: It has been removed

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: I have modified the code as you suggested.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: It is now possible to display results in parallel using GLVis.

miniapps/nurbs/makefile
kaipengliuHIT: It has been updated

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: It has been revised.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: It has been revised.

kaipengliuHIT:
The modifications have been completed based on your suggestions. @dylan-copeland

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,871 @@
+//                                MFEM Example nurbs_ex25p -- modified for NURBS FE
+//
+// Compile with: make nurbs_ex25p
+//
+// Sample runs:  mpirun -np 4 nurbs_ex25p -m ../../data/square-nurbs.mesh
```
dylan-copeland, line 5: The first sample run used to converge in less than 200 iterations. Now it takes 2502 iterations. Is this because the default value of `ref_levels` was changed to 8? It is faster for coarser meshes, and it would be better to have default values to make the sample runs fast for regression testing.

miniapps/nurbs/nurbs_ex25p.cpp
dylan-copeland: Thank you for fixing the glvis output. A minor suggestion, for consistency with other MFEM examples, is to change `refined.mesh.` to `refined.`, `sol_r.gf.` to `sol_r.`, etc.

fem/fe/fe_nurbs.cpp
dylan-copeland: Thanks! This line is long and can be split into 2 lines.

tests/unit/fem/test_nurbs_projectgrad.cpp
```diff
@@ -0,0 +1,418 @@
+// Copyright (c) 2010-2025, Lawrence Livermore National Security, LLC. Produced
```
dylan-copeland, line 1: 2026. Note that we update the copyright year each January, but a new file might cause complications in the merge and be missed.

fem/fe/fe_nurbs.cpp
dylan-copeland: Thank you for adding these tests! It looks like the tests just verify that the code runs. Can they be made stronger by comparing the values of projections? For example, in `test_lin_interp.cpp`, the tests apply an operator to a function for which the exact result is known, and a comparison is made to ensure two GridFunctions match to within a small tolerance. For `ProjectGrad`, the projection can be compared to the exact gradient. For `NURBSProject3D`, maybe compare the projected function to the exact coefficient at some quadrature points?

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,871 @@
...
+   // 6. Determine the list of true (i.e. conforming) essential boundary dofs.
+   //    In this example, the boundary conditions are defined by marking all
+   //    the boundary attributes from the mesh as essential (Dirichlet) and
+   //    converting them to a list of true dofs.
+   Array<int> ess_tdof_list;
+
+   Array<int> ess_bdr(pmesh->bdr_attributes.Max());
+   ess_bdr = 1;
+   fespace->GetEssentialTrueDofs(ess_bdr, ess_tdof_list);
+   cout << "Number of knowns in essential BCs: "
```
dylan-copeland: The rank should be included in this output, or it can be output only on rank 0 like `ex25p`. Also for `Size of linear system` below, or the system size can be removed from output since it is redundant after printing `Number of finite element unknowns`.

miniapps/nurbs/nurbs_ex25p.cpp
```diff
@@ -0,0 +1,871 @@
...
+   }
+   if (myid == 0 )
+   {
+      args.PrintOptions(cout);
+   }
+
+   // Enable hardware devices such as GPUs, and programming models such as
+   // CUDA, OCCA, RAJA and OpenMP based on command line options.
+   Device device(device_config);
+   device.Print();
```
dylan-copeland: `if (myid == 0) { device.Print(); }`

mesh/nurbs.hpp
dylan-copeland: Thanks for simplifying this. At this point, the `virtual` keyword can be removed, but a more important issue is whether these global tables can be avoided entirely. In 2D, we already have `Generate2DElementDofTable`, which sets the table for all active elements (all local elements in the parallel partition of the mesh). This PR introduces `Get2DGlobalElementDofTable`, which is essentially the same function but for all global elements. One concern is the scalability of using all global elements on each MPI rank. The global table is constructed by `GetGlobalElementDofTable` only in a `ParNURBSExtension` constructor. Are all the elements necessary, or would the active elements be sufficient, so that we could just call `GenerateElementDofTable`?

fem/fe/fe_nurbs.cpp
IdoAkkerman: I agree with @dylan-copeland I have some doubt about the projectors, which I'll try to explain in another post.
Most of these doubts could be removed by having a unit test that tests a meaning mathematical property, for example L2 or H1 convergence. See for instance PR #4326 where the lack of H1 convergence of the projector created issues.

I wonder if in the current use of the projectors certain issues could be camouflaged by the krylov solver?

mesh/nurbs.hpp
IdoAkkerman: @dylan-copeland There is indeed some almost doubling up of code going on. Which at this point is unfortunately necessary. Ideally we have  `Generate2DElementDofTable` defined in `NURBSExtension` and one overloaded in `ParNURBSExtension`, this way the global Table could be obtained by calling the routine for the `NURBSExtension::Generate2DElementDofTable` this would require a rewrite of the NURBS classes (see #5105 )

fem/fe/fe_nurbs.cpp
```diff
@@ -817,6 +817,97 @@ void NURBS_HCurl2DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
...
+
+   // Get the integration points (nodes) of this HCurl element
+   const IntegrationRule &ir = Nodes;
+   DenseMatrix dshape(fe.GetDof(), 2);
+   Vector grad_k(fe.GetDof());
+
+   grad.SetSize(dof, fe.GetDof());
+
+   // For each DOF in the HCurl space
+   for (int k = 0; k < dof; k++)
```
IdoAkkerman: Instead of looping over all the dofs in one go, and making a selection on what to do based on relatively complicated logic (r844 and if statement on r846), I think it would be better to have two explicit loops over each of the cases. This would make this routine inline with the other `NURBS_HCurl2DFiniteElement` routines.

fem/fe/fe_nurbs.cpp
```diff
@@ -817,6 +817,97 @@ void NURBS_HCurl2DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
...
 
+void NURBS_HCurl2DFiniteElement::ProjectGrad(const FiniteElement &fe,
+                                             ElementTransformation &Trans,
+                                             DenseMatrix &grad) const
+{
+   MFEM_ASSERT(fe.GetMapType() == VALUE, "ProjectGrad requires H1 element");
+   MFEM_ASSERT(fe.GetDim() == 2, "Dimension mismatch in ProjectGrad");
+
+   // Get the integration points (nodes) of this HCurl element
+   const IntegrationRule &ir = Nodes;
```
IdoAkkerman: Are these `Nodes`  defined in a meaningful way? Or are they all trivial (`=(0,0)`)?

fem/fe/fe_nurbs.cpp
```diff
@@ -817,6 +817,97 @@ void NURBS_HCurl2DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
...
+   DenseMatrix dshape(fe.GetDof(), 2);
+   Vector grad_k(fe.GetDof());
+
+   grad.SetSize(dof, fe.GetDof());
+
+   // For each DOF in the HCurl space
+   for (int k = 0; k < dof; k++)
+   {
+      const IntegrationPoint &ip = ir.IntPoint(k);
```
IdoAkkerman: The first portion of the `Nodes` will be used for the dx component, while the second will be used for the dy component.  This looks wrong, or otherwise unnatural. Please explain and  add some comments to the code.

fem/fe/fe_nurbs.cpp
```diff
@@ -817,6 +817,97 @@ void NURBS_HCurl2DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
...
+void NURBS_HCurl2DFiniteElement::Project(const FiniteElement &fe,
+                                         ElementTransformation &Trans,
+                                         DenseMatrix &I) const
+{
+   if (fe.GetRangeType() == SCALAR)
+   {
+      int sdim = Trans.GetSpaceDim();
+      real_t vk[Geometry::MaxDim];
+      Vector shape(fe.GetDof());
+      constexpr real_t tk_2d[4] = {1.0, 0.0, 0.0, 1.0};  // Unit tangents in x and y
```
IdoAkkerman: This looks like a convoluted way to define 2 normals. Spliting the dof loop in two explicit loops will allow you to avoid this.

fem/fe/fe_nurbs.cpp
```diff
@@ -817,6 +817,97 @@ void NURBS_HCurl2DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
...
+
+      I.SetSize(dof, sdim*fe.GetDof());
+      for (int k = 0; k < dof; k++)
```
IdoAkkerman: Please split loop. Logic on l886 and l895 makes it quit unclear.

fem/fe/fe_nurbs.cpp
```diff
@@ -817,6 +817,97 @@ void NURBS_HCurl2DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
...
+            grad(k, j) = dshape(j, 1);  // dy component
+         }
+      }
+   }
+}
+
+void NURBS_HCurl2DFiniteElement::Project(const FiniteElement &fe,
+                                         ElementTransformation &Trans,
+                                         DenseMatrix &I) const
+{
```
IdoAkkerman, line 1239: Please explain the assumptions/theory underlying this routine, ideally with a meaningful reference to a journal paper. I can imagine that this paper does not exist, in that case I think it would be worth writing it, however, personally I think the approach might have some issues...

Please see   PR #4326 for discussions  on "Nodal projections" of standard NURBS elements.
Likely there are analogies to be seen. In this PR we  ended settling on 3 projection options.
- Global L2 projection
- Local L2 projection -- based on the Bezier projection paper, with favorable and provable convergence properties
- Nodal _projection_ -- based on my intuition, which is formally not a projection, seems to have issues  H1 convergence, but results in monotone solutions

This routine seems closely related to the problematic NodalProjection.

fem/fe/fe_nurbs.cpp
```diff
@@ -1008,6 +1099,108 @@ void NURBS_HCurl3DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
...
+   MFEM_ASSERT(fe.GetDim() == 3, "Dimension mismatch in ProjectGrad");
+
+   // Get the integration points (nodes) of this HCurl element
+   const IntegrationRule &ir = Nodes;
+   DenseMatrix dshape(fe.GetDof(), 3);
+
+   grad.SetSize(dof, fe.GetDof());
+
+   // For each DOF in the HCurl space
+   for (int k = 0; k < dof; k++)
```
IdoAkkerman: Split loop in 3 distinct loops.

fem/fe/fe_nurbs.cpp
```diff
@@ -1008,6 +1099,108 @@ void NURBS_HCurl3DFiniteElement::CalcCurlShape(const IntegrationPoint &ip,
...
+{
+   if (fe.GetRangeType() == SCALAR)
+   {
+      int sdim = Trans.GetSpaceDim();
+      real_t vk[Geometry::MaxDim];
+      Vector shape(fe.GetDof());
+      constexpr real_t tk_3d[9] = {1.0, 0.0, 0.0,  0.0, 1.0, 0.0,  0.0, 0.0, 1.0};  // Unit tangents
+
+      I.SetSize(dof, sdim*fe.GetDof());
+      for (int k = 0; k < dof; k++)
```
IdoAkkerman: Idem.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Thank you for your  @IdoAkkerman input. I will implement these tests as you suggested, as it is indeed very necessary, something I hadn't fully realized before. Also, this works well with Krylov—I find that AMS is much faster than AMG. I will conduct comprehensive tests later and share the results with you for discussion.

mesh/nurbs.hpp
```diff
@@ -1073,6 +1076,13 @@ class ParNURBSExtension : public NURBSExtension
        The @a parent can be either a local NURBSExtension or a global one. */
    ParNURBSExtension(NURBSExtension *parent,
                      const ParNURBSExtension *par_parent);
+
+   /** @brief Create a parallel version of @a parent with partitioning as in
+       @a par_parent based on @a VNURBSExt in Hcurl space; the @a parent object is destroyed.
+       The @a parent can be either a local NURBSExtension or a global one. */
+   ParNURBSExtension(NURBSExtension *parent,
+                     Array<NURBSExtension *> VNURBSExt,
+                     const ParNURBSExtension *par_parent);
```
IdoAkkerman, line 1200: In my opinion this constructor has a strange signature and strange behavior.
I can understand that it seems necessary but this is a consequence of how the `NURBSExtensions` are handled in `ParFiniteElementSpace::ParInit`.
I feel this choice will paint us in a corner, that will become problematic when we will consider the generic multipatch case.

mesh/nurbs.cpp
```diff
@@ -5638,146 +5790,187 @@ ParNURBSExtension::ParNURBSExtension(NURBSExtension *parent,
...
+   Swap(bel_to_IJK, parent->bel_to_IJK);
+
+   Swap(weights, parent->weights);
+   MFEM_VERIFY(!parent->HavePatches(), "");
+
+   delete parent;
+
+   MFEM_VERIFY(par_parent->partitioning,
+               "parent ParNURBSExtension has no partitioning!");
+
```
IdoAkkerman, line 6253: The code below this point looks out of place. Separation of responsibility seems to be violated here, orchestration that, in my opinion, should be done in `ParFiniteElementSpace` is performed here.

IdoAkkerman (changes requested):

IdoAkkerman:
> I believe my modifications are more concise and effective, offering greater scalability and compatibility in the future.

Can you corroborate?

I think the handling of the `NURBSExtensions` in `ParFiniteElementSpace::ParInit`, and the resulting `ParNURBSExtension` will paint us in a corner when trying to extend to the multipatch case.

kaipengliuHIT:
> > I believe my modifications are more concise and effective, offering greater scalability and compatibility in the future.
> 
> Can you corroborate?
> 
> I think the handling of the `NURBSExtensions` in `ParFiniteElementSpace::ParInit`, and the resulting `ParNURBSExtension` will paint us in a corner when trying to extend to the multipatch case.

Actually, in my local code, I have already implemented parallel multipatch for the Hcurl space. I originally planned to submit it after this PR is merged, as it requires adding a considerable amount of code involving projections between two different NURBS patches. If needed, I can upload it together with this PR, but considering the current PR already contains a lot of content, I intend to submit a new PR after this one is completed.

mesh/nurbs.hpp
kaipengliuHIT: Based on this code, I have implemented computations for multiple NURBS patch cases, including parallel computation. The convergence of the current approach can be easily proven, and the present construction does not affect the condition number of the matrix. I originally planned to submit it after this PR is merged, as it requires adding a considerable amount of code involving projections between two different NURBS patches. If needed, I can upload it together with this PR, but considering the current PR already contains a lot of content, I intend to submit a new PR after this one is completed.

IdoAkkerman: Could you please point me to the branch/repo where I could find that code, I would be very interested to see it.

Does it work for "simple" multipatch meshes? Or does it work for generic multipatch cases?
My PhD student did quite some work on getting a C-mesh to work for standard NURBS, complication was the direction of edges and the knotVector associated with them. These direction issues are dealt with in a haphazard way in different location in the code. This is some technical debt that needs (does it?) to be addressed at some point.

IdoAkkerman: I was reluctant to go to multipatch before addressing the technical debt. I could not see a way out using the current approach, I ran out of steam.

Perhaps your work helps me see things in a different light.

kaipengliuHIT: Okay Prof Akkerman, I'll organize it later and create a new branch. Currently, it is possible to handle the multi-patch case with identical NURBS boundaries, meaning that in the mesh, each adjacent pair of NURBS patches share a common NURBS boundary.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: Yes, the large iteration count was caused by the overly large default refinement level. I changed the default `ref_levels` from 8 to 3 to keep the sample run reasonably small for regression testing, while still allowing finer runs through `-r/--refine`. The preconditioner work for H(curl) IGA electromagnetic problems has already been completed on my side. It is more suitable than the current AMS option for the IGA electromagnetic problems targeted here. After this PR is finished, I plan to submit it as a separate follow-up PR, since it is a separate solver/preconditioner topic from the NURBS H(curl) support added in this PR.

tests/unit/fem/test_nurbs_projectgrad.cpp
kaipengliuHIT: Thanks for the note. I updated the copyright header in the new test file accordingly, and the copyright CI now passes.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Addressed partially by adding a projector-level convergence test, independent of the Krylov solve.
The new `NURBSProjectGradConvergence2D` test computes the H(curl) field produced directly by `ProjectGrad` and compares it with the exact gradient of a smooth scalar function in the L2 norm. The test then checks monotone error reduction under uniform refinement.
This is intended to avoid relying only on the electromagnetic solve, where solver behavior could potentially hide projector issues. I agree that this is not a complete theoretical proof of the projector, but it provides a direct mathematical regression check on the projector itself.
The observed errors decrease under refinement:
  err0 = 2.2214414680
  err1 = 2.0882125885
  err2 = 1.1397362748
  err3 = 0.5659108383

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Addressed. The H(curl) NURBS `Nodes` are now explicitly initialized, so they are not left as the default trivial points. The nodes are set as tensor-product reference points following the same component-block ordering used by the H(curl) basis. In 2D, the first block corresponds to the x-directed component space and the second block to the y-directed component space. In 3D, the same structure is used for the x-, y-, and z-directed blocks. I also added unit-test checks to verify that the NURBS H(curl) nodes have the expected size, lie in the reference element, and are not all zero.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Addressed. I rewrote this routine using two explicit component-wise loops. The first loop handles the x-directed H(curl) block with orders `(p_x, p_y+1)`, and the second loop handles the y-directed block with orders `(p_x+1, p_y)`. This removes the previous per-dof branch logic and makes the routine consistent with the other `NURBS_HCurl2DFiniteElement` tensor-product loops.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Addressed. I added explicit comments explaining the component-block ordering and initialized the H(curl) NURBS `Nodes` consistently with that ordering. The ordering is intentional: the NURBS H(curl) basis is stored by component blocks. In 2D, the first block is x-directed with orders `(p_x, p_y+1)`, and the second block is y-directed with orders `(p_x+1, p_y)`. The `Nodes` now follow the same block structure, so the first block is used for the dx/x-directed component and the second block for the dy/y-directed component. I also added unit-test checks to verify that these nodes are explicitly initialized and are not left as trivial default points.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Thank you for the detailed explanation. I agree that this should not be presented as a global or local L2 projection with the same theoretical guarantees. The routine is intended as a local interpolation/projection operator consistent with the tensor-product NURBS H(curl) component spaces. The underlying H(curl) space structure follows the isogeometric de Rham construction described in Buffa, Sangalli, and Vázquez, "Isogeometric analysis in electromagnetics: B-splines approximation", CMAME 199 (2010), 1143-1152. In this implementation, the H(curl) basis is organized in component blocks, and the projector evaluates the corresponding scalar shape data at the component nodes and applies the H(curl) covariant mapping through the element Jacobian.

kaipengliuHIT: I agree that this is closer in spirit to a nodal/interpolation-type projection than to a provably stable L2 projection. To reduce the risk of hidden issues, I added projector-level tests that do not go through the Krylov solver: exact-value tests for linear gradients and constant vector fields, plus a small 2D L2 refinement convergence test for `ProjectGrad`. So this PR now documents the assumptions and adds regression/convergence coverage, but I am not claiming a full theoretical proof for this projector. If a stronger projection with provable L2 stability is preferred, I think that would be a separate follow-up design, similar to the projection options discussed in PR #4326.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Addressed. I split the loop into two explicit component-wise loops, so the previous indirect normal/tangent selection logic is no longer needed. The first loop handles the x-directed block and uses the x reference direction directly. The second loop handles the y-directed block and uses the y reference direction directly.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Addressed. I split this into explicit component-wise loops.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Addressed. I split the 3D routine into three explicit component-wise loops.

fem/fe/fe_nurbs.cpp
kaipengliuHIT: Addressed.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: Addressed. The essential-BC output now includes the MPI rank, since that value is rank-local. The linear-system size output is now restricted to rank 0, following the usual example output pattern.

miniapps/nurbs/nurbs_ex25p.cpp
kaipengliuHIT: Addressed. `device.Print()` is now called only on rank 0.

mesh/nurbs.cpp
kaipengliuHIT: I agree with this concern. The current constructor is doing orchestration for the H(curl) component extensions, and that responsibility would fit better in `ParFiniteElementSpace::ParInit` or in a dedicated helper used from there. For this PR, I added comments and guards clarifying that the current path is limited to the single-patch H(curl) case and should not be treated as the final multipatch design. A cleaner design would move the component-extension orchestration and group/table construction out of `ParNURBSExtension`, preferably avoiding global element tables on each rank. I would prefer to keep that refactor as follow-up work, because it touches the parallel NURBS initialization design more broadly than the single-patch H(curl) support added in this PR.

kaipengliuHIT:
> > I believe my modifications are more concise and effective, offering greater scalability and compatibility in the future.
> 
> Can you corroborate?
> 
> I think the handling of the `NURBSExtensions` in `ParFiniteElementSpace::ParInit`, and the resulting `ParNURBSExtension` will paint us in a corner when trying to extend to the multipatch case.

Thank you for pointing this out. I agree that the current handling of the component `NURBSExtension`s in `ParFiniteElementSpace::ParInit` and the resulting `ParNURBSExtension` is not the final design I would want for generic multipatch H(curl) support. 
For this PR, I have clarified the scope in the code: the current H(curl) extension path is guarded to single-patch NURBS meshes, and the component-wise table construction is documented as a single-patch mechanism, not as a complete multipatch continuity/orientation solution.
I do have follow-up multipatch H(curl) work locally, but I agree that the generic multipatch case needs a cleaner design for tangential DOF matching, orientation, and patch-interface handling. I would prefer to keep that as a separate follow-up PR, so this PR remains focused on parallel single-patch H(curl) NURBS support and the nurbs_ex25p miniapp.

kaipengliuHIT:
Thank you all for the detailed review. I have pushed another update addressing the main implementation and testing comments.

Summary of the latest changes:
  - Rewrote the NURBS H(curl) `ProjectGrad` and matrix-form `Project` routines using explicit component-wise loops in 2D and 3D.
  - Added comments explaining the tensor-product H(curl) component-block ordering.
  - Explicitly initialized the NURBS H(curl) `Nodes` so they are not left as trivial default points.
  - Strengthened the unit tests so they check projected values, not only successful execution.
  - Added a small 2D `ProjectGrad` L2 refinement convergence test.
  - Cleaned up the `nurbs_ex25p` miniapp output/default refinement behavior.
  - Merged current `mfem/master` with a regular merge commit, without force-pushing.

  I also want to clarify the scope of this PR. The current implementation is intended for the single-patch parallel H(curl) NURBS case. I agree with the review comments that a generic
  multipatch H(curl) implementation needs a cleaner design for tangential DOF matching, orientation, patch-interface handling, and possibly avoiding global element-dof tables on each rank.
  I have therefore added comments/guards to make the current limitation explicit rather than presenting this as the final multipatch architecture.

  After this PR is finished, I plan to submit two follow-up PRs:
  1. A multipatch H(curl) NURBS version.
  2. A solver/preconditioner update for H(curl) IGA electromagnetic problems. That preconditioner work has already been completed on my side and is more suitable than the current AMS option for the IGA electromagnetic problems targeted here.

I think keeping those as follow-up PRs will make this review more focused and avoid mixing the single-patch H(curl) support, generic multipatch design, and solver/preconditioner changes in one PR.

kaipengliuHIT:
For additional context, the follow-up solver/preconditioner work is currently in my `opt_em_iga` project:
  https://github.com/kaipengliuHIT/opt_em_iga.
  The multipatch H(curl) NURBS work is currently on the `multipatch_hcurl_fix_backup_20260523` branch. After
  this PR is finished, I plan to clean up these two pieces of work and submit them as separate follow-up
  PRs.

PR #5167, opened 2025-12-18
Nurbs variable 1d dev

### LINKED ISSUE #5165: Support for 1D NURBS mesh with varying orders
I tried defining and loading a 1D NURBS mesh file with segments/curves of different orders, and was not successful.

For example, the following mesh w/ a linear, quadratic and cubic element failed to load:
```
MFEM NURBS mesh v1.0

dimension
1

# Three segments with different NURBS orders
elements
3
1 1 0 1
2 1 2 3
3 1 4 5

# POINT=0 endpoints (boundary of 1D mesh)
boundary
6
1 0 0
1 0 1
1 0 2
1 0 3
1 0 4
1 0 5

edges
3
0 0 1
1 2 3
2 4 5

vertices
6

# Edge 0: linear (order 1, 2 control points)
# Edge 1: quadratic (order 2, 3 control points)
# Edge 2: cubic (order 3, 4 control points)
knotvectors
3
1 2  0 0  1 1
2 3  0 0 0  1 1 1
3 4  0 0 0 0  1 1 1 1

# One weight per control point, in the same order as the control points below; (2 + 3 + 4) = 9 weights total
weights
1
1
1
1
1
1
1.2
1.31
1.32

FiniteElementSpace
FiniteElementCollection: NURBS
VDim: 2
Ordering: 1

0.0 0.0
1.0 1.0
1.0 0.0
2.0 1.0
2.0 0.0
3.0 1.0
1.2 1.2
2.3 1.3
3.3 0.3
```

When I run this against mfem@master, I get
```
Assertion failed: (i>=0 && i<size) is false:
 --> Access element 6 of array, size = 6
 ... in function: T &mfem::Array<int>::operator[](int) [T = int]
 ... in file: /opt/compiler-explorer/libs/mfem/v4.7/linalg/../general/array.hpp:756

Program terminated with signal: SIGSEGV
```

Here's the example w/ godbolt + mfem@4.7:  https://godbolt.org/z/erzdojo53

---

Note: the following w/ uniform order works fine:
```
MFEM NURBS mesh v1.0

dimension
1

# attr=1, SEGMENT=1, topo vertices (0,1)  -> order 1
elements
3
1 1 0 1
2 1 2 3
3 1 4 5

# POINT=0 endpoints (boundary of 1D mesh)
boundary
6
1 0 0
1 0 1
1 0 2
1 0 3
1 0 4
1 0 5

edges
3
0 0 1
1 2 3
2 4 5

vertices
6

knotvectors 
3
2 3  0 0 0 1 1 1
2 3  0 0 0 1 1 1
2 3  0 0 0 1 1 1

# one weight per control point, in the same order as the control points below
weights
1
1
1
1
1
1
1.21
1.22
1.23



FiniteElementSpace
FiniteElementCollection: NURBS
VDim: 2
Ordering: 1

0.0 0.0
1.0 1.0
1.0 0.0
2.0 1.0
2.0 0.0
3.0 1.0
1.0 1.21
2.0 1.22
3.0 1.23
```
See: https://godbolt.org/z/E9Mc6edqW

---

I'm working on a branch to fix this.

kennyweiss:
This PR improves MFEM's support for loading and generating meshes composed of NURBS curves of varying orders.
* It fixes a bug in loading nurbs meshes where the curves have varying orders and the control points are defined using a grid function.  (This was also fixed in parallel in #5166 )
* It also adds support for loading nurbs curve meshes using the "patches" construct
* It adds unit tests and data files for loading these meshes and for procedurally generating patches of 1D NURBS curves embedded in 2D
* Resolves #5165 
* I developed a `readNURBS` example to load these meshes and to degree-elevate them so I can visualize the meshes in VisIt. I'm not sure if it belongs in mfem, so I'll post it here. I'm happy to add it to mfem, if desired/appropriate.

<details>
<summary>Expand for readNURBS.cpp example </summary>

```cpp
#include "mfem.hpp"
#include <string>
#include <sstream>
#include <fstream>
#include <iostream>

{
   using namespace mfem;

   if (argc < 2)
   {
      std::cerr << "Usage: " << argv[0]
           << " <mesh.mesh> [--elevate|-e]\n";
      return 1;
   }

   const bool elevate =
      (argc > 2 &&
       (std::string(argv[2]) == "--elevate" || std::string(argv[2]) == "-e"));

   // Decide where the mesh comes from
   std::unique_ptr<std::istream> mesh_stream;

   {
      // Read mesh from file path given as argv[1]
      auto *ifs = new std::ifstream(argv[1]);
      if (!ifs->good())
      {
         std::cerr << "Error: could not open mesh file '" << argv[1] << "'\n";
         return 1;
      }
      mesh_stream.reset(ifs);
      std::cout << "Reading mesh from file: " << argv[1] << "\n";
   }

   // Load the mesh
   Mesh mesh(*mesh_stream, /*generate_edges=*/1, /*refine=*/1);

   std::cout << "\nMesh loaded successfully\n";
   std::cout << "  Dimension      = " << mesh.Dimension() << "\n";
   std::cout << "  SpaceDimension = " << mesh.SpaceDimension() << "\n";
   std::cout << "  NV             = " << mesh.GetNV() << "\n";
   std::cout << "  NE             = " << mesh.GetNE() << "\n";
   std::cout << "  NBE            = " << mesh.GetNBE() << "\n";
   std::cout << "  NEdges         = " << mesh.GetNEdges() << "\n";

   if (auto *nurbs = mesh.NURBSext)
   {
      std::cout << "\nNURBS extension present\n";
      std::cout << "  NURBS Dimension = " << nurbs->Dimension() << "\n";
      std::cout << "  NURBS NV        = " << nurbs->GetNV() << "\n";
      std::cout << "  NURBS NE        = " << nurbs->GetNE() << "\n";

      if (nurbs->Dimension() == 1)
      {
         const int np = nurbs->GetNP();
         const mfem::Vector &weights = nurbs->GetWeights();

         GridFunction *nodes = mesh.GetNodes();
         if (nodes)
         {
            const FiniteElementSpace *fes = nodes->FESpace();
            const int vdim = nodes->VectorDim();

            Array<const KnotVector *> kv;
            Array<int> dofs;

            std::cout << "\nNURBS curves detail:\n";
            for (int p = 0; p < np; p++)
            {
               nurbs->GetPatchKnotVectors(p, kv);
               MFEM_VERIFY(kv.Size() == 1, "Expected 1D NURBS patch.");

               const KnotVector &kv0 = *kv[0];
               const int order = kv0.GetOrder();

               std::cout << "  Curve " << p << ":\n";
               std::cout << "    Order  = " << order << "\n";
               std::cout << "    Degree = " << order - 1 << "\n";

               std::cout << "    Knots  =";
               for (int i = 0; i < kv0.Size(); i++)
               {
                  std::cout << ' ' << kv0[i];
               }
               std::cout << "\n";

               nurbs->GetPatchDofs(p, dofs);
               std::cout << "    Control points (coords, weight):\n";
               for (int i = 0; i < dofs.Size(); i++)
               {
                  const int dof = dofs[i];
                  std::cout << "      CP " << i << ": (";
                  for (int d = 0; d < vdim; d++)
                  {
                     const int vdof = fes->DofToVDof(dof, d);
                     std::cout << (*nodes)(vdof);
                     if (d + 1 < vdim) { std::cout << ", "; }
                  }
                  std::cout << "), w = " << weights(dof) << "\n";
               }
            }
         }
         else
         {
            std::cout << "  (No NURBS nodes GridFunction available; "
                 << "cannot print control points and weights.)\n";
         }

         if (elevate)
         {
            if (!nodes)
            {
               std::cerr << "\nError: cannot elevate NURBS orders because the mesh "
                    << "does not have a nodes GridFunction.\n";
            }
            else
            {
               const Array<int> &orders = nurbs->GetOrders();
               int max_order = 0;
               for (int i = 0; i < orders.Size(); i++)
               {
                  max_order = std::max(max_order, orders[i]);
               }

               // Elevate all knot vectors so that their orders reach max_order.
               mesh.DegreeElevate(max_order, max_order);

               // Derive output name: <input_basename>-elev.mesh
               std::string in_name(argv[1]);
               std::string base = in_name;
               const std::size_t dot = base.find_last_of('.');
               if (dot != std::string::npos)
               {
                  base.erase(dot);
               }
               const std::string out_name = base + "-elev.mesh";

               std::ofstream ofs(out_name);
               ofs.precision(16);
               mesh.Print(ofs);

               std::cout << "\nElevated NURBS mesh written to: " << out_name << "\n";
            }
         }
      }
   }
   else
   {
      std::cout << "\nNo NURBS extension found (unexpected for this mesh)\n";
   }

   return 0;
}
```

</details>

* Here's the `nurbs-segment2d-patch-multispan.mesh` rendered in VisIt using the MutiResControl operator. (I had to degree-elevate the curves to the max order since VisIt's using an earlier version of mfem). The colors (material attributes of each curve) correspond to the order of each curve:
image

| [#5167](https://github.com/mfem/mfem/pull/5167) | @kennyweiss | @tzanio | @IdoAkkerman + @justinlaughlin + @dylan-copeland | 12/24/25 | 1/14/26 | 1/19/26 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,10 @@
 Version 4.9.1 (development)
 ===========================
 
+Meshing improvements
+--------------------
+- Improved support for 1D NURBS meshes with variable order, including using
+  the patches construct for 1D NURBS meshes.
 
 Version 4.9, released on Dec 11, 2025
 =====================================


data/nurbs-segments2d-patches.mesh
kennyweiss, line 1: I added several 1D NURBS meshes (in 2D and 3D, and using "patches" vs. FES) for this PR. 

They're all pretty small, but please let me know if it's too many to add and we can remove some.

IdoAkkerman:
This PR also allows NURBS to be used to define embedded submanifolds.

Thanks! Looks useful!

mesh/nurbs.cpp
```diff
@@ -2369,13 +2371,22 @@ void NURBSExtension::Load(std::istream &input, bool spacing)
...
+      else
+      {
+         NumOfKnotVectors = 0;
+         for (int i = 0; i < patchTopo->GetNEdges(); i++)
+            if (NumOfKnotVectors < KnotInd(i))
+            {
+               NumOfKnotVectors = KnotInd(i);
+            }
+         NumOfKnotVectors++;
+      }
```
IdoAkkerman, line 2381: In PR #5123 @dylan-copeland suggested that (inadvertently?) allow the use case where the same kv is reused. This addition seems to not allow that for 1D.

Could there be use case where you would like to have different line segments have the same knot?

Does the code work without this addition? If it does I would suggest to remove it.
If not? Why not?

mesh/nurbs.cpp
```diff
@@ -5169,21 +5188,32 @@ void NURBSExtension::Set1DSolutionVector(Vector &coords, int vdim)
    Array<const KnotVector *> kv(1);
    NURBSPatchMap p2g(this);
 
+   const bool d2p = dof2patch.Size() > 0;
+
```
IdoAkkerman: These additions bring `SetSolution1D` inline with its 2D and 3D cousins, thanks for that!

However, `dof2patch` does not seem to be used anywhere in the code...
But this originated from a different PR..
@dylan-copeland @justinlaughlin should we get rid of `dof2patch` at some point?

mesh/nurbs.cpp
```diff
@@ -5273,6 +5303,18 @@ void NURBSExtension::GetPatches(Array<NURBSPatch*> &patches_copy)
...
 }
 
+int NURBSExtension::GetPhysicalDim() const
+{
+   if (patches.Size() > 0)
+   {
+      // Patch dimension includes the weight coordinate.
+      return patches[0]->GetNC() - 1;
+   }
+   // Fallback: for non-patch-based extensions, assume the physical dimension
```
IdoAkkerman: Is this the correct assumption? The function name suggest something what it does not do in case there are no patches. If there are no patches defined there is no notion of a PhyscalDim IMHO.

Perhaps it is better to rename the routine `GetPatchDim` and give an error if there are no patches.

mesh/nurbs.hpp
```diff
@@ -755,6 +755,11 @@ class NURBSExtension
    virtual void SetDofToPatch() { };
 
 public:
+   /** @brief Return the physical dimension of the NURBS geometry,
+       inferred from the first patch when available, i.e. number of coordinates
+       per control point minus one (for the weight). */
+   int GetPhysicalDim() const;
```
IdoAkkerman: Please move this definition to be below `GetDimension`.

mesh/nurbs.cpp
dylan-copeland: The map `dof2patch` was introduced in PR 3807, for NC-patch NURBS meshes. See `NCNURBSExtension::SetDofToPatch()`. It is not used for standard NURBS meshes, just when the patch topology mesh is nonconforming and each DOF should be set only by the patch that constrains it. In the 1D case, it is not possible to have a nonconforming patch topology, so it is unnecessary to check for `dof2patch` in this 1D function.

mesh/nurbs.cpp
kennyweiss: Thanks -- I'm looking into it. I think I can work around this.

mesh/nurbs.cpp
kennyweiss: Thanks @dylan-copeland -- I was mostly pattern matching the other dimensions. I'll try to work around this. It might have been a limitation of an earlier version before some other pieces were in place.

mesh/nurbs.cpp
kennyweiss: Makes sense. I'll try it out.

tzanio:

mesh/nurbs.cpp
kennyweiss: Fixed. I also added a new unit test that exercises the shared KV w/ the same and opposite orientations.

mesh/nurbs.cpp
kennyweiss: I removed the `dof2patch`

mesh/nurbs.cpp
kennyweiss: Done.

mesh/nurbs.hpp
kennyweiss: Done.

mesh/nurbs.cpp
```diff
@@ -3310,14 +3330,43 @@ void NURBSExtension::CheckBdrPatches()
...
+   if (dim == 2)
+   {
+      edges[1] = all_edges[1];
+   }
+   else if (dim == 3)
+   {
+      edges[1] = all_edges[3];
+      edges[2] = all_edges[8];
+   }
+}
```
kennyweiss: I added this utility function to refactor repeated logic related to edge indexing

mesh/nurbs.cpp
```diff
@@ -3510,59 +3509,40 @@ void NURBSExtension::UpdateUniqueKV()
 
 bool NURBSExtension::ConsistentKVSets()
 {
-   // patchTopo->GetElementEdges is not yet implemented for 1D
-   MFEM_VERIFY(Dimension() > 1, "1D not yet implemented.");
```
kennyweiss, line 3527: I added 1D functionality to a few `NURBSExtension` utility functions -- `ConsistentKVSets`, `CheckPatches`, `CheckKVDirection`

data/nurbs-segments2d-patches.mesh
justinlaughlin: Alternatively, some/all of these meshes could be added to `miniapps/nurbs/meshes`, which seems to be where we put nurbs meshses used in testing.

mesh/nurbs.cpp
```diff
@@ -3248,7 +3229,46 @@ void NURBSExtension::MergeGridFunctions(
 
 void NURBSExtension::CheckPatches()
 {
-   if (Dimension() == 1 ) { return; }
+   const int dim = Dimension();
+
+   if (dim == 1)
```
justinlaughlin: If we are going to enforce these constraints (do we need to?) we should add it to `CorrectPatchTopoOrientations()`.

See https://github.com/mfem/mfem/pull/5123/files#diff-92f5f01582f8a1d510fdf0f6a41890effe6439a863d8830bc543b9853e53d858R6666

kennyweiss: Thanks @justinlaughlin. I can back out of these changes to `CheckPatches` if it would be better addressed after your NURBS orientation branch is merged.

justinlaughlin: #5123  is now merged 👍.

In the 2d/3d case, this method checks that the knotvectors are consistently oriented per-patch, based on the expected edge orientations of a quad/hex. If a knotvector needs to be flipped, that flip is encoded in the [sign](https://github.com/mfem/mfem/pull/5123/files#diff-d39187387db654f0a8e77684c09ab8360000881460157e785b4c0af470337470R3262-R3270) of `edge_to_ukv`. For 1d, there is no consistency to check "per-patch", so these checks are new.

My understanding is there are 3 checks here
  1) edge size == patch size (line 3239 and 3248)
  2) edge vertex ordering == element vertex ordering + edge index == patch index (line 3256)
  3) KnotSign == expected_sign (line 3266)

The first condition should just be expected from the data imo.

The second condition is new to me - this is the one I was referring to when I was wondering if we need it.

The third condition I'm wondering if we need it as well - if we do, we could add a correction to `CorrectPatchTopoOrientations()`.

Pinging @dylan-copeland to see if he has any opinions here.

justinlaughlin: The only mandatory change from #5123 is to merge and make the return value a `bool`. Imo it is perfectly fine to place responsibility on the mesh-generator rather than our corrections.

I'm being pedantic here because we seem to have multiple definitions/implementations of knotvector orientation that have accumulated over the years and if we're adding new ones (which are defined/enforced by this method) we should do it carefully IMO.

dylan-copeland: As @justinlaughlin described, in the 1D case this function is checking 3 things, which are not for consistency in patches like 2D and 3D. It makes sense to me to have different expectations in 1D, 2D, and 3D. We already have different logic for 2D and 3D, based on the ordering of vertices in a quad and hex. The question is whether to have the new checks (2) and (3), which ensure (2) that the edge and element vertex orderings match and (3) that the knotvectors are oriented from smaller vertex index to larger vertex index. I think it is fine to require these conditions from the mesh generator, since they are trivial to satisfy in 1D. It would be easy to make the corrections in 1D in MFEM if these conditions are not satisfied, but I don't think that should be necessary. In the worst case, if a mesh generator does not satisfy these conditions, a simple script could make the corrections to the mesh file. I think the code is fine as it is.

mesh/nurbs.hpp
```diff
@@ -820,6 +824,13 @@ class NURBSExtension
    /// Return the dimension of the reference space (not physical space).
    int Dimension() const { return patchTopo->Dimension(); }
 
+   /** @brief Return the physical dimension of the NURBS geometry
+
+       The physical dimension is inferred from the first patch,
+       i.e. number of coordinates per control point minus one (for the weight).
+       This method requires patch data to be present, i.e. HavePatches() == true */
+   int GetPatchDim() const;
```
dylan-copeland: For consistency with the `Mesh` class, should we name this `SpaceDimension()`?

IdoAkkerman: `GetPatchSpaceDimension`? Or is that to verbose?
I think Patch should be in the name, to convey where the dimension is coming from and justify giving an error when the patches are not defined.

tests/unit/mesh/test_nurbs.cpp
```diff
@@ -146,6 +148,291 @@ TEST_CASE("NURBS NC-patch mesh loading", "[NURBS]")
...
+      REQUIRE(a.GetOrder() == b.GetOrder());
+      REQUIRE(a.GetNCP() == b.GetNCP());
+      REQUIRE(a.Size() == b.Size());
+      for (int i = 0; i < a.Size(); i++)
+      {
+         REQUIRE( (a[i]-b[i]) == MFEM_Approx(0.));
+      }
+   };
+
+   const char *mesh_same_orientation = R"(
```
IdoAkkerman: Perhaps it is better to add these meshes  in `tests/data` - or another more appropriate location- and simple read them?

mesh/nurbs.cpp
IdoAkkerman: The routine is intended to check that opposing edges in 2D and 3D have the correct knotvector associated with them. This equivalent does not exist in 1D. Therefor, the (non-trivial) checks performed for the 1D case are (unnecessarily??!!!) more restrictive.

So unless they are deemed necessary I would suggest to remove them.

Test 1 could be made semi-generic (1d/2d/3d) when changed to 
```
if (patchTopo->GetNEdges() > 0)
{
   MFEM_VERIFY(edge_to_ukv.Size() == patchTopo->GetNEdges());
}
else // 1D without edges
{
  MFEM_VERIFY(edge_to_ukv.Size() == patchTopo->GetNE());
}
```
this is a relatively trivial check.

Test 2  seems to check something that does not seem essential, and something that is governed outside to the NURBS class. Should it be checked/enforced? Should it be checked/enforced here?

Test 3 seems pretty analogous to the 2d/3d checks. However, it seems overly strict, which will not cause issues at this point, but might later in the future. My suggestion would be to remove it. Or is this to harsh? 

@justinlaughlin @dylan-copeland I think this discussion is more fuel to the fire ... to the argument that the fundamental assumptions are not super clearly defined, decided and communicated.

mesh/nurbs.cpp
IdoAkkerman: Looks nice and useful. Should `orient` be used? To either change the sign of the `edges` or as separate output? This could also be postponed to when it is needed... just thinking out loud.

mesh/nurbs.cpp
kennyweiss: Thanks everyone. From @IdoAkkerman's comment, it sounds like I should make "Test 1" general and remove "Test 2" and "Test 3" from this PR, and that a follow-up issue/discussion/PR is warranted to make the expectations and checks explicit and consistent.

dylan-copeland: I agree that the checks should be added only if needed for some reason. @kennyweiss is there a reason or example where these assumptions are necessary? If the assumptions are well justified, it should be easy to satisfy them in 1D with edge/element vertex ordering. If the assumptions are unnecessary, then we could just remove the 1D checks.

mesh/nurbs.hpp
kennyweiss: I'm happy to change it to whatever you all feel is best.

mesh/nurbs.cpp
kennyweiss: @dylan-copeland -- I was mostly aiming for consistency. The `CheckPatch` seemed like a good way to centralize expectations, but it seems that it is overzealous compared to the actual requirements.

tests/unit/mesh/test_nurbs.cpp
kennyweiss: Thanks -- will do.  What about the files I added to `mfem/data`? Are they all ok there, or should some of them move to `mfem/test/data`?

mesh/nurbs.hpp
dylan-copeland: I am thinking of this dimension as the physical space dimension, independent of patches. I'm not sure the name needs to include "patch" to indicate the implementation. It is unfortunate that the physical dimension is available only in `NURBSPatch` as `Dim - 1`, and that complication leads to the requirement `HavePatches() == true` along with the documentation and function naming. Maybe a better solution is to add a data member in `NURBSExtension`, e.g. `int spaceDim` as in `class Mesh`, initialized to 0 and set by the first patch. It should be a generally valid assumption that the same space dimension will be used in all patches, which could be verified in `NURBSExtension` constructors, loaders, etc. That way, the patch-dependent implementation is in the constructors, rather than the getter function. In the case that patches are not defined, it can be set to `NURBSExtension::Dimension` as in `ConvertToPatches`. This would enable storing the space dimension independently of `NURBSPatch`, as an intrinsic property of the `NURBSExtension`.

justinlaughlin: I agree with Dylan that something like `SpaceDimension()` makes sense.

IdoAkkerman: As some preliminary work on refactoring the NURBS I have removed the patches as data member from the `NURBSExtension` this is in the branch `rn_rm_compreKV_Patches`. All function calls that require patches, have the patches as argument. This creates - in my opinion- a clean separation between the  topology/connectivity stuff and the actual geometry (or solution) defined using a gridfunction(or that vector converted to patchformat).

In my view the NURBS extension does not have a  spatial dimension. That only becomes relevant if the extension gets associated with a gridfunction (or patches). Therefore `SpaceDimension()` seems wrong, while  `GetPatchSpaceDimension` seems a good (temporary) compromise.
Note: some NURBS extension are only used for  gridfunctions representing solutions.

Perhaps I am unnecessarily stubborn, but I feel we are only adding more technical debt, which will take extra effort to untangle, exacerbated  by the some times long times to get changes/updates accepted.

Granted, a simple name change does not break the bank. It just feels like "we are still moving in the wrong direction".

Perhaps this PR is not the place to have this discussion.

kennyweiss: Thanks for the context. `GetPatchSpaceDimension()` seems like a good compromise for this PR.

dylan-copeland: We can use `GetPatchSpaceDimension()` for now, considering that major refactoring is planned, in which we may rethink the design of these classes. The simplest/minimal implementation can be done for now, as already done in this PR.

mesh/nurbs.cpp
kennyweiss: I updated the 1D `CheckPatch` based on (my understanding of) @IdoAkkerman's request (and following the other requests), keeping the requested version of "Test 1" and removing the other checks which were unnecessarily constraining.

mesh/nurbs.hpp
kennyweiss: I changed the name to `GetPatchSpaceDimension()`

tests/unit/mesh/test_nurbs.cpp
kennyweiss: I outsourced these two test meshes to `./tests/data/`

mesh/nurbs.cpp
kennyweiss: What were you thinking specifically about using `orient`? 
Looks like we'd want to combine it in some way with `CheckKVDirection()` to use it.

I was able to directly plug the returned `edges` into the call sites.

Would it make sense to save this for a future refactor?

IdoAkkerman: No problem. Nice addition! If needed we can add more functionality later.

mesh/nurbs.cpp
IdoAkkerman: I think the size check also makes sense for the 2D/3D cases, so I would personally prefer something like:

```
   // If the patch topology has an explicit `edges` section, require it to be
   // consistent with the 1D element list so patch index == edge index.
   // Otherwise, check for consistency with the number of elements
   const int expected_size = patchTopo->GetNEdges() > 0
                             ? patchTopo->GetNEdges()
                             : patchTopo->GetNE();
   if (edge_to_ukv.Size() != expected_size)
   {
      return false;
   }
```

kennyweiss: Done.

IdoAkkerman: I happy with this result!

justinlaughlin (approved):
Thanks Kenny this looks great!

tests/unit/mesh/test_nurbs.cpp
```diff
@@ -154,6 +156,177 @@ TEST_CASE("NURBS NC-patch mesh loading", "[NURBS]")
...
+      const Array<int> &new_orders = mesh.NURBSext->GetOrders();
+      REQUIRE(new_orders.Size() == orders.Size());
+      for (int i = 0; i < new_orders.Size(); ++i)
+      {
+         REQUIRE(new_orders[i] == max_order);
+      }
+   }
+}
+
+TEST_CASE("NURBS 1D shared KnotVector in patches", "[Mesh]")
```
dylan-copeland: `[Mesh]` -> `[NURBS]`?

tests/unit/mesh/test_nurbs.cpp
```diff
@@ -14,6 +14,8 @@ using namespace mfem;
 
 #include "unit_tests.hpp"
 
+#include <sstream>
```
dylan-copeland: Remove? It compiles without this include.

mesh/nurbs.hpp
```diff
@@ -599,22 +599,26 @@ class NURBSExtension
    /// Throw an error if any boundary patch has invalid KnotVector orientation.
    MFEM_DEPRECATED void CheckBdrPatches();
 
+   /// Return the patch-topology edge indices that define the KnotVectors for
+   /// patch @a p in each parametric direction.
+   void GetPatchKnotVectorEdges(int p, Array<int> &edges);
```
dylan-copeland: This is a very helpful function, but I'm not sure about the name. This function returns local patch edges 0, 1 in 2D, and 3, 8 in 3D, which are representative edges in each parametric direction. These edges are associated with the `KnotVectors` in each direction on the patch, but it may be misleading to say that the edges "define the KnotVectors." Also, I'm not sure the function name should include `KnotVector`. I would remove mention of `Knotvector` from the comment and the name. How about `GetPatchDirectionEdges` or`GetPatchAxisEdges`?

mesh/mesh.cpp
```diff
@@ -12389,6 +12402,44 @@ void Mesh::PrintTopoEdges(std::ostream &os, const Array<int> &e_to_k,
...
+      for (int i = 0; i < ne; i++)
+      {
+         const int *v = elements[i]->GetVertices();
+         int v0 = v[0], v1 = v[1];
+
+         int ki = e_to_k[i];
+         const bool flip = (ki < 0); // desired output vertex order: descending
+         if (flip) { ki = -1 - ki; } // print the unsigned knotvector index
+
+         if (vmap && ncmesh)
```
dylan-copeland: In this case, `(Dim == 1 && ncmesh) == true`. Is that possible?

kennyweiss: Sounds like we do not support/have 1D ncmesh NURBS -- I'll remove this.

mesh/nurbs.hpp
kennyweiss: Renamed.

mesh/mesh.cpp
kennyweiss: Done.

tests/unit/mesh/test_nurbs.cpp
kennyweiss: done

tests/unit/mesh/test_nurbs.cpp
kennyweiss: done

tzanio:
Thanks @kennyweiss!

Is this worth mentioning in `CHANGELOG`?

tzanio:

kennyweiss:
> Is this worth mentioning in `CHANGELOG`?

I wasn't sure.  

Would the following work:
```
- Improved support for 1D NURBS meshes with variable order, including using the patches construct for 1D NURBS meshes.
```

Should I add it? Or is the PR frozen once it's in `next`?

tzanio:
You can still add it. Just please wrap lines at 80 characters.

kennyweiss:
I updated the `CHANGELOG`

tzanio:

PR #5173, opened 2025-12-22
dFEM transpose action using PA cache

jandrej:

PR #5188, opened 2026-01-12
Dfem multiple integrators

victor-decaria-nnl:
I am trying to make it so you can add multiple integrators to a `DifferentiableOperator` in the same way that you can for `NonlinearForm` and `BilinearForm`. I've added a mass option to the diffusion dfem unit test to assemble (u,v) + (rho grad(u), grad(v)). I would like some help with proceeding.

**Status** (more details below):
- **Matrix free** : Works with suboptimal choices of qfunctions where they all use the same parameters.
- **GetDerivative** : Fixed a memory bug and it can run, but assertions fail when including mass.

### Matrix Free
I've found that the matrix free action works provided that the mass qfunction takes all the same parameters as the diffusion one, which isn't quite right but at least shows a working example where the integrators can have different inputs and outputs, i.e., `Gradient<U>` and `Value<U>` in the same operator. The desired behavior is that the qfunctions should be able to take different subsets of parameters from the parameter set that was used to construct the `DifferentiableOperator`. See `mf_mass_qf` vs `desired_mass_qf` here.

### GetDerivative
There seemed to be a memory bug with how the q data is cached for get derivative. See the changes to `doperator.hpp`. Both (u,v) and (rho grad(u), grad(v)) integrators would have the same derivative id, but different sizes on quadrature points so they need different cache vector sizes. Also it looks like the reference to the cache captured by the lambdas dangles when adding multiple integrators, and the code was crashing. I just hackily mapped it by vector size instead of derivative id to get it running, but the assertions don't pass.

### DIFF
--- a/tests/unit/dfem/test_diffusion.cpp
+++ b/tests/unit/dfem/test_diffusion.cpp
@@ -139,6 +139,10 @@ void diffusion(const char *filename, int p, bool include_mass)
          [] MFEM_HOST_DEVICE(const dscalar_t u, const real_t &rho /*Needed for some reason*/,
                              const tensor<real_t, DIM, DIM> &J, const real_t &w)
       { return tuple{u * w * det(J)}; };
+      const auto desired_mass_qf =
+         [] MFEM_HOST_DEVICE(const dscalar_t u,
+                             const tensor<real_t, DIM, DIM> &J, const real_t &w)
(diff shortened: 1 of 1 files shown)


victor-decaria-nnl:
Closing since it should be fixed by #5173

Issue #5162, opened 2025-12-16
Issues when building with CUDA on windows with Visual studio, latest release 2025


horstsoft0815:
Referring to [Issues when building with CUDA on windows with Visual studio](https://github.com/mfem/mfem/issues/4644)

1) The issues trying to build mfem with CUDA still persist, starting with an incorrectly formatted compiler option "/wd4819", as an output of the CMake generation, leading to a compile command line of
"[...] /wd4819 -XCompiler="EHsc""
instead of
"[...] -XCompiler="EHsc, /wd4819""
leading to an immediate failure of the nvcc compiler.

After fixing the above problem manually by editing the build command:

2) In addition also the reported M_PI issue persists, which is most likely caused by including cmath.h somewhere before the macro "_USE_MATH_DEFINES" was defined.

In addition there are issues at multiple code places, of the form

3) "an "asm" declaration is not allowed here", for instance in bilininteg_hcurl_kernel.hpp:3048 --> this is only one example of multiple similar occasions
4) "On Windows, the enclosing parent function ("PAHcurlVecH1IdentityApplyTranspose3D") for an extended __host__ __device__ lambda cannot have internal or no linkage", for instance in bilininteg_interp_pa.cpp:1592 --> this is only one example of multiple similar occasions.

Note that the mfem compilation without CUDA support works fine, the above issues arise when setting
MFEM_USE_CUDA=YES
The INSTALL was followed thoroughly.
Windows version: 11 23H2 (10.0.22631 Build 22631.6060)
Cuda toolkit version is 12.8
MSVSC compiler version is 19.40.33818
Visual Studio version is 17.10.10

Building eg. Hypre with the same configuration leads to a working library without any problems, which was tested outside of mfem.

see the attached CMakeCache.txt
[CMakeCache.txt](https://github.com/user-attachments/files/24189528/CMakeCache.txt)

horstsoft0815:
I made some progress.
Issue 1) can be solved when replacing the following CMake command in the (uppermost) CMakeLists.txt

if (MSVC)
  target_compile_options(mfem PUBLIC "/wd4819")
endif()

by 

if (MSVC)
  add_compile_options(/wd4819)
endif()

Issue 2) can be resolved by adding "-D_USE_MATH_DEFINES" to the CMAKE_CUDA_FLAGS.

The issues 3) & 4) sadly remain unresolved...
[...]

Issue #4449, opened 2024-08-15
add utf-8 option for  MSVC compiler to fix build fail with MSVC and code page 936


yuanzid:
###Trouble
Visual Studio with MSVC compiler compiles sources with **current user code** page usually as  [this page](https://learn.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170). Some special characters have been found  in source code comment, which causes warning C4819#2976.  Even worse when using some code page (for example 936 which is code page for simplified Chinese), the **compilation fails** and the error message is really misleading. 

###Fix
add 
` cmake
if(MSVC)
   add_compile_options("/utf-8")
endif()
`
or something similar in CMakeLists.txt

jandrej:
Doesn't https://github.com/mfem/mfem/issues/2976 and https://github.com/mfem/mfem/pull/3004 take care of that?

yuanzid:
> Doesn't #2976 and #3004 take care of that?
no,  it doesn't .  it's  case by  case. The wd4819 option  suppress 
the warnin  and  hide matters.  Some code  page  would work fine. some code page would fail, msvc build fail and show lots of misleading error message.

PR #5204, opened 2026-01-23
Fix CUDA compilation errors on Windows

### LINKED ISSUE #4644: Issues when building with CUDA on windows with Visual studio
Hi everyone, Merry Christmas.

I have some issues when building MFEM with CUDA on the windows platform. First of all, it seems that the "/wd4819" options is read wrongly by the nvcc compiler causing it to give an error ("A single input file is required for a non-link phase when an outputfile is specified").

This is simple enough fix for a local build as I just remove the line defining that option from the cmakelists.

Furthermore, I get an error for M_PI and some other constants being undefined. The cause seems to be that nvcc includes cmath before the _USE_MATH_DEFINES in read in the code. My current fix for this is to just add -D_USE_MATH_DEFINES in the additional comamand line options of the mfem project when building the library.

There are also some problems directly in the code: 
* Line 90 in isockstream.cpp needing "::" prefix because of "using namespace std"  (I saw an existing issue on this, cannot find it again but this was the fix suggested)
* line 177 in error.hpp "asm("trap;"); doesn't compile on 64-bit platforms. I got around this by changing asm directive with __trap(); I found that another solution could be an assertion but I am not knowledgeable enough to know if it is a viable replacement.

The last problem I cannot fix in a natural way that actually works. I get an error saying "cannot open source file "_config.hpp" ".
The file is in mfem/<build_folder>/config/_config.hpp but the project cannot find it. I've only been able to build the library by manually including the hpp file to mfem project but then I cannot run any example with cuda for some reason with the following error:

image

I am open to do any testing I can do on Windows platform as I understand that most of this was probably done on Linux
Any help would be appreciated. 

Best regards,
Max

adam-sim-dev:
Close https://github.com/mfem/mfem/issues/4644 and https://github.com/mfem/mfem/issues/5162

These fixes are proposed by Google AI, so please review carefully. 

I tested these fixes on Windows 11 with Visual Studio 2022 and CUDA 13.1.1, Ubuntu 24.04.3 with GCC 13.3.0 and CUDA 13.1.1

(1)
```
if (MSVC)
  target_compile_options(mfem PUBLIC "/wd4819")
endif()
```
is added by me in PR https://github.com/mfem/mfem/pull/3004. The /wd4819 flag is MSVC-specific and cannot be parsed directly by nvcc. When compiling CUDA files, this flag must be explicitly passed to the underlying host compiler using the -Xcompiler prefix.

```
if(MSVC)
    add_compile_options(/wd4819)
    if(CMAKE_CUDA_COMPILER)
        add_compile_options($<$<COMPILE_LANGUAGE:CUDA>:SHELL:-Xcompiler /wd4819>)
    endif()
endif()
```

does not work either, I don't know why. Google AI said that "this approach fails because global add_compile_options are often overridden or ignored when a project explicitly sets CMAKE_CUDA_FLAGS or uses target_compile_options. The latter set() commands in the CMakeLists.txt wipe out the flags added here."

The C4819 warnings are very annoying for Windows users. It would be very helpful if the development team could look into a permanent fix to keep the build logs clean and actionable.  Related to https://github.com/mfem/mfem/issues/4449

(2) The second error is about the use of `asm("trap;");` in `error.hpp`.  

https://github.com/mfem/mfem/blob/0e61a94b5f78f80d9514fd7d442a3c2e8f9412f9/general/error.hpp#L177

It causes errors like:
```
fem\integ\bilininteg_hcurl_kernels.hpp(2533): error : an "asm" declaration is not allowed here
2>        { printf("This kernel should only be used on GPU."); asm("trap;"); };
```

(3) https://github.com/mfem/mfem/blob/0e61a94b5f78f80d9514fd7d442a3c2e8f9412f9/linalg/kernels.hpp#L566

`linalg\kernels.hpp(573): error : identifier "mfem::kernels::internal::Epsilon" is undefined in device code`
`linalg\kernels.hpp(597): error : identifier "mfem::kernels::internal::Epsilon" is undefined in device code`

(4) https://github.com/mfem/mfem/blob/0e61a94b5f78f80d9514fd7d442a3c2e8f9412f9/general/reducers.hpp#L471-L472

```
linalg\../general/reducers.hpp(472): error : identifier "__lzclz" is undefined
2>        return sizeof(unsigned) * 8 - __lzclz(N);
2>                                      ^
2>          detected during:
2>            instantiation of "int mfem::internal::reduction_kernel<B, R>::block_log2(unsigned int) [with B=lambda [](int, value_type &)->void, R=mfem::L2Reducer]" at line 549
2>            instantiation of "void mfem::reduce(int, T &, B &&, const R &, __nv_bool, mfem::Array<T> &) [with T=mfem::DevicePair<double, double>, B=lambda [](int, value_type &)->void, R=mfem::L2Reducer]" at line 981 of \linalg\vector.cpp
```

(5) See M_PI related errors (https://github.com/mfem/mfem/issues/4644) only for serial builds （maybe because I build hypre without CUDA,  -I/hypre and -I/metis are included before -I/cccl, metis and hypre include math.h implicitly?）. M_PI is defined in C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\ucrt\math.h and corecrt_math_defines.h if _USE_MATH_DEFINES is defined. google.com/ai said that

```
The root cause of the M_PI undefined error when compiling MFEM with nvcc lies in the intervention of the NVIDIA CCCL (CUDA C++ Core Libraries).
CCCL is designed to provide a standard-compliant library (libcudacxx) that ensures consistent behavior across both CPU and GPU. Since M_PI is not part of the ISO C++ standard—but rather a POSIX extension—the CCCL <cuda/std/cmath> wrappers omit these non-standard macros to maintain strict standards compliance and cross-platform reliability. When nvcc processes source files as CUDA code, it prioritizes CCCL headers, which effectively masks the non-standard constants typically provided by host compilers (like MSVC via _USE_MATH_DEFINES), leading to the "identifier undefined" error in the compilation chain.
```


general/reducers.hpp
```diff
@@ -466,10 +466,15 @@ template<class B, class R> struct reduction_kernel
...
-#if defined(__GNUC__) || defined(__clang__)
+#if defined(__CUDA_ARCH__)
+      return N ? (int)(sizeof(unsigned) * 8 - __clz(N)) : 0;
+#elif defined(__GNUC__) || defined(__clang__)
       return N ? (sizeof(unsigned) * 8 - __builtin_clz(N)) : 0;
 #elif defined(_MSC_VER)
-      return sizeof(unsigned) * 8 - __lzclz(N);
+      if (N == 0) { return 0; }
+      unsigned long index;
+      _BitScanReverse(&index, (unsigned long)N);
```
helloworld922, line 480: Does this need to include `<intrin.h>`?

general/error.hpp
```diff
@@ -174,7 +174,7 @@ __device__ void abort_msg(T & msg)
 #define MFEM_ABORT_KERNEL(...) \
    {                           \
       printf(__VA_ARGS__);     \
-      asm("trap;");            \
+      __trap();            \
```
helloworld922, line 169: It seems that clang cuda doesn't seem to support `__trap();` (on Linux, unsure about Windows).

This seems to work for me with nvcc+cuda and clang+cuda on Linux:

```c++
#if defined(__CUDA_ARCH__)
#if defined(__NVCC__)
#define MFEM_ABORT_KERNEL(...) \
   {                           \
      printf(__VA_ARGS__);     \
      __trap();                \
   }
#elif defined(__clang__)
#define MFEM_ABORT_KERNEL(...) \
   {                           \
      printf(__VA_ARGS__);     \
      asm("trap;");            \
   }
#endif
#elif defined(__HIP_DEVICE_COMPILE__)
#define MFEM_ABORT_KERNEL(...) \
   {                           \
      printf(__VA_ARGS__);     \
      abort_msg("");           \
   }
#else
#define MFEM_ABORT_KERNEL(...) \
   {                           \
      printf(__VA_ARGS__);     \
      MFEM_ABORT("");          \
   }
#endif
```

general/reducers.hpp
adam-sim-dev: While Copilot recommended explicitly including <intrin.h>, I found that MFEM builds successfully without it on my current setup. Anyway, I think it should be included.

PR #5210, opened 2026-01-27
Fix Array non-owning move assignment

hughcars:
The move assignment for an Array which does not own its memory does not assign the values. This can result in for instance DenseMatrix which are views into a DenseTensor, not having values assigned as expected. The fix is for the case of non-owning memory, to fallback to the copy assignment operator by casting to an lvalue reference.

This bug was introduced in https://github.com/mfem/mfem/pull/4028 which changed the backing memory model for DenseMatrix to the Array, from which it inherited this bug. I found it because we were move assigning into slices of DenseTensor, via it's `operator()` method.

An example of this in Palace https://github.com/awslabs/palace/blob/main/palace/models/materialoperator.cpp#L249, this can be addressed with out param equivalents, but is fairly semantically surprising, that assigning to something does not result in the LHS being equal to what the right hand side was after assignment is over. I believe this comes down to owning or not owning the data changing the semantics from a value to a pointer, but only when the RHS is an rvalue. 

As an example `tensor(i) = tensor(j)` would not copy the data from one tensor slice to another without this change.

| [#5210](https://github.com/mfem/mfem/pull/5210) | @hughcars | @tzanio | @helloworld922 + @pazner + @v-dobrev | 1/28/26 | 2/17/26 | 2/24/26 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


pazner:
I think your proposed change makes sense. When the left-hand side of an assignment operation is an alias, the semantics can be a little tricky (even without move semantics).

Supposing `alias` is an alias of `base`. Then, if `alias` and `other` have the same size, `alias = other;` will modify `base`, but if they have different sizes, then `base` will not be modified.

In the case of this PR, the current behavior (on master) is that `alias = other;` will modify `base`, but `alias = std::move(other);` will not modify `base`.

Maybe we could also add an overload to `Array<T>::CopyFrom(const Array<U> &)` that always copies the contents (this is the same semantics as `Array<T>CopyFrom(const U *)`). That would make things explicit.

hughcars:
> I think this makes sense. When the left-hand side of an assignment operation is an alias, the semantics can be a little tricky (even without move semantics).
> 
> Supposing `alias` is an alias of `base`. Then, if `alias` and `other` have the same size, `alias = other;` will modify `base`, but if they have different sizes, then `base` will not be modified.
> 
> In the case of this PR, the current behavior (on master) is that `alias = other;` will modify `base`, but `alias = std::move(other);` will not modify `base`.
> 
> Maybe we could also add an overload to `Array<T>::CopyFrom(const Array<U> &)` that always copies the contents (this is the same semantics as `Array<T>CopyFrom(const U *)`). That would make things explicit.

For the `DenseTensor` piece, it's possible to use `.Set(1, ...)` and out parameters to avoid the assigning, so I have a workaround for the current issue encountered in Palace. If the broader consensus is that this behaviour is the desired, then we can work with it and I'll close this PR, but it was quite confusing for me.

pazner:
Just to clarify, by "I think this makes sense", I was referring to your proposed change :)

hughcars:
> Just to clarify, by "I think this makes sense", I was referring to your proposed change :)

Oh! I was a bit confused by your sentence. If you look at the Palace PR linked here, there are a few more examples of needing to use `.Set` to recover the previous behaviour, so it does prove it can be done if something in the spirit of this PR isn't ultimately supported.

tzanio:

v-dobrev:
I thought about this issue and I think a better solution is to `= delete` the move-assignment operator of `DenseMatrix`. This way assignments to `DenseMatrix` will always be copy-assignments. The reason I think this is a better solution is that re-defining the move-assignment (especially at the level of `Array`) to do a deep copy, in the case when the LHS is an alias, goes against the idea of move-assignment. We will lose the ability to move-assign `DenseMatrix` but that is not a big deal, I think -- this was already the case for a long time before we added it in #4028.

Another option I was thinking about is to add a view-of-`DenseMatrix` class that will be the result of indexing into a `DenseTensor` with a single index. The idea is that then this view-class will not allow the viewed memory to be changed by assignment operators -- i.e. move-assignment will not be allowed, and assignment from `DenseMatrix` will copy the data. This approach may be better ultimately but it will probably require bigger changes.

general/array.hpp
```diff
@@ -114,11 +114,25 @@ class Array
...
+         src.DeleteAll();  // Leave src in valid moved-from state
+      }
+      else
+      {
+         // Default move semantics for owning arrays
+         Swap(src);
+         src.DeleteAll();
+      }
       return *this;
    }
```
pazner: If we decide that we do want to implement this approach, here are a couple small suggestions.

You don't need to explicitly cast to lvalue reference; it will happen automatically. Setting src to the "moved-from" state can be hoisted out of the if.

It only makes sense to use the copy operation if *this doesn't need to grow. If it needs to grow, then it won't be an alias anymore.

```suggestion
   /** If *this is a non-owning view (e.g., from MakeRef()), the data is copied
       so that the base is also modified. */
   Array<T> &operator=(Array<T> &&src)
   {
      if (this == &src) { return *this; }
      // If *this is a non-owning view (alias), and its capacity is sufficient
      // to contain src, then copy into *this so that the alias's base memory is
      // modified.
      if (!OwnsData() && Capacity() >= src.Size())
      {
         *this = src; // Copy assignment.
      }
      else
      {
         Swap(src); // Swap the pointers only.
      }
      src.DeleteAll();
      return *this;
   }
```

hughcars:
> I thought about this issue and I think a better solution is to `= delete` the move-assignment operator of `DenseMatrix`. This way assignments to `DenseMatrix` will always be copy-assignments. The reason I think this is a better solution is that re-defining the move-assignment (especially at the level of `Array`) to do a deep copy, in the case when the LHS is an alias, goes against the idea of move-assignment. We will lose the ability to move-assign `DenseMatrix` but that is not a big deal, I think -- this was already the case for a long time before we added it in #4028.
> 
> Another option I was thinking about is to add a view-of-`DenseMatrix` class that will be the result of indexing into a `DenseTensor` with a single index. The idea is that then this view-class will not allow the viewed memory to be changed by assignment operators -- i.e. move-assignment will not be allowed, and assignment from `DenseMatrix` will copy the data. This approach may be better ultimately but it will probably require bigger changes.

I have thought about the "View" type, for lack of a better term, too. In order that it not break a few things, it would need some careful implicit conversion operators, certainly to DenseMatrix, if not also from, so things like `operator DenseMatrix() {}` that binds a non-owning DenseMatrix onto the view. This so you could pass those slices into functions that operate on DenseMatrix. That or the view needs to be a child, but then modifying `operator=` starts to mess with the "is a" inheritance pattern. It gets fiddly pretty fast. Basically stealing ideas from `span` and `mdspan`. To make lots of methods work, without type erasure or template proliferation, allowing DenseMatrix and this slice view to work together in most methods, will be pretty tricky.

tzanio:

PR #5223, opened 2026-02-05
Partial assembly algorithms for Bernstein basis on simplicial meshes [pa-simplices-dev]

jdongg:
This PR adds partial assembly routines for mass and diffusion integrators on simplicial meshes with Bernstein basis functions (i.e. Positive basis in MFEM), mostly following the algorithms in [1]. The PR includes the following changes:

**Quadrature rules**
- Added `Quadrature1D::GaussJacobi` routine which computes the $n$-point Gauss-Jacobi rule with Jacobi parameters $(\alpha,\beta)$ (with reasonable limitations on the size of $\alpha$ and $\beta$ due to use of asymptotic expansions of the Jacobi roots).
- Added `TriangleStroudIntRules` and `TetrahedronStroudIntRules``IntegrationRules` objects, for use with ragged tensor Bernstein algorithms (they may be used as simplicial quadrature rules in general contexts as well). 
- Added `DuffyTrans` routine which maps the nodes of a given quadrature rule from the reference $d$-cube to simplex (only implemented for 2D and 3D rules). The corresponding inverse transform is performed on-the-fly when constructing `DofToQuad` arrays.

**FE structures**
- Added `RAGGED_TENSOR` `Mode` to the `DofToQuad` class for use with partial assembly routines with the Bernstein basis functions. 
- Added `GetRaggedTensorDofToQuad` routines for the `H1Pos_TriangleElement` and `H1Pos_TetrahedronElement` classes which populates arrays for storing the 1D ragged tensor bases, i.e. $B_{\alpha_1}^{p}(t_1)$, $B_{\alpha_2}^{p-\alpha_1}(t_2)$, $B_{\alpha_3}^{p-\alpha_1-\alpha_2}(t_3)$, etc. and special maps from Bernstein multiindices to lexicographic indices.

**Integrators**
- Added fallback and shared memory kernels for mass and diffusion integrators on triangles and tetrahedrons (`PAMassApplyTriangle`, `SmemPAMassApplyTriangle`, `PAMassApplyTetrahedron`, `SmemPAMassApplyTetrahedron`, `PADiffusionApplyTriangle`, `SmemPADiffusionApplyTriangle`, `PADiffusionApplyTetrahedron`, `SmemPADiffusionApplyTetrahedron`).
- Small modifications to `AssemblePA` routines in `MassIntegrator` and `DiffusionIntegrator` classes to allow calling `GetRaggedTensorDofToQuad` routines on simplicial Bernsteins

**Utilities**
- Added `FiniteElementSpace::IsBernsteinSimplexSpace` routine for checking whether Bernstein basis is being used on simplices. Used to toggle mapping of Stroud nodes between $d$-cube and reference simplex in `AssemblePA` routines. 

[1] Ainsworth, M., Andriamaro, G., & Davydov, O. (2011). Bernstein–Bézier finite elements of arbitrary order and optimal assembly procedures. SIAM Journal on Scientific Computing, 33(6), 3087-3109. [https://doi.org/10.1137/11082539X](https://doi.org/10.1137/11082539X).

| [#5223](https://github.com/mfem/mfem/pull/5223) | @jdongg | @tzanio | @camierjs + @v-dobrev + @pazner | 2/6/26 | 6/16/26 | 6/22/26 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:

fem/fe/fe_base.hpp
```diff
@@ -167,7 +167,14 @@ class DofToQuad
...
           product structure. The ordering of the degrees of freedom is the
           same as TENSOR, but the sizes of B and G are the same as FULL.*/
-      LEXICOGRAPHIC_FULL
+      LEXICOGRAPHIC_FULL,
+
+      /** @brief Ragged tensor product representation using 1D matrices/tensors
+          with dimensions using 1D number of quadrature points and ragged tensor degrees of
+          freedom. Used only for partial assembly of the H1 positive basis. The
+          size of B is d1d x qnpt x dim. Since different Gauss-Jacobi quadrature rules
+          are employed in each dimension, we need to store dim arrays. */
```
pazner, line 177: Maybe split the `@brief` part from the rest of the explanation.

fem/fe/fe_base.hpp
```diff
@@ -198,6 +205,33 @@ class DofToQuad
        - #ndof x #nqpt x dim, for vector elements. */
    Array<real_t> Bt;
 
+   /** @brief Special basis function structures for positive (Bernstein) basis with
```
pazner: This feels like it should be a separate data structure from `DofToQuad`, since none of these members make sense for TENSOR or FULL mode.

fem/integ/bilininteg_diffusion_kernels.cpp
```diff
@@ -19,8 +19,17 @@ namespace mfem
 DiffusionIntegrator::Kernels::Kernels()
 {
    // 2D
+   // Q = P
```
pazner: Do we want to keep these new specializations or is it for testing purposes?

.gitignore
```diff
@@ -443,6 +443,8 @@ miniapps/diag-smoothers/mg-abs-l1-jacobi
 miniapps/contact/contact
 miniapps/contact/ParaView
 
+miniapps/pa-simplices/ex_pa_simplices
+
```
pazner: Do we want to keep this addition?

Presumably this could be rolled into ex1 (i.e. if `pa` is true, and the mesh is simplicial, then use Bernstein)

tests/benchmarks/bench_ceed.cpp
```diff
@@ -86,17 +87,43 @@ static void AddKernelSpecializations()
...
+      {
+         if constexpr (DIM == 2)
+         {
+            return Mesh::MakeCartesian2DWith4TrisPerQuad(nx, ny);
+         }
+         else
+         {
+            return Mesh::MakeCartesian3DWith24TetsPerHex(nx, ny, nz);
+         }
+      }
```
pazner: Why are these ones used rather than the standard `MakeCartesian2D` and `MakeCartesian3D` (with fewer tris/tets per grid cell)?

miniapps/common/mesh_extras.hpp
```diff
@@ -151,8 +151,8 @@ class KershawTransformation : public VectorCoefficient
 // pmesh->Transform(spiralT);
 class SpiralTransformation : public VectorCoefficient
 {
-private:
```
pazner, line 154: Keep `private:`?

mesh/mesh.cpp
```diff
@@ -15804,8 +15804,8 @@ Mesh PartitionMPI(int dim, int mpi_cnt, int elem_per_mpi, bool print,
 
    if (print && dim == 2)
    {
-      int elem_par_x = mpi_x * el0_x * pow(2, par_ref),
-          elem_par_y = mpi_y * el0_y * pow(2, par_ref);
+      int elem_par_x = mpi_x * el0_x * static_cast<int>(pow(2, par_ref)),
```
pazner: Are these changes prompted by something specific? (Same question for similar changes elsewhere).

general/forall.hpp
```diff
@@ -57,8 +57,8 @@ struct DofQuadLimits_CUDA
 
 struct DofQuadLimits_HIP
 {
-   static constexpr int MAX_D1D = 10;
-   static constexpr int MAX_Q1D = 10;
```
pazner, line 61: If these are just because of the new simplex integrators, we probably want to modify the checks in those integrators rather than globally change this.

fem/intrules.hpp
```diff
@@ -407,7 +415,8 @@ class Quadrature1D
       OpenUniform     = 2,  ///< aka open Newton-Cotes
       ClosedUniform   = 3,  ///< aka closed Newton-Cotes
       OpenHalfUniform = 4,  ///< aka "open half" Newton-Cotes
-      ClosedGL        = 5   ///< aka closed Gauss Legendre
+      ClosedGL        = 5,  ///< aka closed Gauss Legendre
+      GaussJacobi     = 6
```
pazner: This is different from the other types because it is not a single rule, but a parameterized family of rules. I don't think it fits this framework.

fem/intrules.hpp
```diff
@@ -461,12 +475,22 @@ class IntegrationRules
    /// The following methods allocate new IntegrationRule objects without
    /// checking if they already exist.  To avoid memory leaks use
    /// IntegrationRules::Get(int GeomType, int Order) instead.
-   IntegrationRule *GenerateIntegrationRule(int GeomType, int Order);
+   IntegrationRule *GenerateIntegrationRule(int GeomType, int Order,
+                                            bool stroud = false);
    IntegrationRule *PointIntegrationRule(int Order);
-   IntegrationRule *SegmentIntegrationRule(int Order);
+   IntegrationRule *SegmentIntegrationRule(int Order, real_t alpha=0,
+                                           real_t beta=0);
```
pazner: This doesn't really make sense: you can only have one alpha and beta in `SegmentIntRules`, which is then used for the quad and hex rules, so only (0,0) would make sense to use here.

fem/intrules.hpp
```diff
@@ -430,9 +439,14 @@ class IntegrationRules
 
    Array<IntegrationRule *> PointIntRules;
    Array<IntegrationRule *> SegmentIntRules;
+   Array<IntegrationRule *> SegmentStroud1IntRules;
```
pazner: These are just Gauss-Legendre points, right? So we can just use `SegmentIntRules` rather than duplicating.

fem/intrules.hpp
```diff
@@ -430,9 +439,14 @@ class IntegrationRules
...
+   Array<IntegrationRule *> SegmentStroud2IntRules;
+   Array<IntegrationRule *> SegmentStroud3IntRules;
```
pazner: Do we really need to cache these?

Perhaps it's simpler to cache only the `TriangleStroudIntRules` and `TetrahedronStroudIntRules`, and rebuild the 1D rules when requested. (I don't think the overhead for the 1D rules should be particularly high).

In any case, the names should ideally be clearer (i.e. Stroud_1_0_Rules and Stroud_2_0_Rules).

fem/intrules.hpp
```diff
@@ -461,12 +475,22 @@ class IntegrationRules
...
    IntegrationRule *PointIntegrationRule(int Order);
-   IntegrationRule *SegmentIntegrationRule(int Order);
+   IntegrationRule *SegmentIntegrationRule(int Order, real_t alpha=0,
+                                           real_t beta=0);
+   IntegrationRule *SegmentStroud1IntegrationRule(int Order, real_t alpha=0,
+                                                  real_t beta=0);
+   IntegrationRule *SegmentStroud2IntegrationRule(int Order, real_t alpha=1,
+                                                  real_t beta=0);
+   IntegrationRule *SegmentStroud3IntegrationRule(int Order, real_t alpha=2,
+                                                  real_t beta=0);
```
pazner: See my other comment: if we don't cache separately the 1D Stroud rules, then these can be removed, and the interface will be simplified.

fem/intrules.hpp
```diff
@@ -478,9 +502,10 @@ class IntegrationRules
                              int type = Quadrature1D::GaussLegendre);
 
    /// Returns an integration rule for given GeomType and Order.
-   const IntegrationRule &Get(int GeomType, int Order);
+   const IntegrationRule &Get(int GeomType, int Order, bool stroud = false);
```
pazner: Somehow this doesn't feel like the right interface. For non-simplex geometries, the `stroud` parameter is meaningless. I don't really have a better suggestion for now though…

fem/intrules.hpp
```diff
@@ -478,9 +502,10 @@ class IntegrationRules
...
 
-   void Set(int GeomType, int Order, IntegrationRule &IntRule);
+   void Set(int GeomType, int Order, IntegrationRule &IntRule,
+            bool stroud = false);
```
pazner: Do we really want to support overwriting Stroud rules? I don't see the purpose of it.

fem/intrules.hpp
```diff
@@ -270,6 +270,12 @@ class IntegrationRule : public Array<IntegrationPoint>
    /// applying this rule on each knot interval.
    IntegrationRule* ApplyToKnotIntervals(KnotVector const& kv) const;
 
+   /// Duffy Transformation of 2D and 3D tensor product rule.
+   void DuffyTrans(int dim);
+
+   /// Inverse Duffy Transformation for 2D and 3D tensor product rules.
+   const IntegrationRule InverseDuffyTrans(int dim) const;
```
pazner: Should write out what the transformation (and its inverse) actually are in coordinates.

fem/intrules.cpp
```diff
@@ -224,6 +224,84 @@ IntegrationRule::ApplyToKnotIntervals(KnotVector const& kv) const
    return kvir;
 }
 
+void IntegrationRule::DuffyTrans(int dim)
+{
+   // todo: add dimension argument, currently implement for triangle
+   IntegrationRule ir_mapped(GetNPoints());
+
```
pazner: Unused

fem/intrules.cpp
```diff
@@ -224,6 +224,84 @@ IntegrationRule::ApplyToKnotIntervals(KnotVector const& kv) const
...
+   if (dim == 2)
+   {
+      for (int i = 0; i < GetNPoints(); i++)
+      {
+         const real_t L1 = IntPoint(i).x;
+         const real_t L2 = IntPoint(i).y * (1 - L1);
+         IntegrationPoint &ip = const_cast<IntegrationPoint &>((*this)[i]);
+         ip.x = L1;
+         ip.y = L2;
+         // corresponds to x1 = (1,0) and x2 = (0,1)
```
pazner: Seems excessive?

```suggestion
         IntPoint(i).y *= (1.0 - IntPoint(i).x);
```

fem/intrules.cpp
```diff
@@ -224,6 +224,84 @@ IntegrationRule::ApplyToKnotIntervals(KnotVector const& kv) const
...
+      {
+         const real_t L1 = IntPoint(i).x;
+         const real_t L2 = IntPoint(i).y * (1 - L1);
+         const real_t L3 = IntPoint(i).z * (1 - L1 - L2);
+         IntegrationPoint &ip = const_cast<IntegrationPoint &>((*this)[i]);
+         ip.x = L1;
+         ip.y = L2;
+         ip.z = L3;
+         // corresponds to x1 = (1,0,0) and x2 = (0,1,0) and x3 = (0,0,0)
+         // and x4 = (0,0,1)
```
pazner: Similarly,


```suggestion
         IntPoint(i).z *= (1.0 - IntPoint(i).x - IntPoint(i).y);
         IntPoint(i).y *= (1.0 - IntPoint(i).x);
```

fem/intrules.cpp
```diff
@@ -224,6 +224,84 @@ IntegrationRule::ApplyToKnotIntervals(KnotVector const& kv) const
...
+      for (int i = 0; i < GetNPoints(); i++)
+      {
+         const real_t L1 = IntPoint(i).x;
+         const real_t L2 = IntPoint(i).y * (1 - L1);
+         IntegrationPoint &ip = const_cast<IntegrationPoint &>((*this)[i]);
+         ip.x = L1;
+         ip.y = L2;
+         // corresponds to x1 = (1,0) and x2 = (0,1)
+      }
+      return;
```
pazner: You can delete the early returns, turn the next `if` into `else if`, and the final `ABORT` can be in the `else` clause

fem/intrules.hpp
pazner: Also a little strange that one of these is in-place, and the other isn't.

fem/intrules.cpp
```diff
@@ -421,6 +499,142 @@ class HP_Quadrature1D
...
+
+   if (alpha <= -1.0 || beta <= -1.0)
+   {
+      MFEM_ABORT("Gauss-Jacobi quadrature only defined for alpha > -1 and beta > -1");
+   }
+   // Jacobi weight function is undefined whenever alpha <= -1 or beta <= -1
+
+   if (alpha > 4.0 || beta > 4.0)
+   {
+      MFEM_ABORT("Current Gauss-Jacobi quadrature implementation only tested for alpha <= 4 and beta <= 4");
```
pazner: Only tested, or only works?

fem/intrules.cpp
```diff
@@ -421,6 +499,142 @@ class HP_Quadrature1D
...
+      // of Gamma function instead, i.e. Gamma(a)/Gamma(b) = exp(lgamma(a) - lgamma(b))
+      ir->IntPoint(n-i).x = 0.5 * xi + 0.5;
+      ir->IntPoint(n-i).weight = 0.5 * c0 * pow(2.0,
+                                                ab + 1) / ((1.0 - xi*xi)*pp*pp) / pow(2, ab);
+      // map nodes and weights to the interval [0,1]
+   }
+
+#else // MFEM_USE_MPFR is defined
+
+   MFEM_ABORT("MPFR implemented of Gauss-Jacobi quadrature not defined yet")
```
pazner: To do?

fem/intrules.cpp
```diff
@@ -421,6 +499,142 @@ class HP_Quadrature1D
...
+         // p1 is Jacobi polynomial
+         pp = n * (alpha - beta - (2 * n + ab) * z) * p1 + 2 * (n + alpha) *
+              (n + beta) * p2;
+         pp = pp / ((2 * n + ab) * (1 - z*z));
+         // derivative of the Jacobi polynomial
+         if (done) { break; }
+
+         dz = p1/pp;
+#ifdef MFEM_USE_SINGLE
+         if (std::abs(dz) < 1e-7)
```
pazner, line 574: Does `std::numeric_limits<real_t>::epsilon()` work in the single case too?

fem/intrules.cpp
```diff
@@ -1654,6 +1973,36 @@ IntegrationRule *IntegrationRules::TriangleIntegrationRule(int Order)
...
+   }
+
+   AllocIntRule(TriangleStroudIntRules, RealOrder); // RealOrder >= Order
+   TriangleStroudIntRules[RealOrder-1] =
+      TriangleStroudIntRules[RealOrder] =
+         new IntegrationRule(*SegmentStroud2IntRules[RealOrder],
+                             *SegmentStroud1IntRules[RealOrder]);
+   // create rule in unit square
+   TriangleStroudIntRules[RealOrder-1]->DuffyTrans(2);
+   // map rule to reference triangle
```
pazner: These comments refer to the preceding lines

fem/fespace.hpp
```diff
@@ -1502,6 +1502,15 @@ class FiniteElementSpace
...
    }
 
+   /// Return whether the space is Bernstein on a simplex
+   bool IsBernsteinSimplexSpace() const
+   {
+      const auto *h1_fec = dynamic_cast<const H1_FECollection *>(fec);
+      return (h1_fec && h1_fec->GetBasisType() == BasisType::Positive &&
+              (GetTypicalFE()->GetGeomType() == Geometry::TRIANGLE ||
+               GetTypicalFE()->GetGeomType() == Geometry::TETRAHEDRON));
+   }
```
pazner: What about Bernstein L2?

Also, rather than check the typical geometry, we probably need to check all the geometries in the mesh.

pazner: It actually seems like `UsesRaggedTensorBasis` does the correct check

fem/integ/bilininteg_mass_pa.cpp
```diff
@@ -164,7 +175,46 @@ void MassIntegrator::AddMultPA(const Vector &x, Vector &y) const
          MFEM_ABORT("OCCA PA Mass Apply unknown kernel!");
       }
 #endif // MFEM_USE_OCCA
-      ApplyPAKernels::Run(dim, D1D, Q1D, ne, B, Bt, D, x, y, D1D, Q1D);
+
+      if (fespace->IsBernsteinSimplexSpace())
+      {
+         const Array<real_t> &Ba1 = maps->Ba1;
```
pazner: Maybe the kernel function should just take `DofToQuad` (or the analogous object if that is refactored) rather than passing the arrays individually?

fem/integ/bilininteg_mass_pa.cpp
```diff
@@ -164,7 +175,46 @@ void MassIntegrator::AddMultPA(const Vector &x, Vector &y) const
...
+            {
+               return internal::OccaPAMassApply2D(D1D,Q1D,ne,B,Bt,D,x,y);
+            }
+            if (dim == 3)
+            {
+               return internal::OccaPAMassApply3D(D1D,Q1D,ne,B,Bt,D,x,y);
+            }
+            MFEM_ABORT("OCCA PA Mass Apply unknown kernel!");
+         }
+#endif // MFEM_USE_OCCA
```
pazner: This looks duplicated

fem/integ/bilininteg_mass_pa.cpp
```diff
@@ -48,17 +50,28 @@ void MassIntegrator::AssemblePA(const FiniteElementSpace &fes)
...
    }
    int map_type = el.GetMapType();
-   dim = mesh->Dimension();
    ne = fes.GetMesh()->GetNE();
    nq = ir->GetNPoints();
-   geom = mesh->GetGeometricFactors(*ir, GeometricFactors::DETERMINANTS, mt);
-   maps = &el.GetDofToQuad(*ir, DofToQuad::TENSOR);
+   if (stroud)
+   {
+      geom = mesh->GetGeometricFactors(ir->InverseDuffyTrans(dim),
```
pazner: Why is the "pulled back" `ir` passed to the geometric factors here?

fem/integ/bilininteg_diffusion_pa.cpp
```diff
@@ -16,6 +16,15 @@
...
 
+#ifdef NVTX_DEBUG_HPP
+#undef NVTX_COLOR
+#define NVTX_COLOR ::nvtx::kNvidia
+#include NVTX_DEBUG_HPP
+#else
+#define db1(...)
+#define dbg(...)
+#endif
+
```
pazner: Delete

fem/integ/bilininteg_diffusion_pa.cpp
```diff
@@ -68,6 +77,21 @@ void DiffusionIntegrator::AddMultPA(const Vector &x, Vector &y) const
       }
 #endif // MFEM_USE_OCCA
 
+      db1("dofs1D: {}, quad1D: {}", dofs1D, quad1D);
```
pazner: Delete

fem/integ/bilininteg_diffusion_kernels.hpp
```diff
@@ -1213,11 +1643,582 @@ inline void SmemPADiffusionApply3D(const int NE,
...
+            idx = lex_map[a1+1][a2][a3];
+            y(idx,e) += p2 * u;
+
+            MFEM_SYNC_THREAD;
+            idx = lex_map[a1][a2+1][a3];
+            y(idx,e) += p2 * v;
+
+            MFEM_SYNC_THREAD;
+            idx = lex_map[a1][a2][a3+1];
+            y(idx,e) += p2 * w;
```
pazner: Maybe I don't understand, but these `MFEM_SYNC_THREAD` don't look right to me.

pazner (commented):
This is a good and important contribution - thanks @jdongg!

tests/benchmarks/bench_ceed.cpp
jdongg: I've fixed this.

fem/integ/bilininteg_diffusion_kernels.cpp
jdongg: @pazner I think so. `DiffusionIntegrator::GetRule` will return a rule with `Q = P` on triangles and tetrahedrons as long as the Jacobian of the reference to physical mapping is constant. Without these additions, the fallback simplex kernels will be called instead of the shared memory versions.

fem/intrules.cpp
jdongg: Removed.

fem/intrules.cpp
jdongg: Fixed.

fem/intrules.cpp
jdongg: Done.

fem/intrules.cpp
jdongg: Fixed.

fem/intrules.hpp
jdongg: I've added some details on the transformation.  The reason for the non-in-place inverse mapping is because it gets called in the `AssemblePA` routines, which have `const` instances of the integration rules. Maybe there's a better way to do this while maintaining consistency of the forward and inverse transformations.

fem/intrules.hpp
jdongg: Yes, I've done ahead and used `SegmentIntRules` for the (0,0) Gauss-Jacobi rule, then removed `SegmentStroud2IntRules` and `SegmentStroud3IntRules`. The (1,0) and (2,0) rules are just constructed within the calls to `TriangleStroudIntegrationRule` and `TetrahedronStroudIntegrationRule` now.

fem/intrules.hpp
jdongg: Done.

fem/intrules.hpp
jdongg: Done.

fem/intrules.hpp
jdongg: Removed.

fem/intrules.cpp
jdongg: Only tested. The initial guess for Newton has error term $\mathcal{O}(n^{-4})$ when $-1/2 < \alpha,\beta < 1/2$. The authors of [2] (https://doi.org/10.1137/120889873) did some testing and found that $-1 < \alpha,\beta \leqslant 2$ converged fine with the choice of initial guess. I found that it converges even with $\alpha,\beta \approx 4$, but didn't do substantial testing beyond that. 

In any case, I'm going to add some unit tests for the Gauss-Jacobi and Stroud rules to be more rigorous.

fem/integ/bilininteg_diffusion_pa.cpp
jdongg: Done. I think these were added when John was doing some tested and accidentally committed

fem/integ/bilininteg_diffusion_pa.cpp
jdongg: Done.

fem/integ/bilininteg_mass_pa.cpp
jdongg: Removed.

fem/intrules.cpp
jdongg: Fixed.

fem/intrules.hpp
jdongg: I've removed Gauss Jacobi as an option here.

fem/integ/bilininteg_mass_pa.cpp
jdongg: I fixed this, but it actually doesn't seem to make a difference. I also missed that there's a call to `fespace->GetElementRestriction` in `GetGeometricFactors`, so right now this routine is assuming `ElementDofOrdering::NATIVE` ordering rather than `ElementDofOrdering::LEXICOGRAPHIC`. Trying to understand why this doesn't cause issues and the test problems in `miniapps/pa-simplices` still return the correct results....

general/forall.hpp
jdongg: I added new variables `MAX_D1D_SIMPLEX` and `MAX_Q1D_SIMPLEX` which are only used in the simplex integrators and restored the old values of `MAX_D1D` and `MAX_Q1D`.

fem/integ/bilininteg_diffusion_kernels.hpp
jdongg: Removed these. Not sure why I had them there to begin with.

fem/fespace.hpp
jdongg: I just removed `IsBernsteinSimplexSpace` entirely and use `UsesRaggedTensorBasis` now (after moving it inside `FiniteElementSpace`...unless there's a good reason to have an inline version of `UsesRaggedTensorBasis` in `fespace.hpp` outside of the class like `UsesTensorBasis`. Figured we could avoid essentially duplicating the function this way.

pazner:
By the way, I thought a bit more about the design with regard to the `IntegrationRules` class. The Stroud rules aren't a clean fit for this class because, at least the way I see it, the `IntRules` object's interface is just designed for the user to give a geometry and degree of exactness, and receive an arbitrary rule that satisfies those criteria. With the global `IntRules`, there isn't an explicit promise that any specific quadrature rule will be used. It's designed to abstract away the choice of quadrature from the user.

On the other hand, in the case of the Stroud rules, we really do want to request a particular rule. In this case, the only function of `IntRules` would be caching. The interface might be cleaner if there was just a separate mechanism, distinct from `IntegrationRules`, to obtain the Stroud rules. One way to do this would be to implement a free function `const IntegrationRule &GetStroudRule(Geometry::Type geom, int order)` (where `geom` must be a simplex type), which implements an internal caching mechanism.

fem/intrules.hpp
jdongg: I've added an abort if the user tries to call this with `stroud = true`.

mesh/mesh.cpp
jdongg: I think this change and the one in `mesh_extras.hpp` from @camierjs were possibly addressing failed tests when we weren't sure what was causing them. I identified the cause of the failed tests and it's unrelated to these changes AFAIK. If they're necessary for some other reason, perhaps they should go in a separate PR?

fem/fe/fe_base.hpp
jdongg: Done.

fem/fe/fe_base.hpp
jdongg: For now, I've moved these to a derived class `RaggedDofToQuad` which holds the Bernstein-specific fields as it was a minimally invasive solution. I guess a loose end would be to override `DofToQuad::Abs()` for the new fields, but is this necessary right now? Are there any instances in which the Bernstein PA routines currently implemented would actually call `AddAbsMult`?

fem/intrules.cpp
```diff
@@ -1788,6 +2037,39 @@ IntegrationRule *IntegrationRules::TetrahedronIntegrationRule(int Order)
...
+   {
+      SegmentIntegrationRule(RealOrder);
+   }
+
+   IntegrationRule *ir_1_0 = new IntegrationRule;
+   int n = RealOrder/2 + 1;
+   QuadratureFunctions1D::GaussJacobi(n, 1.0, 0.0, ir_1_0);
+
+   IntegrationRule *ir_2_0 = new IntegrationRule;
+   QuadratureFunctions1D::GaussJacobi(n, 2.0, 0.0, ir_2_0);
```
pazner, line 2543: `ir_1_0` and `ir_2_0` don't need to be heap-allocated.

fem/integ/bilininteg_mass_kernels.hpp
```diff
@@ -1115,7 +1914,121 @@ inline void PAMassApply3D(const int NE,
...
+                                                                 BASIS_DIM2D, forward_map2d,
+                                                                 inverse_map2d, forward_map3d, Ba1, Ba2, Ba3, T, D, X, Y, d1d, q1d);
+   });
+
+   // mfem::forall_3D(NE, Q1D, Q1D, Q1D, [=] MFEM_HOST_DEVICE (int e)
+   // {
+   //    internal::SmemPAMassApplyTetrahedron_Element<T_D1D, T_Q1D>(e, NE, BASIS_DIM, BASIS_DIM2D, forward_map2d,
+   //                                                 inverse_map2d, forward_map3d, Ba1, Ba2, Ba3, T, D, X, Y, d1d, q1d);
+   // });
+}
```
pazner, line 2028: Delete comment if not needed

fem/intrules.hpp
```diff
@@ -270,6 +270,15 @@ class IntegrationRule : public Array<IntegrationPoint>
...
    IntegrationRule* ApplyToKnotIntervals(KnotVector const& kv) const;
 
+   /// Duffy Transformation of 2D and 3D tensor product rules of the form
+   /// $X(t) = \sum_{i=1}^{d+1} \lambda_i(t) * x_i$, where $x_i$ are the vertices
+   /// of the simplex and $\lambda_i = t_i * (1-\lambda_1-...-\lambda_{i-1})$, with
+   /// $t$ being the coordinates in the unit square/cube.
+   void DuffyTrans(int dim);
+
+   /// Inverse Duffy Transformation for 2D and 3D tensor product rules.
+   const IntegrationRule InverseDuffyTrans(int dim) const;
```
pazner, line 283: Maybe better if these are both free functions that return IntegrationRule (like InverseDuffyTrans already does).

pazner: Potential changes here should wait until we have a consensus on what the interface should look like for accessing and caching the collapsed rules.

It could make sense to cache both the collapsed and "un-collapsed" Stroud rules, since both are needed in the implementations. So there could be an object that caches and returns both versions.

I know @v-dobrev prefers, for consistency, to just have another global `IntegrationRules` object that stores the Stroud rules (e.g. `extern MFEM_EXPORT IntegrationRules StroudIntRules;` in `intrules.hpp`), I think the use-case might be distinct enough that a different class (with very similar interface) could be better.

fem/integ/bilininteg_mass_kernels.hpp
jdongg: Done.

fem/intrules.cpp
jdongg: Done.

fem/intrules.hpp
jdongg: I suppose if we have another `IntegrationRules` object, that entails having the `stroud` flag as an argument to the constructor rather than to the `IntegrationRules::Get` function? If we view the `IntegrationRules` class as being general and abstracting away the choice of rule from the user, I guess it doesn't make much sense to have Stroud bundled with them.

I personally would be fine with another class like `StroudIntegrationRules`. We could store both the collapsed and cube rules and replace the call to `InverseDuffyTrans` in the `AssemblePA` routines with a separate `Get` routine that returns the cached un-collapsed rule. This shouldn't be much work to do. @v-dobrev @pazner  what are your thoughts?

.gitignore
camierjs: We can also turn it into unit tests.

fem/intrules.hpp
jdongg: I've gone ahead and made a `StroudIntegrationRules` class and a global `StroudIntRules` object. Initially, I had a boolean argument to `StroudIntegrationRules::Get` which would return the pullback of the Stroud rules if the user requested it, the idea being to replace the calls to `InverseDuffyTrans` in `MassIntegrator::AssemblePA` and `DiffusionIntegrator::AssemblePA` with `StroudIntRules.Get(...)`, but I think an issue arises if the user supplies their own rule in the construction of the integrators (e.g. like in the CEED benchmarks with different values of `q`), the pullback needs to match the supplied integration rule and not what's stored in `StroudIntRules`. For now, the pullback that's computed in `AssemblePA` with `InverseDuffyTrans` is kept that way.

fem/intrules.cpp
jdongg: Added unit testing for Gauss-Jacobi rules [here](https://github.com/mfem/mfem/pull/5223/changes/7521efeff9659359c5ff214a1f33c32c19e6753d). Since $\alpha$ and $\beta$ are real-valued, I narrowed it down to $1/4$ increments between $-1/4$ and $4$ for up to 50-point rules. If people have thoughts on what use cases are useful here, please do chime in.

mesh/mesh.cpp
camierjs: I removed them in [87168f5](https://github.com/mfem/mfem/pull/5223/commits/87168f51ab0af955c4b78514bc230aa8de3a8910)

.gitignore
jdongg: This has been rolled into `ex1` for the diffusion integrator, and both mass and diffusion are covered in the new unit tests by @camierjs

fem/intrules.hpp
```diff
@@ -487,12 +489,77 @@ class IntegrationRules
...
 extern MFEM_EXPORT IntegrationRules RefinedIntRules;
 
+/// A global object with all Stroud integration rules (defined in intrules.cpp)
+extern MFEM_EXPORT StroudIntegrationRules StroudIntRules;
+
+/// Duffy Transformation of 2D and 3D tensor product rules of the form
+/// $X(t) = \sum_{i=1}^{d+1} \lambda_i(t) * x_i$, where $x_i$ are the vertices
+/// of the simplex and $\lambda_i = t_i * (1-\lambda_1-...-\lambda_{i-1})$, with
+/// $t$ being the coordinates in the unit square/cube.
+const IntegrationRule DuffyTrans(const IntegrationRule ir, int dim);
```
pazner: Maybe add a comment here that the weights are not modified.

`const` is not needed in the return type. `ir` should be passed as const reference rather than by value.

tests/unit/fem/test_pa_simplices.cpp
```diff
@@ -0,0 +1,131 @@
...
+                     "../../data/square-disc-p3.mesh",
+                     "../../data/periodic-annulus-sector.msh"
+                   };
+      auto extra = { "../../data/star-q2.mesh",
+                     "../../data/star-q3.mesh",
+                     "../../data/inline-quad.mesh",
+                     "../../data/klein-donut.mesh",
+                     "../../data/fichera-quad.mesh",
+                     "../../data/square-disc-p2.mesh",
+                     "../../data/square-disc-p3.mesh",
```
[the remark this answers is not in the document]
camierjs: done in [449c9f5](https://github.com/mfem/mfem/pull/5223/commits/449c9f59031261a5c25d19e7294971e8fd47e72d)

fem/intrules.hpp
```diff
@@ -487,12 +489,75 @@ class IntegrationRules
...
+/// A global object with all Stroud integration rules (defined in intrules.cpp)
+extern MFEM_EXPORT StroudIntegrationRules StroudIntRules;
+
+/// Duffy Transformation of 2D and 3D tensor product rules of the form
+/// $X(t) = \sum_{i=1}^{d+1} \lambda_i(t) * x_i$, where $x_i$ are the vertices
+/// of the simplex and $\lambda_i = t_i * (1-\lambda_1-...-\lambda_{i-1})$, with
+/// $t$ being the coordinates in the unit square/cube. This function is used only
+/// in the partial assembly of Bernstein elements on simplices and does NOT
+/// modify the quadrature weights.
+IntegrationRule DuffyTrans(const IntegrationRule &ir, int dim);
```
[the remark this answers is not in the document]
jdongg: Done.

fem/intrules.hpp
jdongg: This has been fixed.

fem/intrules.cpp
```diff
@@ -433,6 +468,142 @@ class HP_Quadrature1D
...
+      // of Gamma function instead, i.e. Gamma(a)/Gamma(b) = exp(lgamma(a) - lgamma(b))
+      ir->IntPoint(n-i).x = 0.5 * xi + 0.5;
+      ir->IntPoint(n-i).weight = 0.5 * c0 * pow(2.0,
+                                                ab + 1) / ((1.0 - xi*xi)*pp*pp) / pow(2, ab);
+      // map nodes and weights to the interval [0,1]
+   }
+
+#else // MFEM_USE_MPFR is defined
+
+   MFEM_ABORT("MPFR implemented of Gauss-Jacobi quadrature not defined yet")
```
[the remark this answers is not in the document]
jdongg: Done.

fem/intrules.cpp
```diff
@@ -433,6 +468,142 @@ class HP_Quadrature1D
...
+
+         dz = p1/pp;
+#ifdef MFEM_USE_SINGLE
+         if (std::abs(dz) < 1e-7)
+#elif defined MFEM_USE_DOUBLE
+         if (std::abs(dz) < std::numeric_limits<real_t>::epsilon())
+            // this seems to cause trouble if we try std::abs(dz) < 1e-16
+#else
+         MFEM_ABORT("Floating point type undefined");
+         if (std::abs(dz) < 1e-16)
```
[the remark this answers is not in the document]
jdongg: Removed the unnecessary if statement.

fem/intrules.cpp
```diff
@@ -433,6 +468,142 @@ class HP_Quadrature1D
...
+         // the weight (1-x)^alpha * (1+x)^beta to [0,1] as well.
+         ir->IntPoint(0).Set1w(0.5 * x + 0.5,
+                               4.0 * w / ((1.0 - x*x) * (alpha + beta + 2) * (alpha + beta + 2)));
+         return;
+   }
+
+   if (alpha <= -1.0 || beta <= -1.0)
+   {
+      MFEM_ABORT("Gauss-Jacobi quadrature only defined for alpha > -1 and beta > -1");
+   }
```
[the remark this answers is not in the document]
jdongg: Moved check for out-of-range alpha and beta before hard-coded cases.

fem/intrules.cpp
```diff
@@ -2362,6 +2533,195 @@ IntegrationRule *IntegrationRules::CubeIntegrationRule(int Order)
...
+{
+   Array<IntegrationRule *> *ir_array = NULL;
+
+   switch (GeomType)
+   {
+      case Geometry::TRIANGLE:    ir_array = &TriangleStroudIntRules; break;
+      case Geometry::TETRAHEDRON: ir_array = &TetrahedronStroudIntRules; break;
+      case Geometry::INVALID:
+      case Geometry::NUM_GEOMETRIES:
+         MFEM_ABORT("Unknown type of reference element!");
```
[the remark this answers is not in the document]
jdongg: Fixed.

fem/intrules.hpp
```diff
@@ -487,12 +489,75 @@ class IntegrationRules
...
+
+public:
+   /// Sets initial sizes for the integration rule arrays, but rules
+   /// are defined the first time they are requested with the Get method.
+   explicit StroudIntegrationRules();
+
+   /// Returns a Stroud integration rule for given GeomType and Order.
+   const IntegrationRule &Get(int GeomType, int Order);
+
+   void SetOwnRules(int o) { own_rules = o; }
```
pazner: I think we can get rid of this (and the member variable own_rules).

pazner (approved):
Thanks for this very nice (and substantial!) contribution, @jdongg!

Looks good to me - just one more outstanding comment: https://github.com/mfem/mfem/pull/5223#discussion_r3321281353

tzanio:

tzanio:

tzanio:

tzanio:

tzanio (approved):
Thank you for this important contribution, @jdongg ! 🚀 

@pazner, @camierjs and @v-dobrev  -- Thank you for reviewing! 🙏

fem/intrules.cpp
```diff
@@ -433,6 +468,142 @@ class HP_Quadrature1D
...
+      // of Gamma function instead, i.e. Gamma(a)/Gamma(b) = exp(lgamma(a) - lgamma(b))
+      ir->IntPoint(n-i).x = 0.5 * xi + 0.5;
+      ir->IntPoint(n-i).weight = 0.5 * c0 * pow(2.0,
+                                                ab + 1) / ((1.0 - xi*xi)*pp*pp) / pow(2, ab);
+      // map nodes and weights to the interval [0,1]
+   }
+
+#else // MFEM_USE_MPFR is defined
+
+   MFEM_ABORT("MPFR implementation of Gauss-Jacobi quadrature not defined yet");
```
v-dobrev, line 600: This abort causes the unit tests to fail when building with MPFR. It is best to remove the error and just fall back to the non-MPFR implementation. We could issue a warning but I don't think it is that important.

tzanio: ping: @jdongg

jdongg: @v-dobrev I opened a PR here: https://github.com/mfem/mfem/pull/5405 Personally, I think a warning is helpful just to avoid user issues down the line if we're falling back to the non-MPFR implementation.

PR #5224, opened 2026-02-07
Avoid early return in structured OpenMP block

Sbozzolo:
Technically, OpenMP does not allow early returns in structured blocks (like `critical`). I found this because my compiler stopped compiling MFEM with OpenMP with this error:
```
     318    fem/fe/fe_base.cpp: In member function 'void mfem::NodalFiniteEleme
            nt::CreateLexicographicFullMap(const mfem::IntegrationRule&) const'
            :
  >> 319    fem/fe/fe_base.cpp:672:9: error: invalid exit from OpenMP structure
            d block
     320      672 |       { return; }
     321          |         ^~~~~~
```

Luckily, there's a very simple way to rewrite the function so that this does not happen: negating the "if" condition.

### DIFF
--- a/fem/fe/fe_base.cpp
+++ b/fem/fe/fe_base.cpp
@@ -663,58 +663,58 @@ const
    #pragma omp critical (DofToQuad)
 #endif
    {
-      // If the new Dof2Quad is already present, e.g. added in a previous call
-      // or added by another omp thread, return.
-      if (DofToQuad::SearchArray(dof2quad_array, ir,
-                                 DofToQuad::LEXICOGRAPHIC_FULL))
-      { return; }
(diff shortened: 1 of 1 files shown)


v-dobrev:
I think this is duplicate of #5222.

Sbozzolo:
> I think this is duplicate of #5222.

Ah yes! I didn't see that!

If you prefer merging this PR, the PR to fix compilation with spack with `+openmp` is already ready :) 

https://github.com/spack/spack-packages/pull/3292

v-dobrev:
I'll close this in favor of #5222 which was created first and has essentially the same fix.

PR #5235, opened 2026-02-13
Dgdiffusion face coefficient

arotem3:
# DG Diffusion Now Correctly Evaluates Piecewise Coefficients.

**Previous Error:** DG Diffusion was evaluating coefficients using FaceQuadratureSpace. This lead to an error when attempting to evaluate piecewise coefficients because interior faces are not labeled with attributes.

**Fix:** The coefficients are now evaluated via the element quadrature points from the left and right elements on the face. This correctly handles piecewise coefficients.

### DIFF
--- a/fem/coefficient.cpp
+++ b/fem/coefficient.cpp
@@ -2195,6 +2195,8 @@ void CoefficientVector::SetConstant(const DenseSymmetricMatrix &constant)
 
 int CoefficientVector::GetVDim() const { return vdim; }
 
+void CoefficientVector::SetVDim(int vdim_) { vdim = vdim_; }
+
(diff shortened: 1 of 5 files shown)


arotem3:
I got a segfault when using this code with MPI during `SetupPA` which does not occur in serial or CUDA. @pazner is it possible that when we evaluate the coefficients there is a communication step missing?

arotem3:
Narrowed down error `bilininteg_dgdiffusion_pa.cpp:589`  (and similarly at `:958`) there is a reference to `T.elem2` which, I believe, may be null for interior faces (but on the boundary of the mesh partition). I don't know how to fix this.

PR #5238, opened 2026-02-18
Fix bug in ParL2FaceRestriction with vdim > 1

pazner:
The layout of the FaceNbrData vector was not handled properly.

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#5238](https://github.com/mfem/mfem/pull/5238) | @pazner | @tzanio | @najlkin + @dylan-copeland | 2/24/26 | 3/17/26 | 3/22/26 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
      - [ ] Add the example code to the `ALL_EXE_SRCS` variable.
      - [ ] Make sure `THIS_TEST_OPTIONS` is set correctly for the new example.
   - [ ] List the new example in `doc/CodeDocumentation.dox`.
   - [ ] If new examples directory (e.g.`examples/pumi`), list it in `doc/CodeDocumentation.conf.in`
[...]

PR #5258, opened 2026-03-04
Fix incorrect (+) state unpacking in ParL2FaceRestriction::DoubleValuedConformingMult

MTCam:
ParL2FaceRestriction::DoubleValuedConformingMult

The shared-face branch assumed FaceNbrData is arranged as (scalar_dof, component), but it is actually equation-blocked within each nbr element block.

This resulted in incorrect (+) state on shared faces after applying the restriction operator, and in some cases an incorrect reshaping of the exchange buffer caused memory faults when using device=hip.

Fixes:
1. Use scalar nsdofs = GetFaceNbrVSize() / vdim when reshaping the exchange buffer (*not* GetFaceNbrVSize).
2. Read shared face values using eq-blocked stride instead of assuming (sdof, component) interleaving.

Results:
 - Correct (+) state values for shared faces independent of MPI rank count
 - Resolved device memory fault associated with incorrect Reshape parameters


v-dobrev:
Hi @MTCam,

Unfortunately, this PR slipped through the cracks.

I see that you closed it -- was the fix you proposed not correct or was there some other reason you closed it?

Thanks,
Veselin

MTCam:
> Hi @MTCam,
> 
> Unfortunately, this PR slipped through the cracks.
> 
> I see that you closed it -- was the fix you proposed not correct or was there some other reason you closed it?
> 
> Thanks, Veselin

Hi @v-dobrev,

Thanks for pinging me about this.  The problem this PR was intending to address was fixed by PR: #5238 .   I was adding this PR and an associated issue when I realized that #5238 was already addressing the problem and was about to merge.

Cheers,
-Mike

PR #5271, opened 2026-03-13
Proteus option for make-based build system

tomstitt:


config/defaults.mk
```diff
@@ -640,6 +641,11 @@ endif
 ENZYME_OPT = -fplugin=$(ENZYME_PLUGIN)
 ENZYME_LIB =
 
+# Protues configuration
+PROTEUS_DIR = @MFEM_DIR@/../proteus
```
johnbowen42, line 645: I think this is ok but we might want to add a note to building instructions about relative paths

johnbowen42 (approved):
LGTM, we will have to change this if we ever add proteus as a sub module

PR #5273, opened 2026-03-17
Add example for the Signorini contact problem using the Proximal Galerkin method

yym1928:
## Overview
This PR adds a new example demonstrating the solution of the Signorini contact problem using the Proximal Galerkin Method. The implementation uses a mixed saddle-point formulation, where auxiliary variables are defined only on the boundary of the domain (contact surface). To support this formulation, additional operators are implemented to couple variables defined on the parent mesh and the boundary submesh, allowing the resulting system to be assembled and solved as a coupled system.

## Key Features
- Showcase the direct use of SubMesh features to construct a monolithic system matrix.
- Custom mixed bilinear form for assembling the coupling matrix between the parent mesh and the boundary submesh.
- Custom integrator for terms of the form (u · n~, v).

_This work originated within the Center for Information Geometric Mechanics and Optimization (CIGMO), a PSAAP-IV Focused Investigatory Center supported by the U.S. Department of Energy, National Nuclear Security Administration under Award Number DE-NA0004261._


examples/ex42.cpp
```diff
@@ -0,0 +1,385 @@
+//                        MFEM Example 42 - serial version
+//
+// Compile with: make ex42
+//
+// Sample runs: ./ex42
+//
+// Description: This example code demostrates the use of MFEM to
+//              solve a bound-constrained energy minimization problem
+//
+//                 minimize ||∇u||² subject to u ≥ ϕ on Γ in H¹₀(Ω).
```
brendankeith: This is not the correct energy functional or constraint

examples/ex42.cpp
```diff
@@ -0,0 +1,385 @@
...
+//              The problem is discretized and solved using the proximal
+//              Galerkin finite element method, introduced by Keith and
+//              Surowiec [1].
+//
+//              This example highlights the use of MFEM's SubMesh and
+//              MixedBilinearForm features to construct a coupled
+//              nonlinear system involving variables defined separately
+//              over the domain and the boundary submesh.
+//
+// [1] Keith, B. and Surowiec, T. (2023) Proximal Galerkin: A structure-
```
brendankeith: This is not the correct reference. Remove this reference or provide a different one

PR #5298, opened 2026-04-10
Parallel LOR reconstruction miniapp.

skang67:
This PR adds a parallel version of the LOR reconstruction method, implemented in `p-reconstruction`. The convergence test script `run_convergence_study.py` is also extended to support both serial and parallel executables.

### Usage

```bash
# Run
mpirun -np 4 ./p-reconstruction -ho 3 -r 2 -m LOR_reconstruction -f exponential

# Convergence study
python run_convergence_study.py --executable ./p-reconstruction --np 4 -m LOR_reconstruction

### CODE REVIEW

File: miniapps/tools/p-reconstruction.cpp [hunk 1]
```diff
@@ -0,0 +1,228 @@
...
+#include "mfem.hpp"
+#include <limits>
+#include <string>
+#include <unordered_map>
+
+using namespace mfem;
+
+using profile_t = std::function<real_t(const Vector&,const Vector&)>;
+
+void L2Reconstruction(const GridFunction& src, GridFunction& dst);
```
cjvogl, line 21: The `L2Reconstruction` name should not be used anymore: instead, `ElementAverageReconstruction` should be used. That said, I wouldn't include the element average reconstruction option in the parallel code b/c it does not support parallel meshes / FE spaces.

miniapps/tools/p-reconstruction.cpp
```diff
@@ -0,0 +1,228 @@
...
+   std::string field_profiles_help = "Profile of field to be reconstructed:";
+   for (const auto& [name, _] : field_profiles)
+   {
+      field_profiles_help += "\n\t" + name;
+   }
+
+   // Parse options
+   OptionsParser args(argc, argv);
+   args.AddOption(&reconstruction_method, "-m", "--method",
+                  "Reconstruction method: \"element_average_reconstruction\" or \"LOR_reconstruction\".");
```
cjvogl, line 56: I wouldn't include the element average reconstruction in the parallel code

miniapps/tools/p-reconstruction.cpp
```diff
@@ -0,0 +1,228 @@
...
+   args.AddOption(&field_ky, "-fy", "--field-ky",
+                  "Value of ky in field profile.");
+   args.AddOption(&visualization, "-vis", "--visualization", "-no-vis",
+                  "--no-visualization", "Enable or disable GLVis visualization.");
+   args.AddOption(&visport, "-visp", "--visualization-port",
+                  "Use custom port number for GLVis.");
+   args.ParseCheck();
+
+   MFEM_VERIFY((reconstruction_method=="LOR_reconstruction"),
+               "Currently only LOR_reconstruction is supported.");
```
cjvogl, line 74: Once the element average reconstruction code is removed, this verify statement can also be removed.

miniapps/tools/p-reconstruction.cpp
```diff
@@ -0,0 +1,228 @@
...
+      serial_mesh_im.UniformRefinement();
+   }
+   ParMesh pmesh_im(MPI_COMM_WORLD, serial_mesh_im);
+   serial_mesh_im.Clear();
+   ParMesh pmesh = ParMesh::MakeRefined(pmesh_im, lref, BasisType::ClosedUniform);
+   int dim = pmesh.Dimension();
+
+   L2_FECollection fec_lo(order_lo, dim);
+   L2_FECollection fec_hi(order_ho, dim);
+   H1_FECollection fec_im(order_im, dim);
```
cjvogl, line 110: Because the intermediate finite element space is L2 in the serial version, I suggest the same be used here for consistency.

miniapps/tools/p-reconstruction.cpp
```diff
@@ -0,0 +1,228 @@
...
+   M_im.Finalize();
+
+   ParBilinearForm M_hi(&pfespace_hi);
+   M_hi.AddDomainIntegrator(new MassIntegrator);
+   M_hi.Assemble();
+   M_hi.Finalize();
+
+   // Set up the right-hand side vector for the exact solution
+   ParLinearForm b_lo(&pfespace_lo);
+   DomainLFIntegrator *lf_integ = new DomainLFIntegrator(u_function_exact);
```
cjvogl, line 137: Managed/smart pointers should be used instead of raw pointers, i.e., use `std::unique_ptr` here.

miniapps/tools/run_convergence_study.py
```diff
@@ -1,4 +1,12 @@
 #!/usr/bin/env python3
+#
+# Example usage:
+#   Serial:   
+#           python run_convergence_study.py -m element_average_reconstruction
+#   Parallel:
+#           python run_convergence_study.py --executable ./p-reconstruction --np 4 -m LOR_reconstruction
```
cjvogl, line 7: Might consider inferring the executable from whether or not `--np` is provided.  In other words, assume the executable is `reconstruction` if `--np` is not provided, and assume `p-reconstruction` otherwise.

miniapps/tools/run_convergence_study.py
```diff
@@ -229,10 +244,21 @@ def main():
         default='element_average_reconstruction',
         help='Method passed to ./reconstruction via -m (default: element_average_reconstruction).'
     )
+    parser.add_argument(
+        '--executable',
+        default='./reconstruction',
+        help='Path to the reconstruction executable (default: ./reconstruction).'
```
cjvogl, line 250: For what it's worth, you can specify for the `ArgumentParser` to show the default values by passing an argument to it during construction, i.e.,
``` python
parser = argparse.ArgumentParser(description="Run or print reconstruction convergence study results.",
  formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
This avoids needing to remember to change the help string if the default value is changed.

cjvogl (changes requested):
This is looking good so far. I've added some inline comments, most of which seem to come from `p-reconstruction.cpp` being built off a version of `reconstruction.cpp` prior to changes made in the last PR.

More generally, we can consider the naming of `p-reconstruction` vs `preconstruction` vs `reconstructionp`.  I see that the `lor-transfer`, `plor-transfer` pair follows the standard of prepending a "p"... but I also see the advantage of adding the "p" to the end so the two files show up next to each other in an alphabetical listing. We can discuss in real time during a meeting.

miniapps/tools/p-reconstruction.cpp
cjvogl: Oops, you can disregard this as I now see the serial version uses H1 for the intermediate finite element space.  I'll leave this thread unresolved for now, though, as a reminder that it would be worth comparing the performance of H1 vs L2... I suspect L2 will be faster for larger problems in parallel.

miniapps/tools/p-reconstruction.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+   // STEP 1: L2 projection onto u_lo
+   std::unique_ptr<HypreParMatrix> M_par_lo(M_lo.ParallelAssemble());
+   CGSolver cg(MPI_COMM_WORLD);
+   cg.SetOperator(*M_par_lo);
+   cg.SetRelTol(1e-16);
+   cg.SetMaxIter(1000);
+   cg.SetPrintLevel(0);
+   u_lo = 0.0;
+   cg.Mult(b_lo, u_lo); // Solve: M * u_lo = b_lo
```
cjvogl, line 158: We should discuss how to make this into a direct solve for large parallel problems.  Perhaps for a future PR.

miniapps/tools/p-reconstruction.cpp
```diff
@@ -0,0 +1,228 @@
...
+   u_lo.SetTrueVector();
+   u_lo.SetFromTrueVector();
+
```
cjvogl, line 161: Sorry for yet another comment, but it's not clear to me what these lines are doing: can you elaborate (either here or in a meeting)?

miniapps/tools/p-reconstruction.cpp
skang67: I have compared the use of H1_FECollection and L2_FECollection for the intermediate space; Contrary to our intuition, H1 turns out to be faster with speedup 3x for serial and 2x for parallel. I will investigate what causes this.

miniapps/tools/p-reconstruction.cpp
cjvogl: Noting this was addressed in [9862291](https://github.com/mfem/mfem/pull/5298/commits/9862291182c04ab6a88cc9e4de677cace8bd83ea)

miniapps/tools/p-reconstruction.cpp
cjvogl: Noting this was addressed in [9862291](https://github.com/mfem/mfem/pull/5298/commits/9862291182c04ab6a88cc9e4de677cace8bd83ea)

miniapps/tools/p-reconstruction.cpp
cjvogl: Noting this was addressed (possibly in [9862291](https://github.com/mfem/mfem/pull/5298/commits/9862291182c04ab6a88cc9e4de677cace8bd83ea))

miniapps/tools/p-reconstruction.cpp
cjvogl: Interesting finding!  Noting here that we hypothesized that L2 could be slower because the more DOFs... but we can revisit this in the future.

miniapps/tools/p-reconstruction.cpp
cjvogl: Nothing this has been addressed across various commits.

miniapps/tools/run_convergence_study.py
cjvogl: This is addressed in [e66c2bc](https://github.com/mfem/mfem/pull/5298/commits/e66c2bc853230ee942c85b8acfdf4146431be200)

miniapps/tools/p-reconstruction.cpp
cjvogl: Decided this will be in a future PR.

cjvogl (approved):
This looks good to me: thanks @skang67 for addressing all the comments!

PR #5193, opened 2026-01-13
Eigen solver interface

SchmidtMat:
This class provides a common interface for all existing and future generalized eigen solver implementations.
- Base class
- Hypre LOBPCG
- Slepc

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#5193](https://github.com/mfem/mfem/pull/5193) | @SchmidtMat | @tzanio | @tzanio + @mlstowell + @dylan-copeland | 1/26/26 | ⌛due 2/9/26 | ⌛due 2/16/26 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
      - [ ] Add the example code to the `ALL_EXE_SRCS` variable.
      - [ ] Make sure `THIS_TEST_OPTIONS` is set correctly for the new example.
   - [ ] List the new example in `doc/CodeDocumentation.dox`.
[...]

PR #5315, opened 2026-04-23
Common eigensolver interface

mlstowell:
This is a draft of an eigensolver interface which should encompass real or complex-valued, standard or generalized eigenvalue problems.

This is a possible, more general, alternative to the interface proposed in PR #5193. Where as, that PR focuses on designing wrappers for particular solver packages this PR focuses on creating common interfaces for the different types of eigenvalue equations. There are pros and cons for either choice.

### DIFF
--- a/examples/ex11p.cpp
+++ b/examples/ex11p.cpp
@@ -321,8 +321,8 @@ int main(int argc, char *argv[])
 
       for (int i=0; i<nev; i++)
       {
-         // convert eigenvector from HypreParVector to ParGridFunction
-         x = lobpcg->GetEigenvector(i);
+         // convert eigenvector from Vector to ParGridFunction
+         x.Distribute(lobpcg->GetEigenvector(i));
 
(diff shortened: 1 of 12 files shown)


lindsayad (commented):
Would it be appropriate for these base classes to inherit from `Solver`? Ultimately, `HypreAME` and `HypreLOBPCG` both define `SetOperator` which is the lone virtual interface in `Solver`. It would help our MOOSE wrapping of these classes if there was the common base class

v-dobrev:
> Would it be appropriate for these base classes to inherit from `Solver`? Ultimately, `HypreAME` and `HypreLOBPCG` both define `SetOperator` which is the lone virtual interface in `Solver`. It would help our MOOSE wrapping of these classes if there was the common base class

`Solver` inherits from `Operator` and I'm not sure eigensolvers make sense as `Operator`s. Maybe a separate `Eigensolver` base class, that defines abstract virtual method `SetOperator`, makes more sense?

PR #5346, opened 2026-05-26
Preemptively delete rownnz if ownership flags set to -1

nmnobre:
Another attempt at #5200. This time, I'm taking the bull by the horns: the issue was always on deallocation, so that's where I'm trying to tackle the problem now. Hopefully the intent is also clearer and the implementation more targeted, which should make the patch easier to understand and less prone to breaking things.

Speaking of breaking things, I did manage to reproduce both failures mentioned [here](https://github.com/mfem/mfem/pull/5327#issuecomment-4451953205) on my CUDA system. I can confirm all tests pass with this new patch though.

For your attention @tzanio, @v-dobrev, @najlkin and @victorapm.

| [#5346](https://github.com/mfem/mfem/pull/5346) | @nmnobre | @tzanio | @helloworld922 + @camierjs + @najlkin | 5/26/26 | 6/22/26 | 7/1/26 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:
Thanks @nmnobre

tzanio:

najlkin:
I see this bug comes from the fact that older versions of hypre allocated `rownnz` on host independently of the `memory_location` attribute, so this fix makes sense to me. However, the whole approach of changing the memory location from device to host is very error prone 💣 . In MFEM, we care only about I, J and data, which are moved back to host, but the hypre object contains other pointers which are potentially allocated on the device (`rowindices`, `rowvalues`, `big_j`), which are not moved. I see some comments about it in the destructors in hypre from @liruipeng . It would be better to avoid this altogether if possible (?), i.e., let hypre correctly delete the internal device objects by not switching the memory location to host 🤔 .

najlkin:
@v-dobrev , I see you have written most lines of the destructor of `HypreParMatrix`, what do you think? ☝️

v-dobrev:
Hi @nmnobre,

I'm trying to understand the issue you described in #5200. Is there a simple way for me to reproduce it? In particular, can you clarify what version of hypre you are using, how is it built (e.g. I assume CUDA is enabled; what about other options like "mixed-int"?), how is MFEM built, and what MFEM device you are running with (I assume `cuda`?)?

Thanks!

nmnobre:
Hi @v-dobrev, thanks for taking the time to look into this.

> I'm trying to understand the issue you described in #5200. Is there a simple way for me to reproduce it?

Yes, you can try running ` mpirun -n 2 ./ex25p -d cuda {-slu,-mumps,-strumpack}`. Hopefully you have mfem configured with one of those three packages ready to go. The key is to exercise any of the three codepaths doing
```
HypreParMatrix *A = Ah.As<ComplexHypreParMatrix>()->GetSystemMatrix();
[...]
delete A;
```
EDIT: so if you don't have mfem configured with any of those three packages, just extract those two lines out of the ifdef to force their execution and you'll still be able to reproduce the seg fault.

> In particular, can you clarify what version of hypre you are using

The latest version, 3.1.0.

> how is it built (e.g. I assume CUDA is enabled; what about other options like "mixed-int"?)

Yes to both, here's a few of the options used: `--enable-gpu-aware-mpi --with-cuda --with-gpu-arch="90" --with-umpire --enable-mixedint=yes`.

> how is MFEM built, and what MFEM device you are running with (I assume `cuda`?)?

Yes, `-DMFEM_USE_CUDA=YES -DCUDA_ARCH=sm_90`.

v-dobrev:
Hi @nmnobre, @najlkin, @tzanio, @helloworld922,

I created a separate branch, https://github.com/mfem/mfem/tree/hypre-v3-gpu-fix, with a test that reproduces the issue and proposes an alternative solution. The solution is to have MFEM delete the owned host pointers `A->{diag,offd}->{i,j,data}` and let hypre destroy the rest of the struct with the memory location set to device. I think this approach will be less likely to require changes in the future due to modifications in hypre. This is also in line with what @najlkin suggested above.

If there's consensus that the alternative solution is better, we can merge the branch here, or create a separate PR from it.

Let me know what you think.

nmnobre:
> I created a separate branch, https://github.com/mfem/mfem/tree/hypre-v3-gpu-fix, with a test that reproduces the issue and proposes an alternative solution.

I think adding the unit test is a no-brainer, so I've cherry-picked that here.

> The solution is to have MFEM delete the owned host pointers `A->{diag,offd}->{i,j,data}` and let hypre destroy the rest of the struct with the memory location set to device.

That works. :) We're effectively swapping the pointers twice in close succession now. I think we can make that more obvious with this change instead, which is just a slight tweak of what you suggested (I'll probably also use `hypre_CSRMatrix{I,J,Data,OwnsData}` for consistency within this function at least):

```diff
index 64db7f090b..7f59a0664d 100644
--- a/linalg/hypre.cpp
+++ b/linalg/hypre.cpp
@@ -2882,14 +2882,33 @@ void HypreParMatrix::Destroy()
 
       MemoryClass mc = (diagOwner == -1 || offdOwner == -1) ?
                        Device::GetHostMemoryClass() : GetHypreMemoryClass();
-      Write(mc, diagOwner < 0, offdOwner <0);
+      Write(mc, diagOwner < 0, offdOwner < 0);
+      if (diagOwner == -1)
+      {
+         mfem_hypre_TFree_host(A->diag->i);
+         if (A->diag->owns_data)
+         {
+            mfem_hypre_TFree_host(A->diag->j);
+            mfem_hypre_TFree_host(A->diag->data);
+         }
+      }
+      if (offdOwner == -1)
+      {
+         mfem_hypre_TFree_host(A->offd->i);
+         if (A->offd->owns_data)
+         {
+            mfem_hypre_TFree_host(A->offd->j);
+            mfem_hypre_TFree_host(A->offd->data);
+         }
+      }
+      Write(GetHypreMemoryClass(), diagOwner < 0, offdOwner < 0);
    }
 #endif
 
    mem_diag.I.Delete();
    mem_diag.J.Delete();
    mem_diag.data.Delete();
-   if (diagOwner >= 0)
+   if (diagOwner >= -1)
    {
       hypre_CSRMatrixI(A->diag) = NULL;
       hypre_CSRMatrixJ(A->diag) = NULL;
@@ -2898,7 +2917,7 @@ void HypreParMatrix::Destroy()
    mem_offd.I.Delete();
    mem_offd.J.Delete();
    mem_offd.data.Delete();
-   if (offdOwner >= 0)
+   if (offdOwner >= -1)
    {
       hypre_CSRMatrixI(A->offd) = NULL;
       hypre_CSRMatrixJ(A->offd) = NULL;
```

najlkin:
@nmnobre , that is nicely concise, but I think you forgot about the `#ifdef` and the `if` block. The deallocation should be modified only for the pathway of GPU-enabled hypre.

nmnobre:
> @nmnobre , that is nicely concise, but I think you forgot about the `#ifdef` and the `if` block. The deallocation should be modified only for the pathway of GPU-enabled hypre.

Hmmm, did I? What I did forget is that `ParCSROwner` may be false, so I'd have to change `if (diagOwner >= -1)` to `if (diagOwner >= 0 || (diagOwner == -1 && ParCSROwner))`, but I'm not sure that would be more readable than what @v-dobrev suggested, despite getting rid of the redundancy... I've pushed @v-dobrev's suggestion and then just updated the comment explaining what's happening so it matches the new implementation and then used `hypre_CSRMatrix{I,J,Data,OwnsData}` where applicable.
EDIT: I've added a different suggestion to avoid the redundancy by just explicitly setting the memory location.

We did forget to fence the new unit test with `#ifdef MFEM_USE_MPI`, so I've also fixed that.

linalg/hypre.cpp
```diff
@@ -2882,7 +2882,31 @@ void HypreParMatrix::Destroy()
...
+      {
+         // Note: mfem_hypre_TFree_host() sets the pointer to NULL too.
+         mfem_hypre_TFree_host(hypre_CSRMatrixI(A->diag));
+         if (hypre_CSRMatrixOwnsData(A->diag))
+         {
+            mfem_hypre_TFree_host(hypre_CSRMatrixJ(A->diag));
+            mfem_hypre_TFree_host(hypre_CSRMatrixData(A->diag));
+         }
+         hypre_CSRMatrixMemoryLocation(A->diag) =
+            GetHypreParMatrixMemoryLocation(GetHypreMemoryClass());
```
najlkin: This needs version fencing as in `Write`.

nmnobre: I was hoping we could start forgoing checks for a version that is now 7 years old, but I've added the fence. :P

Do we think this is better? Or would we prefer to just call `Write()` and repeat the code to set the pointers to null like it was? I think this version shows intent more explicitly, but it's now the same no. of lines anyway, so it also feels a bit pointless.

nmnobre: EDIT, I've switched to `GetHypreMemoryLocation()` so it's cleaner.

v-dobrev (approved):
Looks good to me. Thanks @nmnobre!

tzanio:
@najlkin and @victorapm -- do you have any comments?

linalg/hypre.cpp
```diff
@@ -2882,7 +2882,33 @@ void HypreParMatrix::Destroy()
 
       MemoryClass mc = (diagOwner == -1 || offdOwner == -1) ?
                        Device::GetHostMemoryClass() : GetHypreMemoryClass();
-      Write(mc, diagOwner < 0, offdOwner <0);
+      Write(mc, diagOwner < 0, offdOwner < 0);
+      if (diagOwner == -1)
+      {
+         // Note: mfem_hypre_TFree_host() sets the pointer to NULL too.
```
najlkin:
```suggestion
         // Note: mfem_hypre_TFree_host() sets the pointer to NULL.
```

linalg/hypre.cpp
```diff
@@ -2882,7 +2882,33 @@ void HypreParMatrix::Destroy()
...
+            mfem_hypre_TFree_host(hypre_CSRMatrixJ(A->diag));
+            mfem_hypre_TFree_host(hypre_CSRMatrixData(A->diag));
+         }
+#if MFEM_HYPRE_VERSION >= 21800
+         hypre_CSRMatrixMemoryLocation(A->diag) = GetHypreMemoryLocation();
+#endif
+      }
+      if (offdOwner == -1)
+      {
+         // Note: mfem_hypre_TFree_host() sets the pointer to NULL too.
```
najlkin:
```suggestion
         // Note: mfem_hypre_TFree_host() sets the pointer to NULL.
```

najlkin (approved):
Thanks @nmnobre  and @v-dobrev 👍

victorapm (approved):
Looks good to me. Thanks!

tzanio:

camierjs (approved):

PR #5359, opened 2026-06-10
Don't copy over host/device interface from invalid side

lindsayad:
I'm a little bit surprised that the code was this way without it being intentional, so it's fully possible I'm missing something. But it seems like before this patch we were explicitly testing what side of the base device/host interface was valid and then copying data from the untested side over to the valid side. This bug only triggers when one side is valid and the other is not so maybe users are generally not calling `SyncAlias_` when there is this asymmetry


lindsayad:
Ok I was missing something and the failing sedov test shows it. The method seems premised on the alias flags being the source of truth and not the base flags. This premise is what the sedov tests rely on. My change here goes in the opposite direction of that premise. So the misunderstanding/misuse appears to be on the MOOSE side

PR #5378, opened 2026-06-24
Use a bandmatrix for NURBS interpolations

IdoAkkerman:
This PR creates a band matrix class, that allows easy and efficient use of banded lapack routines. This is particularly useful for NURBS interpolation routines, such as required by LOR. 

Unit tests are provided.

NOTE: A small bug in the DenseMatrix::Det() routine was found. The suggested trivial change (+ instead of -)  seems to fix the issue but might need some scrutiny.


linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,366 @@
...
+
+/// Band matrix storing (2xbandwidth + 1) numbers per row.
+class BandMatrix : public Matrix
+{
+   friend class BandMatrixInverse;
+
+private:
+   Array<real_t> data;
+   int bandwidth, stride;
+   real_t zero = -0.0;
```
dylan-copeland, line 31: Eliminate `zero`?

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,366 @@
...
+
+#include "../config/config.hpp"
+#include "../general/globals.hpp"
+#include "matrix.hpp"
+#include "densemat.hpp"
+
+namespace mfem
+{
+
+/// Band matrix storing (2xbandwidth + 1) numbers per row.
```
dylan-copeland: `2xbandwidth` -> `2 x bandwidth`

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,366 @@
...
+class BandMatrix : public Matrix
+{
+   friend class BandMatrixInverse;
+
+private:
+   Array<real_t> data;
```
dylan-copeland, line 29: Would `Vector` be simpler than `Array<real_t>` or is there an advantage to `Array`?

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,366 @@
...
+   /// Matrix vector multiplication.
+   void Mult(const Vector &x, real_t *y) const;
+
+   /// Matrix vector multiplication.
+   void Mult(const Vector &x, Vector &y) const override;
+
+   /// Matrix vector multiplication with the inverse.
+   /** Uses Inverse() and deletes factorisation afterwards.
+       Unnecesarily inefficient when called multiple times,
+       use Inverse() instead when solving multiple times. */
```
dylan-copeland: Would it be more convenient for the user to have an option for reuse of a stored inverse, rather than calling `Inverse()` and having to manage the `MatrixInverse`? Isn't this already possible with `Invert()` instead of `Inverse()`? If so, this could be mentioned in the documentation.

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,366 @@
...
+   ///  Get the size of the inverse matrix
+   int Size() const { return Width(); }
+
+   /// Factor the current BandMatrix, *a
+   void Factor();
+
+   /// Factor a new BandMatrix of the same size
+   void Factor(const BandMatrix &mat);
+
+   void SetOperator(const Operator &op) override;
```
dylan-copeland, line 334: Document that only operators of type `BandMatrix` are supported?

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,366 @@
...
+   BandMatrixInverse(bool spd_=false) : a(NULL) { Init(0, 0); }
+
+   /** Creates square dense matrix. Computes factorization of mat
+       and stores its factors. */
+   BandMatrixInverse(const BandMatrix &mat);
+
+   /// Same as above but does not factorize the matrix.
+   BandMatrixInverse(const BandMatrix *mat);
+
+   ///  Get the size of the inverse matrix
```
dylan-copeland: Extra space.

linalg/bandmatrix.cpp
```diff
@@ -0,0 +1,579 @@
...
+   height = dm.Height();
+   width = dm.Width();
+   if (bw < 0)
+   {
+      bandwidth = 0;
+      for (int i = 0; i < dm.Height(); i++)
+      {
+         for (int j = 0; j < dm.Width(); j++)
+         {
+            if (dm(i,j)*dm(i,j) > 1e-16)
```
dylan-copeland: Make a `constexpr` in the class for this tolerance 1e-16?

linalg/bandmatrix.cpp
```diff
@@ -0,0 +1,579 @@
...
+   {
+      for (int i = 0; i < height; i++)
+      {
+         dm(i,j) = Elem(i,j);
+      }
+   }
+   return dm;
+}
+
+void BandMatrix::Print(std::ostream &os, int width33_) const
```
dylan-copeland: `width33_` is unused.

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,366 @@
...
+    * @brief Compute the LU factorization of the current matrix
+    *
+    * Factorize the current matrix of size (m x m) using LAPACK
+    *
+    * @param [in] m size of the square matrix
+    * @param [in] TOL optional fuzzy comparison tolerance. Defaults to 0.0.
+    *
+    * @return status set to true if successful, otherwise, false.
+    */
+   bool Factor(int m, real_t TOL = 0.0) override;
```
dylan-copeland, line 284: Upper case `TOL` would fit the MFEM style better as lower case `tol`.

mesh/nurbs.hpp
```diff
@@ -51,20 +51,11 @@ class KnotVector
...
-
-#ifdef MFEM_USE_LAPACK
-   // Data for reusing banded matrix factorization in FindInterpolant().
-   mutable DenseMatrix fact_AB; /// Banded matrix factorization
-   mutable Array<int> fact_ipiv; /// Row pivot indices
-#else
-   mutable DenseMatrix A_coll_inv; /// Collocation matrix inverse
-#endif
-
+   void ComputeDemko(bool force = false) const;
```
dylan-copeland, line 61: Document `force`. Would the argument name `force_recompute` be more descriptive?

linalg/bandmatrix.hpp
IdoAkkerman: On line 360 of the app file there is a return by reference that requires this... Instead of returning a zero we could also return an error, but this is more convenient I think.

linalg/bandmatrix.hpp
IdoAkkerman: Done.

linalg/bandmatrix.hpp
IdoAkkerman: Perhaps... but I just copied stuff from DenseMatrix and SymMatrix, also all the data access routines.
`Memory<real_t> ` is the most appropriate I guess? Or do you get this for free when adapting `Vector`?

At this stage I was just going for the simplest copy past.

linalg/bandmatrix.hpp
IdoAkkerman: I do not really understand the question.
`Invert()` and `Inverse()` are straight analogies of the functionality present in `DenseMatrix`.

On top of that I added the `Solve(x,y)` routine... for convenience.. which would make sense to add to `DenseMatrix` as well  for symmetry (but lets not touch that!).
 

Okay... I think I understand the comment. Yes. use either `Invert()` or `Inverse()`. I modified the comments to point this out.

linalg/bandmatrix.hpp
IdoAkkerman: Done.

linalg/bandmatrix.hpp
IdoAkkerman: Done. But why? Is it MFEM convention to separate the constructors with an additional space?

linalg/bandmatrix.cpp
IdoAkkerman: Done. `tol` could be a parameter int he future..

linalg/bandmatrix.cpp
IdoAkkerman: The Print function was still in debug mode...

linalg/bandmatrix.hpp
IdoAkkerman: Changed to lowercase.

mesh/nurbs.hpp
IdoAkkerman: Done.

linalg/bandmatrix.hpp
dylan-copeland: There are two versions of `BandMatrix::operator()(int i, int j)`, with one being `const`. The `const` version allows access outside the band, returning 0 in that case, whereas the non-`const` one allows access only within the band (for setting values). Would it make sense for the `const` version to return by value (and 0.0 outside the band without needing to define `zero`), with the non-`const` version returning by reference?

linalg/bandmatrix.hpp
dylan-copeland: I would have chosen `Vector` because it's shorter to type, just wondering if there are other pros/cons between these 2 choices. I don't think it really matters.

linalg/bandmatrix.hpp
dylan-copeland: Sorry, my comment was not clear. I just meant there was an extra space between `///` and `Get` (asking to remove extra spaces, not add empty lines).

linalg/bandmatrix.hpp
dylan-copeland: Thanks, but now I notice that `tol` is unused in the function. Can it be removed?

miniapps/nurbs/nurbs_surface.cpp
```diff
@@ -530,8 +531,12 @@ void SurfaceInterpolator::ComputeNURBS(int coordinate,
             (*x[2])[i] = -1.0 + z + s_ij;
          }
 
-         const bool reuse_factorization = j > 0;
-         kv[0].GetInterpolant(x,ugrid[0], reuse_factorization);
+         for (int i = 0; i < 3; i++)
+         {
+            Vector tmp;
```
dylan-copeland: `tmp(*x[i])` should be the same thing in one line.

miniapps/nurbs/nurbs_surface.cpp
```diff
@@ -554,17 +560,20 @@ void SurfaceInterpolator::ComputeNURBS(int coordinate,
             (*x[1])[j] = (*patch)(i,j,k,1);
             (*x[2])[j] = (*patch)(i,j,k,2);
          }
-
-         const bool reuse_factorization = i > 0;
-         kv[1].GetInterpolant(x, ugrid[1], reuse_factorization);
-
+         for (int i = 0; i < 3; i++)
```
dylan-copeland: Variable `i` is already used here.

miniapps/nurbs/nurbs_surface.cpp
```diff
@@ -519,6 +519,7 @@ void SurfaceInterpolator::ComputeNURBS(int coordinate,
       for (int i = 0; i < dim; ++i) { x[i]->SetSize(ncp[0]); }
 
       // Sweep in the first direction
+      MatrixInverse *A_coll_inv = kv[0].GetInverseInterpolationMatrix(ugrid[0]);
```
dylan-copeland, line 522: This file was not added to `miniapps/nurbs/makefile` (before this PR). Could you please add it so that `make` in this directory builds this miniapp?

linalg/bandmatrix.hpp
IdoAkkerman: I guess the return can be by value, I just wanted to be as close as possible to the DenseMatrix case. If you feel the return by value is more appropriate I will change it. Just let me know.

linalg/bandmatrix.hpp
IdoAkkerman: Done

linalg/bandmatrix.hpp
IdoAkkerman: This is possible,however, it is an overloaded the function (hence the override), so I think it might cause confusion.
I have changed the comment to reflect that it is not used.

miniapps/nurbs/nurbs_surface.cpp
IdoAkkerman: Done. FYI: I am far from happy  with how some of there nurbs miniapps are programmed, but that is an aside...

miniapps/nurbs/nurbs_surface.cpp
IdoAkkerman: Changed to j, and merged lines (as before)

miniapps/nurbs/nurbs_surface.cpp
IdoAkkerman: Done?

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,391 @@
...
+
+   /** Creates square dense matrix. Computes factorization of mat
+       and stores its factors. */
+   BandMatrixInverse(const BandMatrix &mat);
+
+   /// Same as above but does not factorize the matrix.
+   BandMatrixInverse(const BandMatrix *mat);
+
+   /// Get the bandwidth of the input matrix
+   int GetBandWidth() const { return bandwidth; } const
```
dylan-copeland: The last `const` can be removed.

linalg/bandmatrix.hpp
dylan-copeland: Thanks for documenting and pointing out the base class. I didn't realize it was `TOL` in the base class, so maybe we should leave it as upper case `TOL`.

linalg/bandmatrix.hpp
dylan-copeland: It would make more sense to me to return by value in the `const` version, also for `DenseMatrix` (which we cannot change), but I suppose we can leave it as is, to be consistent with `DenseMatrix`.

linalg/densemat.cpp
```diff
@@ -3476,7 +3476,7 @@ real_t LUFactors::Det(int m) const
    real_t det = 1.0;
    for (int i=0; i<m; i++)
    {
-      if (ipiv[i] != i - ipiv_base)
+      if (ipiv[i] != i + ipiv_base)
```
dylan-copeland, line 3479: Also line 3439 `ipiv[i] = piv + 1;` could use `ipiv_base` instead of the hard-coded `1`.

linalg/bandmatrix.hpp
IdoAkkerman: Done.

linalg/densemat.cpp
IdoAkkerman: I will try this. But will pay extra attention to reg testing.

linalg/bandmatrix.hpp
IdoAkkerman: Changed back to tol.

linalg/bandmatrix.hpp
IdoAkkerman: I'll keep it consistent

linalg/densemat.cpp
IdoAkkerman: i think it is fine. However the `merge master` created a small lapack-- ncmesh glitch (see issue #5466 )

linalg/matrix.hpp
```diff
@@ -67,6 +67,13 @@ class MatrixInverse : public Solver
    /// Creates approximation of the inverse of square matrix
    MatrixInverse(const Matrix &mat)
       : Solver(mat.height, mat.width) { }
+
+   /// Compute the determinant
```
dylan-copeland, line 71: This comment could clarify whether the determinant is for the matrix inverse or the original matrix.

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,401 @@
...
+       When solving multiple times, please consider using Inverse()
+       and reuse the factorisation, or Invert() which
+       computes an inplace inverse.*/
+   void Solve(const Vector &x, Vector &y) const;
+
+   /// Returns a pointer to the matrix inverse.
+   /// Caller gets ownership.
+   MatrixInverse *Inverse() const override;
+
+   /// Returns a pointer to (an approximation) of the matrix inverse.
```
dylan-copeland, line 206: `of` should be in the parentheses.

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,401 @@
...
+       with a reduced bandwidth. The bandwidth is the minimum bandwidth
+       required to achieve the specified tolerance, when computing the
+       Frobenius norm of inv(approx(A))*A - I. If @a bw is specified this is the
+       bandwidth used, the result is compared with the tolerance.
+       Note: the INPUT matrix of the factorisation is reduced in bandwidth.*/
+   MatrixInverse *Inverse(real_t tol, int &bw) const;
+
+   MatrixInverse *Inverse(real_t tol) const
+   {
+      int bw;
```
dylan-copeland, line 218: Initialize `bw = -1`?

linalg/kernels.hpp
```diff
@@ -185,6 +185,44 @@ void Mult(const int height, const int width, const TA *data, const TX *x, TY *y)
    }
 }
 
+/** @brief Matrix vector multiplication: y = A x, where the matrix A is of size
+    @a height x @a width and @a bandwidth with given @a data, while @a x and @a y specify the
```
dylan-copeland, line 189: Long line.

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,401 @@
...
+      return Inverse(tol, bw);
+   }
+
+   /// Replaces the current matrix with its inverse
+   /** If @a tol is negative the exact inverse is computed, oteherwise the
+       resulting matrix has the minimum bandwidth to achieve the specified
+       tolerance, when computing the Frobenius norm of approx(inv(A))*A - I.
+       If @a bw is specified this is thebandwidth used, the result is compared
```
dylan-copeland, line 226: `the bandwidth`

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,401 @@
...
+   int bandwidth = -1;
+   BandLUFactors * factors = nullptr;
+
+   void Init(int m, int bw);
+   bool own_data = false;
+public:
+   /// Default constructor.
+   BandMatrixInverse(bool spd_=false) : a(NULL) { Init(0, 0); }
+
+   /** Creates square dense matrix. Computes factorization of mat
```
dylan-copeland, line 313: Is it correct that this `creates square dense matrix`?

linalg/bandmatrix.hpp
```diff
@@ -0,0 +1,401 @@
...
+      return Inverse(tol, bw);
+   }
+
+   /// Replaces the current matrix with its inverse
+   /** If @a tol is negative the exact inverse is computed, oteherwise the
+       resulting matrix has the minimum bandwidth to achieve the specified
+       tolerance, when computing the Frobenius norm of approx(inv(A))*A - I.
+       If @a bw is specified this is thebandwidth used, the result is compared
+       with the tolerance. Note that in general the inverse is fully dense.
+       Note: the OUTPUT matrix of the factorisation is reduced in bandwidth.*/
```
dylan-copeland, line 228: If I understand correctly, the only scalable way to solve a system with a band matrix is to create a `BandMatrix` and use `BandMatrix::Solve` to solve with the LU factorization, assuming MFEM is built with LAPACK. There are several other functions and code paths that use dense matrix inversion, which is not scalable. The documentation should try to make it as clear as possible whether each function uses a dense matrix or banded LU factorization. Here it seems clear that a dense matrix is involved, but other places are not as clear. Functions that use a dense matrix should have a warning about scalability. The goal should be for a user to understand easily what to use to solve a system without dense matrix inversion.

mesh/nurbs.hpp
```diff
@@ -51,20 +51,14 @@ class KnotVector
...
-   // Data for reusing banded matrix factorization in FindInterpolant().
-   mutable DenseMatrix fact_AB; /// Banded matrix factorization
-   mutable Array<int> fact_ipiv; /// Row pivot indices
-#else
-   mutable DenseMatrix A_coll_inv; /// Collocation matrix inverse
-#endif
-
+   /** @brief Checks if demko vector is already defined. If vector has the
+       correct size the computation is skipped, unless @a force is set to true.
+       in that case the computation is done regardless.*/
```
dylan-copeland, line 60: Capitalize `In`.

mesh/nurbs.hpp
```diff
@@ -237,12 +231,63 @@ class KnotVector
...
+
+   /** @anchor mfem_KnotVector_interpolation
+       @name Methods for interpolating NURBS curves.
+
+       These methods are intended to be used only once. The factorization of
+       the underlying matrix is build and deleted in each routine. When
+       multiple calls to this routines are made consider using
+       GetInterpolationMatrix() or GetInverseInterpolationMatrix() instead.
+
+       For the knot location one can use  for instance GetBotella, GetDemko or
```
dylan-copeland, line 249: Extra space. How can the point type be chosen here?

miniapps/nurbs/nurbs_surface.cpp
dylan-copeland, line 522: This interface seems to imply that a dense matrix inverse is computed here, but it should be using an LU factorization to solve a system with each `Mult` call. Is this correct, assuming MFEM is built with LAPACK? If so, can we change the interface to make this less confusing? From a miniapp like this, a high-level interface might be something like `kv[0].FactorizeInterpolationMatrix(...)` and `kv[0].Interpolate(...)` instead of `A_coll_inv->Mult()`. To make it even simpler, maybe the first call to `kv[0].Interpolate(...)` can do the factorization and set a flag so that subsequent calls reuse it. Do you agree that would be more readable, and can we make such an interface?

linalg/matrix.hpp
IdoAkkerman: Clarified it is the Determinant of the original matrix

linalg/bandmatrix.hpp
IdoAkkerman: Fixed

linalg/bandmatrix.hpp
IdoAkkerman: Done

linalg/kernels.hpp
IdoAkkerman: Fixed

linalg/bandmatrix.hpp
IdoAkkerman: Done

linalg/bandmatrix.hpp
IdoAkkerman: Indeed that is incorrect. Corrected the comments.

linalg/bandmatrix.hpp
IdoAkkerman: >  Here it seems clear that a dense matrix is involved, but other places are not as clear. Functions that use a dense matrix should have a warning about scalability.

I do not understand this statement?

As the inversion is in place the result is a (technically) a `BandMatrix`. Depending on the input the inverse might have full bandwidth (end effectively be a dense matrix). The routine, after some redirects, uses the same routine to solve, using the factorisation.

Note: Moreover `BandMatrix::Solve`  might be a bad option for scalability. It computes the factorization every time it is called.

mesh/nurbs.hpp
IdoAkkerman: Done

mesh/nurbs.hpp
IdoAkkerman: Removed the space.

Points are chosen by setting the vector u.

miniapps/nurbs/nurbs_surface.cpp
IdoAkkerman: I am a but confused by this comment and the previous.. but perhaps I think it stems for a misunderstanding how the inversion  of bandmatrix is implemented.

The `BandMatrix`  member  `Inverse() ` provide an abstract  `MatrixInverse *` , but it should be noted that the actual object is a `BandMatrixInverse` as defined in bandmatrix.hpp. `BandMatrixInverse` is a wrapper around the scalable LAPACK routines.

IdoAkkerman:
@dylan-copeland The name of the branch is also a bit confusing... apologies

miniapps/nurbs/nurbs_surface.cpp
dylan-copeland: For a user reading the documentation in `linalg/bandmatrix.hpp` and `nurbs.hpp` and trying to implement usage like `miniapps/nurbs/nurbs_surface.cpp`, some things in the interface are unclear. Some important questions to clarify for the user are 

1. Is a dense matrix inversion or band matrix factorization being used?
2. Is the inversion or factorization being reused when possible or recomputed wastefully?

Clearly `BandMatrix` is stored efficiently as a band matrix, and one would expect the inverse to be a band factorization, but `BandMatrix::Inverse()` computes a band inverse only if MFEM is built with lapack (otherwise `DenseMatrixInverse`). Also, `MatrixInverse *KnotVector::GetInverseInterpolationMatrix` is ambiguous, since the returned `MatrixInverse` could be dense or band. One possible way to make it clearer would be making the inverse or solver function names contain the words "Band" or "Dense", and the "Band" version could fail with an error when not built with lapack. There may be better ways, but that is one idea for example. For the question of reuse, your current approach of returning a `MatrixInverse` makes it clear that the same inverse is being reused, which is fine if the user knows whether the inverse is a band or dense matrix.

Another question I had for the interface was whether it is better to have the `KnotVector` class keep the inverse or factorization internal so the user avoids the direct usage of a `MatrixInverse`. Then a member function in `KnotVector` would enable solving with the factorization or inverse. For all of these questions and issues, the goal is for this miniapp to be easily readable, so it is obvious whether a band matrix factorization is used without recomputing.

PR #5334, opened 2026-05-18
VectorFEMassIntegrator ApplyPA improvements

helloworld922:
Extracted from #5323. Depends on #5330.

- Re-arranges AddMultPA/AddAbsMultPA/etc. in VectorFEMassIntegrator to use the kernel dispatch framework
- Optimized the H(curl)-H(curl) 3D mass integrator performance on device.

Performance on matrix:

hcurl_mass_1
hcurl_mass_2
hcurl_mass_3

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#5334](https://github.com/mfem/mfem/pull/5334) | @helloworld922 | @v-dobrev | @mlstowell + @dylan-copeland + @camierjs | 7/6/26 | 8/7/26 | 8/9/26 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
[...]

PR #5386, opened 2026-06-29
Curl Interpolator PA

helloworld922:
Extracted from #5323, depends on #5334

Implemented a "partial assembly" curl interpolator for 3D H(curl) -> H(div).

| [#5386](https://github.com/mfem/mfem/pull/5386) | @helloworld922 | @tzanio | @pazner + @psocratis + @dylan-copeland + @tzanio + @mlstowell | 8/10/26 | 8/12/26 | 8/14/26 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


tzanio:

CHANGELOG
```diff
@@ -92,6 +92,8 @@ GPU computing
 
 - Added device assembly support for 3D H(curl) VectorFEDomainLFIntegrator.
 
+- Added support for device "partial assembly" CurlInterpolator.
```
dylan-copeland: The CHANGELOG has many instances of `partial assembly` without quotation marks.

tests/unit/fem/test_lin_interp.cpp
```diff
@@ -1069,4 +1069,75 @@ TEST_CASE("Exact Sequence Properties: d(df)=0",
    }
 }
 
+TEST_CASE("Partial Assemble Linear Interpolator",
+          "[CurlInterpolator]"
+          "[GPU]")
+{
+   const int maxOrder = 3;
```
dylan-copeland: `constexpr`

tests/unit/fem/test_lin_interp.cpp
```diff
@@ -1069,4 +1069,75 @@ TEST_CASE("Exact Sequence Properties: d(df)=0",
...
+   auto order = GENERATE_COPY(range(1, maxOrder + 1));
+   CAPTURE(order);
+
+   int n = 3, dim = -1;
+   real_t tol = 1e-10;
```
dylan-copeland: Does this work for single-precision builds?

fem/bilininteg.hpp
```diff
@@ -4064,12 +4064,61 @@ class VectorIdentityInterpolator : public IdentityInterpolator
     discrete curl matrix. */
 class CurlInterpolator : public DiscreteInterpolator
 {
+   /// 1D finite elements that generate and own the 1D DofToQuad maps below
```
dylan-copeland, line 4068: Comment that all this data is for PA only?

fem/integ/bilininteg_interp_pa.cpp
```diff
@@ -14,9 +14,218 @@
...
 
+#include "bilininteg_hcurlhdiv_kernels.hpp"
+
 namespace mfem
 {
 
+namespace
+{
+
+void PAHcurlApplyCurl2D(const int c_dofs1D,
```
dylan-copeland, line 25: The unit test is only in 3D. Should a 2D test be added?

CHANGELOG
helloworld922: I put partial assembly in quotations because it's really an interpolation, and it's not really clear to me what the distinction between partial assembly and matrix free should be in this context. It does seem much closer to "matrix free" than partial assembled since the assembly only needs information that's defined for a single reference element, but the other interpolators treat this as "partial assembly" as well.

dylan-copeland: I think we should make this clearer and remove the quotation marks, to avoid confusion. Maybe a longer comment should be made, explaining the change.

helloworld922: Updated the changelog text

tests/unit/fem/test_lin_interp.cpp
helloworld922: fixed

tests/unit/fem/test_lin_interp.cpp
helloworld922: It appears to work. We can loosen the tolerances if this fails on some platforms/compilers.

fem/bilininteg.hpp
helloworld922: updated

fem/integ/bilininteg_interp_pa.cpp
helloworld922: Added 2D in-plane and out-of-plane tests. This also revealed a bug in the full assembly versions which was fixed.

helloworld922:
@dylan-copeland @mlstowell @pazner @psocratis @tzanio @camierjs all the changes in this PR are also in #5400. Feel free to review only that PR if you want to see everything in one place, I will keep that branch in sync with all updates from this PR.

dylan-copeland (approved):

mlstowell (approved):

helloworld922:
@dylan-copeland @mlstowell I reverted the change of removing `ProjectGrad` from RT quads and triangles since this appears to be used by HypreAMS. It's not clear to me if it's ok to change HypreAMS to use a negative CurlInterpolator instead of GradInterpolator for all possible edge fespaces so I left this for compatibility reasons.

tzanio:
> It's not clear to me if it's ok to change HypreAMS to use a negative CurlInterpolator instead of GradInterpolator for all possible edge fespaces so I left this for compatibility reasons.

If this was used for the discrete gradient operator, then the sign doesn't matter and you can use the new (fixed) CurlInterpolator in 2D. (The reason is that we use that for a subspace correction and that depend only on the range of the operator)

helloworld922:
It looks like `BatchedLOR_AMS` also generates the same sign convention for the assembled LOR AMS gradient. Do we also want to change this for RT spaces?

tzanio:
> It looks like BatchedLOR_AMS also generates the same sign convention for the assembled LOR AMS gradient. Do we also want to change this for RT spaces?

Sorry, can you elaborate? The sign (or more generally any rescaling) of the discrete gradient doesn't matter in AMS, but if we want to avoid confusion it is good to be consistent.

tzanio:
🤖 Review by Fable5

1️⃣  ⚠️  Behavioral/API break not called out in the `CHANGELOG`. The 2D RT `ProjectCurl` sign is flipped and the `ProjectGrad` overrides on `RT_QuadrilateralElement`/`RT_TriangleElement` are removed (calls now hit the base-class error). Downstream users of the old rotated-gradient convention (as lor_ams.cpp and hypre.cpp had to be) will get silently sign-flipped results. This deserves an explicit `CHANGELOG` entry under "API changes", not just the PA feature note.

2️⃣  🐛 Possible dangling `IntegrationRule` in the `DofToQuad` cache (`bilininteg_interp_pa.cpp`, 2D `AssemblePA`): `closed_ir`/`open_ir` are stack locals passed to `GetDofToQuad(...)`, which caches the map keyed by the IntRule pointer inside the FE (`closed_dofquad_fe`). After `AssemblePA` returns, the cached maps hold pointers to destroyed rules; a second `AssemblePA` call on the same object could spuriously match (or miss) the cache via a dangling pointer compare. Store the rules as members, or use `IntRules.Get(...)` for persistent rules.

3️⃣  🐛 Test tolerance is double-precision only: `tol = 1e-10` in `TestCurl/CompareCurlPA` will fail under `MFEM_USE_SINGLE` (cf. the recent "Fixed a test for single precision" commit on 4.10-dev). Use a real_t-aware tolerance (e.g. scale by `std::numeric_limits<real_t>::epsilon()`).

4️⃣ ⚡ Extra temporary and full pass in the L2 path of `PADiscreteLinearOperatorExtension::AddMult: tempY.SetSize(y.Size()); MultTranspose(...); y += tempY;` allocates and adds a whole vector per call. `L2ElementRestriction` supports `AddMultTranspose(localTest, y)` directly — that would avoid tempY entirely. Also note tempY needs to be a mutable member or this allocates every apply.

5️⃣ ❓ `NBZ3D` shared-memory heuristic looks inverted (`bilininteg_hcurlhdiv_kernels.hpp`): `smem_req` is computed already multiplied by tmp batches, and then `min(tmp, ceil(48K / smem_req))` divides the 48 KB budget by the total for tmp batches rather than the per-batch footprint — this under-batches whenever `tmp > 1`. If intentional (conservatism), a comment would help; otherwise compute per-batch bytes and divide the budget by that.

6️⃣ 💡 `dim`, `ne`, `ndof_o`, `nquad_o` in `CurlInterpolator` are uninitialized until `AssemblePA` (`bilininteg.hpp`). `AddMultPA` on an un-assembled object reads garbage `dim`. Cheap fix: default-initialize (`int dim = 0, ne = 0; ...`) like the other members, and the existing `MFEM_ABORT("Bad dimension!")` fallbacks then fire cleanly. Same header: an enum for `pa_mode_2d` (0/1/2 are magic numbers) would read better.

7️⃣ Minor: the `error.cpp` null-msg guard and the `MixedVectorWeakCurlIntegrator MQ` abort are unrelated drive-bys — fine to keep but worth a sentence in the PR description; `AssemblePA` 2D could use `dom_fel->GetOrder()` instead of the extra `dynamic_casts` to get the order; the doc typo fix `$H(curl$ → $H(curl)$` is appreciated.

general/error.cpp
```diff
@@ -171,7 +171,14 @@ void mfem_error(const char *msg)
 #ifdef MFEM_USE_EXCEPTIONS
    if (mfem_error_action == MFEM_ERROR_THROW)
    {
-      throw ErrorException(msg);
+      if (msg)
+      {
+         throw ErrorException(msg);
```
pazner: Maybe just `throw ErrorException(msg ? msg : "");` (nitpick; sorry).

fem/bilinearform_ext.cpp
```diff
@@ -2036,7 +2045,18 @@ void PADiscreteLinearOperatorExtension::AddMult(
    }
    else
    {
-      mfem_error("In this setting you need a real ElementRestriction!");
+      const L2ElementRestriction* l2_elem_restrict =
+         dynamic_cast<const L2ElementRestriction*>(elem_restrict_test);
+      if (l2_elem_restrict)
```
pazner: Better to get rid of the `if` and use `MFEM_VERIFY(l2_elem_restrict, …`

fem/bilinearform_ext.cpp
```diff
@@ -1997,7 +1997,16 @@ void PADiscreteLinearOperatorExtension::Assemble()
...
-      mfem_error("A real ElementRestriction is required in this setting!");
+      const L2ElementRestriction* l2_elem_restrict =
+         dynamic_cast<const L2ElementRestriction*>(elem_restrict_test);
+      if (l2_elem_restrict)
+      {
+         test_multiplicity = 1.0;
+      }
+      else
+      {
+         mfem_error("A real ElementRestriction is required in this setting!");
```
pazner: Ditto previous comment

fem/lor/lor_ams.cpp
```diff
@@ -94,10 +94,10 @@ void BatchedLOR_AMS::Form2DEdgeToVertex_RT(Array<int> &edge2vert)
          const int iv0 = ix + iy*op1;
          const int iv1 = ix1 + iy1*op1;
 
-         // Rotated gradient in 2D (-dy, dx), so flip the sign for the first
-         // component (c == 0).
```
pazner, line 98: OK, turns out I did think about this rot/curl issue at some point in the past 🤔

helloworld922:
> 2️⃣ 🐛 Possible dangling `IntegrationRule` in the `DofToQuad` cache (`bilininteg_interp_pa.cpp`, 2D `AssemblePA`): `closed_ir`/`open_ir` are stack locals passed to `GetDofToQuad(...)`, which caches the map keyed by the IntRule pointer inside the FE (`closed_dofquad_fe`). After `AssemblePA` returns, the cached maps hold pointers to destroyed rules; a second `AssemblePA` call on the same object could spuriously match (or miss) the cache via a dangling pointer compare. Store the rules as members, or use `IntRules.Get(...)` for persistent rules.

This is a real bug with spurious failures, and looks like it is present in multiple places outside of just CurlInterpolator. I'll have to think more about what's the best way to fix this, do we want to defer fixing this to a future PR @tzanio? The AI suggested fix isn't sufficient in all cases.

pazner (approved):
Thanks @helloworld922! Just some minor comments…

general/error.cpp
helloworld922: fixed

fem/bilinearform_ext.cpp
helloworld922: fixed

fem/bilinearform_ext.cpp
helloworld922: fixed

tzanio:
> I'll have to think more about what's the best way to fix this, do we want to defer fixing this to a future PR @tzanio?

Yes, we can defer it, thanks @helloworld922 !

tzanio (approved):
Thanks @helloworld922 !

tzanio:

psocratis (approved):

tzanio:
Note that this PR changes the column order in the AMS discrete gradient, which can affect AMG coarsening. The results should be similar overall, but they will be more than round-off and in particular the iteration count can differ slightly.

PR #5389, opened 2026-07-01
Fix cuDSS build for mpich

hughcars:
The cuDSS build always baked in the prebuilt Open MPI communication layer (`libcudss_commlayer_openmpi.so`) whenever MPI was enabled, regardless of the linked MPI. cuDSS needs a layer matching the MPI, so MPICH builds pick up the Open MPI layer and abort in `CuDSSSolver::CuDSSSolver(MPI_Comm)` with `cudssSetCommLayer` returning `CUDSS_STATUS_INVALID_VALUE`.

Detect the MPI at configure time (`MFEM_MPI_IS_OPENMPI` / `MFEM_MPI_IS_MPICH`, from the `OMPI_MAJOR_VERSION` / `MPICH_VERSION` macros, same `check_cxx_source_compiles` pattern as `FindHYPRE`) and only use the Open MPI layer when the MPI isn't MPICH-family. NVIDIA ships a prebuilt layer for Open MPI only, so for MPICH we look for a user-built one and otherwise leave `MFEM_CUDSS_COMM_LIB` unset, deferring to the `CUDSS_COMM_LIB` env var (buildable via cuDSS's `cudss_build_commlayer.sh`). We only deviate from the Open MPI default on a positive MPICH match, so Open MPI builds are unchanged, and an explicit `CUDSS_COMM_LIB` still wins. `config/defaults.mk` mirrors this for the Make build.

Validation: configured/built with CUDA 13.1 + cuDSS 0.7.1 against Open MPI 5.0 and MPICH 5.0, both CMake and Make. Open MPI bakes the layer as before; MPICH leaves it unset (or honors `CUDSS_COMM_LIB`). `CuDSSSolver` constructs and solves under MPICH with a matching layer, where it previously aborted.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -86,6 +86,11 @@ Miscellaneous
   using the new method ApplyDofSigns() in class ParFiniteElementSpace: the
   method will return immediately if no sign flips are needed.
 
+- Fixed the cuDSS build to select the communication layer matching the linked
+  MPI implementation instead of always using the Open MPI layer. MPICH builds
+  previously failed at CuDSSSolver construction with cudssSetCommLayer returning
+  CUDSS_STATUS_INVALID_VALUE.
+
 
 Version 4.9, released on Dec 11, 2025
 =====================================
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,6 +320,37 @@ endif()
 if (MFEM_USE_MPI)
   find_package(MPI REQUIRED)
   set(MPI_CXX_INCLUDE_DIRS ${MPI_CXX_INCLUDE_PATH})
+  # Detect the MPI implementation (vendor) once, for downstream code that needs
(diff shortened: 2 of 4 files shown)


pazner:
@hughcars, the behavior we currently support is that, for non-OpenMPI builds, the user can either set the environment variable `CUDSS_COMM_LIB`, which will override the default which is set at build-time with `MFEM_CUDSS_COMM_LIB`, or, the user can set, via CMake or makefile build system, the configuration variable `MFEM_CUDSS_COMM_LIB` which will set the default comm lib to be used if the environment variable `CUDSS_COMM_LIB` is not set.

Is this sufficient to resolve your issue?

We discussed a more sophisticated system where we could automatically build the comm layer using the same MPI implementation as MFEM is built with, but we decided that (for now) this added more complexity that it is worth.

hughcars:
Hi @pazner, I think that should work fine, i wasn't aware of that flag, I'll close this.

PR #5392, opened 2026-07-02
Remove NURBSext from mesh

IdoAkkerman:
This is a first step in a larger NURBS rewrite. The goal is to undo some spaghetti-stuff and rationalize.

The idea in this PR is that NURBSExt is mostly an extension of fespace, a NURBS mesh will have access to this NURBSExtension via the Nodes Gridfunction.

So all direct access in the mesh or via the mesh, are now redirected (using a function) to point to the Nodes->FESpace NURBS extension. Checking if the mesh is NURBS mesh is also performed via a function. As the direct access of the NURBSextension was all over the place this PR seems bigger then it actually is.


v-dobrev:
Hi @IdoAkkerman, if this is ready for review, please don't forget to add the `ready-for-review` label -- otherwise we don't see the PR in our review app.

mesh/mesh.hpp
```diff
@@ -1375,6 +1369,13 @@ class Mesh
                            real_t &kappa_min, real_t &kappa_max,
                            Vector *Vh = NULL, Vector *Vk = NULL);
 
+   /// Return true if the face is a conforming face.
```
dylan-copeland: This comment is misplaced.

mesh/mesh.hpp
```diff
@@ -1375,6 +1369,13 @@ class Mesh
...
+   bool IsNURBS() const;
+
+   NURBSExtension* NURBSExt();
```
dylan-copeland, line 1382: Please document, including a comment on what is returned if not a NURBS mesh.

dylan-copeland: Also, since this is an extension of the API, others should comment if they have a preference for the name. It would be consistent with `FiniteElementSpace` to use the name `GetNURBSExt()`, which is a little longer.  I am not sure we should change the name, just bringing it up for consideration.

mesh/mesh.cpp
```diff
@@ -11454,7 +11391,7 @@ void Mesh::Swap(Mesh& other, bool non_geometry)
 
    if (non_geometry)
    {
-      mfem::Swap(NURBSext, other.NURBSext);
+      // mfem::Swap(NURBSext, other.NURBSext);
```
dylan-copeland: Should this be commented out?

examples/ex28p.cpp
```diff
@@ -169,7 +169,7 @@ int main(int argc, char *argv[])
    //    (degree elevated) NURBS space associated with the mesh nodes.
    FiniteElementCollection *fec;
    ParFiniteElementSpace *fespace;
-   const bool use_nodal_fespace = pmesh->NURBSext;
+   const bool use_nodal_fespace = pmesh->NURBSExt();
```
dylan-copeland: `NURBSExt()` -> `IsNURBS()`?

fem/fespace.hpp
```diff
@@ -637,9 +641,16 @@ class FiniteElementSpace
...
+   /// Return true if the space is a NURBS space
+   bool IsNURBS() const
+   {
+      return NURBSext;
+   }
+
    const NURBSExtension *GetNURBSext() const { return NURBSext; }
    NURBSExtension *GetNURBSext() { return NURBSext; }
    NURBSExtension *StealNURBSext();
+   void OwnNURBSext(NURBSExtension *NURBSext = NULL);
```
dylan-copeland, line 654: To avoid confusion with the class member `NURBSext`, change the argument name to `ext`?

fem/fespace.cpp
```diff
@@ -2640,6 +2650,18 @@ NURBSExtension *FiniteElementSpace::StealNURBSext()
    return NURBSext;
 }
 
+void FiniteElementSpace::OwnNURBSext(NURBSExtension *ext)
+{
+   MFEM_VERIFY(!own_ext, "FESpace already owns the NURBSExtension");
+
+   if (ext)
```
dylan-copeland: `if (NURBSext)`?

mesh/mesh.hpp
IdoAkkerman: I agree... I was inspired by the `GridFunction` call `FESpace()`, without the Get. 
I would suggest to drop the Get in FESpace? But this should perhaps be discussed in a generic MFEM meeting? Perhaps a claude agent can highlight these type of "inconsistencies"? And make a PR? What is the current state/policy regarding this?

Making things consist is a plus, I guess. At the risk of AI creating a lot of churn with little benefit.

mesh/mesh.cpp
IdoAkkerman: I think so... do you know if the code inside this if statement is part of the regression testing?

I'll investigate.

examples/ex28p.cpp
IdoAkkerman: Done

fem/fespace.cpp
IdoAkkerman: I think either way would work. Changed it any way.

mesh/mesh.cpp
IdoAkkerman: The If branch is part of several tests. 
Moreover, mesh does not have an extension so it can not be copied.
It retrieves the extension for the Nodes, which is swapped.

So swap should not happen.

I removed it.

mesh/mesh.hpp
IdoAkkerman: Corrected.

fem/fespace.cpp
dylan-copeland: My understanding is that it should check whether the class member `NURBSext` has already been set to something other than `nullptr` whereas `ext` is assumed to be a valid pointer, so it should be `NURBSext` that is checked here.

fem/fespace.hpp
dylan-copeland: Ping @IdoAkkerman

dylan-copeland (approved):

PR #5412, opened 2026-07-14
Better support for using `mfem.hpp` in pure C++ sources when MFEM is built with CUDA or HIP

v-dobrev:
When MFEM is built with CUDA/HIP, the class `SparseMatrix` contains member variables that have CUDA/HIP types. This means that we need to include the CUDA/HIP headers even when not using the CUDA/HIP language extensions.

To compile a pure C++ source file (without the use of the CUDA/HIP language extensions), one needs to:
* Add the paths to the CUDA/HIP headers to the compile command; e.g. `-I/usr/local/cuda-12.6/include` or `-I/opt/rocm-6.4.1/include`.
* In pure C++ mode, the required HIP headers need the compiler option `-D__HIP_PLATFORM_AMD__=1`.

Note: features in `mfem.hpp` that require the CUDA/HIP language extensions, such as the `mfem::forall` template function, cannot be used in pure C++ mode and will generate compilation errors that indicate that the code being compiled needs the CUDA/HIP language.

Note: This PR also fixes a few compiler warnings in GSLIB-related source files -- when comparing signed and unsigned integers, explicitly cast the signed version to unisigned (in all cases the signed number is >= 0).

| [#5412](https://github.com/mfem/mfem/pull/5412) | @v-dobrev | @tzanio | @najlkin + @jandrej + @pazner | 7/28/26 | 8/13/26 | 8/15/26 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>

### DIFF
--- a/general/backends.hpp
+++ b/general/backends.hpp
@@ -14,7 +14,7 @@
 
 #include "../config/config.hpp"
 
-#if defined(MFEM_USE_CUDA) && defined(__CUDACC__)
+#if defined(MFEM_USE_CUDA)
 #include <cusparse.h>
 #include <library_types.h>
 #include <cuda_runtime.h>
@@ -22,7 +22,7 @@
 #endif
 #include "cuda.hpp"
 
-#if defined(MFEM_USE_HIP) && defined(__HIP__)
+#if defined(MFEM_USE_HIP)
 #include <hip/hip_runtime.h>
 #endif
 #include "hip.hpp"
@@ -45,15 +45,17 @@
 #endif
 
 #if !defined(MFEM_USE_CUDA_OR_HIP)
-constexpr bool mfem_use_gpu = false;
-#define MFEM_DEVICE
-#define MFEM_HOST
-#define MFEM_LAMBDA
-// #define MFEM_HOST_DEVICE // defined in config/config.hpp
 // MFEM_DEVICE_SYNC is made available for debugging purposes
 #define MFEM_DEVICE_SYNC
 // MFEM_STREAM_SYNC is used for UVM and MPI GPU-Aware kernels
 #define MFEM_STREAM_SYNC
+#endif
+
+#if !defined(MFEM_USE_CUDA_OR_HIP_LANG)
+#define MFEM_DEVICE
+#define MFEM_HOST
+#define MFEM_LAMBDA
+// #define MFEM_HOST_DEVICE // defined in config/config.hpp
 #define MFEM_LAUNCH_BOUNDS(...)
 #endif
 
@@ -107,6 +109,19 @@ MFEM_HOST_DEVICE T AtomicAdd(T &add, const T val)
    add += val;
    return old;
 #endif
+}
+
+namespace mfem::internal
+{
+
+template <bool using_cuda_or_hip_language>
+void StaticAssertCudaOrHipLanguage()
+{
+   static_assert(
+      using_cuda_or_hip_language,
+      "The calling function needs to be compiled with CUDA/HIP language!");
+}
+
 }
 
 #endif // MFEM_BACKENDS_HPP
--- a/general/backends.hpp
+++ b/general/backends.hpp
@@ -115,16 +115,16 @@ namespace mfem::internal
 {
 
 #if defined(MFEM_USE_CUDA_OR_HIP) && !defined(MFEM_USE_CUDA_OR_HIP_LANG)
-static constexpr bool using_cuda_or_hip_language = false;
+static constexpr bool can_compile_kernels = false;
 #else
-static constexpr bool using_cuda_or_hip_language = true;
+static constexpr bool can_compile_kernels = true;
 #endif
 
-template <bool using_cuda_or_hip_language=using_cuda_or_hip_language>
-void RequireCudaOrHipLanguage()
+template <bool can_compile_kernels = can_compile_kernels>
+void RequireKernelCompilation()
 {
    static_assert(
-      using_cuda_or_hip_language,
+      can_compile_kernels,
       "The calling function needs to be compiled with CUDA/HIP language!");
 }
 
--- a/general/backends.hpp
+++ b/general/backends.hpp
@@ -114,8 +114,14 @@ MFEM_HOST_DEVICE T AtomicAdd(T &add, const T val)
 namespace mfem::internal
 {
 
-template <bool using_cuda_or_hip_language>
-void StaticAssertCudaOrHipLanguage()
+#if defined(MFEM_USE_CUDA_OR_HIP) && !defined(MFEM_USE_CUDA_OR_HIP_LANG)
+static constexpr bool using_cuda_or_hip_language = false;
+#else
+static constexpr bool using_cuda_or_hip_language = true;
+#endif
+
+template <bool using_cuda_or_hip_language=using_cuda_or_hip_language>
+void RequireCudaOrHipLanguage()
 {
    static_assert(
       using_cuda_or_hip_language,
--- a/general/forall.hpp
+++ b/general/forall.hpp
@@ -1051,6 +1051,10 @@ inline void ForallWrap(const bool use_dev, const int N,
    MFEM_CONTRACT_VAR(d_body);
    if (!use_dev) { goto backend_cpu; }
 
+#if defined(MFEM_USE_CUDA_OR_HIP) && !defined(MFEM_USE_CUDA_OR_HIP_LANG)
+   internal::StaticAssertCudaOrHipLanguage<false>();
+#endif
+
 #if defined(MFEM_USE_RAJA) && defined(RAJA_ENABLE_CUDA) && defined(__CUDACC__)
    // If Backend::RAJA_CUDA is allowed, use it
    if (Device::Allows(Backend::RAJA_CUDA))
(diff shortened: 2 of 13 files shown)


jandrej (approved):

Pennycook (approved):

tzanio:

tests/benchmarks/bench_ceed.cpp
```diff
@@ -32,7 +32,9 @@
 // Custom benchmark arguments generator
 static void CustomArguments(bm::Benchmark *b) noexcept
 {
-   constexpr int MAX_NDOFS = 16 * 1024 * (mfem_use_gpu ? 1024 : 8);
+   const bool mfem_use_gpu =
+      Device::Allows(Backend::DEVICE_MASK & ~Backend::DEBUG_DEVICE);
+   const int MAX_NDOFS = 16 * 1024 * (mfem_use_gpu ? 1024 : 8);
```
hughcars: `CustomArguments` runs while the benchmarks are registered, before `main()` reads `--benchmark_context=device=...` and creates `Device`. The CUDA/HIP CTest runs therefore still look like CPU runs here, reducing the largest registered problem from `n=240` to `n=48`. This suggestion restores the previous compile-time behavior. If the limit is meant to follow the runtime device instead, registration has to happen after `Device` is configured.

```suggestion
#if defined(MFEM_USE_CUDA_OR_HIP_LANG)
   constexpr int MAX_NDOFS = 16 * 1024 * 1024;
#else
   constexpr int MAX_NDOFS = 16 * 1024 * 8;
#endif
```

general/forall.hpp
```diff
@@ -1051,6 +1051,10 @@ inline void ForallWrap(const bool use_dev, const int N,
    MFEM_CONTRACT_VAR(d_body);
    if (!use_dev) { goto backend_cpu; }
 
+#if defined(MFEM_USE_CUDA_OR_HIP) && !defined(MFEM_USE_CUDA_OR_HIP_LANG)
+   internal::StaticAssertCudaOrHipLanguage<false>();
+#endif
```
pazner: Rather than wrapping `internal::StaticAssertCudaOrHipLanguage<false>()` in the same `#if`s multiple times, can we define a `constexpr bool cuda_or_hip_language_available` in the internal namespace in backends.hpp according to the check `defined(MFEM_USE_CUDA_OR_HIP) && !defined(MFEM_USE_CUDA_OR_HIP_LANG)`, and then use it as the default template parameter in `StaticAssertCudaOrHipLanguage`? This way, the `#if`s aren't needed at the call-sites, and the `<false>` can be omitted. It could be renamed something like `internal::RequireCudaOrHipLanguage()`.

v-dobrev: I don't think the suggestion will work. It was important for the call
```c++
   internal::StaticAssertCudaOrHipLanguage<false>();
```
to be a template instantiation, otherwise you get the error even without trying to instantiate the containing template function.

If you want, give your suggestion a try -- maybe I missed something and the current source can be simplified.

pazner: I'm pretty sure it works (it seems to, I checked for CUDA). There still is a template instantiation, just with default template parameter.

In backends.hpp:
```cpp
#if defined(MFEM_USE_CUDA_OR_HIP) && !defined(MFEM_USE_CUDA_OR_HIP_LANG)
static constexpr bool using_cuda_or_hip_language = false;
#else
static constexpr bool using_cuda_or_hip_language = true;
#endif

template <bool using_cuda_or_hip_language=using_cuda_or_hip_language>
void RequireCudaOrHipLanguage()
{
   static_assert(
      using_cuda_or_hip_language,
      "The calling function needs to be compiled with CUDA/HIP language!");
}
```

And in e.g. forall.hpp
```diff
--- a/general/forall.hpp
+++ b/general/forall.hpp
@@ -1051,9 +1051,7 @@ inline void ForallWrap(const bool use_dev, const int N,
    MFEM_CONTRACT_VAR(d_body);
    if (!use_dev) { goto backend_cpu; }

-#if defined(MFEM_USE_CUDA_OR_HIP) && !defined(MFEM_USE_CUDA_OR_HIP_LANG)
-   internal::StaticAssertCudaOrHipLanguage<false>();
-#endif
+   internal::RequireCudaOrHipLanguage();

 #if defined(MFEM_USE_RAJA) && defined(RAJA_ENABLE_CUDA) && defined(__CUDACC__)
    // If Backend::RAJA_CUDA is allowed, use it
```

v-dobrev: Ok, go ahead and push these changes here. I'll run some of the tests I ran before to double check.

pazner: Pushed in 3c9ee8ff42270e59d07a349d17ff583c256bc40c.

general/backends.hpp
```diff
@@ -109,4 +111,23 @@ MFEM_HOST_DEVICE T AtomicAdd(T &add, const T val)
 #endif
 }
 
+namespace mfem::internal
+{
+
+#if defined(MFEM_USE_CUDA_OR_HIP) && !defined(MFEM_USE_CUDA_OR_HIP_LANG)
+static constexpr bool using_cuda_or_hip_language = false;
+#else
+static constexpr bool using_cuda_or_hip_language = true;
```
v-dobrev: @pazner, this variable name is not very appropriate -- right now, when building without CUDA/HIP, we get the definition: `using_cuda_or_hip_language = true`.

How about renaming it to `can_compile_all_backends`, or `can_compile_kernels`, or something like that?

general/backends.hpp
```diff
@@ -109,4 +111,23 @@ MFEM_HOST_DEVICE T AtomicAdd(T &add, const T val)
...
+#endif
+
+template <bool using_cuda_or_hip_language=using_cuda_or_hip_language>
+void RequireCudaOrHipLanguage()
```
v-dobrev: Similarly here, rename `using_cuda_or_hip_language` as above.

Also, a better name for the function is `RequireKernelCompilation` or something like that.

tests/benchmarks/bench_ceed.cpp
v-dobrev: > If the limit is meant to follow the runtime device instead, registration has to happen after Device is configured.

I think this is what makes sense. @camierjs, do you know of a simple way to do this?

general/backends.hpp
v-dobrev: Another thing: it is probably best to not have variables like `using_cuda_or_hip_language` whose value changes based on the compiler and or its options -- I'm not sure what happens when different translation units define this in different ways. Instead, it is probably best to use a preprocessor macro.

general/backends.hpp
v-dobrev: Also, maybe it will be clearer if the template parameter is not hidden but used explicitly when calling this:
```c++
   RequireKernelCompilation<MFEM_CAN_COMPILE_KERNELS>();
```
where `MFEM_CAN_COMPILE_KERNELS` is the potential macro replacing `using_cuda_or_hip_language`, as I suggested in the other comment.

general/backends.hpp
pazner: > @pazner, this variable name is not very appropriate -- right now, when building without CUDA/HIP, we get the definition: `using_cuda_or_hip_language = true`.
> 
> How about renaming it to `can_compile_all_backends`, or `can_compile_kernels`, or something like that?

Yes, I agree the name is bad. I just copied the same name as the template argument. I considered changing it, but didn't come up with anything better. `can_compile_kernels` is fine by me.

> Another thing: it is probably best to not have variables like `using_cuda_or_hip_language` whose value changes based on the compiler and or its options -- I'm not sure what happens when different translation units define this in different ways. Instead, it is probably best to use a preprocessor macro.

I don't think this is an issue. `constexpr` implies `const`, which implies internal linkage, meaning each TU gets its own copy of the variable, and the values can be different across different TUs. See here: https://stackoverflow.com/a/50489130

general/backends.hpp
pazner: > Similarly here, rename `using_cuda_or_hip_language` as above.
> 
> Also, a better name for the function is `RequireKernelCompilation` or something like that.

Agreed to both.

> Also, maybe it will be clearer if the template parameter is not hidden but used explicitly when calling this:
> 
> ```c++
>    RequireKernelCompilation<MFEM_CAN_COMPILE_KERNELS>();
> ```
> 
> where `MFEM_CAN_COMPILE_KERNELS` is the potential macro replacing `using_cuda_or_hip_language`, as I suggested in the other comment.

The way I thought about it is that the template parameter is an implementation detail, and it's always intended to be called with the default template parameter. In this case, `MFEM_CAN_COMPILE_KERNELS` is just boilerplate.

pazner: @v-dobrev, do you want to make the edits (since it's your PR) or should I (since I committed the code)?

v-dobrev: Um, I'll go ahead and make the changes. I don't feel that strongly that `RequireKernelCompilation` should have the template parameter explicitly, so I won't change that.

pazner (approved):
Thanks @v-dobrev!

tzanio (approved):
thanks @v-dobrev !

tzanio:

tests/benchmarks/bench_ceed.cpp
camierjs: That looks good to me.

v-dobrev: After talking to @camierjs, we'll postpone the change to make the problem size selection based on the runtime device configuration. This will be done in a later PR.

For now, I'll merge your suggestion, @hughcars. Thanks for pointing out this issue!

tzanio:

PR #5111, opened 2025-11-13
Add MultTranspose for DSmoother and GSSmoother

pazner:
- For non-symmetric matrices, this requires forming the transpose matrix.
- For l1-Jacobi and lumped Jacobi, we don't implement the transpose action for non-symmetric matrices.

This allows DSmoother and GSSmoother to be used in Multigrid, which requires the transpose action for the post-smoothers.

 | PR | Author | Editor | Reviewers | Assignment | Approval | Merge|
 | --- | --- | --- | --- | --- | --- | --- |
| [#5111](https://github.com/mfem/mfem/pull/5111) | @pazner | @tzanio | @dylan-copeland + @psocratis | 11/13/25 | 12/25/25 | 12/29/25 | |

<details>
<summary>PR Checklist</summary>
    
- [ ] Code builds.
- [ ] Code passes `make style`.
- [ ] Update `CHANGELOG`:
    - [ ] Is this a new feature users need to be aware of? New or updated example or miniapp?
    - [ ] Does it make sense to create a new section in the `CHANGELOG` to group with other related features?
- [ ] Update `INSTALL`:
    - [ ] Had a new optional library been added? If so, what range of versions of this library are required? (*Make sure the external library is compatible with our BSD license, e.g. it is not licensed under GPL!*)
    - [ ] Have the version ranges for any required or optional libraries changed?
    - [ ] Does `make` or `cmake` have a new target?
    - [ ] Did the requirements or the installation process change? *(rare)*
- [ ] Update continuous integration server configurations if necessary (e.g. with new version requirements for each of MFEM's dependencies)
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
    - [ ] Check if `make distclean; git status` shows any files that were generated from the source by the project (not an IDE) but we don't want to track in the repository.
    - [ ] Add new patterns (just for the new files above) and re-run the above test.
- [ ] New examples:
    - [ ] All sample runs at the top of the example source file work.
    - [ ] Update `examples/makefile`:
      - [ ] Add the example code to the appropriate `SEQ_EXAMPLES` and `PAR_EXAMPLES` variables.
      - [ ] Add any files generated by it to the `clean` target.
      - [ ] Add the example binary and any files generated by it to the top-level `.gitignore` file.
    - [ ] Update `examples/CMakeLists.txt`:
      - [ ] Add the example code to the `ALL_EXE_SRCS` variable.
[...]

PR #5413, opened 2026-07-14
NURBS mesh physical spacing

dylan-copeland:
A new miniapp `nurbs_spacing` recomputes the control points of a NURBS mesh so that the relative spacing of elements in each patch, in physical space, is approximately the spacing given by the corresponding knot vector in reference (or parameter) space. Extrusion of NC-patch NURBS meshes is also enabled. The unit test is for NC-patch NURBS construction and extrusion, not for physical spacing (which is covered by the new miniapp). Note that `NURBSExtension::PhysicalSpacing` uses `patchCP`, the control points representing each patch as a single element. This could be changed in the future so that meshes without `patchCP` can still use `PhysicalSpacing`.

| [#5413](https://github.com/mfem/mfem/pull/5413) | @dylan-copeland | @tzanio | @tzanio + @kmittal2 + @v-dobrev + @IdoAkkerman | 7/28/26 | ⌛due 8/11/26 | ⌛due 8/18/26 | |

<details>
    
- [ ] Code builds.
- [ ] Update `CHANGELOG`:
- [ ] Update `INSTALL`:
    - [ ] `.github`
    - [ ] `.appveyor.yml`
- [ ] Update `.gitignore`:
- [ ] New examples:
- [ ] New miniapps:
- [ ] New capability:
</details>


mesh/mesh.cpp
```diff
@@ -5246,6 +5248,14 @@ void Mesh::Loader(std::istream &input, int generate_edges,
...
          MFEM_VERIFY(ident == "patch_cp", "Invalid mesh format");
          NURBSext->ReadCoarsePatchCP(input);
+
+         // Check for the optional section "patch_w"
+         if (input.peek() == 'p')
+         {
+            input >> ident;
+            MFEM_VERIFY(ident == "patch_w", "Invalid mesh format");
+            NURBSext->ReadCoarsePatchWeights(input);
+         }
```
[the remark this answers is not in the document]
dylan-copeland: Done in 0434220bd88527bbde986a154390e48cfa70b2d3

miniapps/nurbs/nurbs_spacing.cpp
```diff
@@ -0,0 +1,108 @@
...
+      delete mesh2D;
+   }
+   else
+   {
+      mesh = new Mesh(mesh_file, 1, 1);
+   }
+
+   // Solve for the physical spacing of elements.
+   mesh->NURBSext->PhysicalSpacing(*mesh->GetNodes(), sweep1D);
+
```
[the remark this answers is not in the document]
dylan-copeland: Done in 0434220bd88527bbde986a154390e48cfa70b2d3

tzanio:

linalg/sparsesmoothers.hpp
```diff
@@ -51,9 +51,9 @@ class GSSmoother : public SparseSmoother
 public:
    enum GSType
    {
-      SYMMETRIC, ///< Forward Gauss-Seidel, then backward.
-      FORWARD, ///< Forward Gauss-Seidel ($L^{-1}$).
-      BACKWARD ///< Backward Gauss-Seidel ($U^{-1}$).
+      GS_SYMMETRIC, ///< Forward Gauss-Seidel, then backward.
+      GS_FORWARD, ///< Forward Gauss-Seidel ($L^{-1}$).
+      GS_BACKWARD ///< Backward Gauss-Seidel ($U^{-1}$).
```
pazner: These changes, which technically break backwards compatibility (but probably don't break user code, since the names were introduced at the end of last year in #5111), were made because of a conflict in an application, likely because the [ACIS](https://en.wikipedia.org/wiki/ACIS) software uses `FORWARD` (and `REVERSED`) for its [`REVBIT` typedef](http://www.q-solid.com/ACIS_Docs_R17/online/refman/generated/refman/SPAcstr/typedef_revbit.htm#460), and it seems those are implemented as preprocessor macros.

This means that instead of specifying `GSSmoother::FORWARD` we need `GSSmoother::GS_FORWARD`, which is a little redundant.

While maintaining compatibility for applications that use ACIS is good, it's a little unfortunate to be forced into a name change because of a third-party unprefixed preprocessor macro. Any code that includes the header would not be able to use `FORWARD` as any kind of identifier (unless first doing `#undef FORWARD`).

I wonder if an application-level fix would be preferable rather than a change within MFEM itself.

One more option would be to use non-all-caps identifiers for this enum (since all-caps is sometimes used to mean a macro). Many but not all enums in MFEM use all-caps for the names.

dylan-copeland: It is usually preferred to use `enum class`, but it turns out that is not sufficient to fix the conflict. The name `FORWARD` has to be changed. Since this is an `enum`, `GS_FORWARD` can be used, which is not very long or redundant. It is true that lower-case `forward` also works.

pazner: This was discussed briefly in the PR meeting, the consensus seemed to be to go with `Symmetric`, `Forward`, and `Backward`.

dylan-copeland: Done in ed74d509b871bfbf51ed68bf0cbb7baa9bd261cb

IdoAkkerman: Perhaps I am a bit late and this issue might be resolved.... but I have two comments/questions regarding this:
1) Relying solely on lowercase vs uppercase to make a distinction between these two things seems confusing and fragile.
2) Shouldn't the namespace `GSSmoother::` and a ACIS-based namespace avoid this clash? As in using a `using namespace` is often convenient, but not the gold standard. Exactly because of the above mention clash?

Long story short: Perhaps we should discourage the use of `using namespace`.

tests/unit/mesh/test_ncp_nurbs.cpp
```diff
@@ -0,0 +1,458 @@
...
+   Vector x(dim);
+
+   // Define a grid of points at which to check the radius.
+   constexpr real_t h = 0.25;
+   constexpr real_t tol = 1.0e-2;
+
+   // Radius min/max on each element side.
+   const int numSides = dim == 2 ? 4 : 6;
+   std::vector<std::array<real_t, 2>> side(numSides);
+
```
IdoAkkerman, line 44: Are these values set at creation? On all platforms? Or does this create the possibility of "non initialized" errors on line 19?

tests/unit/mesh/test_ncp_nurbs.cpp
```diff
@@ -0,0 +1,458 @@
...
+// CONTRIBUTING.md for details.
+
+#include "mfem.hpp"
+using namespace mfem;
+
+#include "unit_tests.hpp"
+
+void SetMinMax(real_t v, std::array<real_t, 2> &m)
+{
+   if (m[0] < 0.0)
```
IdoAkkerman, line 21: Is zero a correct init value? What if the max is lower? or the Min is higher? Or is this impossible? Perhaps justify this in a comment?

IdoAkkerman: I see this routine is used for radius, so the assumption is both min and max are >=0. This answers my question. Still worth commenting that.

tests/unit/mesh/test_ncp_nurbs.cpp
IdoAkkerman: Init on line 98

tests/unit/mesh/test_ncp_nurbs.cpp
```diff
@@ -0,0 +1,458 @@
...
+   // Call UpdateNURBS(), enabling radius check.
+   mesh3D->NURBSUniformRefinement(1);
+   CheckRadius(*mesh3D);
+
+   /*
+   mesh3D->NURBSext->ConvertToPatches(*mesh3D->GetNodes());
+   std::ofstream mesh_ofs("test.mesh");
+   mesh_ofs.precision(8);
+   mesh3D->Print(mesh_ofs);
+   */
```
IdoAkkerman, line 457: Remove?

tests/unit/mesh/test_ncp_nurbs.cpp
```diff
@@ -0,0 +1,458 @@
...
+   CheckRadius(*mesh);
+
+   Mesh *mesh3D = Extrude2D(mesh, 3, 1.0);
+
+   // Call UpdateNURBS(), enabling radius check.
+   mesh3D->NURBSUniformRefinement(1);
+   CheckRadius(*mesh3D);
+
+   /*
+   mesh3D->NURBSext->ConvertToPatches(*mesh3D->GetNodes());
```
IdoAkkerman, line 453: Remove comment?

mesh/nurbs.hpp
```diff
@@ -1124,18 +1137,64 @@ class NURBSExtension
...
+       elements approximately matches the reference spacing of knots.
+
+       If patches are not already set, then they will be set by using control
+       point data from the input @a Nodes. If @a sweep1D is true, an approximate
+       1D sweeping solver is used; otherwise, a more expensive full-dimensional
+       system solver is used on each patch. The 1D sweeping solver is more
+       efficient, usually with a small loss of accuracy. The input tolerance is
+       used for solver convergence. */
+   void PhysicalSpacing(const GridFunction &Nodes, bool sweep1D = true,
+#ifdef MFEM_USE_SINGLE
```
IdoAkkerman: This looks a bit more low-level then I am comfortable(happy?) with.

In would have expected a MFEM_EPS or MFEM_TOL parameter to be available. But It is not...

Perhaps consider using 
`const mfem::real_t tol =    100 * std::numeric_limits<mfem::real_t>::epsilon();`
but this is pure personal preference wrt aesthetics... although this would also work if a third or fourth type was introduced.

PS: the 100 value is the subjective fudge factor.

mesh/nurbs.hpp
```diff
@@ -515,6 +518,12 @@ class NURBSPatch
    inline       real_t &operator()(int i, int j, int k, int l);
    inline const real_t &operator()(int i, int j, int k, int l) const;
 
+   /// Copy the control points and weights from another NURBSPatch @a p to this.
+   void SetControlPoints(const NURBSPatch &p);
```
IdoAkkerman, line 522: This seems like a doubling up of existing capability in obfuscatingly named routines (I ll get back with that)

mesh/nurbs.hpp
```diff
@@ -515,6 +518,12 @@ class NURBSPatch
...
+
+   /// Divide weights from control points.
```
IdoAkkerman, line 524: Idem

mesh/nurbs.hpp
IdoAkkerman: `SetCoordsFromPatches` and `SetSolution`

mesh/nurbs.hpp
IdoAkkerman: `ConvertToPatches` and `GetPatchNets`   (or the other way around)

IdoAkkerman: Perhaps the functionality is not 100% the same... but I think we are creating more technical debt by introducing these functions, instead of improving and rationalizing what we have. So don;t know the correct way forward. Perhaps parallel to the existing NURBSExt there should be NURBSExt2 that "does the same" but a mayor refactor...

IdoAkkerman:
I am perhaps a bit confused with the overall aim of the PR.
What is the use case? Could TMOP achieve the same? Or something similar? Is the geometry of the boundary changed?

tests/unit/mesh/test_ncp_nurbs.cpp
dylan-copeland: Right, they are initialized on line 98.

linalg/sparsesmoothers.hpp
dylan-copeland: We have no control over the ACIS code and its global definition of `FORWARD`, so we have to work around it. Unfortunately, `enum class` with `GSSmoother::FORWARD` does not avoid the conflict (I think the compiler complained in the definition of `FORWARD` in the `enum`). The only workaround I see is renaming the variable, so I proposed `GS_FORWARD` initially, but we reached a consensus on `Forward`. You made a fair point that having `FORWARD` and `Forward` differ only by case can be confusing, but this should only be a problem if someone mistakenly uses `FORWARD` with ACIS in their build, which I think is unlikely to be a serious issue.

tests/unit/mesh/test_ncp_nurbs.cpp
dylan-copeland: I included this for easy output of the mesh. It is an interesting mesh, and visualizing it helps to understand the test. Normally, it would be trivial to print a mesh, but in this case it is necessary to call `ConvertToPatches` first, and it would be hard to figure that out without this comment. This is an unfortunate complication of the NURBS classes, which we should simplify later. For now, I propose leaving this in comments.

mesh/nurbs.hpp
dylan-copeland: This function is similar to `NURBSExtension::Set2DSolutionVector` and `NURBSExtension::Set3DSolutionVector` but there are enough differences (and different classes) that I don't see a good way to avoid slight code duplication by refactoring.

mesh/nurbs.hpp
dylan-copeland: I agree that a major refactoring would be a good idea, but for now this function is needed. A positive way to view the NURBS classes is that the correct refactoring is unknown until the features and capabilities are fully known, and we have made progress toward more complete NURBS support, putting us into a good position to make a better design for these classes.

dylan-copeland:
> I am perhaps a bit confused with the overall aim of the PR. What is the use case? Could TMOP achieve the same? Or something similar? Is the geometry of the boundary changed?

Thanks @IdoAkkerman for your review! I have looked briefly into using TMOP for this purpose, but it is an unclear topic for future research. It may be challenging to formulate a TMOP method that works on the patch level instead of element level, to preserve the overall patch structure. The goal for the NURBS patch is to change the relative spacing of elements across the entire patch, while preserving the geometry and curvature of edges and faces of elements in the patch. There are clear objectives at the patch level, but not the element level. This PR is a first attempt at fixing the relative physical spacing without changing geometry. The methods are very basic and leave room for improvement, but all of my tests are working with this PR, so I am proposing this as an initial solution that could be improved later. Maybe a TMOP type of method will replace this in the future. The best way to understand the aim of the PR is to see some examples of improved spacing. There is the example with circles in the unit test and the new miniapp, and I have other meshes with good results.

mesh/nurbs.hpp
dylan-copeland: Done in 049fd20cdf4ce7d8cfd8132949bd691276fe9d61

tests/unit/mesh/test_ncp_nurbs.cpp
dylan-copeland: Comment added in 049fd20cdf4ce7d8cfd8132949bd691276fe9d61

linalg/sparsesmoothers.hpp
pazner: @IdoAkkerman, the issue is that (I believe), ACIS defines a _macro_ called `FORWARD` (and `REVERSED`). Macros in C++ always have global scope, and namespaces, class scoping, enum class, etc. cannot prevent clashes with macros. This is why it is good practice to always add a prefix to macro names (like we do with `MFEM_`). If we want compatibility for applications using ACIS (without asking users to use `#undef`), we cannot use `FORWARD` for any symbol defined in a header file (constant, class name, function, etc., no matter at what scope). Not ideal, but the fix is minor and low-impact. (To be clear, we aren't using case to distinguish to different things, only to avoid a conflict with an externally defined macro).

